PR #8, opened 2017-06-15
Feature/article list

fabiosantoscode:

PR #12, opened 2017-06-20
[Component] Image

foliveira:
Pulled from #8


foliveira:
Do not merge yet!

packages/image/image.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import { View, Image } from "react-native";
+import merge from "lodash/merge";
+
+export default class extends React.Component {
+  constructor(...args) {
+    super(...args);
```
sericaia: we can use only `super(props)` here :)

sericaia: (afaik we are not using `args` anywhere else)

packages/image/image.stories.js
```diff
@@ -0,0 +1,29 @@
...
+  halfWidthView: {
+    width: "100%"
+  },
+  customSize: {
+    width: "100px"
+  }
+});
+
+storiesOf("Image", module)
+.add("Stretched to full width keeping aspect ratio", () =>
```
sericaia: add missing identation here

packages/image/image.stories.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import { StyleSheet, View, Text } from "react-native";
+import { storiesOf } from "@storybook/react";
```
craigbilner: needs to be `react-native` or you'll get RN Document errors

packages/image/image.stories.js
```diff
@@ -0,0 +1,29 @@
...
+  uri:
+    "https://www.thetimes.co.uk/imageserver/image/methode%2Fsundaytimes%2Fprod%2Fweb%2Fbin%2F9242e576-4dfc-11e7-a20e-a11097d3353d.jpg?crop=1463%2C975%2C293%2C12&resize=320"
+};
+
+const styles = StyleSheet.create({
+  halfWidthView: {
+    width: "100%"
+  },
+  customSize: {
+    width: "100px"
```
craigbilner: `px` won't work on native

packages/image/image.stories.js
craigbilner: don't worry about commenting on these types of things - prettier should be in the pipeline

packages/image/image.stories.js
```diff
@@ -0,0 +1,29 @@
...
+  }
+});
+
+storiesOf("Image", module)
+.add("Stretched to full width keeping aspect ratio", () =>
+  <View style={styles.halfWidthView}>
+    <Image source={exampleImage} />
+  </View>
+).add("With custom width of 100px", () =>
```
craigbilner: not necessarily `px`?

packages/image/image.test.js
```diff
@@ -0,0 +1,30 @@
...
+  expect(tree).toBeTruthy();
+});
+
+it("lays out image with correct aspect ratio", done => {
+  const comp = new Image({
+    getSize: (_, cb) => {
+      cb(10, 20);
+    },
+    source: {
+      uri: "http://example.com/image.jpg"
```
craigbilner: can we test it loads the right url(i)?

packages/image/image.js
```diff
@@ -0,0 +1,39 @@
...
+        width: containerWidth,
+        height: containerWidth * height / width
+      });
+    });
+  }
+
+  render() {
+    return (
+      <View onLayout={this.handleLayout}>
+        <Image
```
craigbilner: should we add an `onload` and `onerror` event for analytics etc.?

packages/image/image.stories.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import { StyleSheet, View, Text } from "react-native";
```
craigbilner: awaiting an agreed `eslint` config but `Text` is unused

packages/image/image.js
```diff
@@ -0,0 +1,39 @@
...
+          {...this.props}
+          style={merge(this.props.style, {
```
craigbilner: should this merge with `undefined`?

packages/image/image.js
```diff
@@ -0,0 +1,39 @@
...
+    this.handleLayout = this.handleLayout.bind(this);
+    this.state = { width: 0, height: 0 };
+  }
+
+  handleLayout(event) {
+    const containerWidth = event.nativeEvent.layout.width;
+
+    const { getSize = Image.getSize } = this.props;
+
+    getSize(this.props.source.uri, (width, height) => {
```
craigbilner: this chain is not guaranteed?

packages/image/image.test.js
```diff
@@ -0,0 +1,30 @@
+import "react-native";
+import React from "react";
+import Image from "./image";
+import renderer from "react-test-renderer";
+
+it("renders correctly", () => {
+  const tree = renderer.create(<Image />).toJSON();
+
+  expect(tree).toBeTruthy();
```
craigbilner: can we add a snapshot test?

packages/image/image.js
foliveira: `merge` will still output an object if passed an undefined (or only undefined values). If you're referring to `this.props.style` being undefined the second object passed to `merge` will prevail.

packages/image/image.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import { View, Image } from "react-native";
+import merge from "lodash.merge";
+
+export default class extends React.Component {
+  constructor(...args) {
```
aronshamash: Can this still work with `props`? This is the standard way I have seen this done.

packages/image/image.test.js
foliveira: Added

packages/image/image.js
foliveira: Yes. `args` is a rest parameter (in fact an array containing [`props`, `context` and `updater`](https://github.com/facebook/react/blob/master/src/isomorphic/modern/class/ReactBaseClasses.js#L23)) that will get spread on the call to `super(...)` (which in fact is calling [this](https://github.com/facebook/react/blob/master/src/isomorphic/modern/class/ReactBaseClasses.js#L23)).

It's not entirely necessary in this case, since we're only interfacing with `props` but it was done like this following other examples that have been developed before (prev. PoC).

craigbilner: @foliveira AFAICT the docs only mention `props` and show code passing `props`, so I wouldn't both catering for the undocumented values in case they change etc.

foliveira: Happy to change it (across all components) 👍

packages/image/image.test.js
craigbilner: ah, I meant _instead of_ `toBeTruthy`

packages/image/image.test.js
```diff
@@ -0,0 +1,57 @@
...
+});
+
+it("renders snapshot correctly correctly", () => {
```
craigbilner: typo - `correctly`

packages/image/image.stories.js
```diff
@@ -0,0 +1,27 @@
...
+    "https://www.thetimes.co.uk/imageserver/image/methode%2Fsundaytimes%2Fprod%2Fweb%2Fbin%2F9242e576-4dfc-11e7-a20e-a11097d3353d.jpg?crop=1463%2C975%2C293%2C12&resize=320"
+};
+
+const styles = StyleSheet.create({
+  halfWidthView: {
+    width: "50%"
+  }
+});
+
+storiesOf("Image", module)
```
craigbilner: can we add an error state?

packages/image/image.test.js
```diff
@@ -0,0 +1,57 @@
...
+      width: 200,
+      height: 75
+    }
+  };
+  const tree = renderer.create(<Image {...props} />).toJSON();
+
+  expect(tree).toMatchSnapshot();
+});
+
+it("lays out image with correct aspect ratio", done => {
```
craigbilner: should this be a true integration test maybe with `enzyme`?

packages/image/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@timescomponents/image",
+  "version": "0.0.1",
+  "main": "image.js",
+  "scripts": {
+    "flow": "node_modules/flow-bin/cli.js",
+    "test:dev": "jest --bail --verbose --watch",
```
craigbilner: need to specifically point at test so `jest` doesn't run all the tests and needs the `rooDir` config

craigbilner:
Desired error state from apak:
img_7492

foliveira:
Is that placeholder image available somewhere? We can make it the default image for the component and have a successful load event replace it by the actual image.

As for the remaining changes, they're all done apart from the test for the error state/placeholder.

packages/image/image.stories.js
fampinheiro: I believe you were refering to https://github.com/newsuk/times-components/pull/12/files#diff-98bdad816ac1d214afb2d6aabbf739eaR33.

packages/image/image.test.js
```diff
@@ -0,0 +1,79 @@
+import "react-native";
+import React from "react";
+import Image from "./image";
+import renderer from "react-test-renderer";
+
+function testSetState(width, height, done) {
+  return ({ width, height }) => {
+    expect(width).toEqual(width);
+    expect(height).toEqual(height);
```
trevorah: eh?

fampinheiro:
Code changed according to comments.

PR #14, opened 2017-06-20
[Component] Article Summary

foliveira:
As per #8 last comments


packages/article-summary/article-summary.js
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+import { StyleSheet, Text, View } from "react-native";
+
+const styles = StyleSheet.create({
```
craigbilner: for me at least the font sizes are being overridden in the web storybook so it's hard to visually check/demo

craigbilner: On native there's no padding which seems odd?

packages/article-summary/article-summary.test.js
```diff
@@ -0,0 +1,10 @@
+import "react-native";
+import React from "react";
+import ArticleSummary from "./article-summary";
+import renderer from "react-test-renderer";
+
+it("renders correctly", () => {
+  const tree = renderer.create(<ArticleSummary />).toJSON();
+
+  expect(tree).toBeTruthy();
```
craigbilner: Can we add a snapshot test?

packages/article-summary/article-summary.js
foliveira: We need to find a way to inject that into the storybook as a global style

foliveira: `storybook-native` doesn't seem to support head scripts injection as `storybook` does, so unless a way around this is found, for the mean time default fonts will appear.

PR #23, opened 2017-06-22
chore: Introduces hoisting for dependencies of lerna packages (components)

foliveira:
With this PR if a component dependency version matches the version of the main module (or another package) dependency it gets installed into the `root/node_modules` folder instead of the component folder.

This removes duplicated dependencies living under each component folder, reducing time and space when doing an install.

Keep in mind that for this to work, dependency versions must match for those to be hoisted.

**The process works as follows:**

- Common dependencies will be installed only to the top-level node_modules, and omitted from individual package node_modules.
- Mostly-common dependencies are still hoisted, but outlier packages with different versions will get a normal, local node_modules installation of the necessary dependencies.
- In this instance, lerna bootstrap will always use npm install with the --global-style flag, regardless of client configuration.
- Binaries from those common packages are symlinked to individual package node_modules/.bin directories, so that package.json scripts continue to work unmodified.
- Well-behaved Node-based software should continue to work unmodified.

### DIFF
--- a/lib/cli/package-json.js
+++ b/lib/cli/package-json.js
@@ -1,7 +1,7 @@
 "use strict";
 
 module.exports = variables => ({
-  name: "@times-components/" + variables.packageName,
+  name: "@timescomponents/" + variables.packageName,
   version: variables.version || "0.0.1",
   description: variables.packageDescription,
   main: variables.packageName + ".js",
--- a/lib/cli/package-json.js
+++ b/lib/cli/package-json.js
@@ -1,21 +1,21 @@
 "use strict";
 
 module.exports = variables => ({
-  name: "@timescomponents/" + variables.packageName,
+  name: "@times-components/" + variables.packageName,
   version: variables.version || "0.0.1",
   description: variables.packageDescription,
   main: variables.packageName + ".js",
   scripts: {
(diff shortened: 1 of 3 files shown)


lib/cli/package-json.js
```diff
@@ -1,21 +1,21 @@
 "use strict";
 
 module.exports = variables => ({
-  name: "@timescomponents/" + variables.packageName,
+  name: "@times-components/" + variables.packageName,
```
craigbilner: it's timescomponents I'm afraid...

foliveira: I was just following the naming we have in place for the main module: `times-component`

See [here](https://github.com/newsuk/times-components/blob/master/package.json#L2)

craigbilner: I can't change the org name now :-(

foliveira: Ah fair enough. Although this will require changes to other PR's as well.

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "times-components",
+  "name": "timescomponents",
```
foliveira: @craigbilner Updated both main package name and the template for components

PR #25, opened 2017-06-23
Run prettier --write on pre-commit hook

JGAntunes:
Runs `prettier --write` on the staged .js files using [lint-staged](https://github.com/okonet/lint-staged).
Also moved the `pre-commit` dep to dev deps.

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -4,7 +4,6 @@ const cp = require("child_process");
 const fs = require("fs");
 const path = require("path");
 
-const _ = require("lodash");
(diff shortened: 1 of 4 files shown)


craigbilner:
Thanks @JGAntunes, if a naughty developer goes around the git commit hook...it won't fail on CI?

JGAntunes:
@craigbilner on the CI you can still run `prettier:diff` to make sure everything is ok. I just removed it from the pre-commit hooks because it felt like kind of redundant since we're already using `--write` before.

PR #26, opened 2017-06-23
[Component] GPT component

JGAntunes:
An ad component and its related parts, already supporting header biding also. It's still a work in progress, probably lacking further tests for the `ad-manager`.


sericaia:
@craigbilner can you please review and merge?

packages/gpt/gpt.test.js
```diff
@@ -0,0 +1,19 @@
...
+    networkId: "25436805",
+    adUnit: "d.thetimes.co.uk",
+    section: "article"
+  });
+
+  const tree = renderer
+    .create(<GPT adManager={adManager} code="ad-header" />)
+    .toJSON();
+
+  expect(tree).toBeTruthy();
```
craigbilner: can we snapshot test?

packages/gpt/gpt.web.js
```diff
@@ -0,0 +1,26 @@
...
+    this.props.adManager.registerAd(this.props.code);
+  }
+
+  componentWillUnmount() {
+    this.props.adManager.unregisterAd(this.props.code);
+  }
+
+  render() {
+    const props = this.props;
+    return <div id={props.code} />;
```
craigbilner: `this.props.code`?

packages/gpt/ad-manager.js
```diff
@@ -0,0 +1,122 @@
+import series from "run-series";
+import parallel from "run-parallel";
+
+import { getSlotConfig } from "./generate-config";
+import gptManager from "./gpt-manager";
+import pbjsManager from "./pbjs-manager";
+
+function AdManager(options = {}) {
+  if (!(this instanceof AdManager)) {
```
sericaia: image
image

Should we use this anyway?

craigbilner: I think we're quite far from serving ES2015+ code to the browser? So there'll be a transpile step - `https://www.npmjs.com/package/babel-plugin-transform-class`

packages/gpt/ad-manager.js
```diff
@@ -0,0 +1,122 @@
...
+AdManager.prototype.isReady = function isReady() {
+  return this.initialised;
+};
+
+AdManager.prototype.init = function init(callback) {
+  // Load scripts if needed be
+  gptManager.loadScript();
+  pbjsManager.loadScript();
+
+  series(
```
sericaia: Well.. `async` helps to avoid callback hell and to better structure and organise your code and thinking. So, I would say yes, we need.

packages/gpt/ad-manager.js
```diff
@@ -0,0 +1,122 @@
...
+        pbjsManager.setConfig.bind(pbjsManager)
+      ]),
+      // Init services
+      parallel.bind(parallel, [
+        gptManager.init.bind(gptManager),
+        pbjsManager.init.bind(pbjsManager, this.adQueue)
+      ])
+    ],
+    err => {
+      if (err) throw new Error(err);
```
sericaia: Are we going to have New Relic or some other tool? We should throw the error anyway, I'm not sure yet what's the guidelines for this particular project regarding that.

JGAntunes:
I think all the needed changes have been addressed, can you confirm @craigbilner?

packages/gpt/ad-manager.js
```diff
@@ -0,0 +1,130 @@
...
+  registerAd(code, { width = 1024 } = {}) {
+    this.adQueue.push(this.getSlotConfig(this.section, code, width));
+  }
+
+  unregisterAd(code) {
+    this.adQueue = this.removeItemFromQueue(this.adQueue, code);
+  }
+
+  display(callback) {
+    if (!this.initialised) {
```
craigbilner: I think we can make an assumption that this won't happen? Can we remove the defensive code to keep things cleaner? Can add a README if devs need to understand how to use it better

packages/gpt/ad-manager.js
```diff
@@ -0,0 +1,130 @@
...
+      throw new Error("Ad manager needs to be initialised first");
+    }
+
+    this.gptManager.googletag.cmd.push(() => {
+      const slot = this._createSlot(adSlotId, this.section);
+      slot.addService(this.gptManager.googletag.pubads());
+      slot.defineSizeMapping(this._generateSizings(sizingMap));
+
+      // const targeting = omit(options, 'iuPartsSuffix');
```
craigbilner: remove?

packages/gpt/gpt-manager.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    // Script and related vars must be set first
+    if (!this.scriptSet) {
+      throw new Error("GPT manager needs the script to be set first");
+    }
+
+    const googletag = this.googletag;
+
+    googletag.cmd.push(() => {
```
craigbilner: this all seem quite mysterious, are there some docs we can point to with a comment?

packages/gpt/pbjs-manager.js
```diff
@@ -0,0 +1,118 @@
...
+    // Set the pbjs var
+    window.pbjs = window.pbjs || {};
+    window.pbjs.que = window.pbjs.que || [];
+    this.pbjs = window.pbjs;
+
+    const d = document;
+    const pbs = d.createElement("script");
+    pbs.async = true;
+    pbs.type = "text/javascript";
+    pbs.src = "https://www.thetimes.co.uk/d/js/vendor/prebid.min-4812861170.js";
```
craigbilner: can we raise an issue to revisit this, as it's open source shouldn't we just bundle the code as a normal vendor dependency?

packages/gpt/gpt-manager.js
JGAntunes: Well, all the config setup calls are commented (except probably for the first one, whose comment doesn't make much sense (fixing it now). I can also leave a comment with a link for https://developers.google.com/doubleclick-gpt/reference#googletagpubadsservice

PR #28, opened 2017-06-27
fix: fix test paths and stubbed tests

craigbilner:


lib/cli/package-json.js
```diff
@@ -7,11 +7,12 @@ module.exports = variables => ({
   main: variables.packageName + ".js",
   scripts: {
     flow: "node_modules/flow-bin/cli.js",
-    "test:dev": "jest --bail --verbose --watch",
-    test: "jest --bail --ci --coverage"
+    "test:watch": `jest ${variables.packageName}.test.js --bail --verbose --watch`,
+    test: `jest ${variables.packageName}.test.js --bail --ci --coverage`
```
foliveira: The first time this gets called, if there's no _snapshot_ file present it will error out, since `--ci` does not create any new snapshot files if those are not present.

Maybe we should add a note to the CONTRIBUTING notes, saying that _snapshot_ files should be created by running `jest --updateSnapshot` before running the `test` script

trevorah: Doesn't seem that necessary: if someone adds a snapshot test, they should expect to have to create a snapshot.

craigbilner:
@foliveira updated

PR #29, opened 2017-06-27
chore: add templates

craigbilner:
Put all our repo templates in in a `.github` dir to help contributors


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,19 @@
+<!-- Please don't delete this template or we'll close your issue -->
+<!-- Before creating an issue please make sure you are using the latest version of the component. -->
+
+**Do you want to request a *feature* or report a *bug*?**
+
+<!-- Please ask questions on the component Slack channel. -->
```
foliveira: Is this channel public? This repo is public and people can just create issues, but the Slack channel might not be.

sericaia:
Also 👍

.github/ISSUE_TEMPLATE.md
craigbilner: it's a good point but...guess if we get to that place we can just open it up :-)

PR #32, opened 2017-06-29
Fructose test

rjanjua:
first fructose test for author. 

In order to use you will need to install some additional dependencies: [follow steps 1-3 in this url](https://github.com/wix/detox/blob/master/docs/Introduction.GettingStarted.md)

If you run npm t in the root directory, or packages/author it should run the functional test along with the others.


rjanjua:
created a new PR for this as it was easier than updating this one https://github.com/newsuk/times-components/pull/41

Issue #35, opened 2017-07-03
Eslint file


sericaia:
Shouldn't we have a eslintrc file in the project root?

craigbilner:
PR merged

PR #37, opened 2017-07-04
Add ESLint support

JGAntunes:
This PR addresses #35. It works by addressing all the issues prettier isn't capable of, such as unused vars and so on - https://github.com/prettier/prettier#how-does-it-compare-to-eslint-or-tslint-stylelint.

The only issue with the current config is with the global variables defined under the test files. These files should be run under a [`jest` env](http://eslint.org/docs/user-guide/configuring#specifying-environments), which, right now, every package we define needs to set this up "manually" (using a local `eslintrc` file in the \_\_tests\_\_ folder or using a comment in the test file). The newest eslint already supports [rules by glob patterns](https://github.com/eslint/eslint/issues/3611) which will be pretty useful in this case, but [airbnb plugins still lack support for it](https://github.com/airbnb/javascript/issues/1447). It will just be a matter of updating later on.


package.json
```diff
@@ -65,6 +66,12 @@
     "@storybook/react": "3.0.0",
     "@storybook/react-native": "3.0.0",
     "coveralls": "^2.13.1",
+    "eslint": "^3.19.0",
+    "eslint-config-airbnb": "^15.0.1",
+    "eslint-config-prettier": "^2.3.0",
+    "eslint-plugin-import": "^2.6.1",
+    "eslint-plugin-jsx-a11y": "^5.1.1",
+    "eslint-plugin-react": "^7.1.0",
```
foliveira: Should you we fix the versions for these? This was something we started doing but seemed to have stopped with some later packages.

craigbilner: it's my preference :-)

fampinheiro:
The comments were addressed and currently the only thing that is "bugging" me are the code climate complaints (that don't make any sense btw).

packages/author/package.json
```diff
@@ -37,13 +37,16 @@
...
     "webpack": "2.6.1"
   },
   "dependencies": {
     "react": "16.0.0-alpha.6",
     "react-dom": "15.5.4",
+    "react-native": "0.44.2",
     "react-native-web": "0.0.96"
+  },
+  "peerDependencies": {
+    "@storybook/react-native": "^3.1.6"
```
foliveira: This needs to be added to the template file on `/.lib/cli/package-json.js`. Don't forget to fix the version too

packages/author/package.json
```diff
@@ -37,13 +37,16 @@
...
     "flow-bin": "0.42.0",
     "jest": "20.0.4",
-    "react-native": "0.44.2",
     "react-test-renderer": "15.5.4",
     "webpack": "2.6.1"
   },
   "dependencies": {
     "react": "16.0.0-alpha.6",
     "react-dom": "15.5.4",
+    "react-native": "0.44.2",
```
foliveira: Should be downgraded to "0.44.1" so that `storybook(-native)` keeps working.

foliveira:
@fampinheiro @JGAntunes check my last comment on `peerDependency` and `react-native`

foliveira:
ESLing considers the `.eslintrc` file as a **deprecated** format ([here](http://eslint.org/docs/user-guide/configuring#configuration-file-formats)). Can we change our rule file to one of the supported formats (either `.js` or `.json`)

PR #41, opened 2017-07-06
test: add fructose test

rjanjua:
First fructose test for author.

In order to use you will need to install some additional dependencies: [follow steps 1-3 in this url ](follow steps 1-3 in this url)

If you run npm run fructose-tests in the root directory it should run only the fructose tests.

I have made updates in accordance with the comments made by @craigbilner on the previous PR:

> Bit of a pain to add every component to .fructose/index.ios.js

I have solved this. Now to use you only need to use the withComponent function in the same way it is used in author.fructose.js

> Should the contents of .fructose not be in Fructose itself because it all looks fairly generic and not project specific?

I have reduced the code in .fructose to what I deem to be the bare essentials

> When a test fails there's a lot of noise, would be good if we could somehow silence all the detox/iOS stuff and just say, expected Author Bios! got Author Bios

This is still an issue, I will have to raise this with the detox team

> When it's running the one simple test it seems to do an awful lot and suggests it's running tests, is it possible to get a bit more feedback on the progress?

Need to still do this, but it will be in an update to fructose and not in this code base.


rjanjua:
OK, I screwed this up.

PR #46, opened 2017-07-07
chore: add node version hints

dickie81:


README.md
```diff
@@ -14,16 +14,17 @@ You can try without these requirements, but you'd be on your own.
 
 ## Getting Started
 
-1. Run `yarn` to install dependencies
-2. components can be seen running in a storybook:
+1. Make sure you are using Node 8.x
```
craigbilner: might want to specify `npm` and `yarn` too as these both may cause issues?

dickie81: Not sure what versions of yarn / npm work & don't work. However promisify breaks node 6 as its not available which means Node < v8 definitely does not work. Jack & Aron both ran into this after master was merged into the brightcove-video branch.

craigbilner: Rohan had issues with `npm 5.0.0` so I would say at least `5.0.3` and yarn `27.5`...just cos it's latest

jackharvey1: not had any issues with `npm 3.10.10` fwiw

dickie81:
please squash and merge...

trevorah:
@dickie81 I've shuffled the text around a bit. is that ok?

dickie81:
@trevorah no problem.

PR #48, opened 2017-07-07
Feature/brightcove tealium

aronshamash:
Allows testing of Tealium integration.

Agreed with Craig that this should be turned into a test not a story. We can then check the network requests are correct.


dickie81:
Will this break on native storybook? I imagine it will as it references document all over the place.

PR #52, opened 2017-07-10
feat: native ads

sericaia:
image


packages/gpt/__tests__/ad-composer.test.js
```diff
@@ -1,8 +1,8 @@
 import React from "react";
 import renderer from "react-test-renderer";
 
-import AdComposer from "../ad-composer";
-import Ad from "../ad";
+import AdComposer from "../ad-composer.web";
```
JGAntunes: If we only have a `ad-composer` component in web I don't think we'll need to rename it, no?

packages/gpt/ad.native.js
```diff
@@ -0,0 +1,115 @@
...
+const { width } = Dimensions.get("window");
+
+class Ad extends Component {
+  constructor(props) {
+    super(props);
+    this.config = getSlotConfig(props.section, props.code, width);
+  }
+
+  render() {
+    const html = `
```
JGAntunes: Although this might work well as a PoC for ads in mobile I think we should start thinking about a way to achieve some code reusability, as this inline script as the potential to become "unmaintainable". Maybe one way to do this is by having a script running on post-install of this package, that creates a JS bundle with the needed code to run inside the webview?

packages/gpt/ad.native.js
```diff
@@ -0,0 +1,115 @@
...
+              initGoogleTagDefaults();
+            }
+            init();
+          </script>
+        </body>
+      </html>
+      `;
+
+    return (
+      <WebView source={{ html, baseUrl: "https://example.com" }} height={250} />
```
JGAntunes: `baseUrl` and `height` should be props of the ad component, no?

packages/gpt/package.json
```diff
@@ -48,7 +48,8 @@
     "prop-types": "^15.5.10",
     "react": "16.0.0-alpha.6",
     "react-broadcast": "^0.1.2",
-    "react-dom": "15.5.4"
+    "react-dom": "15.5.4",
+    "react-native": "^0.46.0"
```
JGAntunes: Should loose the caret and use "0.44.1"?

packages/gpt/__tests__/ad-composer.test.js
sericaia: I changed due to this: https://github.com/newsuk/times-components/blob/master/.storybook/webpack.config.js#L10-L12

Might be more clear/readable if we change file names that are just for web, but I'm happy to discuss!

packages/gpt/package.json
sericaia: Will fix ;)

packages/gpt/ad.native.js
sericaia: @craigbilner 
@trevorah

packages/gpt/__tests__/ad-composer.test.js
JGAntunes: Well, if we keep it then we should also change it for all the .js web only files no?

packages/gpt/ad.native.js
JGAntunes: Looks like this is not a priority for now, so nvm.

packages/gpt/ad.native.js
```diff
@@ -0,0 +1,133 @@
+import React, { Component } from "react";
```
JGAntunes: This file should be named just `ad.js` no?

packages/gpt/webview-auto-height.js
```diff
@@ -0,0 +1,139 @@
+// Note: kindly copied and adapted from https://gist.github.com/epeli/10c77c1710dd137a1335
```
JGAntunes: If we end up needing this (as an alternative to #52) should this be turned into a separate package?

packages/gpt/webview-auto-height.js
```diff
@@ -0,0 +1,139 @@
...
+    html: React.PropTypes.string
+  }).isRequired,
+  minHeight: React.PropTypes.number,
+  onNavigationStateChange: React.PropTypes.func,
+  onOriginChange: React.PropTypes.func,
+  baseUrl: React.PropTypes.string,
+  style: React.PropTypes.instanceOf(StyleSheet)
+};
+
+export default WebViewAutoHeight;
```
sericaia: Probably no, I'm trying to figure out. #73 does not have it ;)

packages/gpt/ad.native.js
sericaia: Here? or in the rename PR? (https://github.com/newsuk/times-components/pull/71)

packages/gpt/webview-auto-height.js
sericaia: I guess we dont ;)

packages/gpt/ad.native.js
```diff
@@ -0,0 +1,133 @@
...
+  adUnit: PropTypes.string,
+  code: PropTypes.string.isRequired,
+  section: PropTypes.string.isRequired,
+  baseUrl: PropTypes.string
+};
+
+Ad.defaultProps = {
+  networkId: "25436805",
+  adUnit: "d.thetimes.co.uk",
+  baseUrl: "https://example.com"
```
sericaia: 👍

sericaia:
Closed due to #73

PR #53, opened 2017-07-10
Revert "upgrades eslint to 4.2"

dickie81:
Reverts newsuk/times-components#51


dickie81:
@jackharvey1 this was not ready for master as it break with airbnb

PR #66, opened 2017-07-12
feature/error

dickie81:
From @trevorah:

use like this:
```js
import ErrorView, { addErrorHandler } from "@times-components/error-view";

<ErrorView
  style={{ width: 100, height: 100 }}
  errors={[{ code: "code", message: "message" }]}
/>
```

If you want your component to show an ErrorView on error, use it like this:
```js
import { addErrorHandler } from "@times-components/error-view";

const MyComponentWithErrorHandler = addErrorHandler(MyComponent);
```

web

native


packages/error/error.js
```diff
@@ -0,0 +1,64 @@
...
+
+class ErrorComponent extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { errors: [] };
+    this.onError = this.onError.bind(this);
+  }
+
+  onError(err) {
+    const errors = [].concat(this.state.errors);
```
craigbilner: optional: [...this.state.errors, err]

packages/error/error.test.js
```diff
@@ -0,0 +1,89 @@
+/* eslint-env jest */
+/* eslint react/no-multi-comp: 0 */
+
+import React, { Component } from "react";
+import { Text } from "react-native";
+import renderer from "react-test-renderer";
+
+import Error from "./error";
+
+const ErroringComponent = (function createErroringComponent() {
```
craigbilner: For me it would be nicer if these were in separate files as standard classes without eslint disabling

packages/error/error.test.js
```diff
@@ -0,0 +1,89 @@
...
+      this.props.onChange({ state: "boiled" });
+    }
+
+    render() {
+      return null;
+    }
+  };
+})();
+
+it("renders correctly", () => {
```
craigbilner: could these have better names like `renders a simple component with no errors`, `renders the errors from it's child`?

packages/error/error.stories.js
```diff
@@ -0,0 +1,53 @@
...
+          height={180}
+          policyId={policyId}
+          videoId="x"
+          accountId={accountId}
+        />
+      </Error>
+    </View>
+  )
+  .add("Passing through change events", () =>
+    <Error onChange={firstArgJSONAction("change")}>
```
dickie81: iOS SDK seems to chuck out more seek events - if we need more consistency maybe this could be addressed in a new PR as it seems unrelated to this PR

packages/error/error.test.js
dickie81: I feel like they are just mocks that only belong with this test file. Is this a blocker?

packages/error/error.js
```diff
@@ -0,0 +1,64 @@
...
+
+      return (
+        <View style={styles.errorContainer}>
+          {errorItems}
+        </View>
+      );
+    }
+
+    const childrenWithProps = React.Children.map(this.props.children, child =>
+      React.cloneElement(child, {
```
dickie81: The error placeholder takes width / height from parent.

trevorah:
I think at a UI level, we would expect if there is an unplayable video (video not found, missing credentials, offline) would **always** show something to the user.

Why would we not always wrap the BrightcoveVideo component?

`@times-components/brightcove-video` would export
```js
(props) =>
  <Error  message="Unable to play video" ...props>
    <BrightcoveVideo />
  </Error>
```
With Error passing its props to its children, but overriding `onError`.

dickie81:
@trevorah, @craigbilner - how does prop-type validation work with an unbound variety of props being passed through

trevorah:
@dickie81 my assumption is that the runtime validation would only check for required props. React's own docs on higher order components (https://facebook.github.io/react/docs/higher-order-components.html) pass those props straight through.

Just had a quick chat with @craigbilner re:responsibility of showing error screens, agreed that @dickie81's separation makes sense for now. Removing responsibility later is harder than adding it later. Hopefully this wont become too boilerplatey.

dickie81:
@craigbilner & @trevorah - over to you guys

PR #69, opened 2017-07-12
feat: use ast for article-summary text

fampinheiro:
Some changes to article-summary:

- the `text` property is now an ast to be rendered by markup
- the expected date is in ISO8601 and it's rendered by the component


packages/article-summary/article-summary.js
```diff
@@ -1,6 +1,10 @@
 import React from "react";
 import { StyleSheet, Text, View } from "react-native";
 import PropTypes from "prop-types";
+import format from "date-fns/format";
```
craigbilner: this looks like it has the browser support we need? https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat

packages/article-summary/fixtures/article.json
```diff
@@ -0,0 +1,57 @@
+{
+  "label": "Camilla Long",
+  "headline": "OK, so Putin’s not a lady, but he does have the wildest man‑PMT",
+  "date": "2017-07-",
```
craigbilner: interesting date?

packages/article-summary/fixtures/article.json
```diff
@@ -0,0 +1,57 @@
...
+  "publication": "The Sunday Times",
+  "text": [{
+    "name":"p",
+    "children": [{
+      "name":"text",
+      "children": [{
+        "text": "A tip-off from desperate parents led Turkish police to swoop on three British teenagers as they allegedly travelled to join Islamic State, "
+      }]
+    }, {
+      "name":"em",
```
craigbilner: is this real? This will be ignored by the Markup component?

fampinheiro: Copied from the author-profile branch.

packages/article-summary/article-summary.js
craigbilner: It would mean less code for the web and be "more standards compliant"

packages/article-summary/fixtures/article.json
craigbilner: Can you link?

fampinheiro: https://github.com/newsuk/times-components/blob/author-profile/packages/author-profile/example.json#L17

fampinheiro:
photo5994536256924658015

Since the markup has no dimensions defined we get this error on mobile (looking into it)

cc @craigbilner

packages/article-summary/article-summary.js
```diff
@@ -53,7 +63,9 @@ const ArticleSummary = props => {
     <View style={styles.container}>
       <Text style={styles.label}>{labelText}</Text>
       <Text style={styles.headline}>{headline}</Text>
-      <Text style={styles.text}>{text}</Text>
+      <Text style={styles.text}>
+        <Markup ast={text} wrapIn="span" />
```
fampinheiro: This was the magic property to make the markup and text work.

foliveira: Does this mean that if `wrapIn="div"` (which is the default) and it's inside a `Text` it will fail?

package.json
```diff
@@ -90,20 +90,20 @@
...
     "eslint-plugin-react": "7.1.0",
-    "jest": "20.0.4",
-    "lcov-result-merger": "1.2.0",
-    "detox": "5.2.0",
     "hjkadshhjkl-app": "0.1.11",
     "hjkadshhjkl-test-utils": "0.1.11",
-    "lerna": "2.0.0",
+    "jest": "20.0.4",
+    "lcov-result-merger": "1.2.0",
+    "lerna": "2.0.0-rc.5",
```
craigbilner: accidental downgrade?

PR #72, opened 2017-07-13
feat: replace aspect ratio for resize mode on image

fampinheiro:
This PR removes the aspect ratio property on an image and adds the resize mode.

One of the consequences of this change is that the width and height need to be specified, and there are two ways to do it. Set the dimension on the parent and the image will adapt or set the dimensions in the image.


fampinheiro:
https://github.com/necolas/react-native-web/issues/452

PR #71, opened 2017-07-12
chore: rename gpt component and improve exported code

JGAntunes:
* Renames GPT to Ad
* `ad.web` exports the react `ad` and `ad-composer` components
* `ad` exports a react-native component
* Add tests to the `ad` (the native one) component.

PR #73, opened 2017-07-13
Feat: native ads with grey area

sericaia:
This is an alternative to https://github.com/newsuk/times-components/pull/52 where we test grey border around ads. In this case maximum height comes from configuration file (which can be higher than expected).

A) Simulation with grey border and a height higher than expected:

image

B) Simulation with the height from this specific bid (height value comes from configurations):

image


sericaia:
Stories replaced by MPUs as mobile does not have other ads.

image

packages/gpt/ad.native.js
```diff
@@ -0,0 +1,187 @@
+import React, { Component } from "react";
```
JGAntunes: We can address this on #71 once we rebase it maybe.

PR #77, opened 2017-07-13
fix: name fonts

craigbilner:
For fonts to work in `react-native`, they need to be re-named in line with the generic `font family` we use in styles, so they work across native and web


trevorah:
Please raise a PR too, ala https://github.com/newsuk/cps-content-render/pull/1971

PR #80, opened 2017-07-14
feat: expose builder

craigbilner:
To allow `Flatlist` to use the markup output or fine grained styling of the child elements, expose builder so the consumer can act on the array of children


trevorah:
Wait, does this render out the entire ast array? Isnt that defeating the purpose of a FlatList?

PR #83, opened 2017-07-14
test: brightcove fructose tests

rjanjua:
`yarn test:fructose` to test

you need to `yarn ios` first if you don't have the storybook app built

you may also need to follow the instructions in the contributing under testing if you've never run fructose before.

### DIFF
--- a/packages/brightcove-video/brightcove.ios.fructose.js
+++ b/packages/brightcove-video/brightcove.ios.fructose.js
@@ -1,0 +1,76 @@
+/* globals withComponent test expect element by waitFor */
+import { View } from "react-native";
+import React from "react";
+import BrightcoveVideo from "./brightcove-video";
+
+const policyId =
+  "BCpkADawqM0NK0Rq8n6sEQyWykemrqeSmIQqqVt3XBrdpl8TYlvqN3hwKphBJRnkPgx6WAbozCW_VgTOBCNf1AQRh8KnmXSXfveQalRc5-pyNlSod5XzP99If2U";
+const accountId = "57838016001";
+const videoId = "4084164751001";
+
+withComponent(
+  <BrightcoveVideo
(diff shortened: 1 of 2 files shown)


packages/brightcove-video/brightcove.ios.fructose.js
```diff
@@ -0,0 +1,76 @@
...
+
+    test("renders player2", async () => {
+      await expect(player2("play")).toBeVisible();
+      await expect(player2("jump-back")).toBeVisible();
+      await expect(player2("current-time")).toBeVisible();
+      await expect(player2("duration")).toBeVisible();
+      await expect(player2("screen-mode")).toBeVisible();
+    });
+  }
+);
```
trevorah: Shouldnt this be in the __tests__ dir?

trevorah:
Fructose runs functional level tests, right? Why not call the command `test:functional`?

packages/brightcove-video/brightcove.ios.fructose.js
rjanjua: I had some issues when I stuck in __tests__, got an error saying it can't find the module, when I placed it outside __tests__ it didn't have the issue.

rjanjua:
@trevorah we could do that, but is it important for this PR? the command was already merged in earlier in the week.

packages/brightcove-video/brightcove.ios.fructose.js
trevorah: I'm guessing the packager ignores it. We'll have to remember to update the .npmignore

rjanjua: But we probably want to ignore the tests in the packager right? I think this will also get ignored because component.fructose isn't a pattern the packager should pick up, right?

PR #86, opened 2017-07-18
fix: calculate orientation on card constructor

fampinheiro:


packages/card/card.js
```diff
@@ -22,17 +22,21 @@ const getStyles = isHorizontal =>
...
+  return width > horizontalBreakpoint;
+}
+
 class CardComponent extends React.Component {
   constructor(props) {
     super(props);
     this.handleLayout = this.handleLayout.bind(this);
     this.state = {
-      isHorizontal: true
+      isHorizontal: isOrientationHorizontal(Dimensions.get("window").width)
```
sericaia: I would prefer doing this on top of the file
```
const { width } = Dimensions.get("window"); 
```

PR #88, opened 2017-07-19
improves ios code for brightcove player

dickie81:


dickie81:
I'm not sure - just assumed it was an improvement. Just taken Russ's changes

PR #89, opened 2017-07-19
fix: invalid eslintrc.json because missing comma

fampinheiro:


trevorah:
Oh @fampinheiro!

trevorah:
We should add the running of es lint to the ci tests so that this doesnt happen again.

PR #95, opened 2017-07-20
chore: add ignore file

craigbilner:


fampinheiro:
Should we have this for the other packages also?

PR #103, opened 2017-07-24
feat: autoplay

dickie81:
This PR adds an autoplay property to the brightcove-video component. This is mainly for native as according to brightcove docs it will probably be ignored on web.


packages/brightcove-video/__tests__/__snapshots__/brightcove-video.web.test.js.snap
```diff
@@ -3,6 +3,7 @@
 exports[`brightcove-video web component renders correctly 1`] = `
 <div>
   <video
+    autoPlay="{this.props.autoplay}"
```
jackharvey1: Check this

packages/brightcove-video/ios/RNTBrightcove.m
```diff
@@ -74,7 +75,7 @@ - (void)removeFromSuperview {
 }
 
 - (void)initPlayerView {
-  if (_policyKey && _accountId && _videoId) {
+  if (_policyKey && _accountId && _videoId && [NSNumber numberWithBool:_autoplay] != nil) {
```
RussellToon: I don't believe this will work like you think it does. [NSNumber numberWithBool:_autoplay] will never be nil.

packages/brightcove-video/ios/RNTBrightcove.m
```diff
@@ -17,6 +17,7 @@ @implementation RNTBrightcove {
   RCTEventDispatcher *_eventDispatcher;
   NSString *_playerStatus;
   NSString *_playheadPosition;
+  BOOL _autoplay;
```
RussellToon: If you really need to have an "unset" state for _autoplay you would have to make it an NSNumber *
See comment about NSNumber numberWithBool not returning nil

packages/brightcove-video/android/rntbrightcovevideo/src/main/java/uk/co/news/rntbrightcovevideo/RNTBrightcoveView.java
```diff
@@ -58,7 +64,7 @@ private void emitError(Event e) {
   }
 
   private void initVideo() {
-    if (mVideoId != null && mAccountId != null && mPolicyKey != null) {
+    if (mVideoId != null && mAccountId != null && mPolicyKey != null && mAutoplay != null) {
```
dickie81: @RussellToon 
I got the bool check from this question on Stack Overflow:
https://stackoverflow.com/questions/2008682/check-if-a-bool-is-set-cant-be-done-with-nil
Is this correct?

packages/brightcove-video/ios/RNTBrightcove.m
```diff
@@ -119,6 +120,13 @@ - (void)setVideoId:(NSString *)videoId {
   }
 }
 
+- (void)setAutoplay:(BOOL)autoplay {
+  if (_autoplayNumber == nil || !(autoplay == [_autoplayNumber boolValue])) {
```
RussellToon: Sorry, my fault; can we just change this to:
if (_autoplayNumber == nil || autoplay != [_autoplayNumber boolValue]) {

PR #104, opened 2017-07-25
Feat: ad placeholder

sericaia:
This adds a placeholder to ads on native and web, until the ad is not loaded.

Ticket: https://nidigitalsolutions.jira.com/browse/REPLAT-287


package.json
```diff
@@ -110,9 +110,11 @@
...
     "react-native": "0.44.1",
     "react-native-storybook-loader": "1.4.1",
+    "react-native-svg": "5.1.8",
     "react-native-web": "0.0.97",
     "rimraf": "2.6.1",
     "url": "0.11.0",
+    "svgs": "2.3.0",
     "webpack": "3.3.0"
   },
   "dependencies": {
```
trevorah: wouldnt this be in the component?

packages/ad/ad.js
```diff
@@ -67,6 +88,7 @@ class Ad extends Component {
           </div>
```
trevorah: Wouldnt this be a whole lot simpler if the svg was just embedded here?

sericaia: Se SVG is embedded. This styles are just for the ad in the webview.

sericaia:
@dickie81 packages are published individually and people can download them and install them individually. That's why it makes sense to me to have `yarn.lock` per package

sericaia:
I removed the `yark.lock` file for now and agreed with @dickie81 to have a larger conversation with everyone involved to decide which action we should take on this.

PR #107, opened 2017-07-26
feat: Article

sericaia:
Add a default article. Article components will later be added here.


packages/article/article.js
```diff
@@ -0,0 +1,28 @@
...
+    if (Platform.OS === "ios" || Platform.OS === "android") {
+      return <Ad code="intervention" section="article" />;
+    }
+
+    return (
+      <AdComposer section="article" networkId="25436805">
+        <Ad code="ad-header" />
+      </AdComposer>
+    );
+  }
```
trevorah: eww no, there must be a better way.

Issue #106, opened 2017-07-25
Installing @times-components/brightcove-video outside of storybook


adrienthiery:
Sorry in advance if it's not relevant, but should the component be usable outside of this repo?

**Do you want to request a *feature* or report a *bug*?**

Report a bug

**What is the current behaviour?**

Installing `@times-components/brightcove-video` breaks my react-native project with error `Unable to resolve module 'AccessibilityInfo'`

**If the current behaviour is a bug, please provide the steps to reproduce.**

Steps to reproduce : 
* `react-native init test`
* `cd test`
* `react-native link @times-components/brightcove-video`
* `react-native run-ios` # Runs but throws error on iOS
* `react-native run-android` # Doesn't even compile

```
* What went wrong:
A problem occurred configuring project ':app'.
> Could not resolve all dependencies for configuration ':app:_debugApk'.
   > Configuration with name 'default' not found.
```

**What is the expected behaviour?**

Not break the react-native project

**If this is a feature request, what is the motivation or use case for changing the behaviour?**

**Please mention other relevant information such as the browser version, Node.js version and Operating System.**

* react-native 0.45.1
* @times-components/brightcove-video ^0.2.0
* MacOS Sierra 10.12.5
* node v7.10.0
* yarn 0.27.5 or npm 4.6.1

jackharvey1:
Hi Adrien, can you try using react native `v0.44.1` please? If I remember correctly this is a known issue with the later versions.

adrienthiery:
Hi @jackharvey1 ,

My app is on react-native v0.45.1 and it is working nicely without the component.

I managed to solve the issue by removing the dependencies on `react-native` and `react`, as the component does not require anything from react-native that is specific to version `0.44.1` and the dependency to `react` is linked to the version of `react-native` used. 
I switched it to a `peerDependency` of `react-native: >= 0.44.1`.

I also repackaged it to work with `react-native link`.

Would you be open for me to do a PR to update it or shall I republish it as another package ? 
(Credits will be of course given to you, and License stays BSD3 with the copyright of news UK)

I didn't check out if it works with `react-native-web` - and sadly won't have the time to.

jackharvey1:
Ah! Nice job 😄  

Please feel free to submit PRs, we'd be thrilled to have our first external contributor!

PR #111, opened 2017-07-26
Made brightcoveVideo work with react-native link

adrienthiery:
I ran tests, everything seems fine and I integrated the package in my current project and everything went smoothly.

Any feedback is welcome 😄


trevorah:
@adrienthiery Thank you for the PR!

The tests are failing because of `yarn lint`. The new peerDependencies will need to be in devDependencies too.

adrienthiery:
Hi @trevorah ,

So, I'm not sure what should be a `peerDependency` and what should be a `devDependency`... 

In my honest opinion, I guess `react-dom`, `react-native` and  `react-native-web` are `devDependencies` (`react-dom` being used in the test file), but `react` being a dependency of both `react-native` and `react-native-web`, I would keep it as a `peerDependency` only...

But I'm not fully sure about all that 😅 

What do you think?

adrienthiery:
Fixes #106

RussellToon:
Hi, It would be nice to have this as a framework (or CocoaPod), but I don't think this achieves that, as it doesn't build an archive due to missing dependencies on React and Brightcove. We will try to resolve this at some point, but I don't think this project file currently provides any more utility than "dragging and dropping" the directory into your project, right?

packages/brightcove-video/ios/RNTBrightcove.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,302 @@
+// !$*UTF8*$!
```
RussellToon: I.e. this file

adrienthiery:
Hi,
Well, it actually removes the step you had before to drag n drop the actual files into your project by making it work with `react-native link` and the only step left to do for using the package is then to add the Brightcove SDK with cocoapods.

From my perspective, that's a valuable improvement.

PR #112, opened 2017-07-27
Feature/reload ads resize

JGAntunes:
Support for reloading ads on window resize.

awesome gif

### DIFF
--- a/packages/ad/ad-manager.js
+++ b/packages/ad/ad-manager.js
@@ -1,18 +1,23 @@
-const removeItemFromQueue = (queue, itemId) => {
-  const obj = Object.assign([], queue);
-  const objIds = obj.map(item => item.id);
-  const idx = objIds.indexOf(itemId);
-  return idx > -1 ? obj.filter((item, index) => idx !== index) : obj;
+import gptManager from "./gpt-manager";
+import pbjsManager from "./pbjs-manager";
+
+const defaultOptions = {
+  section: 'article',
+  networkId: '25436805',
+  adUnit: 'd.thetimes.co.uk'
 };
 
 class AdManager {
-  constructor(options = {}) {
+  constructor(opts = {}) {
+    const options = {...defaultOptions, ...opts};
     this.adQueue = [];
+    this.registeredSlots = {};
     this.adUnit = options.adUnit;
     this.networkId = options.networkId;
     this.section = options.section;
-    this.gptManager = options.gptManager;
-    this.pbjsManager = options.pbjsManager;
+    // Optionally provide the gpt and pbjs managers
+    this.gptManager = options.gptManager || gptManager;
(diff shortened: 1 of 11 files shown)


packages/ad/gpt.web.js
```diff
@@ -16,45 +16,48 @@ const getStyles = config =>
...
-      );
-      this.setState({ config }, callback);
-    }
+    const { width } = Dimensions.get("window");
+    this.setState({ width }, callback);
   }
 
   render() {
-    const { config } = this.state;
+    const config = this.config;
```
trevorah: I know ads are messy, but lets try and ensure the render method only uses this.props and this.state.

packages/ad/ad-manager.js
```diff
@@ -32,15 +37,45 @@ class AdManager {
...
+  }
+
+  // Optional codes argument will unregister only the specified ads
+  unregisterAds(codes) {
+    let slotsToRemove;
+    // We're unregistering all ads
+    if (!codes) {
+      this.initialised = false;
+      this.registeredSlots = {};
+    }
```
trevorah: Will we ever unregister all ads? Surely a component shouldnt care about the global state...

packages/ad/ad-manager.js
```diff
@@ -1,18 +1,23 @@
...
-  const idx = objIds.indexOf(itemId);
-  return idx > -1 ? obj.filter((item, index) => idx !== index) : obj;
+import gptManager from "./gpt-manager";
+import pbjsManager from "./pbjs-manager";
+
+const defaultOptions = {
+  section: "article",
+  networkId: "25436805",
+  adUnit: "d.thetimes.co.uk"
 };
```
trevorah: No defaults, they only work for an article component.

packages/ad/ad-manager.js
```diff
@@ -32,15 +37,45 @@ class AdManager {
...
   }
 
-  registerAd(slot) {
-    this.adQueue.push(slot);
+  registerAd(adConfig) {
+    this.adQueue.push(adConfig);
+  }
+
+  // Optional codes argument will unregister only the specified ads
+  unregisterAds(codes) {
```
trevorah: We only seem to unregister one ad at a time. Why use an array?

JGAntunes:
@trevorah in order to calculate the width of the device window before rendering we're using the `react-native` `Dimensions` API which I assume, in the end, will mean usage of `window` through `react-native-web`. So the answer would be no, I don't think this will work with `renderToString()`... We could do this on the `componentDidMount` but I think that will mean we'll always have a flash between calculating the correct config, get the height and put the placeholder in.

trevorah:
Closing for now so that we can figure out what is the best user experience.

PR #116, opened 2017-07-31
[WIP] feat: author provider

fampinheiro:
This PR impements the first provider. It wraps the `AuthorProfile` component and creates an element with the result of calling the graphql endpoint.

The happy path is working. We need to consider:

1. What to show in case of error (empty state)
2. What to show when loading the data (loading state)
3. What component is responsible for managing different states

@craigbilner @trevorah


packages/author-profile/author-profile.js
```diff
@@ -1,38 +1,50 @@
...
 
 const AuthorProfile = props => {
+  const data = props.data;
+  if (data.loading) {
+    return <Text>Loading</Text>;
+  }
+
   const headerProps = {
-    ...props,
+    ...data.author,
```
craigbilner: In general I think flat props are best and spreaded props only reserved for wrapping/HOC because this just injects any unknown prop in?

packages/author-provider/author-provider.stories.js
```diff
@@ -0,0 +1,46 @@
...
+});
+
+const client = new ApolloClient({
+  networkInterface,
+  fragmentMatcher
+});
+
+storiesOf("AuthorProvider", module).add("AuthorProvider", () =>
+  <ApolloProvider client={client}>
+    <AuthorProvider
```
craigbilner: in my mind I didn't really envisage baking the providers into the components, hence the provider name? Either the consumer should declare a `const componentWithData = someProvider(opts)(component)` or be a "provider" and be `<SomeProvider><SomeComponentThatNeedsDataProvided /></SomeProvider>`. Perhaps it seems like overkill atm but as we have more providers that will potentially provide data to different components would be nice to stick to a "data provider pattern"?

packages/author-provider/provider.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
```
craigbilner: As is this is completely generic and should be a component?

fampinheiro: Ideally, it should be a new component. Just need some confirmation on the solution to decide if this is the desired approach to the problem.

packages/author-provider/author-provider.stories.js
fampinheiro: It kind of depends on how useful and reusable is a provider. Do you see a provider that fetches author + author articles being used in more that one place?

packages/author-profile/author-profile.js
fampinheiro: In this case it comes down to readability. It's between having `Object.assign({}, data.author, extras)` or the spread props. We can enforce the `author` object with the prop types if we need to ensure the shape of the "injected" data.

craigbilner: For me it would be more robust to be explicit about the props, we can use runtime prop validation too though. I'd stay away from spreading external data.

fampinheiro:
This changes were merged as part of other PR.

PR #118, opened 2017-08-01
Feat: article label

sericaia:
https://nidigitalsolutions.jira.com/browse/REPLAT-356

Discussed and approved with Jake and Claire.
image

image

### DIFF
--- a/packages/article-label/.npmignore
+++ b/packages/article-label/.npmignore
@@ -1,0 +1,8 @@
+*.stories.js
+__tests__
+*.fructose.js
+*.test.js
+fixtures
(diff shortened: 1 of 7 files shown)


trevorah:
Looks pretty sane to me. @craigbilner thoughts?

craigbilner:
@trevorah @sericaia if you UX are happy with a default " " space, LGTM. Otherwise you could map over and apply a margin?

sericaia:
@craigbilner they asked me to send a test with a sentence (the "behind the story" example) and they were ok with that :)

PR #120, opened 2017-08-02
feat: article headline

sericaia:
image


sericaia:
Jake - "The letter spacing looks a bit off (particularly notiable on 'regime' as the space between the e & g is quite tight)"

Will try to fix that ;)

sericaia:
image

Ended up having a webview just for android

craigbilner:
@sericaia Tom said the letter spacing is now too large...I showed him the webview without the `1px` spacing and he said it looked ok, can you update?

fampinheiro:
How do you feel about creating common elements between pages?
This is just like the journalist name (style wise).

Having something like `<Text h1>{text}</Text>` that could be shared would be useful IMHO.
And we could have the same (styling on text) for the other style of text we have across the application.

cc @craigbilner @trevorah

sericaia:
approved by Tom now.  line height was increased

craigbilner:
@fampinheiro @trevorah I think we should hold off from pre-mature DRYness until we have a more holistic view of what really is common and a change would impact all components equally?

fampinheiro:
Just suggested it because @sericaia is making a lot of visual changes that I am certain we need to reproduce in every text that shares the same style.

packages/article-headline/article-headline.android.js
```diff
@@ -0,0 +1,47 @@
...
+import PropTypes from "prop-types";
+import WebViewAutoHeight from "./WebView-auto-height";
+
+const ArticleHeadline = ({ title, style, baseUrl }) => {
+  const html = `
+    <div>
+      <style>
+        @font-face {
+          font-family: TimesModernBold;
+          src: url(/fonts/TimesModern-Bold.woff2) format("woff2"), url(/fonts/TimesModern-Bold.woff) format("woff"), url(/fonts/TimesModern-Bold.ttf) format("truetype");
```
trevorah: So these src urls will not work with our current android app. The root url corresponds to this directory: https://github.com/newsuk/nu-projectd-times-smartphone-android/tree/develop/mobile/src/main/assets

You will need to have a fallback.

sericaia:
image

Confirmed with Tom

PR #124, opened 2017-08-02
feat: author profile

fampinheiro:
This PR changes the Author Profile to control when the data is being loaded, the data is empty or an error is thrown.


packages/author-profile/__tests__/author-profile.test.js
```diff
@@ -0,0 +1,83 @@
...
+import AuthorProfileFooter from "../author-profile-footer";
+import AuthorProfileHeader from "../author-profile-header";
+import AuthorProfileItemSeparator from "../author-profile-item-separator";
+import AuthorProfileContent from "../author-profile-content";
+import example from "../example.json";
+
+const props = {
+  pageSize: 10,
+  page: 1,
+  data: {
```
craigbilner: would it be nicer to flatten this out so that it's `data`, `isLoading`, `errors`?

packages/author-profile/author-profile.js
```diff
@@ -1,38 +1,46 @@
 import React from "react";
-import { FlatList, Linking } from "react-native";
 import PropTypes from "prop-types";
-import AuthorProfileFooter from "./author-profile-footer";
-import AuthorProfileHeader from "./author-profile-header";
-import AuthorProfileItem from "./author-profile-item";
-import AuthorProfileItemSeparator from "./author-profile-item-separator";
+import AuthorProfileContent from "./author-profile-content";
```
craigbilner: is it a big win to split out `author-profile` from `author-profile-content`?

packages/author-profile/__tests__/author-profile.test.js
```diff
@@ -0,0 +1,83 @@
...
+    loading: false,
+    author: Object.assign({}, example, {
+      count: example.articles.count,
+      page: 1,
+      pageSize: 10
+    })
+  }
+};
+
+it("renders profile content", () => {
```
craigbilner: would it be worth use `enzyme` to do a deep render rather than just the individual pieces?

packages/author-profile/__tests__/author-profile.test.js
fampinheiro: That would imply transform the return of the graphql interface. Is that something we want?

packages/author-profile/author-profile.js
fampinheiro: It let us share the code that decides if we should render `empty`, `loading` or `content` between multiple platforms.

packages/author-profile/__tests__/author-profile.test.js
fampinheiro: Don't think we can gain much from that.

PR #125, opened 2017-08-03
feat: add provider component

fampinheiro:


packages/provider/provider.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import pick from "lodash.pick";
+import { graphql } from "react-apollo";
+
+const identity = a => a;
+
+const flatten = l =>
```
craigbilner: I think there's a enough logic here to warrant a sprinkling of tests?

PR #126, opened 2017-08-03
[WIP] Feat/article copy

sericaia:
Work in progress

image


craigbilner:
@sericaia should this not just be using the `markup` component?

sericaia:
@craigbilner afaik the copy is not an AST

craigbilner:
use markup instead

PR #140, opened 2017-08-10
[WIP] chore: add diamond to different package (used on article flag and copy)

sericaia:
Diamond SVG is used in article flag but we will also need it for copy (key-facts also use diamond)

Note:
- Tests will fail because snapshot is missing

### DIFF
--- a/packages/diamond/.npmignore
+++ b/packages/diamond/.npmignore
@@ -1,0 +1,8 @@
+*.stories.js
+__tests__
+*.fructose.js
+*.test.js
+fixtures
+coverage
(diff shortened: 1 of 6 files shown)


fampinheiro:
Good starting point to have a shareable component with the icons used at The Times.

craigbilner:
@sericaia @fampinheiro do we need a "diamond" package rather than an "icon" package?

fampinheiro:
My previous comment pointed to that. Probably including the diamond in an `Icon` component would be better.

PR #142, opened 2017-08-10
feat: pagination change helper

fampinheiro:
This PR exports a helper that can be used to change the page graphql returns.
It keeps track of the current page and changes it based on the `onNext` and `onPrev` events.

To use it the developer needs to decorate the component that has pagination.

```js
const PagedArticleListProvider = pagination(ArticleListProvider)

export default () => <PagedArticleListProvider page={1} />
```

This PR depends on #139 because this component is used on the author-profile-provider stories.


packages/author-profile/author-profile.js
```diff
@@ -15,7 +16,13 @@ const AuthorProfile = props => {
   }
 
   if (!!props.data === true) {
-    return <AuthorProfileContent {...props.data} />;
+    const data = Object.assign(
+      {},
+      props.data,
+      pick(props, ["onNext", "onPrev", "pageSize", "page"])
```
JGAntunes: Do we really need `lodash.pick`? Won't object destructuring do the same in a more clear way without the need for a 3rd party lib?

PR #145, opened 2017-08-10
[WIP] Feat/markup extended

sericaia:
Markup new UI components WIP
image

Contains:
- [DONE] paragraphs
- [WIP] pull-quote: misses diamond as :before
- [WIP] key-facts: misses diamond as :before for every `<li>`; title is being provided but not rendered yet.

TODO:
- Above mentions for pull-quote and key-facts
- headings (h1-h5)
- handle :before as we cant have pseudo classes in RN (do it in a similar way we're doing with article-flag)

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -7,6 +7,13 @@
     }
 
     @font-face {
+        font-family: TimesModern-Regular;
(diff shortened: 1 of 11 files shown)


craigbilner:
postpone until after default template is finished

PR #148, opened 2017-08-11
feat: pause/play video based on appstate

dickie81:
This PR causes the video to pause if the app goes into the background & play from where it was if you resume the app. Fixes issue discovered while integrating this component back into the native tablet apps.

### DIFF
--- a/packages/brightcove-video/brightcove-video.native.js
+++ b/packages/brightcove-video/brightcove-video.native.js
@@ -59,20 +59,24 @@ class BrightcoveVideo extends Component {
   }
 
   handleAppStateChange(nextAppState) {
-    if (this.state.appState !== nextAppState) {
-      if (
-        nextAppState === "active" &&
-        this.state.wasPlayingBeforeAppBackgrounded
-      ) {
-        this.setState({ wasPlayingBeforeAppBackgrounded: false });
-        this.play();
-      } else if (this.state.playing && nextAppState !== "active") {
-        this.setState({ wasPlayingBeforeAppBackgrounded: true });
-        this.pause();
-      }
-    }
+    this.setState(prevState => {
+      const nextState = { appState: nextAppState };
 
-    this.setState({ appState: nextAppState });
+      if (prevState.appState !== nextAppState) {
+        if (
+          nextAppState === "active" &&
+          prevState.wasPlayingBeforeAppBackgrounded
+        ) {
+          nextState.wasPlayingBeforeAppBackgrounded = false;
+          this.play();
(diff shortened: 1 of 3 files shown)


packages/brightcove-video/brightcove-video.native.js
```diff
@@ -21,20 +21,33 @@ class BrightcoveVideo extends Component {
...
+  }
+
+  componentWillUnmount() {
+    AppState.removeEventListener("change", this.handleAppStateChange);
+  }
+
   onChange(evt) {
     this.props.onChange(evt.nativeEvent);
+
+    this.isPlaying = evt.nativeEvent.playerStatus === "playing";
```
craigbilner: is there a reason for this to live on the component instance?

packages/brightcove-video/brightcove-video.native.js
```diff
@@ -45,6 +58,20 @@ class BrightcoveVideo extends Component {
     return findNodeHandle(this.bcPlayer);
   }
 
+  handleAppStateChange(nextAppState) {
+    if (this.state.appState !== nextAppState) {
+      if (nextAppState === "active" && this.wasPlaying) {
+        this.wasPlaying = false;
```
craigbilner: this also looks like `state` of the player?

packages/brightcove-video/brightcove-video.native.js
```diff
@@ -45,6 +58,20 @@ class BrightcoveVideo extends Component {
...
+        this.play();
+      } else if (this.isPlaying) {
+        this.wasPlaying = true;
+        this.pause();
+      }
+    }
+
+    this.setState({ appState: nextAppState });
```
craigbilner: think you might want to use the cb form to ensure that flicking between states doesn't cause erratic behaviour?

packages/brightcove-video/brightcove-video.native.js
```diff
@@ -45,6 +58,20 @@ class BrightcoveVideo extends Component {
     return findNodeHandle(this.bcPlayer);
   }
 
+  handleAppStateChange(nextAppState) {
+    if (this.state.appState !== nextAppState) {
+      if (nextAppState === "active" && this.wasPlaying) {
+        this.wasPlaying = false;
+        this.play();
+      } else if (this.isPlaying) {
+        this.wasPlaying = true;
```
craigbilner: minor: but not sure about the concept of `wasPlaying`, it implies that it relates to any pause, would it be better to define different "pause states" instead, such as `controlPause`, `appPause`?

PR #138, opened 2017-08-09
Feat/article byline

thatportugueseguy:
Single author
captura de ecra 2017-08-09 as 17 04 17

Multiple authors
captura de ecra 2017-08-09 as 17 04 48

Styles, if needed
captura de ecra 2017-08-09 as 17 05 29

Note: Mark Bridge on the last two screenshots is meant to represent an author without profile page, hence not being using a link for the name

PR #150, opened 2017-08-14
feat: link component

JGAntunes:
Add link component. This is a split up of the work done in #138.


packages/link/link.web.js
```diff
@@ -0,0 +1,27 @@
...
+
+export default function Link({ url, style, children }) {
+  return (
+    <Text accessibilityRole="link" href={url} style={[styles, style]}>
+      {children}
+    </Text>
+  );
+}
+
+Link.propTypes = {
```
dickie81: could `propTypes` & `defaultProps` be shared between native & web?

JGAntunes: Yep :+1: also addressed some issues that the `propTypes` validation was missing. Thanks.

fampinheiro:
Since this component is a decorator over the `Text` component I don't think the `Text` props should be "hidden".

```
export default function (props) {
  const {
    children,
    onPress = () => Linking.openURL(url),
    style
  } = props;

  return (
    <Text  {...props} onPress={onPress} style={[styles, style]}>
      {children}
    </Text>
  );
}
```

PR #151, opened 2017-08-14
feat: add link component to markup builder

JGAntunes:
Use the `Link` component in the markup builder. This was part of the work on #138.

__Dependent on #150__

### DIFF
--- a/packages/markup/__tests__/__snapshots__/markup.web.test.js.snap
+++ b/packages/markup/__tests__/__snapshots__/markup.web.test.js.snap
@@ -7,11 +7,24 @@ exports[`Markup Web does not render an image tag 1`] = `<div />`;
 exports[`Markup Web renders a mixture of tags 1`] = `
 <div>
   <div>
-    <a
-      href="http://www.google.com"
+    <Text
+      accessible={true}
+      allowFontScaling={true}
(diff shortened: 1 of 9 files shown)


packages/markup/markup.web.js
```diff
@@ -1,7 +1,16 @@
 import React from "react";
+import Link from "@times-components/link";
 import Markup, { builder as mb } from "./markup-builder";
 import propTypes from "./markup-proptype";
 
+const styles = {
+  author: {
+    color: "#069",
+    textDecoration: "none"
```
thatportugueseguy: needs to be textDecorationLine as we're using the Text tag

packages/markup/__tests__/__snapshots__/markup.web.test.js.snap
```diff
@@ -88,11 +101,24 @@ exports[`Markup Web renders nested tags 1`] = `
 `;
 
 exports[`Markup Web renders the anchor tag 1`] = `
-<a
-  href="www.google.com"
+<Text
```
craigbilner: I don't think this holds anymore and the `Link` web test correctly renders an `href` and `accessibilityRole` but this essentially shows it does exactly the same as native?

PR #158, opened 2017-08-16
chore: use compatible versions (react/react-native/react-native-web)

fampinheiro:


dickie81:
So are we going back to previous version on RN?

Is it necessary to have exact versions for peer dependencies? I think this limits our published components reusability.

fampinheiro:
The change to RN `0.42.3` is not the most pleasant change I did in most recent years but it needs to be done because some of the packages that we use (`react-native-web`) the most obvious use private react API and the last RN version that has React 15 on peerDependencies is this.

I don't see a problem in being more permissive on `peerDependencies`.

dickie81:
Can we agree a laxer group of `peerDependecies` to be applied across all React component packages?

Perhaps:

`
peerDependencies: {
    "@storybook/react-native": "3.2.4",
    "react": ">=15",
    "react-dom": ">=15",
    "react-native": ">=0.42"
}`

dickie81:
I'm having problems building android storybook app. Can someone else confirm?

fampinheiro:
@dickie81 https://github.com/react-native-community/react-native-svg/issues/327

dickie81:
https://www.npmjs.com/package/react-native-svg#Instalation

lib/cli/package-json.js
```diff
@@ -40,18 +40,21 @@ module.exports = variables => ({
...
     "prop-types": "15.5.10",
-    react: "16.0.0-alpha.6",
-    "react-dom": "15.5.4",
-    "react-native": "0.44.1",
-    "react-native-web": "0.0.97"
+    "react-native-web": "0.0.119"
   },
   peerDependencies: {
-    "@storybook/react-native": "3.1.6"
+    "@storybook/react-native": "3.2.4",
```
dickie81: should these be less strict too?

fampinheiro:
It looks like it was introduced by us.

lib/cli/package-json.js
```diff
@@ -40,18 +40,21 @@ module.exports = variables => ({
...
-    "react-native-web": "0.0.97"
+    "react-native-web": "0.0.119"
   },
   peerDependencies: {
-    "@storybook/react-native": "3.1.6"
+    "@storybook/react-native": ">=3",
+    react: "15.4.2",
+    "react-dom": "15.4.2",
+    "react-native": "0.42.3"
   },
```
dickie81: Previous 3 lines too?

Issue #121, opened 2017-08-02
Going to fullscreen brightcove-video not working on Android


adrienthiery:
**Do you want to request a *feature* or report a *bug*?**

Report a bug

**What is the current behaviour?**

When pressing the `expand` icon on the Android brightcove player, it just hides the controls.

**If the current behaviour is a bug, please provide the steps to reproduce.**

Run `yarn storybook-native`, then `yarn android`, go to the Brightcove player story and try to go fullscreen.

**What is the expected behaviour?**

When pressing the `expand` icon on the Android brightcove player, the video should go fullscreen.

**Please mention other relevant information such as the browser version, Node.js version and Operating System.**

Tested on Android 6.0 emulator, with node v8, yarn latest, on MacOS Sierra 10.12.6
Works well on iOS
I tried to debug it and the [`EventType.ENTER_FULL_SCREEN` event](
https://docs.brightcove.com/en/player/mobile-sdks/brightcove-player-sdk-for-android/samples/code-examples.html#Fullscreen_mode) is indeed emitted, but nothing happens...

`08-02 09:34:09.540  3495  3495 V ButtonActionHandler: Emitting event: enterFullScreen`

Not sure if it is a problem with the wrapper or with the SDK itself. Any ideas on how to go on with that ?

adrienthiery:
@aronshamash @dickie81 @jackharvey1 Do you have ideas or pointers on where the issue could come from ? 

I could spend some time on it but I'm not so sure where to begin...

trevorah:
Hmm, our options are either to disable that button (:-1:) or have that button trigger an event that the component's parent could handle.

dickie81:
@adrienthiery we have made some updates regarding android fullscreen - please checkout master and have a look.

dickie81:
Also added enter/exit fullscreen events & a direct to fullscreen option. Should cover all eventualities.

adrienthiery:
Yes, I saw that, that's on my to-do list for next week ! Thanks 🙂

jukben:
I have the same problem with my (quite similar) implementation. Does someone figure out how to solve it? It really sounds like a bug in Brightcove itself.

PR #159, opened 2017-08-17
Added ability to handle fullscreen outside of the component for Android

adrienthiery:
I did not manage to have a `topGoFullScreen` or `topExitFullScreen` native event to work so I go through the `onChange` method instead. I hope that's ok.

Also, there is one problem left that I didn't manage to solve, it is that the internal state of the component does not seem to see that it goes fullscreen, so the button does not change state (=the icon stays the same) and only the `EventType.ENTER_FULL_SCREEN` event is emitted, so I had to add an internal state variable (`mIsFullScreen`) to handle that problem on our level.

About the `ref` handling in `brightcove-video.native.js`, React docs recommend to avoid binding functions in the render
```
We generally recommend binding in the constructor or using the property
initializer syntax, to avoid this sort of performance problem.
```
https://facebook.github.io/react/docs/handling-events.html

As always, feedback is welcome 😄 

Should solve #121 

What do you think @trevorah ?

### DIFF
--- a/packages/brightcove-video/README.md
+++ b/packages/brightcove-video/README.md
@@ -62,6 +62,8 @@ Properties types are defined in `./brightcove-video.proptypes.js`
 | `height` | number | height of the player
 | `onError` | function | Handle errors
 | `onChange` | function |
+| `onGoFullScreen` | function | Handle signal to go full screen in parent (Android)
+| `onExitFullScreen` | function | Handle signal to go out of full screen in parent (Android)
 | `autoplay` | boolean | Should the video autoplay
 
 ## Usage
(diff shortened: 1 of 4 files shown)


trevorah:
Awesome, thanks @adrienthiery!

I'm going to focus on https://github.com/newsuk/times-components/pull/158 before taking a look at this, just so master is more stable.

dickie81:
LGTM

adrienthiery:
I corrected some inconsistencies in the spelling of full screen 😃

trevorah:
@adrienthiery sorry about the delay!

How come you're using `EventType.ENTER_FULL_SCREEN` for entering _and_ exiting full screen? Why not listen to `EventType.EXIT_FULL_SCREEN` as well?

Also, the toggling seems to ignore the view's own internal state, that is probably why the fullscreen icon is not changing.

craigbilner:
Are you still working on this? @adrienthiery

adrienthiery:
Hi @craigbilner,

Yes, I'm still using it. Gonna work a bunch on it a bit more in the next few weeks.

@trevorah Yes, as I was saying with:

> Also, there is one problem left that I didn't manage to solve, it is that the internal state of the component does not seem to see that it goes fullscreen, so the button does not change state (=the icon stays the same) and only the `EventType.ENTER_FULL_SCREEN` event is emitted, so I had to add an internal state variable (mIsFullScreen) to handle that problem on our level.

I don't know how to impact the internal state of the component and I don't find any info on that.

I had expected that dispatching `EventType.ENTER_FULL_SCREEN` would do that, but it doesn't do anything. Do you have any insights or clue on how to solve that problem?

dickie81:
Hi @adrienthiery - we have made some progress regarding Android fullscreen, rendering this PR obsolete. Please checkout master and let us know what you think.

PR #161, opened 2017-08-17
Feat/article byline - single or multiple authors, and styles

thatportugueseguy:
captura de ecra 2017-08-17 as 15 57 24
captura de ecra 2017-08-17 as 15 57 51
captura de ecra 2017-08-17 as 15 58 14


packages/article-byline/fixtures/authors.json
```diff
@@ -0,0 +1,70 @@
+{
+  "fixture": {
```
craigbilner: bit odd having fixtures/fixture?

packages/article-byline/article-byline.stories.js
```diff
@@ -0,0 +1,31 @@
...
+storiesOf("ArticleByline", module)
+  .add("ArticleByline with a single author", () =>
+    story(<ArticleByline ast={authorsAST.singleAuthor} />)
+  )
+  .add("ArticleByline with multiple authors", () =>
+    story(<ArticleByline ast={authorsAST.multipleAuthors} />)
+  )
+  .add("ArticleByline with styles", () =>
+    story(
+      <ArticleByline ast={authorsAST.multipleAuthors} style={bylineStyles} />
```
craigbilner: not sure about the "style API"...the names seem a little off and would we be styling it at this level of granularity?

packages/article-byline/article-byline.js
```diff
@@ -0,0 +1,30 @@
...
+import { builder } from "@times-components/markup";
+import styles from "./article-byline-styles";
+
+import {
+  articleBylinePropTypes,
+  articleBylineDefaultPropTypes
+} from "./article-byline-proptypes";
+
+const ArticleBylineWeb = ({ ast, style }) =>
+  <Text style={[styles.byline, style.byline]}>
```
craigbilner: should byline just return an array of elements rather than with the wrapper?

packages/article-byline/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "flow-bin": "0.42.0",
+    "jest": "20.0.4",
+    "react-test-renderer": "15.5.4",
+    "webpack": "2.6.1"
+  },
+  "dependencies": {
+    "@times-components/link": "0.0.1",
+    "@times-components/markup": "0.1.2",
+    "prop-types": "15.5.10",
+    "react": "16.0.0-alpha.6",
```
craigbilner: needs fixing in line with version changes

packages/article-byline/fixtures/authors.json
thatportugueseguy: I think so too, normally i wouldn't even have the top level fixture prop, i'd have the singleAuthor and multipleAuthors property only. I was only following the pattern on markup, which has fixtures/fixture. Should i change it?

packages/article-byline/article-byline.stories.js
thatportugueseguy: i'm not sure i understand what you mean? What's returning the collection of elements? and where would the the styles be applied? 

the way i see it is that the component is built to have either text or author, and author is link, and text is not. styles.byline affects the text, styles.link affect the author/link. maybe the naming might be changed, but it seems to me that this covers all the use cases.

packages/article-byline/article-byline.js
thatportugueseguy: As far as i understand, this is not possible in react presently. It will be implemented on react 16.

packages/article-byline/package.json
thatportugueseguy: will do

packages/article-byline/fixtures/authors.json
craigbilner: ah, that's because each file is an individual fixture and needs an identifier, this file just contains each named fixture

PR #175, opened 2017-08-22
feat/android-reinit-on-rotate

dickie81:
This PR causes the android brightcove video component to internally re-initialise when the device's orientation is changed. This has been changed because the player was getting into strange states when a playing video was rotated (for example the player controls being displayed in the wrong position).


packages/brightcove-video/android/src/main/java/uk/co/news/rntbrightcovevideo/RNTBrightcoveView.java
```diff
@@ -1,162 +1,112 @@
...
+
+        this.removeAllViews();
+
+        initPlayerView();
+
+        super.onConfigurationChanged(newConfig);
+    }
+
+    public BrightcovePlayerView getPlayerView() {
+        return mPlayerView;
```
trevorah: its pretty nasty to expose your internals like this just for `start()` and `stop()`. Can you create your own start/stop methods and use them to call the mPlayerView's methods please.

trevorah:
@Szymczak @chrishoyes Any thoughts on the java stuff before it reaches your codebase?

PR #177, opened 2017-08-24
fix: ensure no exception is thrown no asset

fampinheiro:


fampinheiro:
Did some changes to `Card` to ensure it plays well with not having an image URL.

@craigbilner what do you mean about video poster image, are they different than giving an image URL?

craigbilner:
@fampinheiro - this is the format

```
{
  "leadAsset": {
    "posterImage": {
      "title": "",
      "crop": {
        "url": "https://image.com/cat"
      }
    }
  }
}
```

PR #178, opened 2017-08-24
twitter username link proposal

trevorah:
DO NOT MERGE

I've done this as a bit of a discussion piece. Web, Android and iOS all have their different navigation styles, so I figured it would be up to something outside of the component to figure out how to navigate to a given uri.

Here's a proposal in the form of an example for author profile twitter username links. Check out and try the storybook.

talking points:
- **buttons vs links:** on web, these "links" feel like buttons. this may help us feel more appy, but it means that people cant open the twitter profiles in new tabs. 👎 
- **prop passing:** It may just be the author profile components, but passing a function through 3 components seems a bit much. fix the nesting or use context hackery?
- **required functions:** buttons that look enabled but are unresponsive are poo poo. I say make them required.
- **fancy links:** article link navigation would be so much easier if ids (not just urls) get passed around. Should we have a catchall function or specialist ones (e.g `onLinkPress` vs `onArticleLinkPress` and`onAuthorLinkPress` etc)

@craigbilner @fampinheiro @nehasri89 @JGAntunes thoughts?

### DIFF
--- a/packages/author-head/author-head.js
+++ b/packages/author-head/author-head.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { View, StyleSheet, Text } from "react-native";
+import { View, StyleSheet, Text, TouchableHighlight } from "react-native";
 import PropTypes from "prop-types";
 
 import Image from "@times-components/image";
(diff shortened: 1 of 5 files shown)


fampinheiro:
- **buttons vs links** Using the link this ends up being just a style change.
- **prop passing** This seems to be the most used approach when dealing with this kind of problem
- **fancy links** we could also have an event emitter on the link component that emit every tap.

PR #181, opened 2017-08-25
test: add author-profile functional test

rjanjua:
A simple first functional test for journalist pages


packages/author-profile/package.json
```diff
@@ -45,6 +45,7 @@
     "babel-plugin-transform-react-remove-prop-types": "0.4.5",
     "babel-preset-flow": "6.23.0",
     "babel-preset-react-native": "1.9.2",
+    "chance": "^1.0.10",
```
fampinheiro: The dependencies normally are locked to a particular version.

packages/author-profile/fixture-generator.js
```diff
@@ -0,0 +1,160 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import Chance from "chance";
+
+const chance = new Chance(88888);
+
+const name = () => "Fiona Hamilton";
```
fampinheiro: Any special reason why these are functions and not simple variable assignments?

rjanjua: We could just use strings. When I started writing it I had something more complicated in my mind using the chance object, but I only ended up using that in a few specific places.

PR #182, opened 2017-08-25
fix: change teaser to content with 100 words

fampinheiro:


craigbilner:
@fampinheiro - needs to change to "100 characters", dependent on API change

PR #183, opened 2017-08-26
BREAKING CHANGE: Links to the root!

trevorah:
UPDATED 7/9/17

This separates the Link component into two, `Link` and `TextLink`. Link now works with blocks (✨ ), whereas `TextLink` is now @fampinheiro's old link component that styles text as well as linking.

This is a breaking change as onPress is now a required prop and TextLink is no longer the default (as it is rarer).

e.g:
```js
import Link, { TextLink } from "@times-components/link";

<Link url="https://thetimes.co.uk" onPress={action("onPress")}>
  <View style={{ width: 100, height: 100, backgroundColor: "pink" }} />
</Link>

<TextLink url="https://thetimes.co.uk/" onPress={action("onPress")}>
  The Times
</TextLink>
```

Both require `onPress` so that native navigation can be hooked up (or a web page that wants to preventDefault) and `url` so that web gets a nice anchor tag.

sep-07-2017 14-43-36

(The default "block" *could* just be an alias for `TouchableOpacity`, but I wanted to limit the props it supported so that it could be replaced if/when react-native changes that the `Touchable*` components.)

## but why?
This big breaking PR is all about exposing navigation to the root component. React native's `Linking.openURL(url)` is nice for opening links outside of the app, but does not help for internal navigation within the app. As we can also will have multiple react native views in our app, we must be able to tie the view owner (an android `Activity` or ios `UIViewController`) to whatever component requested the navigation. If we dont, then every view owner will respond to a navigation event, resulting in double navigation in our "history" stacks.

## Markup changes
```js
import { renderTree, renderTrees } from "@times-components/markup";

const renderers = {
  author(key, attributes, children) {
    const url = `/profile/${attributes.slug}`;
    return (
      <TextLink
        style={styles.link}
        key={key}
        url={url}
        onPress={() => console.error("Whoops! I need a callback")}
      >
        {children}
      </TextLink>
    );
  }
};

const ArticleByline = ({ ast, style }) =>
  <View>
    {renderTrees(ast, renderers)}
  </View>;
```
Markup is no longer a component, just two vanilla functions. It also only renders non-interactive components by default (bold, italic, no links). It can render any named component if you pass in the correct renderer. This way `renderTree` doesnt have to always require a "onLinkPressed" in the off-chance that the ast contains a link.

`renderTrees` exists as the "ast" usually comes in as an array of asts (a forest!), so this acts as a convenience function. It is also no longer a component as the `key` [cannot be passed from component to component](https://facebook.github.io/react/warnings/special-props.html), which would be needed in a safe tree structure.

The tree rendering is also simplified, so it no longer tries to squash text nodes into parents (e.g `bold -> text -> "hello"`` becomes `bold -> "hello"`). This allows us to improve the ast on the server side without too many odd edge cases. You'll notice this in the snapshot changes, but visually it is identical.

## WARNING
I've limited the "breaking change" commits to only apply to single packages. If you squash the commits, then lerna will think that the breaking change applies to all packages, *even the ones that are just adapting to the breaking change!* **Do not squash the commits!**


trevorah:
This breaks the markup component, which I can fix by passing in a `onLinkPressed(event, url)` prop to the markup component. @craigbilner theres an `author` tag in the markup components tag map that generates relative urls. Is that meant to be there?

craigbilner:
@trevorah, an `author` component would directly map to your `TextLink` which is currently used in the `byline`

trevorah:
So we expect every "markupable" bit of content to support `author` tags? This would mean that the Markup component would need to have a `onAuthorPress` prop.

trevorah:
@craigbilner @fampinheiro fresh review please!

fampinheiro:
Thinking about this yesterday couldn't we check if the children is a string and wrap it in a `Text` if that is the case? Otherwise just render the children?

trevorah:
@fampinheiro id like to keep the "interpretation" of the ast from the server to be as literal as possible, but also make implementing a renderer pretty idiot proof if bad data comes in. Making each renderer check the type of the children is prone to failure imo.

The ast format is up for change, so we could solve this problem at the server to make the text nodes easier to handle?

Issue #172, opened 2017-08-22
Article Page Findings | Article Flag support for styles


killanaca:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behaviour?**
doesn't support the style prop
**If the current behaviour is a bug, please provide the steps to reproduce.**

**What is the expected behaviour?**

**If this is a feature request, what is the motivation or use case for changing the behaviour?**
support multiple articleFlag on the same article
**Please mention other relevant information such as the browser version, Node.js version and Operating System.**

PR #188, opened 2017-08-30
Article Flag support for styles

killanaca:
ref #172 

- updated design
- fixed a warning on a missing title prop for the base ArticleFlag (inside the tests)
- use `Stylesheet` for internal component styling instead of plain object to avoid duplication on style attribute
- support external styling on the container and on the text element to avoid extra `<View>` to style the component on the `Article` component

```jsx
<View style={styles.ArticleFlagContainer}>
    <NewArticleFlag />
</View>
```
```jsx
  container: {
    backgroundColor: "#EAFF00"
  },
  text: {
    fontSize: 12
  }
});
<NewArticleFlag containerStyle={styles.container} textStyle={styles.text}/>
```

### DIFF
--- a/packages/article-flag/article-flag.js
+++ b/packages/article-flag/article-flag.js
@@ -43,52 +43,52 @@ const ArticleFlag = ({ title, color, style }) => {
 ArticleFlag.propTypes = {
   title: PropTypes.string.isRequired,
   color: PropTypes.string,
-  style: PropTypes.number
+  style: View.propTypes.style
 };
 
 ArticleFlag.defaultProps = {
   color: "black",
-  style: undefined
+  style: {}
 };
 
 export const NewArticleFlag = props =>
   <ArticleFlag title="new" color="#E34605" style={props.style} />;
 
 NewArticleFlag.propTypes = {
-  style: PropTypes.string
+  style: View.propTypes.style
 };
 NewArticleFlag.defaultProps = {
-  style: undefined
+  style: {}
 };
 
 export const UpdatedArticleFlag = props =>
   <ArticleFlag title="updated" color="#3C81BE" style={props.style} />;
 
 UpdatedArticleFlag.propTypes = {
-  style: PropTypes.string
+  style: View.propTypes.style
 };
 UpdatedArticleFlag.defaultProps = {
-  style: undefined
+  style: {}
 };
 
 export const ExclusiveArticleFlag = props =>
   <ArticleFlag title="exclusive" color="#C51D24" style={props.style} />;
 
 ExclusiveArticleFlag.propTypes = {
-  style: PropTypes.string
+  style: View.propTypes.style
 };
 ExclusiveArticleFlag.defaultProps = {
-  style: undefined
+  style: {}
 };
 
 export const SponsoredArticleFlag = props =>
   <ArticleFlag title="sponsored" color="#4D4D4D" style={props.style} />;
 
 SponsoredArticleFlag.propTypes = {
-  style: PropTypes.string
+  style: View.propTypes.style
 };
 SponsoredArticleFlag.defaultProps = {
-  style: undefined
+  style: {}
 };
 
 export default ArticleFlag;
--- a/packages/article-flag/article-flag.js
+++ b/packages/article-flag/article-flag.js
@@ -1,12 +1,12 @@
 import React from "react";
-import { View, Text } from "react-native";
+import { View, Text, StyleSheet } from "react-native";
 import PropTypes from "prop-types";
 
 import Diamond from "./diamond";
 
 // When changing styles please debug both web, android and ios because
 // some styles are not working correctly on all platforms (namely, android)
-const styles = {
+const styles = StyleSheet.create({
   view: {
     flexDirection: "row",
     alignItems: "center"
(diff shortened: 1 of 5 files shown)


packages/article-flag/article-flag.js
```diff
@@ -1,12 +1,12 @@
...
-import { View, Text } from "react-native";
+import { View, Text, StyleSheet } from "react-native";
 import PropTypes from "prop-types";
 
 import Diamond from "./diamond";
 
 // When changing styles please debug both web, android and ios because
 // some styles are not working correctly on all platforms (namely, android)
-const styles = {
+const styles = StyleSheet.create({
```
sericaia: We were not using Stylesheet.create here for performance reasons

packages/article-flag/article-flag.js
```diff
@@ -20,24 +20,24 @@ const styles = {
...
-};
+});
 
 // apply transformations to add uppercase and letter spacing.
 // letterSpacing CSS prop does not work on android:
 // https://github.com/facebook/react-native/pull/13199
 const beautifyTitle = title => title.toUpperCase().split("").join(" ");
 
-const ArticleFlag = ({ title, color }) => {
+const ArticleFlag = ({ title, color, containerStyle, textStyle }) => {
```
thatportugueseguy: should style be a single prop that gets an object?

This way it keeps in line with html convention and it's normally used like that. I implemented it like this on byline: [link](https://github.com/newsuk/times-components/blob/master/packages/article-byline/article-byline-proptypes.js#L13)

packages/article-flag/article-flag.js
```diff
@@ -46,20 +46,25 @@ const ArticleFlag = ({ title, color }) => {
 
 ArticleFlag.propTypes = {
   title: PropTypes.string.isRequired,
-  color: PropTypes.string
+  color: PropTypes.string,
+  containerStyle: PropTypes.number,
```
thatportugueseguy: if we don't use the StyleSheet.create, you can use Text.proptype.style and View.proptypes.sytle

packages/article-flag/article-flag.js
killanaca: StyleSheet.create avoid to duplicate each style using ref instead

packages/article-flag/article-flag.js
```diff
@@ -46,20 +46,23 @@ const ArticleFlag = ({ title, color }) => {
...
 
-const NewArticleFlag = () => <ArticleFlag title="new" color="#E34605" />;
-const UpdatedArticleFlag = () =>
-  <ArticleFlag title="updated" color="#3C81BE" />;
-const ExclusiveArticleFlag = () =>
-  <ArticleFlag title="exclusive" color="#C51D24" />;
-const SponsoredArticleFlag = () =>
-  <ArticleFlag title="sponsored" color="#4D4D4D" />;
+const NewArticleFlag = props =>
+  <ArticleFlag title="new" color="#E34605" {...props} />;
```
craigbilner: any need to spread the props here?
also could just export each line...

packages/article-flag/article-flag.js
```diff
@@ -17,23 +17,23 @@ const styles = {
...
-};
+});
 
 // apply transformations to add uppercase and letter spacing.
 // letterSpacing CSS prop does not work on android:
 // https://github.com/facebook/react-native/pull/13199
 const beautifyTitle = title => title.toUpperCase().split("").join(" ");
 
-const ArticleFlag = ({ title, color }) => {
+const ArticleFlag = ({ title, color, containerStyle }) => {
```
thatportugueseguy: are we using containerStyle instead of style?

packages/article-flag/article-flag.js
```diff
@@ -17,23 +17,23 @@ const styles = {
...
   if (!title) {
     return null;
   }
 
   return (
-    <View style={styles.view}>
+    <View style={[styles.view, containerStyle]}>
       <View style={styles.diamond}>
```
thatportugueseguy: Do we really need the view around `Diamond`?

packages/article-flag/article-flag.js
killanaca: in this case I've just preferred to be more open on the props validation for a better bundle size for the web. 
However I've updated both

packages/article-flag/article-flag.js
killanaca: done

packages/article-flag/article-flag.js
killanaca: removed

fampinheiro:
What is the reason for exporting multiple components instead of receiving a prop that renders a set of predefined styles?

```
<NewArticleFlag {...props } />
<UpdatedArticleFlag {...props } />
<ExclusiveArticleFlag {...props} />
```
instead of 

```
<ArticleFlag {...props } type="new" />
<ArticleFlag {...props } type="updated" />
<ArticleFlag {...props } type="exclusive" />
```

Not blocking the merge, just wondering the reason behind it. I find the usage of the second approach easier to use and grasp.

packages/article-flag/article-flag.js
```diff
@@ -46,26 +42,53 @@ const ArticleFlag = ({ title, color }) => {
 
 ArticleFlag.propTypes = {
   title: PropTypes.string.isRequired,
-  color: PropTypes.string
+  color: PropTypes.string,
+  style: PropTypes.number
```
craigbilner: is this right?

fampinheiro: You probably want to use [ViewPropTypes.style](https://github.com/facebook/react-native/blob/master/Libraries/Components/View/ViewPropTypes.js#L391)

`import { ViewPropTypes } from "react-native"`

killanaca: not yet, we have to stick with `View.propTypes.style` until we'll update React

killanaca:
AFAIN if we want to use StyleSheet, we can't be strict eg:
```
NewArticleFlag.propTypes = {
        style: PropTypes.shape({
            marginRight: PropTypes.number
        })
};
```

otherwise we get  "Invalid prop `style` of type `number` supplied to `NewArticleFlag`, expected `object`." error message.

StyleSheet pass through a number as a ref not a real style object.
@craigbilner any suggestion?

PR #189, opened 2017-08-30
fix: fetch https images on native

fampinheiro:
On native, the URL resolver prepends the images path with `file`.
This PR ensures we fetch images from the image server using the https scheme.

### DIFF
--- a/packages/provider/author-profile-provider.js
+++ b/packages/provider/author-profile-provider.js
@@ -1,6 +1,5 @@
 import get from "lodash.get";
 import gql from "graphql-tag";
-import { Platform } from "react-native";
 import AuthorProfile from "@times-components/author-profile";
 import connectGraphql from "./connect";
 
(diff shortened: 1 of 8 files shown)


packages/provider/author-profile-provider.js
```diff
@@ -56,6 +57,18 @@ const propsToVariables = ({ slug, pageSize, page, imageRatio }) => ({
 });
 
 const transformResponse = response => {
+  const resolve = url => {
```
craigbilner: this isn't to do with `author-profile-provider` - shouldn't this be more generic provider?

fampinheiro: This is not completely related to provider, is more how the components use the data the provider returns.

From a deeper analysis maybe this should be on the image component. The native component should now that it can't resolve an url without schema?

craigbilner: sound better :-)

PR #190, opened 2017-08-31
Fix:Add Babel transforms to Brightcove Video package.json to enable consumers to use CommonJS

jackharvey1:


craigbilner:
agree with @dickie81, what does this give us @jackharvey1 ?

jackharvey1:
Took a different approach

PR #204, opened 2017-09-08
Fix/bio width

foliveira:
**Web:**
web

**Mobile:**
mobile

**iPhone:**

iphone

**Android:**

android


trevorah:
@foliveira looks good! How does it look on iOS and Android?

fampinheiro:
@trevorah the native images were already uploaded.

craigbilner:
@foliveira if we're following the web, < 768px is left aligned?

foliveira:
@craigbilner the change you mentioned could be easily implemented with the responsive stylesheets feature I'm working on, i.e.: it would avoid spreading checks for the width in components.

Can we move to that once that feature lands?

PR #206, opened 2017-09-11
fix: show images on web

fampinheiro:
On native, the `Image.onLoad` is only called when the image has dimensions so we have a minimum height (1). 

On the web, the image doesn't need to have dimensions and in order for the image to appear to the user, we let the browser decide on the dimensions of the image.

### DIFF
--- a/packages/author-head/__tests__/__snapshots__/author-head.test.js.snap
+++ b/packages/author-head/__tests__/__snapshots__/author-head.test.js.snap
@@ -97,8 +97,8 @@ exports[`renders a snapshot 1`] = `
         style={
           Array [
             Object {
-              "height": undefined,
-              "width": undefined,
+              "height": 1,
+              "width": 750,
             },
             Object {
               "borderColor": "#FFF",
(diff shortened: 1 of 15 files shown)


sericaia:
@foliveira remove yarn.lock and its fine to merge 💃

foliveira:
Latest commit ccb9f11 adds fallback `inherit` value to Image width and height and removes parent View from `render()` method, since ReactNative supports `onLayout` calls ootb and ReactNativeWeb already wraps `img` tags in a `View`/`div` tag.

Check: 

https://github.com/newsuk/times-components/pull/206/commits/ccb9f111136967ff2e443c5f041ba5e1f9266935#diff-78ea4c4e39449a548b4b17c26f052ff3R66

https://github.com/newsuk/times-components/pull/206/commits/ccb9f111136967ff2e443c5f041ba5e1f9266935#diff-78ea4c4e39449a548b4b17c26f052ff3R74

foliveira:
@craigbilner added the SSR story

craigbilner:
@foliveira card story looks broken?

foliveira:
Pushed web image component and fixed some stories to work both on native and web.

Right now the best approach would be to wrap `Image`s with a `View` and then size them directly instead of doing it on the `View`.

foliveira:
Closing this one in favour of #257

PR #207, opened 2017-09-11
export the main entry point without the ext

killanaca:
export the main entry point without the ext otherwise we can't use the specific android component


sericaia:
nice catch!

PR #208, opened 2017-09-14
FIX - Xcode errors

killanaca:
Remove storybooknative-tvOS, storybooknative-tvOSTests, storybooknativeTests and relative targets (and errors) from the Xcode project to deploy the storybooknative app on a real iphone device


nehasri89:
Tested on real iphone and works

PR #209, opened 2017-09-14
FIX - Date Publication line Height

killanaca:
- fix lineHeight 
- update snapshots
- use naming convention on `__test__` and `.native.test.js`

schermata 2017-09-14 alle 5 39 08 pm


packages/date-publication/__test__/__snapshots__/date-publication.ios.native.test.js.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
sericaia: why do you need ios.native.test and android.native.test if the tests are equal?

killanaca: fixed

PR #210, opened 2017-09-14
FIX - Article Label letter spacing and line height

killanaca:
- use unicode space (to mimic the letter spacing just for android)
- fix lineHeight 
- fix font-size
- update snapshots
- add android tests
- use naming convention on `__test__` and `.native.test.js`

schermata 2017-09-14 alle 5 43 58 pm


packages/article-label/__test__/article-label.ios.native.test.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-env jest */
```
sericaia: We dont need this test.. we only need `android` and `others` (web, ios) because android is the only one that is different.

packages/article-label/article-label.js
```diff
@@ -1,24 +1,29 @@
...
+  if (Platform.OS === "android") {
+    return title
+      .toUpperCase()
+      .split("")
+      .join("\u200A");
+  }
+  return title.toUpperCase();
+};
 const ArticleLabel = ({ title, color }) => {
   if (!title) {
```
sericaia: I'll remove this validation as this prop is required

PR #212, opened 2017-09-14
FIX - fetch fonts task

killanaca:
- add fontFamily property to fonts collection inside the fetch-fonts task to fix an issue with bold and italic variants of the same font family
- add TimesModern-Regular font 
- add TimesDigitalW04-Regular font
- add TimesDigitalW04-Italic font 
- add TimesDigitalW04-Bold font


killanaca:
@sericaia fonts added. thanks for the catch

packages/author-head/author-head.js
```diff
@@ -47,7 +50,8 @@ const styles = StyleSheet.create({
     textDecorationLine: "none"
   },
   bio: {
-    fontFamily: "TimesDigital-Regular",
+    fontFamily:
+      Platform.OS === "android" ? fontFamilyAndroid : fontFamilyWebAndIos,
```
sericaia: I would prefer to have Platform.select

.storybook/preview-head.html
```diff
@@ -26,5 +33,26 @@
         font-weight: 400;
```
sericaia: shouldn't this be removed too?

nehasri89: just noticed this is diff than 'TimesDigital-Regular', it is 'TimesDigital-RegularSC'

killanaca:
#227

PR #213, opened 2017-09-14
Feat/article images [DO NOT MERGE]

sericaia:
Styles under confirmation with Jake Cox. > Update: Confirmed!

This PR is not supposed to be merged. [better-links](https://github.com/newsuk/times-components/pull/183) needs to be first merged to master.

image


packages/article-image/article-image.js
```diff
@@ -0,0 +1,114 @@
...
+      paddingTop: 0
+    }
+  }
+};
+
+const ArticleImage = ({ ast }) => (
+  <View>
+    {renderTrees(ast, {
+      image(key, attributes) {
+        const { id, display, caption, credits, ratio, url } = attributes;
```
trevorah: those attributes are coming straight off the api, which doesnt guarantee their individual presence.

You will need to think about what to do when they are missing, otherwise you'll get exceptions all over the place

killanaca:
how do we test the use of `Platform.select`? 
can we use the `__test__` folder?

sericaia:
Closed in favor of https://github.com/newsuk/times-components/pull/222

PR #214, opened 2017-09-15
wip: breakpoints in stylesheets

foliveira:
**Example of resizing adapting layout:**

rw

**Attention:** Tests for some components are failing due to some CSS properties not being supported in React Native.

### DIFF
--- a/packages/responsive-hoc/responsive-hoc.js
+++ b/packages/responsive-hoc/responsive-hoc.js
@@ -1,0 +1,67 @@
+import React from 'react';
+import { View, Dimensions } from 'react-native';
+
+const Breakpoints = {
+  LARGE: Symbol("large"),
+  MEDIUM: Symbol("medium"),
+  SMALL: Symbol("small")
+};
+
+const checkSize = (minWidth, maxWidth) => {
+  return (w) => w >= minWidth && (!maxWidth || w < maxWidth);
+};
+
+const Sizes = {
+  [Breakpoints.SMALL]: checkSize(0, 768),
+  [Breakpoints.MEDIUM]: checkSize(768, 1024),
+  [Breakpoints.LARGE]: checkSize(1024)
+};
+
+const StyleHOC = (WrappedComponent, Styles) => {
+  return class extends React.Component {
+    static propTypes = WrappedComponent.propTypes;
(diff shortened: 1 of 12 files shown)


foliveira:
Check [responsive stylesheet declaration](https://github.com/newsuk/times-components/pull/214/files#diff-113de1101152433e7c06751df1f9ce0fR71), [using the HOC](https://github.com/newsuk/times-components/pull/214/files#diff-113de1101152433e7c06751df1f9ce0fR151) and [actual HOC code to enable applying different stylesheets](https://github.com/newsuk/times-components/pull/214/files#diff-3e75e4ae83e903b94f13c166f311ce19R20)

craigbilner:
Thanks @foliveira 

I've rebased and make a suggestion for "resolving breakpoints" here:

https://github.com/newsuk/times-components/compare/master...fix/author-avatar-notes?expand=1

I've made some notes (not all related to the PR)

* Headline size needs looking at
* Left align bio on small breakpoints
* Title is `TimesDigital204-RegularSC`
* Bio is a bit tight on twitter

* Need to use web select for web only props
* Sizes should work like CSS where it takes "greater than" (see branch)
* Symbols on IE11 😏 
* not sure about the API, `responsive` is perhaps unclear, we've used style(s) in other components
* it might be safe to assume that responsive hackery will be web only for the forseeable future?
* to make the HOC bullet proof it would need to do all the copying of static methods etc. see tracking for an example

packages/responsive-hoc/responsive-hoc.js
```diff
@@ -0,0 +1,67 @@
...
+
+const Breakpoints = {
+  LARGE: Symbol("large"),
+  MEDIUM: Symbol("medium"),
+  SMALL: Symbol("small")
+};
+
+const checkSize = (minWidth, maxWidth) => {
+  return (w) => w >= minWidth && (!maxWidth || w < maxWidth);
+};
```
thatportugueseguy:
```js 
// closure that checks if the argument is within the bound values
const checkSize = (minWidth, maxWidth) => w => (w >= minWidth && (!maxWidth || w < maxWidth));
```

Maybe there's no need for a function body here?

PR #217, opened 2017-09-19
feat/brightcove-reset

dickie81:
Exposes a reset method so player can be reset to poster image. (Also exposes play and pause as they should have been)


trevorah:
Can you give some context as to why this is needed? It would help me understand what the user expects if they press play on a "resetted" video.

dickie81:
This was requested by Mike so that when a video finishes or we navigate away it can be reset to poster image. When the reset video plays, it plays from the start.

packages/brightcove-video/brightcove-video.js
```diff
@@ -1,22 +1,66 @@
...
+    this.state = {
+      isLaunched: props.autoplay
+    };
+
+    this.play = this.play.bind(this);
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    return nextState.isLaunched !== this.state.isLaunched;
+  }
```
trevorah: ???

PR #220, opened 2017-09-19
Fix/article byline fix

nehasri89:
- Changing Line Height for byline as per design
- Updating snapshot
screen shot 2017-09-19 at 12 14 13
screen shot 2017-09-19 at 14 11 20


packages/article-byline/article-byline-styles.js
```diff
@@ -1,10 +1,22 @@
 import { Platform } from "react-native";
 
+const lineHeightStyle = Platform.select({
+  ios: {
+    lineHeight: 14
+  },
+  android: {
+    lineHeight: 17
+  },
+  web: {
```
nehasri89: Needs
whiteSpace: "pre" style for web, not sure lineHeight: 13 resolves to 13px? need to check in storybook.

sericaia: `whiteSpace: "pre"` > where does it come from?

LineHeight resolves, I confirmed and you can check that on snapshots too

killanaca: it resolves to 13px because we use vanilla object attributes (not StyleSheet)

nehasri89: @sericaia the whitespace attr was originally there in the web styles file

sericaia: Fixed

PR #222, opened 2017-09-19
Feat: article images

sericaia:
Same as https://github.com/newsuk/times-components/pull/213 but to master, as https://github.com/newsuk/times-components/pull/183 is already in!

NOTE: We support only Mobile web, ios and android. Desktop to follow after responsive design

the other styles for desktop were removed on purpose for now


packages/article-image/article-image.js
```diff
@@ -0,0 +1,123 @@
...
+    return null;
+  }
+  return (
+    <View style={styles[`${display}Caption`]}>
+      <Caption text={caption} credits={credits} style={captionStyle[display]} />
+    </View>
+  );
+};
+
+const ArticleImage = ({ ast }) => (
```
craigbilner: This doesn't look right? Should an `ArticleImage` component which AFAICT calculates an aspect ratio and sticks a caption on take an `ast` of images? Surely the article will `renderTrees` and handle images in their renderer?

packages/article-image/article-image.js
```diff
@@ -0,0 +1,123 @@
...
+          return null;
+        }
+
+        const source = {
+          uri: url
+        };
+
+        const ratioStyle = getImageRatioBox(ratio);
+        return (
+          <View key={id} style={styles[display]}>
```
craigbilner: ...all the nested views...not great for perf?

packages/article-image/article-image.js
sericaia: changing it from ast to `<ArticleImage image={imageDetails} caption={captionDetails} />`

packages/article-image/article-image.js
sericaia: They are required due to the paddingBottom trick

packages/article-image/article-image.js
```diff
@@ -0,0 +1,171 @@
...
+      }
+    })
+  }
+});
+
+const getImageRatioBox = ratio => {
+  const tokens = ratio.split(":");
+  const padding = tokens[1] / tokens[0] * 100;
+
+  return StyleSheet.create({
```
thatportugueseguy: Maybe we should move Platform.select to this level, or even more up (return early)?

packages/article-image/article-image.js
```diff
@@ -0,0 +1,171 @@
...
+  return (
+    <View key={url} style={styles[display]}>
+      <View style={styles[`${display}Image`]}>
+        <View style={ratioStyle.ratioBox}>
+          <View style={ratioStyle.ratioBoxInside}>
+            <Image source={source} />
+          </View>
+        </View>
+      </View>
+      {renderCaption(display, caption, credits)}
```
thatportugueseguy: do we really need this function? we could have an if statement instead

sericaia: not going to be fixed as we discussed ;)

PR #226, opened 2017-09-20
Feat: article

sericaia:
TODO:
- fix markup paragraphs which no longer work like before


packages/article/article-meta.js
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { View } from "react-native";
+
+import ArticleByline from "@times-components/article-byline";
+import DatePublication from "@times-components/date-publication";
+
+import styles from "./article-style";
+
+const ArticleHeader = ({ byline, publishedTime, publicationName }) =>
```
sericaia: should this file be renamed to article-header?

packages/article/article-style.js
```diff
@@ -0,0 +1,152 @@
+import { StyleSheet, Platform } from "react-native";
+
+const standFirstFontFamily = "TimesModern-Regular";
+const bodyFontFamilyWebIos = "TimesDigitalW04";
+const bodyFontFamilyAndroid = "TimesDigitalW04-Regular";
+const borderColor = "#d0cece";
+const borderWidth = StyleSheet.hairlineWidth;
+
+const globalStyle = StyleSheet.create({
+  Container: {
```
sericaia: Shouldn't we use camelCase in this styles?

packages/article/article.js
```diff
@@ -1,11 +1,153 @@
...
+      get(articleData, "content", []).map(i => ({
+        type: "article_body_row",
+        data: i
+      }))
+    );
+    return ArticleArray;
+  }
+
+  static renderRow(rowData) {
+    if (rowData.type === "ads" && Platform.OS === "web") {
```
sericaia: this "ads" part should be removed for now, right? and afaik it will use what @thatportugueseguy @JGAntunes are doing

nehasri89: actually, those guys are doing the ads within the article and this is the header ad, so this stays but need review from @thatportugueseguy and @JGAntunes as discussed in yesterday's standup

packages/article/article.js
```diff
@@ -1,11 +1,153 @@
...
+      )
+    };
+    return (
+      <View style={styles.PageWrapper}>
+        <ListView
+          dataSource={this.state.dataSource}
+          renderRow={ArticlePage.renderRow}
+          initialListSize={10}
+          scrollRenderAheadDistance={10}
+          pageSize={1}
```
sericaia: should we move this config values somewhere?

packages/article/article.js
```diff
@@ -1,11 +1,153 @@
...
+ArticlePage.propTypes = {
+  data: PropTypes.shape({
+    article: PropTypes.object,
+    loading: PropTypes.boolean
+  })
+};
+
+ArticlePage.defaultProps = {
+  data: {
+    loading: true
```
sericaia: if `article` is not required, we need to add a default prop

packages/article/article.js
nehasri89: confirmed with @craigbilner remove the ad from here and will raise a separate story

packages/article/article.js
```diff
@@ -1,11 +1,153 @@
...
+  render() {
+    if (this.props.data.loading) {
+      return (
+        <View style={styles.Container}>
+          <ActivityIndicator size={"large"} />
+        </View>
+      );
+    }
+
+    this.state = {
```
sericaia: I'm really confused about this. We should not mutate state here and every time we mutate state we should use`this.setState` and not mutating it directly. The only case where we can assign to `this.state` is in the constructor

packages/article/article.js
```diff
@@ -1,11 +1,153 @@
...
+    const ArticleArray = Array.prototype.concat(
+      [
+        { type: "ads", data: adsData },
+        { type: "leadAsset", data: leadAssetData },
+        { type: "header", data: articleHeaderData },
+        { type: "middleContainer", data: articleMidContainerData }
+      ],
+      // NOTE failing gracefully
+      get(articleData, "content", []).map(i => ({
+        type: "article_body_row",
```
sericaia: do we really need this?

packages/article/article.js
sericaia: consider https://facebook.github.io/react/docs/react-component.html#componentwillreceiveprops

packages/article/article.js
```diff
@@ -1,11 +1,146 @@
...
+      //       >
+      //         {React.cloneElement(el, {
+      //           style: StyleSheet.flatten([styles.articleTextElement])
+      //         })}
+      //       </View>
+      //     )}
+      //   </View>
+      // );
+      // TODO :: fix this, builder does not exit now
+      return null;
```
sericaia: this return will never be used

packages/article/CHANGELOG.md
```diff
@@ -3,35 +3,6 @@
 All notable changes to this project will be documented in this file.
 See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-<a name="0.5.3"></a>
```
JGAntunes: Should this PR be interfering with the CHANGELOG like this? (removing releases)

packages/article/article.js
```diff
@@ -1,11 +1,146 @@
...
+    } else if (rowData.type === "article_body_row") {
+      // return (
+      //   <View style={[styles.articleMainContentRow, styles.articleText]}>
+      //     {builder({ ast: [rowData.data] }).map(el =>
+      //       <View
+      //         style={styles.articleTextWrapper}
+      //         key={`paragraph-${Date.now().toLocaleString()}`}
+      //       >
+      //         {React.cloneElement(el, {
+      //           style: StyleSheet.flatten([styles.articleTextElement])
```
sericaia: why do you need StyleSheet.flatten @killanaca ?

packages/article/article.js
nehasri89: yes, but eslint complains of not having a return from a method, so either we can disable eslint for this block or have a null return

killanaca:
#234

PR #227, opened 2017-09-20
FIX - Fetch Fonts

killanaca:
- add fontFamily property to fonts collection inside the fetch-fonts task to fix an issue with bold and italic variants of the same font family
- add TimesModern-Regular font 
- add TimesDigitalW04-Regular font
- add TimesDigitalW04-Italic font 
- add TimesDigitalW04-Bold font

schermata 2017-09-21 alle 10 30 14 am
screen shot 2017-09-21 at 11 03 35

schermata 2017-09-21 alle 10 30 35 am
screen shot 2017-09-21 at 11 02 59

schermata 2017-09-21 alle 10 30 58 am
screen shot 2017-09-21 at 11 03 16

Screenshots of **Android doing the style for italic and bold on Left** and using **discrete fonts on Right**

at_presentactual_fonts


rjanjua:
Tested the fructose tests locally and they are passing. Reason for failure on bitrise is due to us using a cached version of the app, and not having fontforge on the build machine

nehasri89:
The way in which the fonts work is different for different platforms. 

- For ios, we need different fonts for each fontStyle i.e. italic, bold, normal and they should share the same font family. To use it we provide the fontFamily to the text and ios is able to implicitly choose the right font and use it as per the text style.
- For Android, it does not understands the font family, it needs an actual font name, and is able to style the same font looking at the style attribute
- For Web, it can also style the same font similar to Android looking at the fontStyle property, we can either provide the fontFamily OR a particular font of style normal.
 hence the check below:

Platform.OS === "android" ? fontFamilyAndroid : fontFamilyWebAndIos,

PR #228, opened 2017-09-20
Feat/trackingv4

toboid:
PR for discussion of tracking implementation - not ready for merge.

1st commit is the tracking components.
2nd commit is example usage. It's a little rough and just there to show how the tracking could be integrated. It's not yet confirmed exactly what will be sent to Tealium so this is just an example of how the tracking might be used. This falls behind master very quickly, when we are happy to move forward with tracking i'll redo in a seperate PR.

**addTracking()**
Adds tracking to the wrapped component. This can be tracking views, child views (e.g. paragraphs in an article) or events raised by the wrapped component. 

Configuration is by parameters to `addTracking()` - if we create and test a tracked version of a component it won't be re-configurable in different usages via props. The `attrs` configuration option allows hardcoded key/values to be applied to events as well as generation of values from props (for example extract the id from an article).

Since we don't have a way of automatically observing child views on native (that i'm aware of), list components would be expected to expose an `onViewed` event, on web this isn't required. I'm especially interested in any thoughts on this .

**addTrackingContext()**
Adds a tracking context that will decorate events raised any descendent components with requested tracking attributes. The root context on a page will raised a page view event when it's mounted, it also connects a reporter which writes events out to a target such as Tealium or the console.

```javascript
const PaginationWithTracking = addTracking(Pagination, {
  analyticsEvents: ["onNext", "onPrev"],
  attrs: {
    page: props => props.page,
    perPage: props => props.pageSize
  }
});

const ArticleListWithTracking = addTracking(
  ArticleList,
  {
    trackChildViews: {
      id: "id",
      listPath: "articles.list",
      attrs: {
        index: props => props.index,
        progress: props => _round((props.index + 1) / props.total * 100, 2)
      }
    }
  }
);

const AuthorProfileProfileWithTracking = (props) => (
  <View>
    <PaginationWithTracking props={...} />
    <ArticleListWithTracking props={...} />
  </View>
);

const AuthProfileWithTrackingContext = addTrackingContext(
  AuthorProfileWithTracking,
  {
    attrs: {
      name: props => props.data.name,
      siteArea: "profile",
      pageType: "listing"
    }
  }
);

<AuthProfileWithTrackingContext {...props} analyticsStream={reporter} />
```


toboid:
Couple of thoughts:

- The complexity around the base-tracking, tracking and tracking-web originates from the fact that AuthorProfileContent is different for native and web. If we could unify this it would make the tracking implementation cleaner as base-tracking, tracking and tracking-web could be merged. However ScrollView doesn't tell us when components are in view (we would have to try and calculate based on offset) and web doesn't ListView/FlatList.
Alternatively my suggestion is that park child tracking (and therefore scroll depth) for now whilst we think about this more. This would allow us to get started without the class hierarchy.

- As noted by @craigbilner the parameters to addTracking are a little complicated. An alternative could be to split these responsibilities out into separate components e.g. `addViewTracking()`, `addEventTracking()` and `addChildViewTracking()`. I prefer this as they could still be composed if we need more than 1 type of tracking on a component. 

What do you think @craigbilner @trevorah?

trevorah:
Yeah, `trackChildViews` is going to be useless until web uses a scroll view. Leave it for now.

The tricky bit with splitting out the tracking into `trackX(trackY(trackZ(MyComponent)))` is the `attrs`. All of those functions will need the mapping. Might be easier if `attrs` is a function that can be passed in: 
```
const attrs = ({ page, pageSize}) => ({ page, perPage: pageSize })
```

toboid:
I like the idea of attrs being a function, that would be easier to work with. If we split the tracking into separate components then it would still need to be passed to each but that's probably ok (though I suspect most components would only need one type of tracking).

```javascript
const getAttrs = ({ page, pageSize}) => ({ page, perPage: pageSize, foo: "bar" });
trackX(trackY(MyComponent, getAttrs), getAttrs);
```
This would also mean that different attrs could be applied to different types of tracking if required.

I think its preferable to be able to apply attrs with a context or tracking as you might have specific attributes that are only relevant to the wrapped component (e.g. `perPage` for pagination) and would prefer to not have to arbitrarily wrap with the a context to apply them (hope that makes sense).

PR #232, opened 2017-09-21
Feat: inline rendering of the ads coming from the api

thatportugueseguy:
captura de ecra 2017-09-21 as 15 20 34


packages/markup/markup.stories.js
```diff
@@ -32,6 +35,25 @@ const Centered = ({ children }) => (
 storiesOf("Markup", module)
   .addDecorator(story => <Centered>{story()}</Centered>)
   .add("Multiple paragraphs", () => <View>{renderTrees(multiParagraph)}</View>)
+  .add("Multiple paragraphs with ads", () => {
+    if (Platform.OS === "web") {
```
sericaia: markup.stories.web.js?

sericaia:
Remove yarn.lock's files :)

packages/markup/markup.stories.js
thatportugueseguy: we didn't feel that we had the need for a whole file, just because of this. It would be more important if it was a test, i'd say. it's just a story and one usage of the Platform api

thatportugueseguy:
we made changes to the components dependencies, they're needed (the yarn.lock files)

PR #233, opened 2017-09-22
Gradle clean up and studio 3 changes

Szymczak:
We would like to move to android studio 3 and we need to update the gradle files in the project. Also we have done some clean up on those files as some dependencies were using .+ which is dangerous.

### DIFF
--- a/packages/brightcove-video/android/build.gradle
+++ b/packages/brightcove-video/android/build.gradle
@@ -1,20 +1,17 @@
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    compileSdkVersion 26
+    buildToolsVersion "26.0.1"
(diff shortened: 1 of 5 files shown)


packages/brightcove-video/android/build.gradle
```diff
@@ -1,20 +1,17 @@
...
 android {
-    compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    compileSdkVersion 26
+    buildToolsVersion "26.0.1"
 
     defaultConfig {
         minSdkVersion 16
-        targetSdkVersion 22
-        versionCode 1
```
CoreFloDev: You sure this is really not required?

Szymczak: I think the manifest merger will just always use the one thats specified on the app project so we should be fine

PR #234, opened 2017-09-22
Feat/default article replat 304

killanaca:


packages/article/article-header.js
```diff
@@ -0,0 +1,56 @@
...
+  NewArticleFlag,
+  SponsoredArticleFlag,
+  UpdatedArticleFlag,
+  ExclusiveArticleFlag
+} from "@times-components/article-flag";
+import ArticleLabel from "@times-components/article-label";
+import { Text, View } from "react-native";
+import styles from "./article-style";
+
+const flagsMapping = new Map([
```
thatportugueseguy: I think this should be an object, rather than a Map,  as we don't need to iterate the flagsMapping.

packages/article/article-style.js
```diff
@@ -0,0 +1,150 @@
...
+    lineHeight: 26,
+    marginBottom: 20,
+    fontSize: Platform.OS === "android" ? 16 : 17,
+    color: "#333333"
+  }
+});
+
+const styles = {};
+
+if (Platform.OS !== "web") {
```
thatportugueseguy: Maybe it would be clearer with a .web file a native one? Or maybe trying to have a single style object and a helper function, because there are so many overlapping styles?
```js
const fnName = (androidStyle, webAndIOS) => Platform.OS === "android" ? androidStyle : webAndIOS;
```

(needs a lovely, and expressive, but preferably short name, which I can't come up with right now)

packages/article/article.js
```diff
@@ -1,11 +1,143 @@
...
+      "standfirst",
+      "flags"
+    ]);
+
+    const articleMidContainerData = pick(articleData, [
+      "publicationName",
+      "publishedTime",
+      "byline"
+    ]);
+    return Array.prototype.concat(
```
thatportugueseguy: Maybe there's no need to call the array prototype here?
```js
      [
        { type: "leadAsset", data: leadAssetData },
        { type: "header", data: articleHeaderData },
        { type: "middleContainer", data: articleMidContainerData }
      ].concat(
        articleData.content.map(i => ({
          type: "article_body_row",
          data: i
        }))
    );
```
Or maybe moving stuff into variables: 

```js
return headerAssetsData.concat(articleBodyRows);
```

packages/article/article.js
```diff
@@ -1,11 +1,143 @@
...
+        pageSize={listViewPageSize}
+        enableEmptySections
+      />
+    );
+  }
+}
+
+ArticlePage.propTypes = {
+  data: PropTypes.shape({
+    article: PropTypes.object,
```
thatportugueseguy: should this type be a bit more strict?

packages/article/article.js
```diff
@@ -1,11 +1,143 @@
...
+import ArticleMeta from "./article-meta";
 
-export default Article;
+const ds = new ListView.DataSource({ rowHasChanged: (r1, r2) => r1 !== r2 });
+const listViewPageSize = 1;
+const listViewSize = 10;
+const listViewScrollRenderAheadDistance = 10;
+
+class ArticlePage extends React.Component {
+  static prepareDataForListView(props) {
```
thatportugueseguy: I feel that these static methods should actually be private methods or more general external helper functions (if possible).

packages/article/article.js
```diff
@@ -1,11 +1,143 @@
...
+        { type: "middleContainer", data: articleMidContainerData }
+      ],
+      articleData.content.map(i => ({
+        type: "article_body_row",
+        data: i
+      }))
+    );
+  }
+
+  static renderRow(rowData) {
```
thatportugueseguy: Maybe the logic could be extracted into an object that calls some stateless components returning the right type of rowData?

I feel that would look cleaner

packages/article/article-header.js
killanaca: why we can't use Map if we need a mapping logic? there's some overhead?

thatportugueseguy: you can use the Map without any significant overhead, but Map is meant more like a key/value store that would need to be iterated.

Given that we're not iterating anything, just accessing properties values, a plain object feels more in order, but i'm good with this being a Map if you want to keep it that way.

packages/article/article.js
```diff
@@ -1,11 +1,127 @@
...
+        </View>
+      );
+    }
+
+    return null;
+  }
+  constructor(props) {
+    super(props);
+    this.state = {
+      dataSource: ds.cloneWithRows({})
```
sericaia: I guess it can be null in the beggining, no?

packages/article/article-header.js
```diff
@@ -0,0 +1,50 @@
...
+  ["NEW", <NewArticleFlag />],
+  ["UPDATED", <UpdatedArticleFlag />],
+  ["EXCLUSIVE", <ExclusiveArticleFlag />],
+  ["SPONSORED", <SponsoredArticleFlag />]
+]);
+const ArticleHeader = ({ label, title, standfirst, flags }) => (
+  <View style={[styles.articleMainContentRow]}>
+    {label ? (
```
sericaia: Having this syntactic sugar here makes the code less readable. Can you create auxiliary functions?

sericaia: (btw, nice mapping ;) )

sericaia:
remove `yarn.lock` from commit ;)

packages/article/article-header.js
killanaca: done

packages/article/article-style.js
nehasri89: changed the way the styles are done @thatportugueseguy please revew

packages/article/article.js
killanaca: as discussed with @sericaia lets go with the proposed solution with the empty object

packages/article/article-header.js
```diff
@@ -31,15 +45,7 @@ const ArticleHeader = ({ label, title, standfirst, flags }) => (
       <ArticleHeadline title={title} style={styles.articleHeadLineText} />
     </View>
     {standfirst ? <Text style={[styles.standFirst]}>{standfirst}</Text> : null}
```
sericaia: i think all of them should be in a function: standfirst and label also :)

The idea is to change this:
{ x ? : x : null }

to a function 
```
function getX(x) {
  if (!x) {
    return null
  }
  return <MyComponent x={x} />
}```

Comment (killanaca) [reply packages/article/article-header.js]: done

File: packages/article/article-header.js [hunk 10]
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const ArticleHeader = ({ label, title, standfirst, flags }) => (
+  <View style={[styles.articleMainContentRow]}>
+    {getLabel(label)}
+    <View style={[styles.articleHeadline]}>
+      <ArticleHeadline title={title} style={styles.articleHeadLineText} />
+    </View>
+    {getStandfirst(standfirst)}
+    {getFlags(flags)}
```
thatportugueseguy: why the word get, rather than, for instance, render? we're not "getting/fetching" anything.

packages/article/article-header.js
```diff
@@ -0,0 +1,66 @@
...
+          {flagsMapping.get(flag)}
+        </View>
+      ))}
+    </View>
+  );
+};
+
+const getStandfirst = standfirst => {
+  if (!standfirst) return null;
+  return <Text style={[styles.standFirst]}>{standfirst}</Text>;
```
thatportugueseguy: does the text need a view around it?

packages/article/article-style.js
thatportugueseguy: I see it, much better! don't forget to delete this file :)

packages/article/article.js
```diff
@@ -1,11 +1,134 @@
...
+  componentWillReceiveProps(nextProps) {
+    if (!nextProps.isLoading) {
+      this.setState({
+        dataSource: ds.cloneWithRows(listViewDataHelper(nextProps.data.article))
+      });
+    }
+  }
+
+  render() {
+    if (this.props.isLoading) {
```
craigbilner: do we have an issue/ticket to replace this with skeletons?

packages/article/data-helper.js
```diff
@@ -0,0 +1,30 @@
+import pick from "lodash.pick";
+
+const prepareDataForListView = articleData => {
+  const leadAssetData = pick(articleData, ["leadAsset"]);
```
craigbilner: seems gratuitous?

packages/article/data-helper.js
```diff
@@ -0,0 +1,30 @@
...
+    "byline"
+  ]);
+
+  return [
+    { type: "leadAsset", data: leadAssetData },
+    { type: "header", data: articleHeaderData },
+    { type: "middleContainer", data: articleMidContainerData }
+  ].concat(
+    articleData.content.map(i => ({
+      type: "article_body_row",
```
craigbilner: see naming - ideally this would have failed a lint test...

packages/article/package.json
```diff
@@ -60,16 +53,31 @@
...
+    "react": "15.4.2",
+    "react-dom": "15.4.2",
+    "react-native": "0.42.3",
     "react-native-web": "0.0.119"
   },
   "peerDependencies": {
-    "react": ">=15",
-    "react-dom": ">=15",
-    "react-native": ">=0.42"
+    "@storybook/react-native": ">=3",
```
craigbilner: if this is required this should be repo wide?

packages/article/styles/article-header-style.js
```diff
@@ -0,0 +1,47 @@
...
+    lineHeight: 26,
+    fontFamily: standFirstFontFamily,
+    color: "#333333",
+    paddingBottom: 9
+  }
+});
+
+const nativeStyles = StyleSheet.create({
+  articleHeadLineText: {
+    fontSize: Platform.OS === "android" ? 28 : 30,
```
craigbilner: the platform specific stuff is getting a bit messy here?

packages/article/article-header.js
```diff
@@ -0,0 +1,64 @@
...
+} from "@times-components/article-flag";
+import ArticleLabel from "@times-components/article-label";
+import { Text, View } from "react-native";
+import styles from "./styles/article-header-style";
+import {
+  articleHeaderPropTypes,
+  articleHeaderDefaultPropTypes
+} from "./article-header.proptypes";
+
+const flagsMapping = new Map([
```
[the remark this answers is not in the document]
nehasri89: 👍

packages/article/data-helper.js
killanaca: fixed

packages/article/package.json
killanaca: fixed

packages/article/data-helper.js
killanaca: fixed

PR #237, opened 2017-09-22
fix: prevent 'undefined' being used in pagination display name

toboid:
Fixes "undefined" being used in displayName of pagination state hoc in cases where JSX doesn't set the displayName of wrapped component.

Before:
image

After:
image


craigbilner:
https://www.npmjs.com/package/react-display-name looks suitable?

toboid:
@craigbilner updated

PR #240, opened 2017-09-25
feat: adds option to reset on finish

dickie81:
Adds option to reset video to the launch poster on finish. This will save memory when the player is disposed of.


packages/brightcove-video/brightcove-video.js
```diff
@@ -41,6 +44,15 @@ class BrightcoveVideo extends Component {
     this.setState({ isLaunched: false });
   }
 
+  handleFinish() {
+    if (this.props.resetOnFinish) {
+      // calling syncronously here inteferes with player and causes errors to be thrown
+      setTimeout(this.reset, 0);
```
thatportugueseguy: Often times, asyncronous flow on react components can be sorted using the setState callback. Maybe this is one of those times?

packages/brightcove-video/__tests__/brightcove-video.test.js
```diff
@@ -94,6 +94,32 @@ it("will reset properly", () => {
...
+  const rootInstance = root.getInstance();
+
+  rootInstance.play();
+
+  rootInstance.handleFinish();
+
+  setTimeout(() => {
+    expect(root.toJSON()).toMatchSnapshot();
+    done();
+  }, 20);
```
craigbilner: what's this arbitary time?

packages/brightcove-video/brightcove-video.js
craigbilner: agree, and there's no concept of setting state synchronously, so you're just deferring making the asynchronous call which may lead to odd behaviour/bugs

dickie81: Tried the setState callback with no luck. Agree looks very ugly but not sure how to progress. We are working with the Brightcove SDK here and its querks.

PR #241, opened 2017-09-25
chore: add bitrise.yml to repo

rjanjua:


package.json
```diff
@@ -26,7 +26,7 @@
     "storybook:publish": "./lib/publish_storybook.sh",
     "prestorybook-native": "node ./node_modules/.bin/rnstl && node ./lib/fetch-fonts",
     "storybook-native": "storybook -c .storybook.native start -p 7007",
-    "compile-fructose-components": "rnstl --searchDir ./packages --pattern './*/*.fructose.js' --outputFile ./.fructose/components.js",
+    "compile-fructose-components": "node ./node_modules/.bin/rnstl --searchDir ./packages --pattern './*/*.fructose.js' --outputFile ./.fructose/components.js",
```
thatportugueseguy: this seems unnecessary and it's not common at all.. was it erroring because it wasn't called with node? I've looked at the file attributes, it's got execute permission (x), so calling the binary should suffice..

rjanjua: yeah, that's what I thought too. I changed it and it worked - though it had been working previously until earlier today. 

I noticed that prestorybook-native also does the same, so at least that means it's consistent. :/

thatportugueseguy: @rjanjua Do you want to keep it like that? I don't think that either one should be using that "pattern", but might be specific to that file.

And it shouldn't stop working out of the blue..

PR #246, opened 2017-09-25
chore: moved ad rendering into the article

trevorah:
The markup component is meant to only deal with text formatting by default.

This moves the ad renderer into a set of renderers that can be used by the article.


craigbilner:
@JGAntunes @trevorah I think this can wait until we've delivered everything else?

trevorah:
@JGAntunes @craigbilner sure, closing for now.

PR #250, opened 2017-09-27
Refactor article package json 

nehasri89:


packages/article/package.json
```diff
@@ -60,15 +60,12 @@
...
-    "react": "15.4.2",
-    "react-dom": "15.4.2",
-    "react-native": "0.42.3",
     "react-native-web": "0.0.119"
   },
   "peerDependencies": {
-    "react": "15.4.x",
-    "react-dom": "15.4.x",
-    "react-native": "0.42.x"
+    "react": ">=15",
```
trevorah: Are we not limiting this due to flaky react 16 support?

trevorah: (craig just explained it to me)

PR #251, opened 2017-09-27
Drop old text node support

trevorah:
Easier to do this sooner rather than later. This also means that the markup component (and anything that uses it) will be easier to type in the future...

Odd bits:
- i've simplified the markup fixtures whilst I was in there. The json structure for those fixtures no longer exports a "fixture" object.
- The only snapshot change is for some props in a subcomponent that passes the tree onto the markup component.

### DIFF
--- a/packages/article/fixtures/article-no-flags-no-standfirst.json
+++ b/packages/article/fixtures/article-no-flags-no-standfirst.json
@@ -13,11 +13,10 @@
             "children": [
               {
                 "name": "text",
-                "children": [
-                  {
-                    "text": "Rick Broadbent"
-                  }
-                ]
+                "attributes": {
+                  "value": "Rick Broadbent"
+                },
+                "children": []
               }
             ]
           }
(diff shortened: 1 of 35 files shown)


packages/article/fixtures/article-no-flags-no-standfirst.json
```diff
@@ -31,33 +30,30 @@
...
-                        "text": "last year that rote learning has “a lot to be said against it”, but added: “As the phrase ‘learning by heart’ implies, this sort of remembering places the poem at a central place in ourselves and makes it precious.”"
-                      }
-                    ]
+                    "attributes": {
+                      "value": "last year that rote learning has “a lot to be said against it”, but added: “As the phrase ‘learning by heart’ implies, this sort of remembering places the poem at a central place in ourselves and makes it precious.”"
+                    },
+                    "children": []
                   },
                   {
                     "name": "break",
```
craigbilner: this nesting should be fixed now too

PR #256, opened 2017-09-28
Track views

toboid:
`TrackView` and `TrackingContext` higher order components. 
Features:
* Raise event when tracked component is mounted
* Decoration of events raised from descendent components via contexts. Contexts can be nested.
* Raise page view event when root context is mounted


toboid:
WIP - should be possible to do with HOCs and use standard composition

packages/tracking/tracking-context.js
```diff
@@ -0,0 +1,96 @@
...
+) => {
+  const componentName = getDisplayName(WrappedComponent);
+
+  class WithTrackingContext extends Component {
+    getChildContext() {
+      const self = this;
+
+      return {
+        tracking: {
+          analytics(e) {
```
craigbilner: option: deconstruct

packages/tracking/track-view.js
```diff
@@ -0,0 +1,37 @@
...
+
+  class WithTrackView extends Component {
+    componentDidMount() {
+      if (!this.context || !this.context.tracking) {
+        return;
+      }
+
+      this.context.tracking.analytics({
+        component: trackingName || componentName,
+        action: "Viewed",
```
craigbilner: `Rendered`? I think `Viewed` is misleading?

packages/tracking/tracking-context.js
```diff
@@ -0,0 +1,96 @@
...
+              component: e.component,
+              action: e.action,
+              attrs: {
+                ...resolveAttrs(getAttrs, self.props),
+                ...e.attrs
+              }
+            };
+
+            if (e.object || trackingObject) {
+              trackingEvent.object = e.object || trackingObject;
```
craigbilner: what happens if I have no object?

packages/tracking/tracking-context.js
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            if (self.isRootTrackingContext()) {
+              trackingEvent.attrs.eventTime = new Date().toISOString();
```
craigbilner: probably ok for now but if we start getting many events might want a high res timestamp?

packages/tracking/tracking-context.js
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            self.analyticsStream(trackingEvent);
+          }
+        }
+      };
+    }
+
+    componentWillMount() {
+      if (this.isRootTrackingContext()) {
```
craigbilner: you get `nextContext` in three of the lifecycle methods, is the earliest we know that this doesn't exist?

packages/tracking/tracking-context.js
```diff
@@ -0,0 +1,96 @@
...
+        if (!this.props.analyticsStream) {
+          throw new TypeError(
+            "Missing prop analyticsStream of WithTrackingContext"
+          );
+        }
+      }
+    }
+
+    isRootTrackingContext() {
+      return !_get(this.context, "tracking");
```
craigbilner: perhaps a little overkill to use `lodash` here?

packages/tracking/__tests__/resolve-attrs.test.js
```diff
@@ -0,0 +1,25 @@
...
+    expect(attrs).toEqual({});
+  });
+
+  it("returns empty object if getAttrs isn't a function", () => {
+    const attrs = resolveAttrs("not a function");
+
+    expect(attrs).toEqual({});
+  });
+
+  it("forwards props amd eventArgs to getAttrs", () => {
```
craigbilner: typo

craigbilner:
@toboid stories are the main missing piece. They can act as living docs within Tracking

packages/tracking/tracking-context.js
toboid: Within a context that is not the root one, it's ok to not have object as it will added by the root context.
Within the root context `trackingObject` will always have a value as it's validated.

packages/tracking/tracking-context.js
toboid: sure, since high res isn't yet supported on native, would prefer to not have separate logic just for this would could come back to it if we're happy to proceed with this

toboid:
@trevorah rebased and tests re-run

PR #257, opened 2017-09-28
fix: Image component for web

foliveira:
This PR supersedes of #206 

Web image component with fixed stories to work both on Native and Web.

_Right now the best approach would be to wrap `Image`s in a `View` and then size them directly instead of doing so on the `View`._


trevorah:
Closing in favour of https://github.com/newsuk/times-components/pull/272

PR #258, opened 2017-09-28
feat: render ads in the article article body

JGAntunes:
Created a simple wrapper component to solve the issue. Maybe `article-content` isn't the best name, so, open to suggestions :smile:.

image


craigbilner:
@JGAntunes what is "the issue" and why is web different to native?

JGAntunes:
@craigbilner the issue? The lack of ads in the article page for all the platforms  :smile:. The difference between web and native comes from the different approaches each Ad component follows. The native one loads everything it needs inside the webview, the web one requires the `ad-composer` to control all the ads in the page.

craigbilner:
@JGAntunes `ArticleListView` which is conditionally wrapped with the `AdComposer` HOC for web otherwise returns a `ListView` feels a bit cleaner and less ambiguous than `ArticleContent`?

PR #268, opened 2017-10-03
feat: use funcs as children

craigbilner:
chore: update lock
chore: update props
chore: use mocked data for stories and testing

Change providers to call a child function with the Apollo props such as `isLoading` and `error` as well as the given props to the provider and the result of the GraphQL query.

e.g.

```
<ApolloProvider
  <SomeProvider someConfig={1}>
    {({ isLoading, error, someData, someConfig }) => /* render component here */}
  </SomeProvider>
</ApolloProvider>
```

This gives a slightly nicer API in that any child can be placed inside and consume whatever they desire and conditionally render themselves rather than a traditional HOC where the behaviour is baked in.

Slightly bigger change was to move away from using real GraphQL endpoints and remove custom `react-apollo` mocking in favour of test utils from Apollo. These are completely undocumented and were used in a best effort reading the code but they may/should mature with age and we can contribute back if we want anything from them. A `MockedProvider` is a bit neater regarding the tests and provides out of the box stories especially for the public storybook. A small downside is that if you use `__typename` you have to litter it everywhere in your mocks atm _and_ you have to provide one mock for every GraphQL call.

Upgraded storybook to remove the multiple components exported warning which was a `react-docgen` bug via Babel


packages/author-profile/package.json
```diff
@@ -71,10 +71,12 @@
     "@times-components/card": "0.6.5",
     "@times-components/link": "0.8.0",
     "@times-components/pagination": "0.6.4",
+    "@times-components/provider": "0.8.5",
     "jest-cli": "20.0.4",
     "lodash.get": "4.4.2",
     "lodash.merge": "4.6.0",
     "prop-types": "15.5.10",
+    "react-apollo": "1.4.14",
```
sericaia: replace by 1.4.16

trevorah:
I'll merge master into this after https://github.com/newsuk/times-components/pull/272 merges.

All looks good, I'll just fix the author profile mocks to handle the last article list page.

PR #271, opened 2017-10-04
fix markup fixtures imports

thatportugueseguy:
There was a change to the fixtures on the `d14cb15` commit that made the imports wrong.

Just a simple fix.


trevorah:
@thatportugueseguy just saw this, thanks for fixing it!

PR #281, opened 2017-10-09
Feat/android-fullscreen

dickie81:
Fixes android brightcove button state when fullscreen pressed.


dickie81:
closing due to multitude of hacks in this approach

PR #282, opened 2017-10-09
fix: article summary font

rjanjua:
Here are a list of screenshots that demonstrate what the fix has changed (designs [here](https://app.zeplin.io/project/59a6a2c092fb8c2b6944c504) for comparison).

ios old: 
ios-original
ios new:
ios-fix

desktop old: 
desktop-original

desktop new:
desktop-fix

And to show that android is the same (as it was already using the correct font):
old:
android-original
new:
android-fix


rjanjua:
closing this PR because there's a bigger issue around fonts that needs to be addressed #284

PR #283, opened 2017-10-10
feat: article with provider

sericaia:
This PR adds the provider to the article: article was adapted and new stories added.

Questions/Notes:
1. Should all fixtures have typenames?
2. Is empty state required here?
3. React 16 introduces componentDidCatch.. we should add a ticket to add that and remove error components inside article and author profile. What do you think?
4. Loading is also repeated everywhere, can we have a separated component?


killanaca:
LGTM

craigbilner:
@sericaia 

1) If you're using a union type and therefore need to provide a `typename` you will have to litter them with `typenames` for now. Otherwise, you can just disable `typenames` with the `removeTypename` flag

2) I don't believe there'd be a concept of an empty article?

3) `componentDidCatch` isn't for the errors returned from a provider? And even if we add error boundaries typically you'd still present some form of error state. However, yes we can potentially add these in the future

4) If we introduce "skeleton" loading states, different components will implement their own representation so it may be unlikely we'd have a "loading component" beyond the current RN implementation?

craigbilner:
@killanaca - can you review my changes? ta

killanaca:
@craigbilner 👍

PR #285, opened 2017-10-11
fix: image caption on article when primary image

sericaia:
image

image


nehasri89:
I thought we wanted to wait for Jake to review and then change the values @sericaia

craigbilner:
will be picked up at the end of the sprint

sericaia:
Reopened as @killanaca checked that primary image padding is still an issue

sericaia:
will create a separate PR

PR #289, opened 2017-10-12
chore: upgraded yarn to 1.2.0

damnhipster:
[Changelog](https://github.com/yarnpkg/yarn/releases)

Ran locally, looks like it works fine.


killanaca:
should we update the `bitrise.yml` file also?

PR #292, opened 2017-10-12
feat/brightcove-android-fullscreen-events

dickie81:
Fires fullscreen events to js layer to handle fullscreen layout change.

Also improves rotation handling by firing a configuration change event on to the player view:
`mPlayerView.getEventEmitter().emit(EventType.CONFIGURATION_CHANGED);`

This forces the player controls to resize / position correctly. Also used on enter / exit fullscreen for same reason.

Another important change to note; the brightcove player view has to be instantiated with the Activity context & not a ThemedReactContext for the fullscreen button to accept any clicks.

### DIFF
--- a/android/app/src/main/java/com/storybooknative/MainApplication.java
+++ b/android/app/src/main/java/com/storybooknative/MainApplication.java
@@ -3,8 +3,6 @@ package com.storybooknative;
 import android.app.Application;
 
 import com.facebook.react.ReactApplication;
-import com.facebook.react.ReactRootView;
-import com.facebook.react.views.text.ReactTextView;
 import com.facebook.stetho.Stetho;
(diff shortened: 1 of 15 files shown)


trevorah:
Cool, will iOS work the same way?

android/app/src/main/java/com/storybooknative/MainApplication.java
```diff
@@ -3,6 +3,9 @@
 import android.app.Application;
 
 import com.facebook.react.ReactApplication;
+import com.facebook.react.ReactRootView;
+import com.facebook.react.views.text.ReactTextView;
```
CoreFloDev: I think some import are unused here

dickie81:
iOS seems to just work, no need for extra work on react side

PR #293, opened 2017-10-12
fix: update coverage patterns

craigbilner:
Coverage is currently running on every `js` file, narrow to only the link package in order to restore sanity to the build

Also - the svg version were out of date with the other packages


craigbilner:
@killanaca what was special about `5.1.7`?

PR #301, opened 2017-10-16
feat: article summary joining and ellipses

damnhipster:


packages/article-summary/article-summary.js
```diff
@@ -38,12 +39,16 @@ const styles = {
 const ArticleSummary = props => {
   const { label, headline, text, date, publication } = props;
 
+  const teaser = text.pop();
```
craigbilner: ideally we wouldn't want to mutate props in our components which should be immutable

damnhipster:
@craigbilner I've added a story which contains richer content for the article summary (basically a TV Review).

Until line breaks are supported, I'm not sure it's worth writing test cases which assert the wrong markup, otherwise we're just duplicating the tests already written for the markup component.

On the first point, I feel like using CSS to override what a paragraph tag is meant to look like is more confusing than changing it to a span, even if it's slightly more performant on native.

craigbilner:
thanks @damnhipster. Not sure about the screenreading...they are actually new paragraphs, so the text is likely to be in the context of new paragraphs, we just present them conjoined for visual symmetry or the like...but it is an edge case as generally speaking article summary doesn't cross paragraphs

damnhipster:
- [ ] Add coverage for author profile

packages/article-summary/__tests__/article-summary-renderer.test.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-env jest */
+
+import "react-native";
```
craigbilner: do we need this level of testing? We can test the behaviour rather than the implementation?

packages/article-summary/__tests__/article-summary.test.js
```diff
@@ -4,16 +4,24 @@ import "react-native";
 import React from "react";
 import renderer from "react-test-renderer";
 import ArticleSummary from "../article-summary";
-import props from "../fixtures/article.json";
+import articleFixture from "../fixtures/article.json";
```
craigbilner: can we have web tests if we're going to do special web things?

packages/article-summary/article-summary-renderer.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { Text } from "react-native";
+
+export default {
+  paragraph(key, attributes, renderedChildren) {
+    return <Text key={key}>{renderedChildren} </Text>;
+  },
+  sentence(key, attributes, renderedChildren) {
```
craigbilner: raised a PR on TPA to address this issue

packages/article-summary/article-summary.js
```diff
@@ -35,15 +36,43 @@ const styles = {
   }
 };
 
+const format = {
```
craigbilner: without sentence I think this becomes a lot cleaner?

packages/article-summary/article-summary.js
```diff
@@ -35,15 +36,43 @@ const styles = {
...
+  teaser(item) {
+    return Object.assign({}, item, { name: "teaser" });
+  }
+};
+
+const summarise = text => {
+  let summary = [];
+  if (text.length) {
+    const initial = text.slice(0, text.length - 1);
+    const tail = text[text.length - 1];
```
craigbilner: typically we wouldn't call this `tail`, it would be `last` or something similar, `tail` is everything that is not `head`

packages/article-summary/__tests__/article-summary.test.js
damnhipster: We haven't added any special behaviour for web.

damnhipster: #332

craigbilner:
Raised the TV review stuff with editorial - ideally needs fixing before go live

PR #304, opened 2017-10-16
chore: fixed time on CI to Europe/London

damnhipster:
Accommodates for daylight savings when using recent articles for snapshots, where the articles contain dates since March 26th.

You can see the failing tests [here](https://travis-ci.org/newsuk/times-components/jobs/288511877).

Passing tests should appear [here](https://travis-ci.org/newsuk/times-components/builds/288576526).


damnhipster:
Rather than changing this in Travis, we should set the tests to use UTC.

PR #305, opened 2017-10-16
Add tracking to pagination component

toboid:
Adds event tracking to pagination control. No visual changes.

A couple of points to consider:
- Are we happy to export 2 versions of the component - one with and one without tracking. 
- If exporting with and without tracking, which should be the default export? In this PR I have made the untracked version the default as it then matches the file name. However we could argue that the tracked version should be the default so that developers don't forget to use the tracked version.
- If the tracked component is a named export, are we happy with the naming convention`<ComponentName>Tracked`?


trevorah:
🗣 from convo: default export will be tracked.

toboid:
Update to make tracked version default export

trevorah:
@toboid this looks cool, but the event after tapping "Previous Page" is a little bit confusing:
```
analytics-event: {
  component: "Pagination"
  action: "Pressed"
  attrs: {
    nextPage: 1
    direction: "back"
    eventTime: "2017-10-17T13:39:19.210Z"
  }
 object: "Story"
}
```
How about `direction` being "previous" or "next", and `nextPage` -> `destinationPage`?

toboid:
@trevorah yep that's better

PR #306, opened 2017-10-16
Add tracking to author twitter link

toboid:
Adds tracking to twitter link on author profile


toboid:
will update as per https://github.com/newsuk/times-components/pull/305 - default export will be tracked

toboid:
Updated to make tracked version default export

PR #310, opened 2017-10-17
feat/brightcove-android-direct-to-fullscreen

dickie81:
Adds new property for android to allow player to go directly to fullscreen.


packages/brightcove-video/android/src/main/java/uk/co/news/rntbrightcovevideo/BrightcovePlayerActivity.java
```diff
@@ -0,0 +1,39 @@
...
+
+        super.onCreate(savedInstanceState);
+
+        BrightcovePlayerView playerView = new BrightcovePlayerView(this);
+
+        playerView.initVideo(videoId, accountId, policyKey, true, true);
+
+        setContentView(playerView);
+    }
+}
```
trevorah: Seems legit. Any lifecycle events needed here? Does the video pause when it goes into the background?

packages/brightcove-video/README.md
```diff
@@ -66,6 +66,7 @@ Properties types are defined in `./brightcove-video.proptypes.js`
 | `onProgress` | function | Handles progress events
 | `onFinish` | function | Handles video finish events
 | `autoplay` | boolean | Should the video autoplay? (default false)
+| `directToFullscreen` | boolean | Should the video play directly in fullscreen? (default false) (Android only)
```
CoreFloDev: Can we make it default to true?

dickie81:
Some known issues with this approach:

- Events no longer being passed out
- Fullscreen button can't be hidden

packages/brightcove-video/android/src/main/java/uk/co/news/rntbrightcovevideo/BrightcovePlayerActivity.java
dickie81: video appears to pause when backgrounded

packages/brightcove-video/README.md
dickie81: I think it would be better to wait for iOS & web support otherwise the default behaviour on different platforms will be very different

PR #314, opened 2017-10-19
fix scrollRenderAheadDistance

sericaia:
I changed `scrollRenderAheadDistance` from 10 to 500 (estimate of a device pixels height). The idea is to make loading ahead faster.


craigbilner:
deprecated in favour of `FlatList`

PR #324, opened 2017-10-20
feat: article loading state

fampinheiro:


trevorah:
😍

android/app/build.gradle
```diff
@@ -144,6 +144,7 @@ repositories {
 }
 
 dependencies {
+    compile project(':react-native-linear-gradient')
```
CoreFloDev: implementation please

fampinheiro: What do you mean by implementation?

CoreFloDev: compile is obsolete in gradle 4.0, you need to change it to implementation instead. Check for more info https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html#new_configurations

craigbilner:
@fampinheiro looks good.

Can we:

- [ ] fix the paging `n - zero`
- [ ] add avatar to stories
- [ ] remove the redundant empty state story

packages/author-profile/author-profile-author-head.js
```diff
@@ -0,0 +1,57 @@
+import AuthorHead from "@times-components/author-head";
+import Gradient from "@times-components/gradient";
+import React from "react";
+import { StyleSheet, View } from "react-native";
+
+const styles = StyleSheet.create({
+  container: {
+    alignItems: "center",
+    backgroundColor: "#F9F8F3",
+    flexDirection: "column",
```
craigbilner: default?

packages/author-profile/author-profile-content.js
```diff
@@ -1,116 +1,115 @@
...
-            name={name}
-            bio={biography}
-            uri={uri}
-            title={jobTitle}
-            twitter={twitter}
-            onTwitterLinkPress={onTwitterLinkPress}
+    const data = articlesLoading
+      ? Array(pageSize)
+          .fill()
+          .map((_, id) => ({
```
craigbilner: minor: but the Haskellesque placeholder if used as a pattern can be confused for `lodash`?

packages/author-profile/author-profile-author-head.js
fampinheiro: It was here when we had conditional styles. Not the case anymore (removing).

packages/author-profile/author-profile-content.js
fampinheiro: Can change to a meaningful name. 👍

fampinheiro:
@craigbilner regarding the pagination issue.
We didn't reach a conclusion if the pagination should be attached to the author-head query or a separated one, or try to hack something with state on the `author-profile` component.

craigbilner:
@fampinheiro thought we were going to look at the Apollo update function which will retain the given page size?

fampinheiro:
I read about the [update](http://dev.apollodata.com/react/cache-updates.html#directAccess) but this is used with a mutation.

PR #326, opened 2017-10-23
chore: upgrade lint

craigbilner:
Eslint is now prehistoric in JS terms, before we drift too far away, this is the upgrade.

Notable changes:

## a bit buggy

I'll try and recreate and raise PRs for these, but the workarounds are straightforward

[react/no-typos](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-typos.md) - kind of annoying referencing component `propTypes` now. You have to declare the reference separately...made a bit more consistent with `react-style-proptype`
[jsx-a11y/anchor-is-valid](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md) - a bit of a bug with the `Link` component in that it enforces `href` but ours is cross-platform so `href` isn't really appropriate, just renamed for distinction
[react/no-unused-state](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-state.md) - a small bug with setting state with a function (a la brightcove) so had to ignore there
[react/prop-types](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md) - complains with `Object.assign` but `object spread` is fine...

## quick fix

* [react/default-props-match-prop-types](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/default-props-match-prop-types.md) - `defined for isRequired propType` | `has no corresponding propTypes declaration`. I've largely just said if a default was provided then it probably shouldn't be required, please review

## simple change

* [prefer-promise-reject-errors](https://eslint.org/docs/rules/prefer-promise-reject-errors)
* [prefer-destructuring](https://eslint.org/docs/rules/prefer-destructuring)
* [react/jsx-curly-brace-presence](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md) - @fampinheiro


packages/ad/__tests__/pbjs-manager.test.js
```diff
@@ -58,7 +58,7 @@ describe("PrebidManager", () => {
       pbjsManager.loadScript();
       const prebid = pbjsManager.pbjs;
       pbjsManager.setConfig();
-      adserverTargeting = prebid.bidderSettings.standard.adserverTargeting;
+      ({ bidderSettings: { standard: { adserverTargeting } } } = prebid);
```
damnhipster: Not sure if this is better or worse lol

craigbilner: haha - we just follow the rules to avoid debate...

packages/author-profile/author-profile-item.js
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import get from "lodash.get";
 import { View } from "react-native";
 import Card from "@times-components/card";
-import Link from "@times-components/link";
+import LinkXP from "@times-components/link";
```
damnhipster: Maybe we should disable the href assert.

The following rule should work

```json
{
    "rules": {
        "jsx-a11y/anchor-is-valid": [ "error", {
            "components": [ "Link" ],
            "specialLink": [ "hrefLeft", "hrefRight" ],
            "aspects": [ "invalidHref", "preferButton" ]
          }]
    }
}
```

packages/brightcove-video/brightcove-player.web.js
```diff
@@ -51,7 +52,6 @@ class BrightcoveVideo extends Component {
       errors: [].concat(BrightcoveVideo.globalErrors),
       isPlaying: "paused",
       isFinished: false,
-      fullscreen: false,
```
damnhipster: Is this necessary?

packages/author-profile/author-profile-item.js
craigbilner: I looked at that...but it felt like a more dangerous work around in terms of global `Link` fudging, really the only mandatory element should be an `anchor`

trevorah:
Wait, whats up with the Link linting rules? Does it mean that anything named `Link` has to have a `href` prop?

packages/brightcove-video/brightcove-player.web.js
damnhipster: Nevermind

packages/ad/ad.js
```diff
@@ -1,9 +1,9 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import { View, WebView, Dimensions, Linking, StyleSheet } from "react-native";
+import StylePropTypes from "react-style-proptype";
```
fampinheiro: I would suggest that we use `ViewPropTypes.style` instead of an external package.

craigbilner:
@trevorah atm, yes

packages/ad/ad.js
craigbilner: happy to change back as long as it's consistent - bear in mind that prop-types will be stripped and are only for dev time.

The pain points are around `View` having `ViewPropTypes` but `Text` doesn't have the equivalent or `WebView` for example, so consistency falls down and the extra reference gets a bit messy

damnhipster:
Should we apply the jsx-a11y/anchor-is-valid rule to only web components?

craigbilner:
@damnhipster @trevorah I've overridden the rule instead and enforced it for our own component. A developer will have to update the `eslintrc` when they use the `Link` component

PR #327, opened 2017-10-23
fix: style warning on android. Styles are not used in WebView

sericaia:
Fix these warnings:

image

image


CoreFloDev:
😢 (about the component not the PR)

PR #328, opened 2017-10-23
fix: responsive author avatar

fampinheiro:
## desktop

image

## tablet

image

## mobile

image


packages/author-head/__tests__/author-head.test.js
```diff
@@ -38,6 +38,7 @@ it("tracks twitter link interactions", () => {
 
   component
     .dive()
+    .dive()
```
damnhipster: dive²

trevorah:
I'm not the biggest fan of having a responsive component like this.

I'd rather have:
```js
withWindowBreakpoints(Component, { size: 768, component: ComponentA },  { size: 1024, component: ComponentB })
```

packages/author-head/__tests__/author-head.test.js
fampinheiro: With the tracking HoC we need to dive deeper to access the links

fampinheiro:
@trevorah, in that case, we need different components for different screen sizes, right?

craigbilner:
needs a rethink

PR #333, opened 2017-10-25
feat: gradient package

fampinheiro:
Like the name says it creates a gradient 👍 
This is built on top of [react-native-linear-gradient](https://github.com/react-native-community/react-native-linear-gradient) to add a property to define the degrees of the gradient.

image

### DIFF
--- a/packages/gradient/gradient.js
+++ b/packages/gradient/gradient.js
@@ -4,6 +4,7 @@ import { StyleSheet, View } from "react-native";
 import LinearGradient from "react-native-linear-gradient";
 import StylePropTypes from "react-style-proptype";
 
+const { supportingArrays } = StylePropTypes;
 const styles = StyleSheet.create({
   container: {
     flex: 1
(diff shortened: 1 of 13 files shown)


packages/gradient/gradient.js
```diff
@@ -0,0 +1,66 @@
+import PropTypes from "prop-types";
+import React from "react";
+import { StyleSheet, View } from "react-native";
+import LinearGradient from "react-native-linear-gradient";
```
damnhipster: Damn, thought you did this from scratch xD

trevorah:
Any chance you could make this less configurable?

damnhipster:
@trevorah It depends on what this will be used for. I'd argue that the hardcoded values should be in the components making use of this.

But at the moment, I think it's only the loading state.

I don't think we have many gradients across the site, all of the styles I've seen are flat. It's worth asking design/thinking about if this should be visible on the storybook, or just a package for other components to make use of.

fampinheiro:
@trevorah @damnhipster I changed the default gradient so it matches the most used angle on the skeletons.

The main reason for having the angle configurable is that the skeleton designs use different angles depending on the screen size.

PR #336, opened 2017-10-26
fix: watermark view box is customizable based on width and height

sericaia:
image

Styles discussed and adapted taking into account Jake Cox notes
https://app.zeplin.io/project/59f1bfe7475a900a398b648f/screen/59f1c0b373801718925cebf3

We might need to move TimesWatermark to a **new package** and make `viewbox` a prop, so @fampinheiro can use it in error pages


packages/watermark/watermark.stories.js
```diff
@@ -0,0 +1,11 @@
+import "react-native";
+import React from "react";
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { storiesOf } from "@storybook/react-native";
+import Watermark from "./watermark";
+
+storiesOf("Watermark", module)
+  .add("MPU", () => <Watermark width={300} height={250} />)
+  .add("Full", () => (
+    <Watermark width={970} height={250} viewBox="0 0 1000 300" />
```
trevorah: Why are you passing in a viewbox? Is there no way to get the svg to cover the container?

damnhipster: The viewbox is used to crop a certain part of the watermark SVG to use. You can see it being used [here](https://github.com/newsuk/times-components/blob/fix/placeholder-watermark/packages/ad/ad-watermark.js#L43-L45). 

I don't think it's worth moving that logic to watermark. Ads and error pages should specify the viewbox of the watermark themselves.

trevorah: I disagree. The viewbox logic should be encapsulated by this component, so It becomes much simpler to use.

sericaia: @trevorah you mean that we should have all possibilities for ads (3, in the image) + error pages with different sizes, all considered inside watermark component?! I disagree..

fampinheiro: The watermark is to be used in multiple places and situations and from the designs that I have seen the parts of the watermark (view box) is not always the same for different screen sizes.

My suggestion is to keep the watermark component as is until a pattern comes from its usage. We can always simplify the component API later.

trevorah: 👍

PR #340, opened 2017-10-27
Compilation fails after RN0.49 upgrade

adrienthiery:
Hi there,

Ran into an issue today after upgrading React Native to v0.49.* and `@times-components/brightcove-video` to the latest version (but I think it's also related to the version of the android build-tools.

I could not build my app anymore because of this error:
```
* What went wrong:
A problem occurred evaluating project ':@times-components/brightcove-video'.
> Could not find method api() for arguments [com.brightcove.player:exoplayer:5.2.0] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

found out that solution : https://stackoverflow.com/questions/45615474/gradle-error-could-not-find-method-implementation-for-arguments-com-android 

Do you know more about that?
I'm not sure if this is going to create the opposite problem to someone with an older version of the build-tools, though ⚠️

### DIFF
--- a/packages/brightcove-video/android/build.gradle
+++ b/packages/brightcove-video/android/build.gradle
@@ -25,6 +25,6 @@ repositories {
 }
 
 dependencies {
-    implementation "com.facebook.react:react-native:0.49.3"
-    api "com.brightcove.player:exoplayer:5.2.0"
+    compile "com.facebook.react:react-native:0.49.3"
+    compile "com.brightcove.player:exoplayer:5.2.0"
 }


CoreFloDev:
that's not the good solution, you need to update gradle to version 3.0 instead

adrienthiery:
Ok, I'll look at that, thanks!

adrienthiery:
Hey @CoreFloDev,

I'm looking into how to upgrade gradle for React Native, but RN is using `gradlew` (their own wrapper around gradle) to do that, and it is hardcoded to 2.14.1 in RN right now...

How do you upgrade that ? Do you replace the `gradlew` in the project?
Or do you always build with android studio and a manual version of gradle?

CoreFloDev:
Gradlew is only an executor script.

If you want to update your project, you need to set this to your gradle-wapper.properties
distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip

and
update your gradle tools to 3.0.0 in your app build.gradle
dependencies {
   ...
   classpath 'com.android.tools.build:gradle:3.0.0'
   ...
}

adrienthiery:
I'm stuck with an error saying that `Could not find com.android.tools.build:gradle:3.0.0`

But it seems latest version is `2.2.3` on bintray... (https://bintray.com/android/android-tools/com.android.tools.build.gradle)

adrienthiery:
Ok, I think I got it, I checked out https://github.com/newsuk/times-components/blob/master/android/build.gradle and saw I needed to add the `google()` repository to my `build.gradle`.

I'm gonna add that to the documentation.

PR #342, opened 2017-10-30
Updating contributing with fonts findings

nehasri89:

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -109,6 +109,42 @@ Follow these steps to deploy storybook native to a real android device.
 - If still struggling; shake the device and `debug js remotely`. Open a console on your computer for more info on the error
 - If you're still struggling verify that you are able to run `yarn storybook` and that it works in web view.
 
+### Fonts Naming Convention
+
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -109,6 +109,42 @@ Follow these steps to deploy storybook native to a real android device.
...
+But the FontName will have the weight details, whatever it is. eg.
+
+TimesDigitalW04-Regular
+TimesDigitalW04-Italic
+TimesDigitalW04-Bold
+
+- Family Name format (meta of the font) : <fontname>
+The font Family name should be constant across all of the above: eg.
+
+TimesDigitalW04
```
aronshamash: is the `04` needed? Can we remove for a nicer font name?

nehasri89: this is a real fontname we are using in the times component

aronshamash: and the `W04` part is needed and can't be removed ?

thatportugueseguy: I just looked into the fonts on the fonts server and at the moment we're using the fonts without  the w04, but it's actually there on the metadata.

given that W04 is just a weight indication, maybe for simplicity we can ditch the W04??

https://www.thetimes.co.uk/fonts/TimesDigital-Regular ==>
screen shot 2017-10-30 at 13 55 56

craigbilner:
@nehasri89 updated the wording
@aronshamash perhaps we keep the `W04` for now to be consistent and make it easier to replace as is?

aronshamash:
yes lets leave `W04` so it is consistent.

PR #343, opened 2017-10-30
Add gradle update step to README

adrienthiery:
No technical impact


damnhipster:
Are there plans to upgrade Gradle across the board to 3.0?

damnhipster:
Actually, we're already using a version of gradle which supports brightcove

https://github.com/newsuk/times-components/blob/master/android/build.gradle#L9
https://github.com/newsuk/times-components/blob/master/android/gradle/wrapper/gradle-wrapper.properties#L5

I tested it out, and the brightcove player works on android.

adrienthiery:
Hi @damnhipster,

newsuk's project already indeed uses Gradle 3.0, but when adding only one of the components to another project, the Android build fails with an error without upgrading. (See #340)

PR #347, opened 2017-10-31
chore: update prettierignore

fampinheiro:
Edit: Fixing failing build in master.


killanaca:
why do we need an update in yarn.lock? @damnhipster @fampinheiro

damnhipster:
This change was to fix the broken build on master, which was holding up a few other pull requests.

The fix was cherry-picked, which contained the change in the yarn.lock.

We tested it locally and didn't see any issues.

PR #351, opened 2017-11-01
update image fixture

killanaca:
**Inline primary**
screen shot 2017-11-01 at 11 49 02

**Inline secondary**
screen shot 2017-11-01 at 11 49 10

**Inline portrait**
screen shot 2017-11-01 at 11 49 15

**Inline landscape**
screen shot 2017-11-01 at 11 49 20


fampinheiro:
@killanaca should we also change the fixtures from the components that use the image?

killanaca:
@fampinheiro was just the one inside the article image a little bit weird

PR #352, opened 2017-11-01
Feature/tap 103 jest android issue

jimhunty:
SPIKE - DO NOT MERGE

The point of this work was to investigate the issue of using platform specific styling files and how they would render in snapshot testing. I can confirm this now works as expected with the below toolset, give it a try.

It is using a mixture of Jest projects and haste mapping to specifically target certain environments. In addition, I have added more granularity around the kind of tests that can be run to make it useful for component development. I've tried to keep things DRY with a lot of duplicate config moved towards a lib file. 

Next steps:

1. Try it out and make sure it follows all the use cases, you will know better than I at this stage.
2. Review the approach and see if you like it.
3. If its all good, i'll create another ticket to... 
    1. Implement this across all of the packages
    2. There seems to be a cli generator? Therefore that would need updating
    3. Provide a test approach doc to support this going forward.


jimhunty:
Travis is having some problems on files I haven't touched. The problems seem consistent though. Is there a known solution here?

lib/jestConfigurator.js
```diff
@@ -0,0 +1,36 @@
...
+
+const config = (platform) => ({
+    preset: "react-native",
+    ...platform === 'web' ? webSpecific() : nativeSpecific(platform),
+    rootDir: "../../../../",
+    transformIgnorePatterns: ["node_modules/(?!@times-components)/"],
+    testMatch: [
+        `<rootDir>/packages/article/__tests__/${platform}/article.${platform}.test.js`
+    ],
+    moduleFileExtensions: [`${platform}.js`, "js", "json", ]
```
trevorah: ios and android also pick up `*.native.js`

packages/article/jest.config.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * This is the shared configuration across all environments
+ */
+
+module.exports = {
+    projects: [ "__tests__/android", "__tests__/ios", "__tests__/web" ]
```
trevorah: be careful with this, there are some odd race conditions that i've found with jest projects.

lib/jestConfigurator.js
```diff
@@ -0,0 +1,36 @@
...
+    testEnvironment: "jsdom",
+});
+
+const config = (platform) => ({
+    preset: "react-native",
+    ...platform === 'web' ? webSpecific() : nativeSpecific(platform),
+    rootDir: "../../../../",
+    transformIgnorePatterns: ["node_modules/(?!@times-components)/"],
+    testMatch: [
+        `<rootDir>/packages/article/__tests__/${platform}/article.${platform}.test.js`
```
trevorah: if you could match on `.test.${platform}.js`, you will have less of a problem of jest's default config running platform specific tests. Also, jest assumes by default that *all* files in `__tests__` are test files.

packages/article/jest.config.js
trevorah: Also, if you have projects as `*/**/jest.config.js`, you can start running *all* test configs across the monorepo.

trevorah: (in a good way)

damnhipster:
@jimhunty Travis is failing on linting one of the changed files. Try running `yarn format`.

jimhunty:
Took a while to find, but the command needed was `yarn fmt`.

The project needs and editor config file as this was all down to "spacing vs tabs". I'll raise an issue.

lib/jestConfigurator.js
jimhunty: Thanks Andy, added support for these.

packages/article/jest.config.js
jimhunty: In order to reduce the race conditions, I have limited the use of projects to `yarn test:native`, `yarn test:native:watch` and `yarn test:watch`. Therefore this can only be used in the context of development and not in CI. `yarn test` is independent of this feature.

It would be cool if the whole monorepo did run on a single test runner. I think we would need to implement this structure first and then test it. But if the race condition bug exists perhaps its best to wait for further stability?

lib/jestConfigurator.js
jimhunty: I have changed this now to make it less specific so the "configurator" (_terrible name_) can actually be used elsewhere. Im not sure what you mean about the matching though?

The tricky thing was, in order to get projects to run, the config for each platform needed to be in its own folder. Essentially adding sub packages. As this was only relating to tests it made sense for these to sit in the tests folder to not pollute the rest of the component. I have added an ignore path so that Jest will ignore them.

packages/article/package.json
```diff
@@ -5,10 +5,16 @@
   "main": "article",
   "scripts": {
     "flow": "node_modules/flow-bin/cli.js",
-    "test:watch": "npm run test:native -- --bail --verbose --watchAll",
-    "test:native": "jest --config='./jest.config.json'",
-    "test:web": "jest --config='./jest.config.web.json'",
-    "test": "npm run test:native -- --ci --bail && npm run test:web -- --ci --bail --coverage",
+    "test:android": "jest --config='./__tests__/android/jest.config.js'",
```
craigbilner: If we make `package.json` changes, need to update the `cli` too

packages/article/package.json
```diff
@@ -5,10 +5,16 @@
...
+    "test:ios": "jest --config='./__tests__/ios/jest.config.js'",
+    "test:web": "jest --config='./__tests__/web/jest.config.js'",
+    "test:android:watch": "yarn test:android -- --bail --verbose --watchAll",
+    "test:ios:watch": "yarn test:ios -- --bail --verbose --watchAll",
+    "test:web:watch": "yarn test:web -- --bail --verbose --watchAll",
+    "test:native": "jest --projects __tests__/android __tests__/ios ",
+    "test:native:watch": "yarn test:native -- --bail --verbose --watchAll",
+    "test": "yarn test:android -- --ci --bail && yarn test:ios -- --ci --bail && yarn test:web -- --ci --bail --coverage",
```
craigbilner: if we go for this pattern, usually devs get confused/find it hard to update the snapshots, so we'd need a script for that

packages/article/__tests__/android/jest.config.js
```diff
@@ -0,0 +1,3 @@
+const config = require("../../../../lib/jestConfigurator");
+
+module.exports = config("article", "android");
```
craigbilner: are these not prettiered?

packages/article/package.json
jimhunty: Agreed as stated in the original message as a next step after this spike.

packages/article/__tests__/android/jest.config.js
jimhunty: It seems not, Prettier is ignoring any android folder. Maybe should change the folder names for these. E.g. `android-tests` ?

packages/article/package.json
jimhunty: Snapshots should only be updated when they need updating. Each test failure should be checked carefully that the snapshot being updated is correct. Providing a script feels like an easy way to add regression and put more pressure on the PR reviewer to pick this things up.

If you did want this though, we would need to use projects and go for something like:
`"test:snapshots:update": "jest --projects __tests__/android __tests__/ios __tests__/web -u",

craigbilner: completely agree about creating _enough_ friction, suggesting that this has created a lot...

packages/article/package.json
craigbilner: if we're happy with this can we just put straight into the `cli`?

jimhunty: If its ok, id like to treat this as a separate ticket along with updating all the current packages to use the format? This being originally a spike and if merged will work fine in isolation. I do think it should be prioritised and this will help enable some of the other "improvement" tasks.

jimhunty:
Added a Utilities package which has the wonderfully named `jestConfigurator`. This will fail Travis because the package hasn't been published yet. I have used `index` naming in folders which looks like its against naming convention? Happy to change if it is.

To test you will need to npm link, so perhaps if this looks ok from a code perspective we publish that module and I can add in the test style code as requested above?

packages/article/styles/const/index.js
```diff
@@ -0,0 +1,3 @@
+import sharedConst from "./shared";
```
jimhunty: Shared constants might be a useful package on its own that can be consumed by other components. It will attempt to keep the packages in sync and can be versioned separately so changes can be cascaded by choice.

packages/article/styles/article-style/native.android.js
```diff
@@ -0,0 +1,19 @@
...
+const { fontFamilyBody } = constants;
+
+const androidStyles = StyleSheet.create({
+  leadAsset: Object.assign({}, sharedStyles.leadAsset, {
+    marginBottom: 6
+  }),
+  articleTextElement: Object.assign({}, sharedStyles.articleTextElement, {
+    fontFamily: fontFamilyBody,
+    fontSize: 16,
+    fontStyle: "normal",
```
jimhunty: Unavoidable duplication due to the import of `.android` as an extension. Could be separated out as another file, however, there is a fair amount of abstraction already.

jimhunty:
@craigbilner This now contains an reworked example of styling and how snapshots can use it. I took a more DRY approach to the styles to attempt to avoid some reusability. There is some deep merging with Object Assign to enable the cascade. The entry point for the component consuming the styles remains the same, but with some logic and RN's extension capabilities to bring the correct stylesheet forward.

I have also started creating some constants that may be useful as a separate package? This gives us some synchronicity between the packages. As this can be separately versioned it should mean opt in updates.

packages/article/styles/article-header-style/native.js
```diff
@@ -0,0 +1,13 @@
+import { StyleSheet } from "react-native";
```
sericaia: shouldn't this file be native.ios.js?

sericaia: I got it, its just a bit confusing ;)

jimhunty: I agree it is a bit confusing. The structure here is designed as a catch all. Android will not always be just a cascade of styles from iOS.

packages/article/styles/article-header-style/native.js
```diff
@@ -0,0 +1,13 @@
...
+const nativeStyles = StyleSheet.create({
+  articleHeadLineText: Object.assign({}, sharedStyles.articleHeadLineText, {
+    fontWeight: "700"
+  }),
+  standFirst: Object.assign({}, sharedStyles.standFirst, {
+    lineHeight: 25
+  })
+});
+
+export default Object.assign({}, sharedStyles, nativeStyles);
```
sericaia: something is a bit confusing here.. 
we're doing `Object.assign({}, sharedStyles, nativeStyles);` in `article-header-style/native.js` and `Object.assign(styles, globalStyle, nativeStyles);` on `article-header-style/index.js` 

Can't sharedStyles be added always in the Object.assign() inside `article-header-style/index.js` and the Object.assign() here be removed? ;)

jimhunty: It could possibly be moved. But the idea here is that the `native.js` and `web.js` files encompass all the styles for those platforms. So moving it down means they are incomplete.

jimhunty:
Fixed some conflicts this morning with updated fonts from master.

packages/jest-configurator/README.md
```diff
@@ -0,0 +1,15 @@
+# Jest
+
+Across the components in the monorepo, we reuse Jest config in a number of places. It makes sense to centralise this. As a result the Jest Configurator was created.
+
+## Usage
+
+Create a file called `jest.config.js` in a platform specific folder under tests. E.g. `<rootDir>/packages/${component}/__tests__/${platform}/jest.config.js`. Then add platform specific code like below.
+
+```js
+const Utilities = require("@times-components/utilities");
```
craigbilner: changed

lib/cli/templates.js
```diff
@@ -26,21 +29,124 @@ const testFile = handlebars.compile(`/* eslint-env jest */
...
+  handlebars.compile(`/* eslint-env jest */
+
+import shared from "../shared";
+
+jest.mock("react-native", () => {
+  const reactNative = require.requireActual("react-native");
+  reactNative.Platform.OS = "${platform}";
+  jest
+    .spyOn(reactNative.Platform, "select")
+    .mockImplementation(obj => obj.${platform} || obj.default);
```
jimhunty: Mixture of styles between template literals and handlebars, but the current code structure doesn't really support platforms. I feel I would have to change a lot to support it and I think there are better ways to produce the whole cli. So in this case im ignoring it.

packages/article/__tests__/android/__snapshots__/article.android.test.js.snap
```diff
@@ -1458,6 +1571,7 @@ exports[`Article test on android renders article no label 1`] = `
               "color": "#333333",
               "fontFamily": "TimesDigitalW04",
               "fontSize": 16,
+              "fontStyle": "normal",
```
jimhunty: @nehasri89 This fontStyle addition, should this be removed now for the new fonts in master?

packages/article/package.json
jimhunty: Its now in the CLI.

PR #360, opened 2017-11-06
Feat: fonts fixes

thatportugueseguy:
Creates a unified way to use the fonts on all platforms (web, ios, android) and updates the fetch-fonts script accordingly.

The fonts in the TimesModern font family are still used in different way, because they have different weights on the regular + bold and the bold version.

Still uses font-forge, as per the ticket spec ([TAP-122](https://nidigitalsolutions.jira.com/browse/TAP-122))

_Note_ : The approach has ben tested on storybooks, native apps and render

### DIFF
--- a/packages/author-head/author-head.js
+++ b/packages/author-head/author-head.js
@@ -30,8 +30,8 @@ const styles = StyleSheet.create({
     overflow: "hidden"
   },
   name: {
-    fontFamily: "TimesModern",
-    fontWeight: "bold",
+    fontFamily: "TimesModern-Bold",
+    fontWeight: "400",
     fontSize: 45,
     color: "#000",
     paddingTop: 32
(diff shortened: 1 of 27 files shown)


packages/author-head/author-head.js
```diff
@@ -34,6 +31,7 @@ const styles = StyleSheet.create({
   },
   name: {
     fontFamily: "TimesModern-Bold",
+    fontWeight: "400",
```
nehasri89: This should be an existing rule @ze we would not have to add anything new

packages/markup/markup.stories.js
```diff
@@ -62,7 +62,8 @@ storiesOf("Markup", module)
               style={{
                 margin: 10,
                 color: "red",
-                fontFamily: "TimesModern-Bold"
+                fontFamily: "TimesModern-Bold",
```
nehasri89: same as above

packages/author-head/author-head.js
thatportugueseguy: I added the fontWeight: 400 to keep it in line with the majority of the other usages.

I actually think we don't need to use it, but if we're using it, we should use it everywhere.

nehasri89: hmm, I think it is actually wrong to set a different font weight, than the inherent weight of a font in first place. We are waiting for the right font to replace this style. we would not want to have more places with incorrect styles

PR #362, opened 2017-11-06
fix: prevent tracked component from unmounting when HOC re-renders

jonnywyatt:
tracking-context HOC was returning another HOC (track-render) from its render method, which was unmounting / remounting when the outer HOC re-rendered, instead of simply re-rendering itself. This PR fixes that issue.


packages/tracking/__tests__/track-render.test.js
```diff
@@ -5,7 +5,7 @@ import renderer from "react-test-renderer";
 import { withTrackRender } from "../tracking";
 import withTestContext from "./test-tracking-context";
 
-describe("WithTrackRender", () => {
+describe.skip("WithTrackRender", () => {
```
toboid: Think we can remove this whole file

PR #372, opened 2017-11-09
Tracking article

jonnywyatt:
https://nidigitalsolutions.jira.com/browse/TAP-87

Track page views only, initially


toboid:
i'll create an issue to update the access of the byline data structure when we have flow

PR #373, opened 2017-11-09
Feat/article flatlist

sericaia:
This is work in progress.

FlatList is being used for ios and android and ListView for web.

image


packages/article/article-content.js
```diff
@@ -0,0 +1,24 @@
...
+  <FlatList
+    testID="scroll-view-article"
+    keyExtractor={item => item.type + item.index || item.type}
+    data={data}
+    renderItem={({ item }) => renderRow(item)}
+  />
+);
+
+ArticleContent.propTypes = {
+  data: PropTypes.arrayOf(PropTypes.object),
```
sericaia: .isRequired?

packages/article/article-content.web.js
```diff
@@ -0,0 +1,41 @@
...
+    renderRow={renderRow}
+    initialListSize={initialListSize}
+    scrollRenderAheadDistance={scrollRenderAheadDistance}
+    pageSize={pageSize}
+    enableEmptySections
+  />
+);
+
+ArticleContent.propTypes = {
+  data: PropTypes.arrayOf(PropTypes.object),
```
sericaia: .isRequired?

packages/article/article-content.js
```diff
@@ -0,0 +1,19 @@
...
+  <FlatList
+    testID="flat-list-article"
+    keyExtractor={item => item.type + item.index || item.type}
+    data={data}
+    renderItem={({ item }) => renderRow(item)}
+  />
+);
+
+ArticleContent.propTypes = {
+  data: PropTypes.arrayOf(PropTypes.object).isRequired,
```
jimhunty: Can we add a shape of the object here?

packages/article/article.fructose.ios.js
```diff
@@ -21,7 +21,9 @@ withComponent(
     });
 
     test("default Article should render correctly", async () => {
-      await expect(element(by.id("listView"))).toBeVisible();
+      await expect(
```
jimhunty: Async await seem redundant in this test.

packages/article/article-content.web.js
```diff
@@ -0,0 +1,16 @@
...
+import { ScrollView } from "react-native";
+
+const ArticleContent = ({ data, renderRow }) => (
+  <ScrollView testID="scroll-view-article">
+    {data.map(item => renderRow(item))}
+  </ScrollView>
+);
+
+ArticleContent.propTypes = {
+  data: PropTypes.arrayOf(PropTypes.object).isRequired,
```
jimhunty: Again it would be useful to have a shape here

packages/article/article.js
```diff
@@ -29,14 +29,15 @@ class ArticlePage extends React.Component {
   static renderRow(rowData) {
     if (rowData.type === "leadAsset") {
       return (
-        <View testID="leadAsset" style={styles.leadAsset}>
+        <View testID="leadAsset" key={rowData.type} style={styles.leadAsset}>
```
jimhunty: Is the type guaranteed to be unique?

sericaia: yes, there's always only one leadAsset

packages/article/article.js
```diff
@@ -29,14 +29,15 @@ class ArticlePage extends React.Component {
...
           <Image uri={rowData.data.crop.url} aspectRatio={16 / 9} />
```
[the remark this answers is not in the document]
sericaia: @nehasri89

packages/article/article.js
```diff
@@ -28,9 +28,11 @@ const withAdComposer = (children, section = "article") => (
 class ArticlePage extends React.Component {
   static renderRow(rowData) {
     if (rowData.type === "leadAsset") {
+      const [ratioWidth, ratioHeight] = rowData.data.crop.ratio.split(":");
```
craigbilner: does this need to happen every row?

killanaca: it's not every row, it's just for the row of type `leadAsset` (one per article). Maybe at some point we can have some utils package (like the `jest-configurator`)  where we can share this kind of code.

PR #375, opened 2017-11-09
fix: don't point at dev imageserver

dickie81:
This story pointed at dev image server which has a bad https cert & is not available outside office network.


dickie81:
Have replaced all references to dev image server across whole repo.

PR #379, opened 2017-11-10
fix: stops using image component for aspect ratio side effect

dickie81:


craigbilner:
@dickie81 is this what we want? Wouldn't it be nicer to reuse the loading state of the image so it's in one place but just not fake it with a blank `uri`?

dickie81:
I think its more honest & easier to read to just expose the placeholder & use it directly. Also we would have to complicate the image component for this one use case.

craigbilner:
going to pick this up separately

PR #383, opened 2017-11-10
fix: bad property name

dickie81:
Fixes error view story for handling brightcove video errors.


craigbilner:
@dickie81 I'd expect the error handler to not choke on an invalid prop otherwise why have it?

dickie81:
@craigbilner its more of a problem with initialisation of the video - it never initialised as it did not get enough props (it did not error as such). We should probably make policyKey required for native so we at least see a warning in this scenario.

PR #384, opened 2017-11-10
fix: add ignore files

craigbilner:
Make package consistent with the rest of the repo and don't prettier `package.json`


jimhunty:
👍

PR #386, opened 2017-11-13
fix: add missing required property of article

dickie81:
At the moment all Article stories are broken - this fixes them.


toboid:
@aronshamash think the tests should pass if you rebase onto latest master

PR #388, opened 2017-11-13
fix: stop limiting future yarn versions

trevorah:
Super annoying if you've updated yarn or just installed it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "engines": {
     "node": ">=8.0.0",
-    "yarn": "1.2.1"
+    "yarn": ">=1.2.1"
   },
   "bin": {
     "times-components": "times-components"


damnhipster:
Yarn released a breaking change when they went live with 1.0, and I remember another breaking change after a minor release that happened. 

Although, things seem to have become more stable since 1.0 was released, and 1.3.2 is working fine at the moment.

craigbilner:
@trevorah  also, whenever we change yarn stuff, don't forget Travis and Bitrise

trevorah:
:-1: I'm against locking to a specific version. When can we rely or yarn to not break between versions?

craigbilner:
@trevorah - this is the status quo and makes things easier for everyone

https://yarnpkg.com/blog/2017/05/31/determinism/

> Yarn is fully deterministic as long as all your teammates are using the same Yarn version

damnhipster:
Is it worth bundling the project with the correct version of yarn? 

Developers then wouldn't have to worry about switching their version of yarn locally to match the project, especially since different projects may use different versions of yarn due to the number of releases yarn goes through.

The convention would then be to run `npx yarn`, which isn't too far off from running `yarn`

craigbilner:
closing in favour of bumping yarn

PR #393, opened 2017-11-15
Fix/watermark svg android perf

killanaca:
at the moment complex svg (like the watermark) take 8 sec to render on low end android devices (eg samsung s4). This PR create a specific (simpler) placeholder android component

screenshot_2017-11-14-17-27-12-5


fampinheiro:
@killanaca is this being applied to every android device?

killanaca:
@fampinheiro yes at the moment

fampinheiro:
Could we do this just for screens that have a pixel ratio lesser than a sensible value? Don't like to level the experience with the crappiest device.

craigbilner:
@fampinheiro @killanaca could we agree some metrics with which we "cut the mustard"?

fampinheiro:
I am happy to merge it the way it is.
Let's just agree that we need to work on this as soon as we have the time.

PR #394, opened 2017-11-16
fix: static rendering of card component

toboid:
Fixes server rendering of the card component by using media queries for layout. Previously layout was done with JS but this doesn't work for server rendered code since the size of the screen is unknown.
 
react-native-web does not support css media queries so another solution was required to get this working correctly on web.

## Website (with client-side react)
Before 
image

After
image

## Website (server render, no client-side react)
Before
image

After
Note that the placeholder overlays the image, see next screenshot for how it looks without the placeholder. There is a separate ticket to fix this, however in practice it's less of an issue than it appears to be in a static shot as the placeholder will be removed by js. The fact that the layout is now correct means that a user with js enabled will no longer see the page flash with very large images that then get small as the page layout changes.
image

After (placeholder removed)
image

## Loading
Before
image

After
image

## Default
Before
image

After
image

## Without image
Before
image

After
image

## Static render (storybook)
Before
N/A

After
This is a client-side static render so dimensions are correct since they are available at render time. However changing the window size would not previously change the layout, it will now. This also highlights an issue with the image placeholder  - it's currently removed with JS so here it overlays the the image. There is another ticket to address this.
image


packages/card/card-styles.web.js
```diff
@@ -0,0 +1,35 @@
...
+const breakpoint = "500px";
+
+export const ImageContainer = styled(View)`
+  flex-grow: 1;
+  flex-shrink: 1 !important;
+
+  margin-bottom: 10px;
+  @media (min-width: ${breakpoint}) {
+    flex-grow: 2;
+    flex-basis: 0 !important;
```
toboid: `!important` is required for these few rules as react-native-web is applying competing styles with `!important`

PR #398, opened 2017-11-16
Chore/apply font smoothing

craigbilner:
The website uses `font-smoothing` therefore to make it easier to compare with Zeppelin/the current site when doing design reviews, blat storybook with it

Minor: fix console warning


damnhipster:
Should the storybook have cross-browser support?

e.g.
screen shot 2017-11-17 at 11 12 34

PR #399, opened 2017-11-17
fix: use ImageBackground wrapped in View

craigbilner:
RN 50 deprecated nested children within an `Image`, however, there is an undocumented component called `ImageBackground` which can drop in and provides similar functionality. The existing solution doesn't particularly work for `Android` which doesn't support `position: absolute` very well


packages/image/image.web.js
```diff
@@ -58,6 +60,14 @@ class TimesImage extends Component {
...
-TimesImage.propTypes = imagePropTypes;
+TimesImage.defaultProps = {
+  style: {}
+};
+
+TimesImage.propTypes = {
+  uri: PropTypes.string.isRequired,
+  aspectRatio: PropTypes.number.isRequired,
+  style: ViewPropTypesStyle
+};
```
trevorah: do they not share the same propTypes now?

packages/image/image.js
```diff
@@ -30,25 +22,40 @@ class TimesImage extends Component {
...
   render() {
-    const { uri: dirtyUri, aspectRatio, style } = this.props;
+    const { uri: dirtyUri, style, aspectRatio } = this.props;
     const { isLoaded } = this.state;
-    // web handles missing protocols just fine, native doesnt. This evens out support.
+    // web handles missing protocols just fine, native doesn't. This evens out support.
     const uri = addMissingProtocol(dirtyUri);
 
+    const props = {
+      style: [{ width: "100%", height: "100%" }].concat(style),
```
trevorah: why not `flex-grow` and `align-items: stretch`?

packages/image/__tests__/web/image.web.test.js
```diff
@@ -26,7 +26,7 @@ it("allows styling", () => {
 
 it("prepends https schema", () => {
   const tree = renderer
-    .create(<Image uri="//example.com/image.jpg" aspectRatio={1 / 1} />)
+    .create(<Image uri="//example.com/image.jpg" aspectRatio={3 / 2} />)
```
trevorah: why?

packages/image/__tests__/image.native.test.js
```diff
@@ -0,0 +1,34 @@
...
+    expect(tree).toMatchSnapshot();
+  });
+
+  it("prepends https schema", () => {
+    const tree = renderer
+      .create(<Image uri="//example.com/image.jpg" aspectRatio={3 / 2} />)
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+};
```
trevorah: What platform does this test run with?

packages/image/__tests__/web/image.web.test.js
trevorah: 🗣 easier to test

packages/image/__tests__/image.native.test.js
trevorah: 🗣 not actually a test, just a func that supplies tests

craigbilner: just common test code

packages/image/image.js
craigbilner: same concept

trevorah: 🗣 same difference

PR #400, opened 2017-11-17
Enable yarn cache in travis (WIP)

nehasri89:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
+cache: yarn
 before_install:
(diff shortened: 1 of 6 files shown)


damnhipster:
I did a spike around this earlier, and found that the results were inconclusive because of the inconsistent duration of the `yarn` step on Travis.

On top of that, I noticed the following error which I'm also seeing in the build logs of the commits above:

screen shot 2017-11-20 at 11 58 36

My guess is that the postinstall process is adding the inconsistency. Splitting the task into two steps in Travis will allow us to see the time taken for each step and hopefully get rid of the inconsistency.

PR #403, opened 2017-11-17
fix: remove JS

craigbilner:
Web images should work without JS enabled which also makes SSR pages more elegant

n.b. still uses JS for placeholder

### DIFF
--- a/packages/image/image.web.js
+++ b/packages/image/image.web.js
@@ -11,7 +11,6 @@ const TimesImage = ({ uri, aspectRatio, style }) => {
       paddingBottom: `${100 / aspectRatio}%`
     },
     placeholder: {
-      flex: 1,
       position: "absolute",
       top: 0,
       bottom: 0,
--- a/packages/image/image.web.js
+++ b/packages/image/image.web.js
@@ -1,62 +1,38 @@
-import React, { Component } from "react";
+import React from "react";
 import { View } from "react-native";
 import Placeholder from "./placeholder";
 import { defaultProps, propTypes } from "./image-prop-types";
 
-class TimesImage extends Component {
-  constructor(props) {
-    super(props);
-
-    this.state = {
-      isLoaded: false
-    };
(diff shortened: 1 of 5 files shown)


packages/image/image.web.js
```diff
@@ -1,62 +1,38 @@
...
-}
+      right: 0,
+      zIndex: 0
+    },
+    img: { display: "block", width: "100%", zIndex: 1, position: "absolute" }
+  };
+
+  const boundedImg = (
+    <div style={styles.wrapper}>
+      <img src={uri} style={styles.img} alt="" />
```
jonnywyatt: Could alt be passed as a prop?

packages/image/image.web.js
```diff
@@ -1,62 +1,38 @@
...
+    wrapper: {
       height: 0,
       overflow: "hidden",
       paddingBottom: `${100 / aspectRatio}%`
-    };
-
-    const placeholderStyle = {
+    },
+    placeholder: {
       flex: 1,
```
jonnywyatt: Placeholder component already sets this property

packages/image/image.web.js
```diff
@@ -1,62 +1,38 @@
...
-
-    // divs cannot be styled with the output of Stylesheet.create()
-    // only react native Views accept those ids
-    return style ? <View style={style}>{boundedImg}</View> : boundedImg;
-  }
-}
+      right: 0,
+      zIndex: 0
+    },
+    img: { display: "block", width: "100%", zIndex: 1, position: "absolute" }
```
jonnywyatt: Does height need 100%?

packages/image/image.web.js
craigbilner: It should be! Will do as a separate PR as need to agree the data too

packages/image/image.web.js
craigbilner: 👍

PR #408, opened 2017-11-20
fix: browser issue

sericaia:
image
From left to right: chrome, FF, safari

image
From left to right: android, ios


sericaia:
Do not merge yet

jimhunty:
Its not what I would use `display: table` for but works well, so its a 👍 from me.

sericaia:
@jimhunty i could not find another way to solve this issue in FF

damnhipster:
We should be able to use a `clearfix` for this issue.

I'll give it a try.

PR #411, opened 2017-11-22
Use Yarn Workspaces

jonnywyatt:
https://yarnpkg.com/lang/en/docs/workspaces/

Yarn will manage the installation / hoisting of dependencies, instead of Lerna.

Lerna will still handle publishing of packages.

Have to make the repo `private` in order to work with workspaces to prevent publishing top level...?


package.json
```diff
@@ -36,9 +36,7 @@
     "prefmt": "prettier --write '+(storybook|.storybook.native|lib)/**/*.+(js|json)' '*.md' '.github/**/*.md'",
     "fmt": "lerna run fmt",
     "packages:publish": "lerna publish --conventional-commits --yes --concurrency=1 --exact",
-    "packages:publish-dry-run": "npm run packages:publish -- --skip-git --skip-npm",
-    "update-deps": "lerna bootstrap --hoist --npm-client=yarn --concurrency=1",
-    "postinstall": "npm run update-deps --silent"
+    "packages:publish-dry-run": "npm run packages:publish -- --skip-git --skip-npm"
```
jimhunty: Should this be `yarn packages:publish...` ?

package.json
```diff
@@ -36,9 +36,7 @@
     "prefmt": "prettier --write '+(storybook|.storybook.native|lib)/**/*.+(js|json)' '*.md' '.github/**/*.md'",
     "fmt": "lerna run fmt",
     "packages:publish": "lerna publish --conventional-commits --yes --concurrency=1 --exact",
-    "packages:publish-dry-run": "npm run packages:publish -- --skip-git --skip-npm",
-    "update-deps": "lerna bootstrap --hoist --npm-client=yarn --concurrency=1",
```
jimhunty: Will these lines be missed?

package.json
jonnywyatt: Yes I should probably replace all npm commands with yarn

rogeriopvl:
I'm all for this change 👍

I'm just wondering if we can we completely replace lerna with yarn workspaces?

jonnywyatt:
@rogeriopvl I don't think yarn workspaces can provide the publish function that lerna has?

rogeriopvl:
@jonnywyatt yes you're correct. I just noticed that for publishing it has to be used alongside lerna.

package.json
jimhunty: 👍

craigbilner: @jonnywyatt can we keep the dry-run which is useful to debug publishing or test stuff out?

killanaca:
@jonnywyatt do we still need lerna for the `run` command or we can migrate that too to Yarn workspace?

lib/cli/package-json.js
```diff
@@ -34,11 +34,11 @@ module.exports = variables => ({
   },
   homepage: "https://github.com/newsuk/times-components#readme",
   devDependencies: {
-    "@times-components/jest-configurator": "0.0.1",
+    "@times-components/jest-configurator": "0.0.8",
```
craigbilner: @jimhunty FYI this will struggle to stay in sync

jimhunty: Why is that?

craigbilner: @jimhunty `jest-configurator` will auto-bump but `package-json.js` is a flat file and will not get the same treatment

jimhunty: Ah of course, sounds like we need a better way of doing the CLI 😉

jimhunty:
I can confirm everything works as expected. I cleared all node modules from root and each of the packages, yarn installed and ran the web, iOS and Android environments successfully.

Good to go in I reckon.

PR #413, opened 2017-11-23
Fructose/android ci

kaliabadi:


packages/brightcove-video/brightcove-player.fructose.android.js
```diff
@@ -0,0 +1,57 @@
+/* globals withComponent test beforeEach driver */
+import React from "react";
+import BrightcovePlayer from "./brightcove-player";
+import VideoWithExternalControls from "./fixtures/video-with-external-controls";
+
+const policyKey =
```
L0wry: are these okay being here?

packages/brightcove-video/brightcove-player.fructose.android.js
```diff
@@ -0,0 +1,57 @@
...
+  <BrightcovePlayer
+    fructoseID="brightcove video renders"
+    policyKey={policyKey}
+    videoId={videoId}
+    accountId={accountId}
+  />,
+  "video with external controls",
+  async fructose => {
+    beforeEach(async () => {
+      await fructose.loadComponent();
```
L0wry: we're not resetting the app here like in the other tests, is this on purpose?

kaliabadi: Addressed

packages/brightcove-video/brightcove-player.fructose.android.js
kaliabadi: Yeah checked with Jack and he said it’s ok

PR #414, opened 2017-11-23
Speedup Travis Part 2

nehasri89:
As part of this Pr we have removed fetch fonts step from lint and test steps, this was tested in a WIP PR https://github.com/newsuk/times-components/pull/412.

Another Pr for yarn cache will go in once the workspace changes are merged


package.json
```diff
@@ -19,7 +19,7 @@
     "prelint": "prettier --list-different '+(.storybook.native|storybook|lib)/**/*.+(js|json)' '*.md' '.github/**/*.md'",
     "lint": "lerna run lint --since",
     "fetch-fonts": "node ./lib/fetch-fonts",
-    "prestorybook": "yarn fetch-fonts && yarn storybook:build-vendor",
+    "prestorybook": "yarn storybook:build-vendor",
```
craigbilner: if you do it this way, it does mean people will have to manually fetch fonts...could we just split out the test storybook build from a "real" one?

PR #416, opened 2017-11-27
feat: add pending

craigbilner:
The current naïve implementation of lazy images does a lot of work on scroll which can cause jank on some machines/browsers. This effectively debounces any pending changes until the observations/scrolling stops before doing the JS work and reflowing to give better UX.

A small tweak to the threshold improved loading on mobile views where the cards are laid vertically and the image starts at the top.

Here are the profiles of scrolling fast down a list of 20 items

Before (lots of work and long frames): 

screen shot 2017-11-25 at 18 16 58

After (less work with one hit at the end):

screen shot 2017-11-25 at 18 18 10

### DIFF
--- a/packages/author-profile/__tests__/author-profile-content.web.test.js
+++ b/packages/author-profile/__tests__/author-profile-content.web.test.js
@@ -377,27 +377,7 @@ it("disconnects from the IntersectionObserver when unmounting", async () => {
     />
   );
 
-  const makeEntries = nodes =>
-    [...nodes].map((node, indx) => ({
-      target: node,
-      intersectionRatio: indx === 0 ? 0.75 : 0
-    }));
-  window.IntersectionObserver.dispatchEntriesForInstance(1, makeEntries);
-
-  await delay(20);
-
-  const makeNewEntries = nodes =>
-    [...nodes].map((node, indx) => ({
-      target: node,
-      intersectionRatio: indx === 0 ? 0.25 : 0.75
-    }));
-  window.IntersectionObserver.dispatchEntriesForInstance(1, makeNewEntries);
-
-  await delay(0);
-
   component.unmount();
-
-  await delay(100);
 
   expect(disconnectSpy).toHaveBeenCalled();
 
--- a/packages/author-profile/__tests__/author-profile-content.web.test.js
+++ b/packages/author-profile/__tests__/author-profile-content.web.test.js
@@ -9,6 +9,8 @@ import authorProfileFixture from "../fixtures/author-profile.json";
 import articleListFixture from "../fixtures/article-list.json";
 import pagedResult from "./paged-result";
 
+const delay = ms => new Promise(res => setTimeout(res, ms));
+
 test(AuthorProfileContent);
 
 const results = {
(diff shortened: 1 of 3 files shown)


packages/author-profile/__tests__/author-profile-content.web.test.js
```diff
@@ -197,3 +212,196 @@ it("renders good quality images if there is no IntersectionObserver", () => {
...
+      onArticlePress={() => {}}
+    />
+  );
+
+  const makeEntries = nodes =>
+    [...nodes].map((node, indx) => ({
+      target: node,
+      intersectionRatio: indx === 0 ? 0.75 : 0
+    }));
+  window.IntersectionObserver.dispatchAll(makeEntries);
```
toboid: Doesn't look like we need any calls to `dispatchAll` for this test?

packages/author-profile/__tests__/author-profile-content.web.test.js
```diff
@@ -197,3 +212,196 @@ it("renders good quality images if there is no IntersectionObserver", () => {
...
+
+  await delay(0);
+
+  component.unmount();
+
+  await delay(100);
+
+  expect(disconnectSpy).toHaveBeenCalled();
+
+  disconnectSpy.mockRestore();
```
toboid: This `mockRestore` is redundant as `window.IntersectionObserver` is being deleted in an after each hook above

packages/author-profile/__tests__/author-profile-content.web.test.js
```diff
@@ -197,3 +212,196 @@ it("renders good quality images if there is no IntersectionObserver", () => {
...
+
+  await delay(100);
+
+  // No work was done
+  expect(spy).not.toHaveBeenCalled();
+
+  spy.mockRestore();
+});
+
+it("does not set state after unmounting", async () => {
```
toboid: I wonder if we're getting much from this test as there is nothing that would cause a setState happening (we have to manually trigger with `window.IntersectionObserver.dispatchAll` in the other tests. Removing it doesn't seem to affect coverage

craigbilner: This would be a danger when doing this kind of optimisation, and something we want to ensure doesn't happen if any changes are made in the future. Coverage should just be a barometer

packages/author-profile/__tests__/author-profile-content.web.test.js
craigbilner: At the moment these tests are quit messy and can run in parallel, think it's a nice safety net to ensure another test hasn't managed to hold onto the reference and asserted on it itself?

packages/author-profile/__tests__/author-profile-content.web.test.js
craigbilner: We want to test a proper lifecycle being unmounted?

packages/author-profile/__tests__/author-profile-content.web.test.js
toboid: 🗣  agreed this does have value as observations of elements entering the viewport are debounced for 100ms, this ensures that we don't process any pending observation if the component is unmounted

PR #421, opened 2017-11-30
Feature/replat 604 article templates responsive

jimhunty:
article responsive

This was tested on chrome across the breakpoints set out in the production website against its code.

We are using Styled Components to make this responsive. We are using React Native Stylesheet first and only applying the required styled in Styled Components where needed.

Our aim was to keep everything to do with making the article responsive, in the article package. Keeping all other child packages pure. We have aimed to reduce the number of `.web.js` files.

We wanted to put Styled Components and React Native StyleSheets in the same folder but this upset `React-Native-Web` for unknown reasons.


packages/article-byline/article-byline-proptypes.js
```diff
@@ -8,10 +8,12 @@ export const articleBylinePropTypes = {
     container: ViewPropTypes.style,
     byline: Text.propTypes.style,
     link: Text.propTypes.style
-  })
+  }),
+  WrapperComponent: PropTypes.func
```
fourlastor: shouldn't this (and other `WrapperComponent`s) be `PropTypes.element`?

packages/article/article.web.js
```diff
@@ -0,0 +1,207 @@
...
+      standfirst,
+      label,
+      byline,
+      publishedTime,
+      publicationName,
+      leadAsset
+    } = articleData;
+    const LeadAssetMedia = ArticlePage.renderLeadAsset(leadAsset);
+    const contentArray = articleData.content.map((i, index) => ({
+      data: i,
```
fourlastor: maybe just rename `i` to `data`? 😄

packages/article/article.web.js
```diff
@@ -0,0 +1,207 @@
...
+      index
+    }));
+    const BodyView = contentArray.map(i => ArticlePage.renderBody(i));
```
fourlastor: what does `i` represent here?

packages/article/article.web.js
```diff
@@ -0,0 +1,207 @@
...
+  static getImageContainer(imageType) {
+    switch (imageType) {
+      case "primary":
+        return PrimaryImg;
+      case "secondary":
+        return SecondaryImg;
+      case "inline":
+        return InlineImg;
+      default:
+        return null;
```
fourlastor: is this an expected case? or can we throw to catch the error soon in development?

packages/article-byline/article-byline-proptypes.js
jimhunty: This causes proptype errors in React 16 - the suggestion is to now use func.

packages/article/article.web.js
jimhunty: Renamed this,

packages/article/article.web.js
jimhunty: There isnt really a default case so will just choose Primary.

aronshamash:
Couple of things while looking at it integrated into the website (render):
1. Headline font size too small
2. Headline padding needs increasing
3. Article Meta needs to line up with the top of the lead asset

image

jimhunty:
@aronshamash When testing this I see the issue with padding, we are missing the article container for the site. We should be able to resolve this.

The body text has definitely gone strange, but there are no missing fonts.

The pages doesn't scroll either...

I got the following message in my console when running Render:
`[BABEL] Note: The code generator has deoptimised the styling of "{path to site}/Times/cps-content-render/src/article-page.bundle.js" as it exceeds the max of "500KB".`

Im wondering could this be the cause of the last two problems?

toboid:
@jimhunty this error message is raised during the creation of a server-side bundle for server rendering of the article. The way this is done needs to be reviewed however I don't think it's the cause of your issue; its babel saying that it won't trim extra whitespace and line terminators because the file size is >500kb. It's the `compact` option explained on [this](https://babeljs.io/docs/usage/api/) page

jimhunty:
@aronshamash Those issues have now been addressed.

@toboid Thanks for that feedback. My original comment was based on not knowing a few things about `Render`, and therefore we can negate those now.

packages/article-byline/article-byline-proptypes.js
craigbilner: @jimhunty I don't understand `WrapperComponent`, why not just wrap-the-component?

packages/article-image/component.js
```diff
@@ -0,0 +1,73 @@
...
+import { View, StyleSheet } from "react-native";
+import Image from "@times-components/image";
+import Caption from "@times-components/caption";
+
+import {
+  articleImagePropTypes,
+  articleImageDefaultPropTypes
+} from "./article-image-proptypes";
+
+// NOTE: We support only Mobile web, ios and android. Desktop to follow after responsive design
```
craigbilner: not relevant anymore?

craigbilner:
merged mega PR, raised tech debt issues offline

PR #424, opened 2017-12-01
Chore/upgrade apollo

craigbilner:
Before we get too far behind with Apollo (which has changed significantly) and to ensure we are pushing forward with 2.x on each platform, this is the upgrade for `times-components`.

Testing still remains undocumented and is rather fiddly, so the shared code has been moved to a new `utils` package in the `graphql` sub-section. This also meant the schema in the `provider` package which is used for linting has been moved and the `FragmentMatcher`, which is exported for others to use when consuming a `provider`.

Dependencies have been fixed and cleaned up too.

Outstanding:
- [ ] Make `util` `PropTypes` stricter
- [ ] Look into linting issues


jimhunty:
Can we have the outstanding issues captured in a JIRA ticket?

craigbilner:
@jimhunty @fourlastor made prop types slightly stricter and `apollo` linting issues fixed [here](https://github.com/apollographql/apollo-client/commit/77d0bd77a0f935d72fc58e74775d130af0d54bee).

Also raised #428 to tidy up linting further

PR #425, opened 2017-12-01
REPLAT-671: Removing the Webview in Android

jimhunty:
screen shot 2017-12-01 at 14 38 48

I removed all the webview code and which meant that the native styles kicked in on Android.

This is waiting for a design review from Jake before WIP can be turned off.


packages/article-label/style/index.android.js
```diff
@@ -0,0 +1,11 @@
+import sharedStyles from "./shared";
+
+const styles = {
+  title: {
+    ...sharedStyles.title,
+    paddingTop: 5,
+    marginBottom: -4
```
jimhunty: Required to position text for the design review. Only applied on Android.

packages/article/styles/header/index.android.js
```diff
@@ -12,7 +12,9 @@ const androidStyles = {
   },
   standFirst: {
     ...sharedStyles.standFirst,
-    lineHeight: 29
+    lineHeight: 29,
+    paddingBottom: 15,
+    marginTop: -7
```
jimhunty: Required to position text for the design review. Only applied on Android.

PR #427, opened 2017-12-03
REPLAT-922: Making flag font lowercase and changing height of diamond

nehasri89:
Snapshot of **before** and **after** for flags
before
after

### DIFF
--- a/packages/article-flag/article-flag.js
+++ b/packages/article-flag/article-flag.js
@@ -29,17 +29,17 @@ const styles = {
 const beautifyTitle = title => {
   if (Platform.OS === "android") {
     return title
-      .toUpperCase()
+      .toLowerCase()
       .split("")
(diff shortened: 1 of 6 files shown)


packages/article-flag/article-flag.js
```diff
@@ -29,17 +29,17 @@ const styles = {
 const beautifyTitle = title => {
   if (Platform.OS === "android") {
     return title
-      .toUpperCase()
+      .toLowerCase()
```
fourlastor: are we sure we don't want it capitalised?

aronshamash: the lowercase version of the font is still CAPITALISED!

fourlastor: oh small caps, 👍

PR #428, opened 2017-12-04
chore: customise eslint dev paths

craigbilner:
The `eslint` disabling was getting a little out of hand. While the storybook is strictly not a dev dependency it certainly acts like one is a lot tidier to be made one. Also establishes a directory pattern for our packages and allows other dev deps to be pulled in without `eslint` worry.


craigbilner:
@fourlastor yeah, I think as it expands having a `times-components/lint` might be sensible

PR #433, opened 2017-12-04
Chore: REPLAT-922: Change fixtures as per design review

nehasri89:
Addressing the comments in design review

- Fixture image ratio should be 16:9 or 3:2 (Inline primary image)
- Fixing fixture for provider data story, which was related to the iamge


jimhunty:
As these are all fixture changes primarily, has the API had an update?

aronshamash:
@jimhunty I wouldn't have thought there would need to be a change to TPA as the crops are decided in Methode.

craigbilner:
@aronshamash don't think TPA provided these?
@nehasri89 I don't think resize should be provided by the API? We don't provide any client information for it to make this sort of decision?

nehasri89:
@craigbilner yes, you are right, I will remove the resize.

PR #434, opened 2017-12-05
Author head styling

toboid:
* Move to new cross platform testing style. This is in the first commit with no other changes so the subsequent commits will have just the relevant snapshot updates for that change
* Update display of author title - reduce size by 1 and add antialiasing on web
* Add twitter icon
* Fix author-head stories on android (were showing blank)
* Add responsive max-width for bio. This includes adding `withResponsiveSyles` if we're happy with this approach I can extract to be it's own package either here or in the next PR so that I can then use in author-profile package

Changes to move profile picture below bio on mobile and to equalise padding between elements will be done in a subsequent PR.

# Screenshots:
### Before
image
image

### After
image
image

### Before
image

### After
image

### Before
(blank)
image

### After
image


packages/author-head/author-head.js
```diff
@@ -33,26 +37,21 @@ const styles = StyleSheet.create({
...
   },
-  title: {
-    fontFamily: "TimesDigitalW04-RegularSC",
-    fontSize: 15,
-    color: "#696969"
-  },
   twitter: {
+    paddingTop: 16,
+    flexDirection: "row",
+    ...Platform.select({
```
jimhunty: This isn't necessary anymore. A seperate android stylesheet should be added. See article package for details.

packages/author-head/author-head.js
```diff
@@ -1,11 +1,15 @@
...
 import { TextLink } from "@times-components/link";
 import { withTrackEvents } from "@times-components/tracking";
-import { renderTrees, treePropType } from "@times-components/markup";
+import { treePropType } from "@times-components/markup";
+
+import AuthorTitle from "./author-title";
+import TwitterIcon from "./twitter-icon";
+import Bio from "./author-bio";
 
 const styles = StyleSheet.create({
```
jimhunty: Again personal preference, but it would be good to get this into a styles file like other packages.

packages/author-head/author-bio.web.js
```diff
@@ -0,0 +1,11 @@
+import Bio from "./author-bio.base";
```
jimhunty: I think we did this on the Article, but I think you can return the content of this file in `author-bio.js`. Because RN doesnt have the concept of media queries, they are ignored and you can still pass the style to your Styled Component as its a Text.

packages/author-head/__tests__/ios/__snapshots__/author-head.ios.test.js.snap
```diff
@@ -125,45 +127,68 @@ exports[`renders with data 1`] = `
...
+        }
+      >
+        @
+        carolmidgley
+      </Text>
+    </View>
     <Text
       accessible={true}
       allowFontScaling={true}
+      className=""
```
jimhunty: Blank class names

packages/author-head/author-bio.web.js
toboid: I'm not sure it's desirable to from a code maintenance point of view to be applying media queries to components used on native, wouldn't that be confusing? However we can save that for another day as there is now a different impl for native/web

packages/author-head/author-title.base.js
```diff
@@ -0,0 +1,34 @@
...
+    {title.toLowerCase()}
+  </Text>
+);
+
+Title.defaultProps = {
+  title: "",
+  className: ""
+};
+
+Title.propTypes = {
```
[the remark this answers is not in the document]
toboid: It's defaulting to empty string which is the existing behaviour

packages/author-head/author-title.web.js
```diff
@@ -0,0 +1,7 @@
+import styled from "styled-components";
+
+import AuthorTitle from "./author-title.base";
+
+export default styled(AuthorTitle)`
```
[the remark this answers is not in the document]
toboid: The convention being used elsewhere in this project is to colocate styles with the component rather than trying to separate based on type (separation of concerns not categories). In general I think this is preferable unless things are becoming unwieldy or there is a specific reason such as re-using for web/native. In this case I think it would be overkill to pull one declaration into a separate file.

packages/author-profile/__tests__/__snapshots__/author-profile-content.native.test.js.snap
```diff
@@ -273,45 +273,68 @@ exports[`renders profile content 1`] = `
               accessible={true}
               allowFontScaling={true}
               aria-level="2"
+              className=""
```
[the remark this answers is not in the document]
toboid: no longer using className

packages/author-head/author-title.base.js
```diff
@@ -0,0 +1,34 @@
...
+    fontSize: 14,
+    color: "#696969"
+  }
+});
+
+const Title = ({ title, className }) => (
+  <Text
+    className={className}
+    accessibilityRole="heading"
+    aria-level="2"
```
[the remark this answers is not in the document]
toboid: Yes agreed we should start doing this, it was mentioned a few weeks ago but not sure if it's on a board

packages/author-head/author-head.js
toboid: I would prefer to leave this as-is as I'm not sure it would improve readability and I don't think it's really in the scope of this PR. Interested to see what @craigbilner thinks though

packages/author-head/__tests__/ios/__snapshots__/author-head.ios.test.js.snap
toboid: Gone now

packages/author-head/author-title.base.js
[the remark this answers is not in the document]
toboid: actually yeah I agree, will add

packages/author-head/author-bio.web.js
craigbilner: I would agree we should have either universal components or separate "wet" components both for grokking and maintenance

packages/author-head/author-head.js
craigbilner: @jimhunty @toboid consistency vs readability...case by case basis for me. I think consistency is important for productivity by reducing cognitive overload but at less than 150 lines having everything in the same file is probably easier to onboard onto?

packages/author-head/author-head.js
craigbilner: @jimhunty I think this is the one @toboid asked about and similar to the other comment, perhaps a separate stylesheet is overkill?

packages/author-head/author-bio.web.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { View } from "react-native";
+import Bio from "./author-bio.base";
+import withResponsiveStyles from "./responsive-styles";
+
+const BioContainer = withResponsiveStyles(View, {
```
craigbilner: Makes me think of nuclear waste or a yoghurt pot...

packages/author-head/author-title.base.js
```diff
@@ -0,0 +1,34 @@
...
+
+const styles = StyleSheet.create({
+  title: {
+    fontFamily: "TimesDigitalW04-RegularSC",
+    fontSize: 14,
+    color: "#696969"
+  }
+});
+
+const Title = ({ title = "", style }) => (
```
craigbilner: minor: defaulting a required prop?

packages/author-head/author-title.web.js
```diff
@@ -0,0 +1,6 @@
+import React from "react";
+import AuthorTitle from "./author-title.base";
+
+export default props => (
+  <AuthorTitle {...props} style={{ WebkitFontSmoothing: "antialiased" }} />
```
craigbilner: do the story styles need updating to add the `auto` overrides for consistency?

toboid: Make sense, i'll add the extra h*n* styles from render

packages/author-head/author-bio.web.js
toboid: ha yeah me too

PR #438, opened 2017-12-06
Adding left padding for inline primary image

nehasri89:
As per design review, the inline primary image should have 10px padding on mobile web and native
 
**Before** 
screen shot 2017-12-06 at 16 23 10

**After**
screen shot 2017-12-06 at 16 23 21


jimhunty:
@nehasri89 It would be worth waiting until #421 is merged into master before this gets reviewed as this depends on it.

nehasri89:
yes @jimhunty I agree

nehasri89:
Opened another Pr just with the caption changes

PR #442, opened 2017-12-07
feat: make inline image captions responsive

nehasri89:
As part of design review, fix the left padding for primary inline images for mob web and native. 

As part of this change we have introduced media queries for caption to add a 10px left padding for mobile web. We have used styled components for caption when it is associated with inline Primary image only.
The style change is applied to `article-image` package which is the consumer of Caption component. Inline images have three variations `inline`, `secondary`, `primary`. We have created a `caption-component` file in `article-image` package which handles the image variations for `secondary` and `inline` and a .web variation for a `primary` inline image where we had to apply the media query for web

**Before**
screen shot 2017-12-06 at 16 23 10

**After**
screen shot 2017-12-07 at 15 16 06

### DIFF
--- a/packages/article-image/__tests__/android/__snapshots__/article-image.android.test.js.snap
+++ b/packages/article-image/__tests__/android/__snapshots__/article-image.android.test.js.snap
@@ -1,8 +1,14 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Article Image test on android does not render Article Image display is not received 1`] = `
+exports[`Article Image test on android Renders Article Image with no caption 1`] = `
 <View
-  style={undefined}
+  style={
+    Object {
+      "flexDirection": "column",
+      "paddingBottom": 25,
+      "width": "100%",
+    }
+  }
 >
   <View
     style={undefined}
@@ -99,6 +105,13 @@ exports[`Article Image test on android does not render Article Image display is 
       </View>
     </View>
   </View>
+</View>
+`;
+
+exports[`Article Image test on android does not render Article Image if display is not received 1`] = `
+<View
+  style={undefined}
+>
   <View
     style={undefined}
   >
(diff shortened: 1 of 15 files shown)


packages/article-image/captionComponent.js
```diff
@@ -0,0 +1,39 @@
+import Caption from "@times-components/caption";
```
toboid: filename should be `caption-component.js`

packages/article-image/caption-primary.web.js
```diff
@@ -0,0 +1,32 @@
+import Caption from "@times-components/caption";
+import styled from "styled-components";
+import { View } from "react-native";
+import PropTypes from "prop-types";
+import React from "react";
+
+const mediumBreakpoint = "768px";
```
craigbilner: need to align with @toboid's responsive package - potentially it exports enums?

packages/article-image/component.js
```diff
@@ -1,7 +1,8 @@
 import React from "react";
 import { View, StyleSheet } from "react-native";
 import Image from "@times-components/image";
-import Caption from "@times-components/caption";
+import Caption from "./captionComponent";
```
craigbilner: so we were using the caption component and now using a different caption implementation?

nehasri89: Added more info in description reg the same

packages/article-image/__tests__/android/jest.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = Object.assign(jestConfigurator("article-image", "android"), {
   collectCoverageFrom: [
     "**/packages/article-image/*.js",
     "!**/packages/article-image/*stories.js",
-    "!**/packages/article-image/*.web.js"
+    "!**/packages/article-image/*.web.js",
+    "!**/packages/article-image/caption-primary.*"
```
craigbilner: I don't think we do want to exclude `caption-primary`?

PR #454, opened 2017-12-11
chore: fix coverage

craigbilner:
Many packages had gone awry with their coverage and testing. Components will need to specify which files they'd like coverage from based on the platform and will need to ignore certain files, this will make the coverage more accurate and the CI run faster


jimhunty:
In general it all seems fine. There are patterns in the file patterns that we collect and not collect coverage from which could be used to reduce some of the code going in. But it doesnt need to be done as part of this PR.

PR #457, opened 2017-12-11
BREAKING CHANGE: add error boundaries

craigbilner:
Change the `error-view` component to be more reusable (and safe) by allowing the consumer to provide the child based on the error boundary

Add example to `author-profile` where an error rendering the card will result in the card being removed

Todo:

- [ ] Add more tests
- [ ] Add error images from Zeppelin


toboid:
Rebased, fixed tests and added missing tests for the existing functionality. I suggest that we merge this and address error states for failed author and failed articles requests as a separate PR.

PR #465, opened 2017-12-12
feat: make author name responsive on web

toboid:
* Make author name responsive on web and fixed (smaller size) on native to match the web small size

# Web
## Desktop
- No change
### Before
image

### After
image

## Mobile 
- Smaller font size for author name
### Before
image
image

### After 
image
image

# Android
- Smaller font size for author name
## Before
image
image

## After
image
image

# iOS
- Smaller font size for author name
## Before
image
image

## After
image
image


packages/author-head/author-name.js
```diff
@@ -0,0 +1,30 @@
...
+    fontSize: 30,
+    color: "#000",
+    paddingTop: 32
+  }
+});
+
+const AuthorName = ({ name }) => (
+  <Text
+    testID="author-name"
+    accessibility-label="author-name"
```
jimhunty: Cant have a hyphen here?

packages/author-head/author-name.web.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
```
jimhunty: You can actually get away with only using one file here. If you take the web version and pass through the native styles as a prop, it will work as expected across all three platforms.

packages/author-head/author-name.web.js
```diff
@@ -0,0 +1,31 @@
...
+  `,
+  mediumUp: () => "font-size: 45px;"
+});
+
+const AuthorName = ({ name }) => (
+  <AuthorNameWrapper
+    testID="author-name"
+    accessibility-label="author-name"
+    accessibilityRole="heading"
+    aria-level="1"
```
jimhunty: Hypens shouldnt work?

packages/author-head/author-name.web.js
craigbilner: @jimhunty would we want to make native components `styled-components`?

jimhunty: To reduce the duplication and get the same result.

packages/author-head/author-name.js
toboid: Good spot, this was copied over from author-head as-is but needs correcting

packages/author-head/author-name.web.js
toboid: Do mean use the `.web` component (which has been styled with styled components) but rely on those styles not being applied on native and then apply the native styles via a style prop with RN styles? 

That would work. I can see it both ways but I'd prefer to keep them separate as I think it's clearer exactly what is happening and what styles are being applied where and the duplication is minimal. Also since styled components can support RN i'm hesitant to rely on these styles always being ignored - it's conceivable that in future there could be different versions of `withResponsiveStyles` for web and native.

packages/author-head/author-name.web.js
toboid: This was copied from author-head. From React 16 unknown attributes are passed along to the DOM so this will be present on web. However, looking at the author profile, we have aria level on the author name (level 1)  and title (level 2), since these are already using `h1` and `h2` respectively I don't think we need them so happy to remove (this attribute isn't supported by RN as far as I can see from the docs)

toboid:
Updated accessibility attributes

packages/author-head/author-name.web.js
jimhunty: It is what I meant. Because the styles are inline, they should always be considered a higher importance than the Styled Components, leaving styled components picking up the rest.

It could be considered confusing and separation does make it easier to understand.

PR #467, opened 2017-12-13
fix: remove unnecessary padding on article component

aronshamash:
For the wide breakpoint on desktop there was unnecessary padding.

Easier to show when integrated into render.

# Before

Notice alignment of marketing header and article lead image.

image

# After

image


jimhunty:
@craigbilner It can be. It seemed unnecessary at the time to add it in, but that should help improve the coverage.

jimhunty:
Manually testing in Latest Firefox, Chrome, Safari, Edge and IE11. The expected behaviour was observed

PR #468, opened 2017-12-13
feature/video-error-handler

dickie81:
giphy

### DIFF
--- a/packages/brightcove-video/__tests__/brightcove-video.test.js
+++ b/packages/brightcove-video/__tests__/brightcove-video.test.js
@@ -179,3 +179,24 @@ it("will call native fullscreen player if 'directToFullscreen option passed'", (
 
   expect(mockNativeModule.playVideo.mock.calls).toHaveLength(1);
 });
+
+it("will handle an error properly", () => {
+  const root = renderer.create(
+    <BrightcoveVideo
+      policyKey={policyKey}
+      videoId={videoId}
+      accountId={accountId}
+      poster={{
+        uri:
+          "data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs="
+      }}
+      autoplay
+    />
+  );
+
+  const rootInstance = root.getInstance();
+
+  rootInstance.handleError({ code: "XxX", message: "a bad booboo occured" });
+
+  expect(root.toJSON()).toMatchSnapshot();
+});
--- a/packages/brightcove-video/__tests__/brightcove-video.test.js
+++ b/packages/brightcove-video/__tests__/brightcove-video.test.js
@@ -200,3 +200,31 @@ it("will handle an error properly", () => {
 
   expect(root.toJSON()).toMatchSnapshot();
 });
+
+it("the player can trigger errors", () => {
+  const root = renderer.create(
+    <BrightcoveVideo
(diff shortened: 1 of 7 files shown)


packages/brightcove-video/__tests__/brightcove-video.test.js
```diff
@@ -179,3 +179,24 @@ it("will call native fullscreen player if 'directToFullscreen option passed'", (
...
+        uri:
+          "data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs="
+      }}
+      autoplay
+    />
+  );
+
+  const rootInstance = root.getInstance();
+
+  rootInstance.handleError({ code: "XxX", message: "a bad booboo occured" });
```
craigbilner: @dickie81 can we not generate an error more genuinely than this? We can use the programmatic play on a broken video?

packages/brightcove-video/brightcove-video.js
```diff
@@ -17,19 +18,23 @@ class BrightcoveVideo extends Component {
     super(props);
 
     this.state = {
-      isLaunched: props.autoplay
+      isLaunched: props.autoplay,
```
craigbilner: minor: can destructure

packages/brightcove-video/brightcove-video.js
```diff
@@ -69,16 +74,27 @@ class BrightcoveVideo extends Component {
     this.props.onFinish();
   }
 
+  handleError(err) {
+    this.setState({ error: err });
```
craigbilner: minor: can use shorthand

packages/brightcove-video/error.js
```diff
@@ -0,0 +1,97 @@
...
+);
+
+VideoError.defaultProps = {
+  poster: null
+};
+
+VideoError.propTypes = {
+  onReset: PropTypes.func.isRequired,
+  poster: SourcePropType,
+  width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
```
craigbilner: could these reuse the native proptypes?

packages/brightcove-video/__tests__/brightcove-video.test.js
dickie81: @craigbilner - I have added another test to make sure errors in the `brightcove-player` trigger errors in the wrapping `brightcove-video` component but I don't think it would be a great idea to write tests that rely on Brightcove's API & therefore a stable internet connection.

packages/brightcove-video/error.js
dickie81: I'm not sure they can as only height and width are passed not a whole style property.

craigbilner: sorry, just meant instead of `oneOfType` which is quite custom, resuse the `width` type from FB?

packages/brightcove-video/__tests__/brightcove-video.test.js
craigbilner: @dickie81 wouldn't need an internet connection? You have tests/stories which create "broken videos" where you could invoke play on them and test the error was handled? I'd just rather test behaviour rather than implementation unless absolutely necessary?

kaliabadi:
test

PR #474, opened 2017-12-19
feat: toggle author profile images

craigbilner:
DON'T SQUASH

There is a new requirement to be able to toggle images off for certain authors who effectively don't have lead assets for their articles or they're not of high enough quality.

`hasLeadAssets` has been added to TPA already.

This PR introduces a new provider which is chosen based on the author's `hasLeadAssets` flag which will ignore lead assets and select summaries of different length. For the web, depending on the width of the viewport, either long or short text will be shown. This is 360 chars for tablet up and 220 otherwise, for cards with images, the char count remains at 145.

To achieve this decoupling the card component has been refactored to be agnostic to it's textual content, and author profiles just happen to use article summaries.

In addition a few bugs have been fixed in article summary and pagination. Article summary was concatenating the teaser onto the end of the previous paragraphs and pagination wasn't limited to the total count.


packages/author-profile/author-profile-item.js
```diff
@@ -45,15 +52,19 @@ const AuthorProfileItem = item => {
...
-          date={publishedTime}
-          label={label}
-          publication={publicationName}
-        />
+          showImage={showImage}
+        >
+          <ArticleSummary
+            label={label}
+            headline={headline}
+            text={showImage ? summary : shortSummary}
```
fourlastor: I think this should be inverted (short summary with image, long without)

packages/author-profile/author-profile.js
```diff
@@ -31,11 +36,36 @@ const AuthorProfile = ({
     return <AuthorProfileError {...error} />;
   }
 
-  const { biography, name, image: uri, jobTitle, twitter, articles } =
+  if (isLoading) {
+    return (
+      <AuthorProfileContent
+        isLoading={isLoading}
```
fourlastor: isn't this always true in here?

packages/author-profile/author-profile.js
```diff
@@ -31,11 +36,36 @@ const AuthorProfile = ({
...
+        showImages
+        pageSize={initPageSize}
+        imageRatio={ratioTextToFloat("3:2")}
+        articlesLoading
+        onTwitterLinkPress={() => {}}
```
fourlastor: shouldn't this be exposed as a prop?

packages/card/card-loading.web.js
```diff
@@ -44,11 +44,13 @@ const Loading = ({ imageRatio }) => {
 };
 
 Loading.defaultProps = {
-  imageRatio: 3 / 2
+  aspectRatio: 3 / 2,
+  showImage: PropTypes.bool
```
fourlastor: shouldn't this be `true/false` since it's the default prop value?

craigbilner: oops, ta

packages/author-profile/author-profile.js
craigbilner: atm it's required but not necessary for the loading state

packages/author-profile/author-profile.js
craigbilner: yep :+1:

packages/author-profile/author-profile-item.js
craigbilner: think that might get even more confusing having shortSummary, summary, longSummary?

fourlastor: I think I misexplained myself, with long summary I meant `summary`, the idea is that (logically speaking), I thought you'd show `summary` without image (because you have more space, and `shortSummary` with the image, but here the logic says the opposite (`summary` with image, `shortSummary` with no image)

toboid:
todo
- [ ] fix author profile - loading (native) story

packages/author-profile/author-profile-item.js
craigbilner: will rename to `image` and `noImage` variants in separate PR

PR #480, opened 2017-12-19
Support preview images

fourlastor:
This PR introduces support (on React Native only, for now) for preview images.

The main issue with `Image` and `ImageBackground` is that multiple sources urls are defined to support different viewports, and there's no standard way to have a downsampled images as preview images out of the box.

The solution we came up with is: we have 2 `ImageBackground`s and switch between them whenever the higher resolution one is loaded.

The way a preview image is tagged right now is adding `&preview=true` to the url, which is something up for discussion (we might have it as a parameter of `ActivatePreviewImages` as a transform function, for example).

This implies that on unconfigured clients, there will be twice the number of requests, so we created `ActivatePreviewImages` which will provide a context to enable this feature, and leaving the current functionality in place if not used.

A way of activating the thing is:

```javascript
const MyRootPage = activatePreviewImage(MyRootView)

// use MyRootPage
```

Testing is currently missing, can we work together to see what we can do for that? Not a lot of experience testing react 👅 

Paired with @tuncaulubilge


L0wry:
Dextrose has taken screenshots of all your changed component stories. You can see them here 
`https://www.bitrise.io/build/3181876624e7f4a3`

From a quick review it looks like 
`AuthorProfileLoading.ios.png`
is not working as expected

Also, as this is the first PR that dextrose has run against we need to add the error stories to Dextrose ignored stories. 
Can you grab me when you see this and we can do it together!

PR #481, opened 2017-12-19
Bump storybook version to 3.2.18

BernieSumption:
3.2.18 includes a fix for a bug that broke storybook in IE11: https://github.com/storybooks/storybook/pull/2428

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,9 +93,9 @@
   },
   "homepage": "https://github.com/newsuk/times-components#readme",
   "devDependencies": {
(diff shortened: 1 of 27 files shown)


L0wry:
Dextrose has taken screenshots of all your changed component stories. You can see them here
https://www.bitrise.io/build/32043da1a0b02429

As this is one of the first PR's that dextrose has run against we need to add the error stories to Dextrose ignored stories.
Can you grab me when you see this and we can do it together!

Issue #482, opened 2017-12-20
Date Publication component issue


killanaca:
**Do you want to request a _feature_ or report a _bug_?**

**What is the current behaviour?**
bug
**If the current behaviour is a bug, please provide the steps to reproduce.**
http://components.thetimes.co.uk/?selectedKind=DatePublication&selectedStory=standard%20DatePublication&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel

**What is the expected behaviour?**
the date publication component should use this format 
"July 7 2016, 12:01am, The Times"

**If this is a feature request, what is the motivation or use case for changing
the behaviour?**
as per design
**Please mention other relevant information such as the browser version, Node.js
version and Operating System.**

killanaca:
new design

PR #485, opened 2017-12-20
Update date-publication component format

killanaca:
ref https://github.com/newsuk/times-components/issues/482


craigbilner:
@killanaca, "fix" is quite subjective. This format was purposefully chosen because the time is garbled and doesn't mean an awful lot. It's also inconsistent across the app/site. In fact it's manually generated I believe for each edition.

killanaca:
fix it's just related to the design guidelines that we should follow.

killanaca:
documented the new logic in the ticket

PR #487, opened 2017-12-21
Android Ripple effect for links

tuncaulubilge:
Replaced TouchableOpacity in Link package with TouchableNativeFeedback on Android. 

Updated tests for link to support platform specific testing.

TouchableNativeFeedback enables native ripple animations instead of the current opacity animation. See the gif below:

ripple


craigbilner:
@tuncaulubilge should we use [this](https://github.com/react-community/react-native-platform-touchable)?

PR #488, opened 2017-12-22
fix: ensure consistent scrolling behaviour on author profile

BernieSumption:
The Next and Prev buttons should scroll to the top of the page when
clicked. Previously, they would scroll upwards, but either scroll too
far or not far enough. This was because the layout was recalculated
while the scroll was happening. This change delays the scroll until
after the layout so that it always scrolls to the top.

**iOS before fix after clicking "Next"**

**iOS after fix after clicking "Next"**

**android before fix after clicking "Next"**

**android after fix after clicking "Next"**

**web after fix after clicking "Next"**

(web never had an issue, but this shows it doesn't regress)

### DIFF
--- a/packages/author-profile/author-profile-content.js
+++ b/packages/author-profile/author-profile-content.js
@@ -104,7 +104,7 @@ class AuthorProfileContent extends React.Component {
           offset: 0,
           animated: true
         });
-      }, 0);
+      });
     };
 
     const paginationComponent = (hideResults = false) => (
--- a/packages/author-profile/author-profile-content.js
+++ b/packages/author-profile/author-profile-content.js
@@ -38,6 +38,10 @@ class AuthorProfileContent extends React.Component {
     this.onViewableItemsChanged = this.onViewableItemsChanged.bind(this);
   }
 
+  componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


times-tools:
/Users/vagrant/deploy

times-tools:
https://www.bitrise.io/build/bebb8269adf21512

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/952b5c72aa9470c1

packages/author-profile/author-profile-content.js
```diff
@@ -94,23 +98,26 @@ class AuthorProfileContent extends React.Component {
       );
     }
 
+    const scrollToTopNextFrame = () => {
+      this.scrollAnimationFrame = global.requestAnimationFrame(() => {
+        this.listRef.scrollToOffset({
+          offset: 0,
+          animated: true
+        });
+      }, 0);
```
craigbilner: does this take a second arg?

BernieSumption: Good lord no, no it doesn't.

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/eea22845eda84f09

PR #493, opened 2018-01-03
chore: use component for story

craigbilner:
Help out dextrose and people alike with a working story. On the whole stories should use the component of the package rather than a specific potion of it


times-tools:
SOME COMMENT

times-tools:
SOME COMMENT

times-tools:
http://google.com

PR #494, opened 2018-01-04
Update link test strategy to follow the recommended pattern

jimhunty:
Due to the Link package being very different depending on whether you are on web or iOS, it seemed prudent to make sure the test strategy matched it and that we had snapshot tests that would cover all the scenarios for all the platforms.


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/afdf9fb2f7d115e9

packages/link/__tests__/shared.js
```diff
@@ -5,6 +5,17 @@ import renderer from "react-test-renderer";
 
 export default (Link, TextLink, Text) => {
   describe("Link", () => {
+    it("renders correctly", () => {
```
fourlastor: isn't this the same test of `renders correctly with children`?

jimhunty: Link vs TextLink

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/7cddbb4146394d1e

PR #496, opened 2018-01-04
Author profile design changes

tuncaulubilge:
Updated Author profile page based on the design review with Jake. (REPLAT-1126)
Major change is that author head is moved above the bio and author name for all platforms.

screen shot 2018-01-04 at 14 43 18


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/3f1eed6942518a7c

packages/author-head/author-bio.js
```diff
@@ -1,13 +1,13 @@
...
 import Bio from "./author-bio.base";
 
 const styles = StyleSheet.create({
-  bio: { maxWidth: "88%", paddingBottom: 32 }
+  bio: { paddingLeft: 10, paddingRight: 10, paddingBottom: 32 }
 });
 
 export default props => (
-  <Text style={styles.bio}>
+  <View style={styles.bio}>
```
aronshamash: Out of interest, why did you need to change to `<View>` ?

tuncaulubilge: This view is just used as a container and the actual <Text> view is inside the <Bio> component. On android, using <Text> as the parent was causing a bug where the text styles were not passed to the child Text view.

aronshamash: Ok thanks

PR #497, opened 2018-01-04
REPLAT-1167 Article Snapshots

jimhunty:
Jest by default in its haste will set the `defaultPlatform` to iOS. This means when having `index.js`, `index.ios.js` and `index.android.js`, when the web tests are run, they will use the `index.ios.js` file.

As a result this PR is to address this problem and have the correct snapshots running.


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/27d89885aff4fde4

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/42a7955ef32a5cd6

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/ce7e3e42a9c65bc3

PR #499, opened 2018-01-05
Feat: GMT support

killanaca:
- REPLAT-655
- REPLAT-1155

### DIFF
--- a/packages/date-publication/__tests__/date-publication.test.js
+++ b/packages/date-publication/__tests__/date-publication.test.js
@@ -10,9 +10,17 @@ import DatePublication from "../date-publication";
 
 Enzyme.configure({ adapter: new React16Adapter() });
 
-describe("Date Publication test", () => {
+function gmtTests(gmtOffset, dateFormat) {
+  const { getTimezoneOffset } = Date.prototype;
+
+  beforeEach(() => {
+    global.Date.prototype.getTimezoneOffset = jest
+      .genMockFunction()
+      .mockReturnValue(gmtOffset); // eslint-disable-line no-extend-native
+  });
+
   const props = {
-    date: new Date("2017-07-01T14:32:00.000Z"),
+    date: new Date("2017-07-01T14:35:00.000Z"),
     publication: "TIMES"
   };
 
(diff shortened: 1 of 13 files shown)


packages/date-publication/date-publication.js
```diff
@@ -1,13 +1,34 @@
...
+  constructor(props) {
+    super(props);
+    const offset = new Date().getTimezoneOffset();
+    this.state = {
+      gmt: offset !== 0 ? " GMT," : ","
+    };
+  }
+  render() {
+    const { date, publication } = this.props;
+    // use the date as an UTC/GMT date (due to date parse behaviour)
```
craigbilner: whaht is UTC/GMT?

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/b3cca069744d85c0

packages/date-publication/date-publication.js
```diff
@@ -1,13 +1,34 @@
...
 };
-
-const DatePublication = ({ date, publication }) =>
-  `${format(date, "dddd MMMM DD YYYY")}, ${publications[publication]}`;
+class DatePublication extends React.Component {
+  constructor(props) {
+    super(props);
+    const offset = new Date().getTimezoneOffset();
+    this.state = {
+      gmt: offset !== 0 ? " GMT," : ","
```
craigbilner: is this the correct behaviour during BST?

packages/date-publication/__tests__/date-publication.test.js
```diff
@@ -34,7 +42,14 @@ describe("Date Publication test", () => {
...
     const component = shallow(<DatePublication {...props} />);
-    expect(component.text()).toContain(format(props.date, "dddd MMMM DD YYYY"));
+    const UTCDate = new Date(
+      props.date.getUTCFullYear(),
+      props.date.getUTCMonth(),
+      props.date.getUTCDate(),
+      props.date.getUTCHours(),
+      props.date.getUTCMinutes()
+    );
+    expect(component.text()).toContain(format(UTCDate, dateFormat));
```
toboid: Looks like we are reimplementing the formatting logic in the test here. For something like this is usually better to assert against a fixed value, so that it's absolutely clear what the correct value is and we're not implementing the same logic twice in different places

killanaca:
follow a new PR

PR #503, opened 2018-01-08
chore: refactor code coverage

craigbilner:
It's currently very onerous to to get coverage right for a package. While it's not completely generic and a little Time monorepo specific atm, this abstracts the coverage config away from the developer removing mistakes and PR back-and-forth.

`coverage.js` maps out the packages files, splits them by platform then groups them back together for the specific platform plus any "universal" files. This then makes the `collectCoverageFrom` array for the specific `jest` config.

Minor tweaks:

* Top level doesn't need to lint packages as well as themselves
* No value in publishing config files?
* If we stream the output from lerna child processes then we can see the actual running times for the tests on Travis
* lerna doesn't run the lifecycle scripts for all packages so forward `postinstall` onto `prepublishOnly` to get desired effect
* Rather than jsdoc added first piece of flow to times-components in `jest-configurator`
* Change old style scripts in `pull-quote`


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/2a015a5de61bac72

packages/jest-configurator/src/coverage.js
```diff
@@ -0,0 +1,77 @@
...
+    ...[...(fileMap.universal || [])]
+      .filter(file => !fileMap[platform] || !fileMap[platform].has(file))
+      .map(fileToCoveragePath(component, "universal"))
+  ];
+};
+
+// will return an array of paths which coverage will be chosen from
+// currently does not group and use wildcards with exclusions i.e.
+// is file specific and inclusive only
+module.exports = (
```
fourlastor: could we use `export default` instead?

craigbilner: ah yeah, I now transpile :-)

packages/jest-configurator/src/coverage.js
```diff
@@ -0,0 +1,77 @@
+// @flow
+
+const path = require("path");
+const recursive = require("recursive-readdir-synchronous");
+
+// it's quite hard to match on things that start with dot seemingly?
+// these would be config files we'd never check for coverage
+const configFiles = file => path.basename(file)[0] === ".";
```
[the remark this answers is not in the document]
craigbilner: ...better get that IE11 polyfill ready....

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/d97f66b71cfaa6fa

packages/jest-configurator/src/coverage.js
```diff
@@ -20,28 +24,50 @@ const staticFilesAndFolders = [
...
+const toValidExt = (ext: string): AllPlatform | null => {
+  switch (ext) {
+    case ".web":
+      return "web";
+    case ".android":
+      return "android";
+    case ".ios":
+      return "ios";
+    default:
+      return null;
```
fourlastor: shouldn't this be `universal` and remove the `| null` from the return type?

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/3695984b2676e5e5

packages/jest-configurator/src/coverage.js
```diff
@@ -20,28 +24,50 @@ const staticFilesAndFolders = [
...
+const toPlatformMap = (filePath: string) => (
+  fileMap: { [AllPlatform]: Set<string> },
+  file: string
+) => {
   const { dir, name } = path.parse(path.relative(filePath, file));
   const { name: subName, ext } = path.parse(name);
-  const platform = ext ? ext.substr(1) : "universal";
-  const files = fileMap[platform] || new Set();
+  const platform = toValidExt(ext);
+  const files = fileMap[platform || "universal"] || new Set();
```
fourlastor: if you remove the `| null` you can remove the `|| "universal"` here

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/508376c285bbe8db

PR #504, opened 2018-01-08
fix: Revert an accidentally deleted link.android.js file.

jimhunty:
@tuncaulubilge mentioned that I had accidentally removed a feature he had added. This was to add a ripple effect for Android links. This has now been reverted with the tests being updated.


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/29425bf9e8b4a1b0

packages/link/__tests__/web/link.web.test.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-env jest */
 import test from "../shared";
-import Link, { TextLink } from "../../link.web.js";
+import Link, { TextLink } from "../../link";
```
rogeriopvl: Shouldn't this be `import Link, { TextLink } from "../../link.web";` ?

jimhunty: I removed it because it works without the suffix and gives greater consistency between the code. The con is it could be confused.

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/f8e2c2f7ddafc00a

PR #508, opened 2018-01-08
Add refetch tests and refactor

toboid:
- Add tests for refetch
- Change `Wrapper` back to a functional component as the use of `retry` means that the wrapped component will now be rendered on refetch so this is no longer requried


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/fb751759f0d5df64

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/4de6e62dac16ea6d

PR #509, opened 2018-01-09
Chore/conventional commits

colincclark:
Running `yarn commit` instead of `git commit`ensures developers are guided through a commit message that follows the conventional commit standard: https://conventionalcommits.org/

Kept it optional rather than adding it to a pre-commit hook, although this can be done later if popular


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/6603a8fc6fd195d2

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/501e27485698244b

PR #510, opened 2018-01-09
Fix: Removed fetch-fonts from the build scripts so travis stops erroring

colincclark:
Removing the fetch-fonts step from the publish storybook step so travis stops failing at this point. As this occurs at the deploy stage, we can only test this once merged I think


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/a883d3d37907e37e

package.json
```diff
@@ -27,7 +27,7 @@
     "fetch-fonts": "node ./lib/fetch-fonts",
     "prestorybook": "yarn fetch-fonts && yarn storybook:build-vendor",
     "storybook": "start-storybook -p 9001 -c .storybook -s ./dist/public",
-    "prestorybook:build": "yarn prestorybook",
+    "prestorybook:build": "yarn storybook:build-vendor",
```
craigbilner: I think this has got quite confusing now - only the static storybook script uses this so doesn't need to be a pre script?

colincclark: I find all the pre scripts quite confusing. The other option here is:

`"storybook:build": "yarn storybook:build-vendor && build-storybook -c .storybook -s ./dist/public",`

is that what you had in mind @craigbilner ?

craigbilner: yah

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/62192d268eac9c0a

colincclark:
@craigbilner changes have been implemented

package.json
colincclark: This change was implemented

PR #511, opened 2018-01-09
chore: make Tealium a peer dependency so not installed on web

toboid:
Moved react-native to be a peer dependency inline with the other packages. This will mean it won't need to be installed on web


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/dcd84079b2864b87

PR #517, opened 2018-01-10
chore: refactor fixture generator

craigbilner:
Move the author profile fixture generator into the provider package which is responsible for data things. Use the mocks to run tests on the new providers and increase coverage.

n.b. made mock delay configurable


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/538751556483fda0

PR #519, opened 2018-01-11
Fix: Separate Responsive from the Native Components

jimhunty:
During the review of code coverage, we noticed that the Native components were needlessly using the Styled Components.

As a result, I have separated them out from the native code and made sure they only relate to the web specific code. I have also changed some files and folder names so that they all follow the same conventions.

Bonus: A small change to line height on Meta Text as it looked wrong

#### Web
screen shot 2018-01-11 at 12 31 29

#### Tablet
screen shot 2018-01-11 at 12 31 46

#### Mobile
screen shot 2018-01-11 at 12 31 56


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/d378b654df45cc13

packages/article/article-body/article-body-row.base.js
```diff
@@ -1,31 +1,13 @@
...
-    case "inline":
-      return InlineImg;
-    default:
-      return PrimaryImg;
-  }
-};
-
-const ArticleRow = props => {
-  const { data, index } = props.content;
+const ArticleRow = ({ content, imageWrapperFunction, PullQuoteWrapper }) => {
```
colincclark: these Props could benefit from PropTypes

packages/article/article-body/article-body-row.base.js
```diff
@@ -35,9 +17,11 @@ const ArticleRow = props => {
       );
     },
     image(key, attributes) {
-      const ImageContainer = getImageContainer(attributes.display);
+      const ImageWrapper = imageWrapperFunction
+        ? imageWrapperFunction(attributes.display)
```
colincclark: Perhaps destruct `attributes` as it's properties are accessed numerous times below

packages/article/article-body/article-body-row.base.js
```diff
@@ -49,21 +33,21 @@ const ArticleRow = props => {
               credits: attributes.credits
             }}
           />
-        </ImageContainer>
+        </ImageWrapper>
       );
     },
-    pullQuote(key, { content, caption: { name, twitter } }) {
+    pullQuote(key, attributes) {
+      const { name, twitter } = attributes.caption;
```
colincclark: attributes could also be destructed within this function

packages/article/article-body/article-body-row.base.js
```diff
@@ -49,21 +33,21 @@ const ArticleRow = props => {
...
+      const PW = PullQuoteWrapper || View;
```
colincclark: can this be named something more explicit?

packages/article/article-body/article-body-row.js
```diff
@@ -0,0 +1,6 @@
+import React from "react";
```
colincclark: do these new files / components need tests?

packages/article/article-body/article-body-row.base.js
```diff
@@ -1,31 +1,13 @@
...
-      return InlineImg;
-    default:
-      return PrimaryImg;
-  }
-};
-
-const ArticleRow = props => {
-  const { data, index } = props.content;
+const ArticleRow = ({ content, imageWrapperFunction, PullQuoteWrapper }) => {
+  const { data, index } = content;
```
craigbilner: minor: could just destructure and remove the brace block

packages/article/article-body/article-body-row.base.js
```diff
@@ -1,31 +1,13 @@
 import React from "react";
+import { View } from "react-native";
```
craigbilner: how much value is there in this file? Seems like web and native have diverged so much this now provides most web functionality which has to be made conditional?

packages/article/article-body/article-body-row.js
craigbilner: This abstraction has put a dint in the coverage because of previous comment

packages/article/article-body/article-body-row.base.js
craigbilner: As a pattern for "defaulting" I think I would use default params or prop types?

packages/article/article-header/article-header.base.js
```diff
@@ -46,40 +49,52 @@ const renderStandfirst = standfirst => {
   );
 };
 
-const renderLabel = label => {
+const renderLabel = (label, LabelWrapper) => {
```
craigbilner: think we should steer clear of this pattern, it's hard to understand/maintain. Think it's better for consumers to consume?

jimhunty: I disagree, its pretty simple to understand and maintain. It defaults to a View unless you pass in a different component for it to wrap with. I have also made it so that you can only pass through a LabelContainer or a View. What would you suggest to do differently?

packages/article/article-body/article-body-row.base.js
jimhunty: No, they are the same except for the need to wrap around the image and the pullquote in the renderTrees function. I can put them in two separate files but it means a lot of duplication

packages/article/article-header/article-header.base.js
craigbilner: it's called in once place, the wrapper should just be in situ?

packages/article/article-body/article-body-row.base.js
jimhunty: Agreed - moment of madness

packages/article/article-body/article-body-row.base.js
jimhunty: 👍

packages/article/article-body/article-body-row.js
jimhunty: Should be resolved in refactoring

packages/article/article-body/article-body-row.base.js
jimhunty: Now removed.

packages/article/article-body/article-body-row.base.js
jimhunty: 👍

packages/article/article-body/article-body-row.base.js
jimhunty: 👍

packages/article/article-header/article-header.base.js
jimhunty: So pass a prop to say that its web and use this within the file? That could be done. We couldnt use the suffix in this way, so it would have to be a prop.

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/17fdee126662d37a

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/a066fb15ce57d252

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/477ddbcc26f9209a

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/357fce30e48ef6f7

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/9df02eae4bf6faa6

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/5e3dca0b08c238ea

packages/article/article-header/article-header-label.js
```diff
@@ -0,0 +1,24 @@
...
+    </View>
+  );
+};
+
+HeaderLabel.propTypes = {
+  label: PropTypes.string
+};
+
+HeaderLabel.defaultProps = {
+  label: null
```
colincclark: type checking still occurs when defaultProps are called, should be empty string

packages/article/article-header/article-header-label.web.js
```diff
@@ -0,0 +1,29 @@
...
+    </LabelContainer>
+  );
+};
+
+HeaderLabel.propTypes = {
+  label: PropTypes.string
+};
+
+HeaderLabel.defaultProps = {
+  label: null
```
colincclark: should be `''` (more below)

packages/article/article-meta/article-meta-base.js
```diff
@@ -0,0 +1,58 @@
...
+
+const TextNode = PropTypes.shape({ text: PropTypes.string });
+
+const nodeShape = {
+  name: PropTypes.string.isRequired,
+  attributes: PropTypes.object.isRequired
+};
+
+nodeShape.children = PropTypes.arrayOf(
+  PropTypes.oneOfType([PropTypes.shape(nodeShape), TextNode])
```
colincclark: nice composition of PropTypes 👍

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/be8958399a52220d

PR #521, opened 2018-01-11
chore: increase tracking coverage

craigbilner:
Tracking was a little out of date in terms of test config so updated and added tests to increase coverage.

Fixed a bug where the child id wasn't being checked correctly.

### DIFF
--- a/packages/tracking/__tests__/tracking.web.test.js
+++ b/packages/tracking/__tests__/tracking.web.test.js
@@ -2,9 +2,13 @@ import { View, Text } from "react-native";
 import React from "react";
 import PropTypes from "prop-types";
 import renderer from "react-test-renderer";
+import Enzyme, { mount } from "enzyme";
+import React16Adapter from "enzyme-adapter-react-16";
 import { withTrackScrollDepth } from "../tracking";
 import withTestContext from "./test-tracking-context";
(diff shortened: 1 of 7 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/b7ec126a0a177f51

packages/tracking/__tests__/tracking.web.test.js
```diff
@@ -2,10 +2,14 @@ import { View, Text } from "react-native";
 import React from "react";
 import PropTypes from "prop-types";
 import renderer from "react-test-renderer";
+import Enzyme, { mount } from "enzyme";
```
colincclark: Might be worth using Jest `setupFiles` array to import these dependencies globally so we don't have to reproduce these imports in each test file? Probably a separate PR though...wdyt?

colincclark:
also, you can use the conventionalcommit type `test` rather than `chore`

packages/tracking/__tests__/tracking.web.test.js
colincclark: https://github.com/newsuk/times-components/pull/524 will finish this off if you think it is worth continuing with?

PR #523, opened 2018-01-12
feat: add debouncing to all graphql queries

BernieSumption:
BREAKING CHANGE: The connectGraphql function now has a mandatory
third parameter debounceTimeMs and will throw an error if it is not
provided.

This has been tested in /storybook(-native)?/, which uses mocked data. A second round of testing should be done when it's integrated into Render.

### DIFF
--- a/packages/provider/.eslintrc.js
+++ b/packages/provider/.eslintrc.js
@@ -10,8 +10,5 @@ module.exports = {
       }
     ]
   },
-  plugins: ["graphql"],
-  env: {
-    jest: true
-  }
+  plugins: ["graphql"]
 };
--- a/packages/provider/.eslintrc.js
+++ b/packages/provider/.eslintrc.js
@@ -10,5 +10,8 @@ module.exports = {
       }
     ]
   },
-  plugins: ["graphql"]
+  plugins: ["graphql"],
+  env: {
+    jest: true
+  }
 };
--- a/packages/provider/__tests__/debounce.test.js
+++ b/packages/provider/__tests__/debounce.test.js
@@ -1,0 +1,142 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Enzyme, { shallow, mount } from "enzyme";
+import React16Adapter from "enzyme-adapter-react-16";
+
+import withDebounce from "../debounce";
+
+Enzyme.configure({ adapter: new React16Adapter() });
+
+jest.useFakeTimers();
+
+// Alias the confusingly misnamed runTimersToTime to advanceTimersByTime
+// Jest has done this in v22, so this can be removed if we upgrade
+jest.advanceTimersByTime = jest.runTimersToTime;
+
+/* eslint react/prop-types: 0 */
+class Inner extends React.Component {
+  constructor(props) {
+    super(props);
+    this.numberOfDebouncedPropsUpdates = 0;
+  }
+
+  componentWillReceiveProps(props) {
+    if (this.props.debouncedProps !== props.debouncedProps) {
+      this.numberOfDebouncedPropsUpdates += 1;
+    }
+  }
+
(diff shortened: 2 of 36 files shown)


packages/provider/__tests__/provider.test.js
```diff
@@ -51,6 +51,10 @@ it("returns query result", done => {
   });
 });
 
+it("complains if you omit the debounceTimeMs parameter to connectGraphql", () => {
+  expect(() => connectGraphql(query)).toThrowError();
+});
+
```
toboid: Can see that the debounce hoc is tested but I think it would be useful to have a test to make sure that the debounce approach actually limits the queries executed by react-apollo, particularly given that it doesn't seem to be the most stable of libraries and behaviour could change. I think this _should_ be straight forward as will know how many queries have been executed by the number of mocks used

packages/provider/__tests__/provider.test.js
```diff
@@ -29,7 +29,7 @@ const mocks = [
   }
 ];
 
-const ConnectedComponent = connectGraphql(query);
+const ConnectedComponent = connectGraphql(query, undefined, 0);
```
toboid: Just an idea but given that this is a breaking change anyway and every call site will need to be updated, we could consider moving `propsToVariables` to be the last parameter so we don't have to pass `undefined`?

packages/provider/connect.js
```diff
@@ -15,7 +27,15 @@ const getQueryVariables = definitions =>
     )
   );
 
-const connectGraphql = (query, propsToVariables = identity) => {
+const connectGraphql = (query, propsToVariables = identity, debounceTimeMs) => {
```
colincclark: maybe default `debounceTimeMs = debounceTimeNone` to help avoid errors

packages/provider/debounce.js
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import getDisplayName from "react-display-name";
+import hoistNonReactStatic from "hoist-non-react-statics";
+
+const withDebounce = (WrappedComponent, debounceTimeMs) => {
```
colincclark: could also add a default to zero here for `debounceTimeMs`

packages/provider/__tests__/provider.test.js
craigbilner: also, humans aren't allowed to use `undefined`, only `null`

packages/provider/__tests__/debounce.test.js
```diff
@@ -0,0 +1,144 @@
...
+
+Enzyme.configure({ adapter: new React16Adapter() });
+
+jest.useFakeTimers();
+
+// Alias the confusingly misnamed runTimersToTime to advanceTimersByTime
+// Jest has done this in v22, so this can be removed if we upgrade
+jest.advanceTimersByTime = jest.runTimersToTime;
+
+/* eslint react/prop-types: 0 */
```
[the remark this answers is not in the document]
BernieSumption: I was doing this because the HOC contains logic relating to PropTypes, so adding them changes behaviour and was giving runtime failures, but I've rewritten some tests to work without having to suppress linter errors.

packages/provider/.eslintrc.js
```diff
@@ -10,5 +10,8 @@ module.exports = {
       }
     ]
   },
-  plugins: ["graphql"]
+  plugins: ["graphql"],
+  env: {
```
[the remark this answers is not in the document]
BernieSumption: Done.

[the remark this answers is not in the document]
colincclark: Standardized across all packages: https://github.com/newsuk/times-components/pull/530

BernieSumption:
Closing this, it's been superseded by #539

PR #527, opened 2018-01-12
feat: adds tealium reporting to article stories

colincclark:
Added an article tracking story so stakeholders can see what events are raised

### DIFF
--- a/packages/article/article.stories.js
+++ b/packages/article/article.stories.js
@@ -7,6 +7,8 @@ import { ArticleProvider } from "@times-components/provider";
 import { MockedProvider } from "@times-components/utils/graphql";
 import { addTypenameToDocument } from "apollo-utilities";
 import { query as articleQuery } from "@times-components/provider/article";
+import { withTrackingContext } from "@times-components/tracking";
+import { action } from "@storybook/addon-actions";
 import Article from "./article";
 
 const fullArticleTypenameFixture = require("./fixtures/full-article-typename.json");
@@ -140,4 +142,15 @@ storiesOf("Article", module)
   ))
   .add("Fixtures - No lead asset", () => (
     <Article {...articleFixtureNoLeadAsset.data} analyticsStream={() => {}} />
-  ));
+  ))
+  .add("Tracking", () => {
+    const ArticleWithTrackingContext = withTrackingContext(Article, {
+      trackingObject: "Story"
+    });
+    return (
+      <ArticleWithTrackingContext
+        {...fullArticleFixture.data}
+        analyticsStream={action("analytics-event")}
+      />
+    );
+  });
--- a/packages/article/article.stories.js
+++ b/packages/article/article.stories.js
@@ -8,7 +8,7 @@ import { MockedProvider } from "@times-components/utils/graphql";
 import { addTypenameToDocument } from "apollo-utilities";
 import { query as articleQuery } from "@times-components/provider/article";
 import { withTrackingContext } from "@times-components/tracking";
-import { action } from "@storybook/addon-actions";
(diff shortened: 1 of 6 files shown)


colincclark:
@toboid

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/1c0e8a279d9db849

packages/article/article.stories.js
```diff
@@ -141,4 +143,15 @@ storiesOf("Article", module)
...
-  ));
+  ))
+  .add("Tracking", () => {
+    const ArticleWithTrackingContext = withTrackingContext(Article, {
+      trackingObject: "Story"
+    });
+    return (
+      <ArticleWithTrackingContext
+        {...fullArticleFixture.data}
+        analyticsStream={action("analytics-event")}
```
craigbilner: did we want to output actual analytics?

colincclark: @craigbilner Not sure what you mean? I was following the pattern of the AuthorHead tracking story, found here: https://github.com/newsuk/times-components/blame/master/packages/author-head/author-head.stories.js#L31

toboid: Ah, I see. I think @craigbilner means that we can actually send this event to Tealium so that the analytics team can see it going all the way through if they want to. There is a helper already written for this it's in the Tealium package and there's an example [here](https://github.com/newsuk/times-components/blob/master/packages/author-profile/author-profile.stories.js#L6). I think the author-head story was written before this was added

toboid: Otherwise looks good

colincclark:
@toboid how do i test this in tealium? or will this "just work"?

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/4967fb895b5b0319

tuncaulubilge:
Article data is empty when tested on storybook "With Provider" story in article component. This is the same bug that's fixed in https://github.com/newsuk/times-components/pull/528 . You might need to implement "isDataReady" on articles to fix this when two branches are merged

colincclark:
This PR is now dependent on https://github.com/newsuk/times-components/pull/528 going in

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/f58808f3f1ae3bde

tuncaulubilge:
Looks good to me 💯

packages/article/package.json
```diff
@@ -39,9 +39,11 @@
   },
   "homepage": "https://github.com/newsuk/times-components#readme",
   "devDependencies": {
+    "@storybook/addon-actions": "3.2.18",
```
toboid: Don't think this one is required any longer

packages/article/article.stories.js
```diff
@@ -96,49 +98,83 @@ storiesOf("Article", module)
...
+      {...articleFixtureNoLeadAsset.data}
+      analyticsStream={storybookReporter}
+    />
+  ))
+  .add("Tracking", () => {
+    const ArticleWithTrackingContext = withTrackingContext(Article, {
+      trackingObject: "Story"
+    });
+    return (
+      <ArticleWithTrackingContext
```
toboid: Since the top-level Article component is already wrapped in a context [here](https://github.com/newsuk/times-components/blob/master/packages/article/article.js#L128) I don't think you need the extra one; it will result in the wrong tracking attributes being raised as described in ticket 629 (see the `attrs` key of the raised page event)

colincclark: Done as discussed

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/cd5c99a3afc29baf

PR #528, opened 2018-01-12
Added isDataReady check for pageView tracking

tuncaulubilge:
Problem: 
The author name for the page view event on author profile was null. This is due to the page view being triggered on `componentDidMount`, but author name is only populated after the api request.

Solution:
Added a `isDataReady` callback on tracking context which only fires page view event when data is ready

### DIFF
--- a/packages/author-profile/author-profile.js
+++ b/packages/author-profile/author-profile.js
@@ -150,7 +150,8 @@ const AuthorProfileWithTracking = withTrackingContext(AuthorProfile, {
     pageSize,
     articlesCount: get(author, "articles.count", 0)
   }),
-  trackingObject: "AuthorProfile"
+  trackingObject: "AuthorProfile",
+  isDataReady: ({ isLoading }) => !isLoading
 });
 
 export default withPageState(AuthorProfileWithTracking);
--- a/packages/tracking/__tests__/tracking-context.test.js
+++ b/packages/tracking/__tests__/tracking-context.test.js
@@ -176,7 +176,7 @@ describe("WithTrackingContext", () => {
   });
 
   it("only tracks page views when data is ready, with multiple updates", () => {
-    isReady = false;
+    let isReady = false;
     const WithTrackingAndContext = withTrackingContext(TestComponent, {
       trackingObject: "AuthorProfile",
       isDataReady: () => isReady
--- a/packages/tracking/__tests__/tracking-context.test.js
+++ b/packages/tracking/__tests__/tracking-context.test.js
@@ -182,8 +182,9 @@ describe("WithTrackingContext", () => {
       isDataReady: () => isReady
     });
     const reporter = jest.fn();
-
-    testRenderer = renderer.create(<WithTrackingAndContext analyticsStream={reporter} />);
+    const testRenderer = renderer.create(
+      <WithTrackingAndContext analyticsStream={reporter} />
+    );
     expect(reporter).not.toHaveBeenCalled();
 
     isReady = true;
--- a/packages/tracking/__tests__/tracking-context.test.js
+++ b/packages/tracking/__tests__/tracking-context.test.js
@@ -166,13 +166,7 @@ describe("WithTrackingContext", () => {
 
     renderer.create(<WithTrackingAndContext analyticsStream={reporter} />);
 
-    expect(reporter).not.toHaveBeenCalledWith(
-      expect.objectContaining({
-        object: "AuthorProfile",
-        action: "Viewed",
-        component: "Page"
-      })
-    );
+    expect(reporter).not.toHaveBeenCalled();
   });
 
   it("only tracks page views when data is ready, with multiple updates", () => {
--- a/packages/tracking/__tests__/tracking-context.test.js
+++ b/packages/tracking/__tests__/tracking-context.test.js
@@ -139,6 +139,60 @@ describe("WithTrackingContext", () => {
     );
   });
 
+  it("tracks page views when data is ready", () => {
+    const WithTrackingAndContext = withTrackingContext(TestComponent, {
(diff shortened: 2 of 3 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/dccffe7fa1b31cbd

packages/tracking/tracking-context.js
```diff
@@ -8,14 +8,16 @@ import resolveAttrs from "./resolve-attrs";
 
 const withTrackingContext = (
   WrappedComponent,
-  { getAttrs = () => ({}), trackingObject } = {}
+  // eslint-disable-next-line no-unused-vars
+  { getAttrs = () => ({}), trackingObject, isDataReady = props => true } = {}
```
toboid: If you change the default to `isDataReady = () => true` then you shouldn't need `eslint-disable-next-line` and it should all still work the same

packages/tracking/tracking-context.js
```diff
@@ -39,11 +41,24 @@ const withTrackingContext = (
     }
 
     componentDidMount() {
-      if (this.isRootTrackingContext()) {
+      this.trackPageEvent(this.props);
+    }
+
+    componentWillReceiveProps(nextProps) {
+      this.trackPageEvent(nextProps);
```
toboid: Could consider renaming to `this.attemptTrackPageEvent(nextProps);`? Just to be clear that the the event may not be raised on this call

packages/author-profile/author-profile.js
```diff
@@ -150,7 +150,8 @@ const AuthorProfileWithTracking = withTrackingContext(AuthorProfile, {
     pageSize,
     articlesCount: get(author, "articles.count", 0)
   }),
-  trackingObject: "AuthorProfile"
+  trackingObject: "AuthorProfile",
+  isDataReady: ({ isLoading }) => !isLoading
```
colincclark: can this just be a boolean?

packages/tracking/tracking-context.js
colincclark: can make `isDataReady` a bool defaulted to `true`

packages/tracking/tracking-context.js
colincclark: I was not getting the issue locally, perhaps because the mocks returned in time to be displayed...will this be called once the authorName is receveived by the `AuthorProfile` component?

packages/author-profile/author-profile.js
colincclark: also, i have a bad feeling that no tests were affected by adding this

packages/tracking/tracking-context.js
toboid: If you go to the __With Provider and Tracking__ story at http://components.thetimes.co.uk/ and look at the event that is written to the action logger, you should see that the authorName is null and articlesCount is 0

packages/author-profile/author-profile.js
tuncaulubilge: This allows the component to provide a callback to the tracking module which receives the props as a params and return if the component's data is ready. If the data is not ready, tracking module just tracks page view with null values.

For author profile, we are checking isLoading prop but for another module this could be a null check or something else.

Agree on the tests, will be looking into it

packages/tracking/tracking-context.js
tuncaulubilge: Check the comment above, we nee this to be a function

tuncaulubilge: Added "props" to enable auto-complete while implementing this function. The variable is unused in the default value but it is still part of the function signature.

packages/tracking/tracking-context.js
tuncaulubilge: Done in 9c0586c, thanks

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/ea321925f1511a13

packages/author-profile/author-profile.js
tuncaulubilge: Added tests on d64c1c8
Let me know if you think of any other scenarios that require testing

packages/tracking/__tests__/tracking-context.test.js
```diff
@@ -139,6 +139,60 @@ describe("WithTrackingContext", () => {
...
+      expect.objectContaining({
+        object: "AuthorProfile",
+        action: "Viewed",
+        component: "Page"
+      })
+    );
+  });
+
+  it("only tracks page views when data is ready, with multiple updates", () => {
+    isReady = false;
```
colincclark: no global variables - `let` is more appropriate here

tuncaulubilge: Fixed 684e5fd , still a js rookie here :) Thanks

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/1d97aaa9aa9ef112

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/f86260f529246d93

aronshamash:
looks good. Don't forget https://conventionalcommits.org/

colincclark:
For conventionalcommits, you can now use `yarn commit` for all of your committment needs

packages/tracking/tracking-context.js
toboid: Cool, I can see the argument for this however guidance from @craigbilner is to avoid using ignore comments unless there is no other way. We'll hopefully be rolling out more and more flow over time which will help with auto-complete

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/5ada77b30d84eea6

PR #532, opened 2018-01-15
chore: remove deps

craigbilner:
Currently the monorepo works as a monorepo but if each package were to be ejected you would neither be able to run lint or the tests. This PR creates more package independence and cleans up erroneous `package.json`s

Adding a `.babelrc` allows `jest` to correctly babel unbabeled code. Sadly some deps rely on `stage-2`....`eslint-config-thetimes` also needs `babel-jest` in order for the linting to work

Babel, flow and webpack aren't need on a package basis

Somehow the ad tests were  passing with illegal test code...now they're not...and now they're fixed...

The `transformIgnorePatterns` needed tweaking to correctly babel `@times-components` as well as other unbabeled dependencies as per the docs

It's not possible to run the tests outside of the repo because `jest-configurator` is too monorepo specific. By refactoring to make it dynamic allows the tests to run anywhere and the config files more generic so can be far more convention based

`react-native-linear-gradient` was setup rather strangely which meant it wasn't installed and the tests couldn't run...

If a package uses `jest-configurator` it is by proxy using the `enzyme` serializer which means that it needs to install it to run the tests...added it as a `peerDep` to `jest-configurator` and to dependent packages


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/9e5a1c788e8b71cf

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/c1b0102d292c5eac

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/50f30c9e4555c0e6

PR #535, opened 2018-01-16
fix: Replace deprecated View.PropTypes.style with ViewPropTypes.style…

tuncaulubilge:
… in card.js

View.PropTypes.style is deprecated so replaced it with ViewPropTypes.style. This was causing an error on release versions


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/9b64915bed68e650

PR #536, opened 2018-01-16
fix: will re-add missing fonts (hopefully) on staging

colincclark:
Reversion of previous 'fix' that stopped Travis CI errors related to fetch font


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/49eda1302e72bd78

PR #538, opened 2018-01-16
chore(Script fix): Removed set -e from jest script

tuncaulubilge:
Jest script was failing when the folder is already there, which is the case if you already have an
older version of the package installed.

### DIFF
--- a/packages/jest-configurator/setup.sh
+++ b/packages/jest-configurator/setup.sh
@@ -1,3 +1,5 @@
 #!/usr/bin/env bash
-mkdir fixtures/all/node_modules
-mkdir fixtures/ignore/node_modules
+set -e
+
+mkdir -p fixtures/all/node_modules
+mkdir -p fixtures/ignore/node_modules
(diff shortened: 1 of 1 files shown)


craigbilner:
@tuncaulubilge bit of a hack? We just need to do an existence check?

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/b38db72b75e883d4

packages/jest-configurator/setup.sh
```diff
@@ -1,5 +1,3 @@
 #!/usr/bin/env bash
-set -e
```
colincclark: not sure about removing the set flag, but you can just do:

```
mkdir -p fixtures/all/node_modules
mkdir -p fixtures/ignore/node_modules
```

to check the files exist

tuncaulubilge: True, that's better. Done in 004f353

times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/9a055dbaeb7be088

PR #543, opened 2018-01-17
Testing/provider

nikhedonia:
This PR adds tests that verifies if the connectGraphql HOC ensures that the latest requested data is being used for rendering regardless in which order the queries are resolved via the graphql-client.

Furthermore it also includes a test that verifies if a change in props re-issue a query even if the variables of the query did not change or if the request has not been resolved yet.


nikhedonia:
currently eslint complains as it does not use my test-specific graphql schema.
Where am I supposed to add the graphql schema so eslint can pick it up?

packages/provider/__tests__/apolloclient-idepotency.test.js
```diff
@@ -0,0 +1,100 @@
+import gql from "graphql-tag";
```
colincclark: the file should be `apolloclient-idempotency-test.js`

packages/provider/__tests__/apolloclient-idepotency.test.js
```diff
@@ -0,0 +1,100 @@
...
+
+const query = gql`
+  query AuthorQuery($slug: Slug!) {
+    author(slug: $slug) {
+      name
+    }
+  }
+`;
+
+it("should resolve correctly", async () => {
```
colincclark: this should be wrapped in a `describe` so the scope is understood in the test runner

packages/provider/__tests__/apolloclient-idepotency.test.js
```diff
@@ -0,0 +1,100 @@
...
+    { type: "request", query: "AuthorQuery", vars: { slug: "1" } },
+    { type: "request", query: "AuthorQuery", vars: { slug: "2" } },
+    { type: "resolving", query: "AuthorQuery", vars: { slug: "1" } },
+    { type: "resolving", query: "AuthorQuery", vars: { slug: "2" } },
+    { type: "resolved", query: "AuthorQuery", vars: { slug: "1" } },
+    { type: "resolved", query: "AuthorQuery", vars: { slug: "2" } }
+  ]);
+});
+
+it("should sent same response if result cached", async () => {
```
colincclark: `should send the same response if result is cached`

packages/provider/__tests__/apolloclient-idepotency.test.js
```diff
@@ -0,0 +1,100 @@
...
+  await Promise.all([q1, q2]);
+
+  expect(getEvents(link)).toEqual([
+    { type: "request", query: "AuthorQuery", vars: { slug: "1" } },
+    { type: "resolving", query: "AuthorQuery", vars: { slug: "1" } },
+    { type: "resolved", query: "AuthorQuery", vars: { slug: "1" } }
+  ]);
+});
+
+it("should return resolveds in the order ", async () => {
```
colincclark: in the same order?

packages/provider/__tests__/provider-execution-order.test.js
```diff
@@ -0,0 +1,199 @@
...
+      isLoading: false,
+      variables: { slug: "3" },
+      networkStatus: 7,
+      author: { name: "3", __typename: "Author" },
+      slug: "3"
+    }
+  ]);
+});
+
+it("should not refetch if same query despite superflourus vars", async () => {
```
colincclark: superfluous

packages/provider/__tests__/provider-testing-utils.js
```diff
@@ -0,0 +1,182 @@
...
+
+export function createProviderTester(
+  requestHandler,
+  Component,
+  defaultProps = {}
+) {
+  const { link, client } = createClientTester(requestHandler);
+
+  let setProps = () => Promise.resolve();
+  class Stateful extends React.Component {
```
colincclark: destruct `Component` on line 6

packages/provider/__tests__/provider-testing-utils.js
```diff
@@ -0,0 +1,182 @@
+import { ApolloClient } from "apollo-client";
+import { InMemoryCache } from "apollo-cache-inmemory";
+import { ApolloLink, Observable } from "apollo-link";
+import renderer from "react-test-renderer";
+import { ApolloProvider } from "react-apollo";
+import React from "react";
+import PropTypes from "prop-types";
+
+function createFuture() {
```
colincclark: is there a better way to do this?

packages/provider/__tests__/provider-testing-utils.js
```diff
@@ -0,0 +1,182 @@
...
+  return {
+    client,
+    link,
+    setProps,
+    component
+  };
+}
+
+function tidyEvent(e) {
+  if (e.type === "render") {
```
colincclark: destruct all the things

packages/provider/__tests__/provider-execution-order.test.js
```diff
@@ -0,0 +1,199 @@
...
+
+const query = gql`
+  query AuthorQuery($slug: Slug!) {
+    author(slug: $slug) {
+      name
+    }
+  }
+`;
+
+it("should resolve in order", async () => {
```
colincclark: wrap in a meaningful `describe`

packages/provider/__tests__/provider-execution-order.test.js
```diff
@@ -0,0 +1,199 @@
...
+      isLoading: true,
+      variables: { slug: "3" },
+      networkStatus: 2,
+      author: { name: "2", __typename: "Author" },
+      slug: "3"
+    },
+    { type: "resolving", query: "AuthorQuery", vars: { slug: "3" } },
+    { type: "resolved", query: "AuthorQuery", vars: { slug: "3" } },
+    {
+      type: "render",
```
colincclark: any chance these tests can be DRYed?

packages/provider/__tests__/provider-testing-utils.js
nikhedonia: We need to give control to the the one who writes the test.
Originally I thought we could use generators & yield to pass control to the user. 
Unfortunately this would not work as ApolloClient#query nor ApolloLink#request  methods are not generators where we need to suspend

Rxjs.Subject and Rxjs.observable.single

packages/provider/__tests__/provider-testing-utils.js
nikhedonia: actually descructuring is a bad idea as a component may have `type` as a `prop` which might lead to some errors.

packages/provider-test-tools/provider-tester.js
```diff
@@ -0,0 +1,68 @@
...
+        );
+    }
+
+    componentWillUnmount() {
+      setProps = () => Promise.resolve();
+    }
+
+    render() {
+      const Child = this.props.children;
+      return <Child {...this.state} />;
```
craigbilner: would you not typically invoke the children with the state as a function?

packages/provider-test-tools/provider-tester.js
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  const component = renderer.create(
+    <ApolloProvider client={client}>
+      {
+        <Stateful>
+          {state => (
+            <Component {...state}>
+              {props => {
+                link.pushEvent({ type: "render", props });
```
craigbilner: If I was a React pedant...don't think you should have side effects in render? You could use a mixture of `componentDidUpdate` and `componentDidMount`?

packages/provider-test-tools/package.json
```diff
@@ -0,0 +1,94 @@
+{
+  "name": "@times-components/provider-test-tools",
+  "version": "0.20.10",
```
craigbilner: the whole `package.json` seems out of date with this odd version number, can you align to the cli version?

packages/provider-test-tools/future.js
```diff
@@ -0,0 +1,14 @@
+export default function createFuture() {
+  let resolve;
+  const promise = new Promise(done => {
+    resolve = done;
+  });
+
+  return {
+    resolve: () => {
+      setTimeout(resolve);
```
craigbilner: don't quite understand this aspect? Promises can only resolve once so you can just resolve and return the promise?

packages/provider-test-tools/__tests__/provider-execution-order.test.js
```diff
@@ -0,0 +1,96 @@
...
+    const { link, setProps } = providerTester(
+      AuthorQueryResolver,
+      connect(query),
+      { slug: "1" }
+    );
+
+    await setProps({ slug: "2" });
+    await link.findByQuery("AuthorQuery", { slug: "2" }).resolve();
+    expect(getRenderedQueries(link).length).toBe(3);
+    expect(getRenderedQueries(link)).toMatchObject([
```
craigbilner: snapshot?

packages/provider-test-tools/__tests__/provider-execution-order.test.js
```diff
@@ -0,0 +1,96 @@
...
+    await link.findByQuery("AuthorQuery", { slug: "2" }).resolve();
+    expect(getRenderedQueries(link).length).toBe(3);
+    expect(getRenderedQueries(link)[2]).toMatchObject({
+      loading: false,
+      variables: { slug: "2" },
+      author: { name: "2" }
+    });
+  });
+
+  it("should render in order even if resolved out of order", async () => {
```
craigbilner: does this really match the test? Does it not need to resolve 1 afterwards?

packages/provider-test-tools/provider-tester.js
nikhedonia: I would usually agree but we want to log an event that states that `render` was called.
We cannot say that a render has happened based on the `componentDidUpdate` method.
What we actually need is a `componentDidRender` hook.

craigbilner: not overly problematic because it's a testing tool but I believe the `componentDidRender` is provided by the two methods above?

nikhedonia: Well, `componentDidUpdate` is called before `render`. We don't know when `render` we only know that it will be called (unless the component gets unmounted before?)
Additionally since fibers I'm not sure if it is guaranteed that `render` is preceded by exactly one `componentDidUpdate`.

In any case it would be a lie to log that a render has happened when it did not.
Lastly we never return a value in render making it a side-effect only procedure.

The whole idea of having side-effect functions is based on the idea of building a computation graph that is independent of the execution order - but the order of execution is exactly what we want to test.

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/3660a3fe20f1b234df8985c726628c817997a609/dextrose-test/index.html

packages/provider-test-tools/__tests__/provider-execution-order.test.js
nikhedonia: it does, please verify here: https://github.com/newsuk/times-components/pull/543/files

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/5d58bb87639dbb0409ed5c048fa2a5d91ce1849f/dextrose-test/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/fdd725e6f6bc3cac13856c81f0fe3806b685324e/dextrose-test/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/94b0094fbce06614032ac4c89dd6e5b1efc75a0a/dextrose-test/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/e8c49d5abfe3dfaf74e6a327233108d40026c007/dextrose-test/index.html

PR #544, opened 2018-01-18
upgrading dextrose and fructose versions

kaliabadi:


times-tools:
Please find visual snapshots of your changed components here: https://www.bitrise.io/build/d6f767ff6eb6c375

PR #539, opened 2018-01-16
feat: debounce author profile paging

BernieSumption:
1. Debounce author profile paging within the UI component - the display page (Showing 1 to 20 of 41) updates immediately, the query is delayed until no page changes have hapened in the last 250ms
2. Refactor pagination component to use onChangePage instead of onNext and onPrev
3. Fixed some errors relating to article publication date formats that had found their way into master after PR 525

PR #548, opened 2018-01-18
feat: add debouncing to all graphql queries

BernieSumption:
This PR replaces #523, and if it is accepted then #539 will become obsolete.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/d02d493a0daecdea0db343c62e287ca8cf463f6d/dextrose-test/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/d02d493a0daecdea0db343c62e287ca8cf463f6d/dextrose-test/index.html

packages/provider/connect.js
```diff
@@ -15,7 +16,17 @@ const getQueryVariables = definitions =>
     )
   );
 
-const connectGraphql = (query, propsToVariables = identity) => {
+export const makeGraphqlOptions = (
+  variableNames,
+  propsToVariables = identity
+) => props => ({
+  variables: _pick(
```
colincclark: why does `_pick` have an underscore? i see it's legacy but y?

packages/provider/debounce.js
```diff
@@ -0,0 +1,68 @@
...
+  WithDebounce.displayName = `WithDebounce(${getDisplayName(
+    WrappedComponent
+  )})`;
+  WithDebounce.propTypes = {
+    debounceTimeMs: PropTypes.number.isRequired,
+    ...WrappedComponent.propTypes
+  };
+  delete WithDebounce.propTypes.debounceTimeMs;
+  delete WithDebounce.propTypes.debouncedProps;
+  delete WithDebounce.propTypes.isDebouncing;
```
colincclark: what is happening here?

packages/provider/debounce.js
```diff
@@ -0,0 +1,68 @@
...
+import hoistNonReactStatic from "hoist-non-react-statics";
+
+const withDebounce = WrappedComponent => {
+  const validateProps = props => {
+    if (typeof props.debounceTimeMs !== "number") {
+      throw new Error("debounceTimeMs prop required");
+    }
+  };
+
+  class WithDebounce extends React.Component {
```
colincclark: destruct `Component`?

packages/provider/connect.js
BernieSumption: I questioned that myself - apparently it's something we do elsewhere when importing a single underscore method, instead of _.pick. Wasn't introduced by my PR.

packages/provider/debounce.js
BernieSumption: debouncedProps and isDebouncing are added by the HOC, so a wrapped component may declare them in their propTypes, in which case you don't want to hoist them to the HOC because then react would complain that you're not providing them to the HOC.

debounceTimeMs actually probably shouldn't be there, as it is provided to the HOC.

packages/provider/debounce.js
BernieSumption: As in, `extends Component`? We're inconsistent in the codebase, we use both.

packages/provider/__tests__/connect.test.js
```diff
@@ -0,0 +1,23 @@
+import { makeGraphqlOptions } from "../connect";
+
+it("Picks only required variables from the supplied props", () => {
```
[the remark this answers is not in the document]
BernieSumption: I agree - I was following the previous code in the package, but in this instance it would have been better just to fix it. Will do so.

packages/provider/__tests__/debounce.test.js
```diff
@@ -0,0 +1,172 @@
...
+    const setState = jest.spyOn(component.instance(), "handleDebounceTimer");
+
+    expect(setState.mock.calls.length).toEqual(0);
+
+    component.setProps({ foo: "updatedFoo1" });
+    jest.advanceTimersByTime(1200);
+    expect(setState.mock.calls.length).toEqual(1);
+
+    component.unmount();
+    component.setProps({ foo: "updatedFoo2" });
```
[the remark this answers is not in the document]
BernieSumption: It's testing componentWillUnmount. That said, setProps should be called before unmount. Will fix.

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/45b6cbe6a3f7aea85118882a5d2f271a11c50241/dextrose-test/index.html

packages/provider/__tests__/debounce.test.js
```diff
@@ -0,0 +1,172 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Enzyme, { shallow, mount } from "enzyme";
+import React16Adapter from "enzyme-adapter-react-16";
```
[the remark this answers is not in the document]
BernieSumption: Can I? I took a look at the package, it looks like it a utility to avoid boilerplate in jest.config.js rather than the test suites themselves?

packages/provider/debounce.js
colincclark: We can push for one over the other 😄

packages/provider/__tests__/debounce.test.js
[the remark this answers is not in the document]
colincclark: This _is_ Jest config boilerplate. https://github.com/newsuk/times-components/blob/master/packages/jest-configurator/setup-jest.js is the salient file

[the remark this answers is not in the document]
colincclark: but if you think it is out of scope, maybe we can standardise on a separate PR...

packages/provider/__tests__/provider.test.js
```diff
@@ -1,4 +1,5 @@
 import React from "react";
```
toboid: I still feel this is missing a test to check that the public API (provider) actually behaves in the desired way with regards to debouncing. There are assumptions made about how the constituent parts (`debounce`, `connect` and `react-apollo`) work together to form the public API but those assumptions need to be tested to give us any assurance that the solution works and will continue to work when changes are made to any of the sub-parts.

packages/provider/debounce.js
BernieSumption: Grand; I push for the version I'm used to ;)

packages/provider/__tests__/provider.test.js
BernieSumption: The behaviour has been manually verified in Dev Tools. @nikhedonia is working on testing something very similar in #543. That PR is currently in review, with changes requested to abstract it and make it reusable. Once that PR is merged, adding this test should be much easier.

packages/provider/debounce.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  WithDebounce.displayName = `WithDebounce(${getDisplayName(
+    WrappedComponent
+  )})`;
+  WithDebounce.propTypes = {
+    debounceTimeMs: PropTypes.number.isRequired,
+    ...WrappedComponent.propTypes
+  };
+  delete WithDebounce.propTypes.debouncedProps;
```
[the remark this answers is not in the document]
BernieSumption: It's `connect` that's wrong here, not debounce. The only reason that connect wasn't failing propTypes linting for using a prop without defining the proptype was that the access to debouncedProps was abstracted into a utility function `makeGraphqlOptions` so didn't get detected.  I've updated connect to define the debouncedProps propType on Wrapper.

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/bc3634425b326f276ae46f90969dba4f93f75d4f/dextrose-test/index.html

killanaca:
I saw that this one doesn't contains the fix for date's props like this one https://github.com/newsuk/times-components/pull/539 , so i did a specific PR https://github.com/newsuk/times-components/pull/556  just in case

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/c6d4d4efce971752cac3aa2ee6f3b1d338fbe1c6/dextrose-test/index.html

packages/provider/connect.js
colincclark: can we standardise and have `pick` then?

toboid:
I think there's still some issues with this PR that I would prefer to be resolved before merging rather than taking on the tech-debt. They've been raised in the comments of this and the prev PR but will summarise here.

There is only one usage of debounce, yet it's written as a generic HoC with the boilerplate that goes with that (copying prop-types, deleting prop-types, hoisting static props etc). `debounce` and `connect` each have their own unit tests; this satisfies code coverage requirements but we don't actually have any tests for the functionality that we are interested in - limiting the number of outgoing requests when props change very quickly.

The issue for me is that in future if I change either `debounce` or `connect` I will have to update these independent unit tests, but that won't give me any assurance that `provider` is working as expected - the tests have a maintenance cost but no value as they don't tell me if I have broken anything. 

As mentioned in one of the comments above it should be possible to build on the work that @nikhedonia has done and add the appropriate tests.

BernieSumption:
@toboid @craigbilner I'll let you two duke out whether the integration test is going to happen. Moving forwards on it is going to require either delaying this PR until the end of the sprint, or adjusting this sprint so that Gaetano or I can spend the time required to get Apollo testing working. It is, as they say, nontrivial

PR #551, opened 2018-01-19
chore: Added babel presets as peer dependency to all packages

tuncaulubilge:
Babel was causing issues on Android (and possible other platforms) without babel preset env and stage-2 installed. Added those as peer dependency on all packages


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/831a6961d372c69a570a139c6e0fdb6b63f574f1/dextrose-test/index.html

PR #552, opened 2018-01-19
chore: Moved jest setup script to pretest instead of postinstall

tuncaulubilge:
Setup script was being run on each install, but it should be running before tests only


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/10d5e3717cdf873266e8e52c493472f109d7ad4b/dextrose-test/index.html

PR #560, opened 2018-01-22
Do not spin up emulator if device connected

fourlastor:
This PR makes sure not to spin an emulator up if there's already a device (real or emulator) connected.

Also:

1. Renamed the vars to be lowercase, as reserved vars will _always_ be uppercase, this way we don't risk name clashes.
2. Used correct path (with `$ANDROID_HOME`) as `emulator` might not be set as an alias.

### DIFF
--- a/lib/start_android_emulator.sh
+++ b/lib/start_android_emulator.sh
@@ -1,14 +1,24 @@
 #!/usr/bin/env bash
 {
-  EMULATOR=$(emulator -list-avds | head -n 1)
-
-  if [[ -z $EMULATOR ]]
+  adb_cmd="$ANDROID_HOME/platform-tools/adb"
+  device=$($adb_cmd devices | tail -n +2)
+  if [[ ! -z $device ]]
   then
-    echo "You don't have an emulator set up"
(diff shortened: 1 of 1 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/30dfa96957b89b5c524341f42d666cc57e46e2e7/dextrose-test/index.html

lib/start_android_emulator.sh
```diff
@@ -1,14 +1,24 @@
 #!/usr/bin/env bash
 {
-  EMULATOR=$(emulator -list-avds | head -n 1)
+  adb_cmd="$ANDROID_HOME/platform-tools/adb"
+  device=$($adb_cmd devices | tail -n +2)
```
craigbilner: does devices return something odd in order to pick the right one?

fourlastor: I'm just counting how many lines there are, excluding the first line ("List of devices attached")

E.g.,
```
List of devices attached
emulator-5554	device
```

PR #561, opened 2018-01-22
feat: add DOMContext implementation

BernieSumption:
In the current Ad component, native and web are two implementations of essentially the same logic. The web implementation was linted and compiled. The native implementation was a 200 line long blob of code stored in a string and injected into a webview.

This component is a tool to avoid this. It provides a DOM context - give it a function and some data, and it will call that function with the data and a DOM element. On web this is trivial. On native, it means creating a WebView, serialising the funciton and data and running them in the WebView. Now we can use the same code for initialising ads on all platforms.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/febf86a173d27f28155afa9db4ec9e51071ecf41/dextrose-test/index.html

PR #563, opened 2018-01-23
Fix/pagination tracking name

tuncaulubilge:
Component displayName is null for release builds so trackingName must be added for all components.

Added trackingName to Pagination and AuthorProfileItem events.

### DIFF
--- a/packages/author-profile/author-profile-item.js
+++ b/packages/author-profile/author-profile-item.js
@@ -75,6 +75,7 @@ export default withTrackEvents(AuthorProfileItem, {
     {
       eventName: "onPress",
       actionName: "Pressed",
+      trackingName: "AuthorProfileItem",
       getAttrs: ({ headline, id }) => ({
         articleHeadline: headline,
         articleId: id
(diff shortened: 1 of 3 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/f9e5e6bfe5ad379f79639514e5ea7144795c405e/dextrose-test/index.html

toboid:
The trackingName should be inferred from the displayName but can be overridden. The website uses the `babel-plugin-add-react-displayname` to ensure that displayName is always present, even when running in prod mode.

Explicitly setting the name might be the better option but if doing that I think we'll need to do it everywhere and make the argument required. Otherwise it will be easy to add tracking elsewhere and think it's working but the displayName will be stripped on Android prod builds. The path of least resistance at the moment might be just to add the babel plugin to the Android build though? Thoughts @craigbilner ?

tuncaulubilge:
@toboid Decided to go with babel plugin, so I'll create another pr to add it as a dev dependency.

PR #564, opened 2018-01-23
fix: re-adds old scripts and fonts back to avoid 404s and CORS issues

colincclark:
- Fixes fonts back to what it was 2 weeks ago...this really needs to work better - we either need a font server or the ability to store fonts locally (they are downloadable anyway, so why not?)

@craigbilner


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/a0c3e3557cfedd8e17d9505f09a40f4b8a2a9aaa/dextrose-test/index.html

craigbilner:
using font server

PR #571, opened 2018-01-25
chore: installs babel-plugin-flow-react-proptype

nikhedonia:
Our codebase is not fully flowtyped. 
Flowtype makes PropTypes redundant as it shifts PropType checking from runtime to flow.
Unfortunately if you use a flowtyped component in a package that does not use flowtypes, you will have no proptype checks.

This PR installs https://github.com/brigand/babel-plugin-flow-react-proptypes
To create automatically proptypes from flowtypes.

You can verify that the plugin does it's job correctly by with the following component:

```js
import {Text} from "react-native";
class FlowTyped extends React.Component {
  props : {foo:string}
  
  render() {
    return <Text>{
      ( FlowTyped.propTypes && FlowTyped.propTypes.foo == PropTypes.string )? "true" : "false"
    }</Text>
  }
}

```

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
 {
-  "presets": ["react-native"]
+  "presets": ["react-native"],
+  "plugins": ["flow-react-proptypes"]
 }
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "@storybook/react-native": "3.3.9",
     "@times-components/fructose": "3.1.12",
     "babel-core": "6.26.0",
+    "babel-plugin-flow-react-proptypes": "14.0.3",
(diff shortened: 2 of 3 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/d4dc7581ad95726352179f89fef2f8e8e2d96f97/dextrose-test/index.html

colincclark:
@nikhedonia awesome! can I use this in my PR for flow in tracking? all I am missing s PropType flow...

nikhedonia:
yes, this will make proptypes redundant.
Since December this should also work for `contextTypes`

PR #574, opened 2018-01-26
Chore/add flowtype proptype transpiler

nikhedonia:
setting `ignoreNodeModules` to true seems to fix the test failure in `package/link`
I ensured that it actually works using `npx lerna run test`

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
 {
-  "presets": ["react-native"]
+  "presets": ["react-native"],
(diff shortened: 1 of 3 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/84a2d58487000235f4bb339568b20f99e3dfb349/dextrose-test/index.html

PR #582, opened 2018-01-30
Feature: Introduce new <Puff /> component

chrishutchinson:
This PR introduces the `<Puff />` component, currently built and maintained by the Editorial Developer team, and not part of core. You can see the _existing_ puff component [here](https://www.thetimes.co.uk/edition/news/theresa-may-faces-growing-calls-to-quit-l3tbwx3xv).

The new component replicates the behaviour of the old UI element.

screen shot 2018-01-30 at 14 49 48

As well as introducing a new component, the following changes have also been made to existing components:
### `<ArticleSummary />`
- the `date` prop is now optional

### Icons
- two new icons have been added, chevron right and left

### Jest configurator
- added `flow-typed` directories to the exclusion list for coverage collection

----

### Outstanding questions / issues worthy of review
- The bar at the top gets coloured based on the passed `section`. I currently have a helper method in `packages/puff/styles/index.js` which generates the style object for each section (allowing me to do ``style={styles[`section${sectionName}`]}`` in my JSX). This works, but I'm not sure I love it. @craigbilner discussed that there may be some work in the future to set up some kind of centralled defined colour palletes, which I think this would benefit from.
- I haven't yet been able to get an Android device up and running, so while all is well on web and iOS, I can't yet say the same for Android. Should have that running shortly
- I have typed the props here using `flow`, but haven't yet typed the `text` prop, which accepts some form of markup. One for @craigbilner potentially, I wasn't sure if this structure had been typed already?


.eslintignore
```diff
@@ -1,5 +1,6 @@
 **/node_modules/*
 **/coverage/*
+**/flow-typed/*
```
craigbilner: if config files are changed need to update the CLI

packages/article-summary/article-summary.js
```diff
@@ -74,17 +74,19 @@ const ArticleSummary = props => {
...
-        accessibilityLabel="datePublication"
-        testID="datePublication"
-      >
-        <DatePublication
-          date={date}
-          publication={publication}
-          showPublication={showPublication}
-        />
-      </Text>
+      {date && (
```
craigbilner: we're refactoring this to make it nicer and avoid this

packages/puff/.npmignore
```diff
@@ -0,0 +1,9 @@
+*.stories.js
```
craigbilner: flow-typed?

packages/puff/__tests__/android/puff.test.js
```diff
@@ -0,0 +1,14 @@
+import shared from "../shared";
+
+jest.mock("react-native", () => {
```
craigbilner: this isn't needed anymore - likewise in the other places

packages/puff/__tests__/shared.js
```diff
@@ -0,0 +1,53 @@
...
+module.exports = () => {
+  it("renders correctly when all props are provided", () => {
+    const tree = renderer.create(<Puff {...puffProps} />).toJSON();
+
+    expect(tree).toMatchSnapshot();
+  });
+
+  it("renders correctly without a section", () => {
+    const tree = renderer
+      .create(<Puff {...puffProps} sectionName={undefined} />)
```
craigbilner: is `undefined` possible? The stories use `null` interestingly

packages/puff/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@times-components/puff",
+  "version": "0.0.1",
+  "description": "Article puff",
```
craigbilner: might want a little more detail?

packages/puff/package.json
```diff
@@ -0,0 +1,67 @@
...
+  "scripts": {
+    "fmt": "prettier --write '**/*.*'",
+    "prettier:diff": "prettier --list-different '**/*.*'",
+    "lint": "eslint . && npm run prettier:diff",
+    "test": "yarn test:android --ci --bail --coverage && yarn test:ios -- --ci --bail --coverage && yarn test:web -- --ci --bail --coverage",
+    "test:android": "jest --config='./__tests__/android/jest.config.js'",
+    "test:ios": "jest --config='./__tests__/ios/jest.config.js'",
+    "test:web": "jest --config='./__tests__/web/jest.config.js'",
+    "test:watch": "jest --projects */**/jest.config.js --watchAll",
+    "flow": "flow"
```
craigbilner: flow won't run on the CI?

.eslintignore
chrishutchinson: Seems like fixing the CLI is a separate PR based on your "we're just starting our Flow journey" comment from the other day, and our Slack discussion last week.

packages/article-summary/article-summary.js
chrishutchinson: Okay, so should I wait for the refactor? An idea of timeline would be good, plus I've never found this an unpleasant way of doing things, what is the alternative? (Curious to learn)

packages/puff/.npmignore
chrishutchinson: Yep, will add

packages/puff/__tests__/android/puff.test.js
chrishutchinson: Er, I'm not clear what you mean? This entire file isn't needed, or just the `jest.mock` call?

chrishutchinson: Either way, this is what the CLI outputs, so based on your comment above, these sort of changes need to be gathered up and put into a ticket to improve the CLI

packages/puff/__tests__/shared.js
chrishutchinson: This is just because I was spreading on `puffProps` and I wanted to give the effect of `sectionName` not being given (`undefined` is technically different to `null`). I think I can use `null` here though - will check

packages/puff/package.json
chrishutchinson: Sure, do you have a general approach to how you do these? Is something like this too much:

>Component for promoting (puffing to) another link or article, with optional image

packages/puff/package.json
chrishutchinson: Again, there wasn't much to go on here as most of the rest of the codebase doesn't have `flow` here, and I also have no idea how the CI process runs. Suggestions welcomed!

chrishutchinson:
Will tidy up the things I've noted, there are a few outstanding questions @craigbilner in the notes above, plus the one below.

Can you elaborate on this? 
>native wasn't working as you had it
It seemed to be working for me, and I was using the fact that `styled-components` doesn't work on native to get the effect I needed.

In future I'd rather comments were left that explain the problem (easier to learn from mistakes, and to be better in future) and the PR author makes the appropriate changes, rather than having someone jump else and push to the branch.

.eslintignore
craigbilner: indeed, but as a general rule we should keep the CLI aligned until that point

packages/article-summary/article-summary.js
craigbilner: it should land in a day or two! We're changing the summary API to be more declarative and remove the need for consumers to bake in if/ternary logic

packages/puff/__tests__/android/puff.test.js
craigbilner: this was removed from the CLI some time ago?

packages/puff/__tests__/shared.js
craigbilner: `undefined !== null` is the main issue as well as the API shouldn't provide `undefined`

packages/puff/package.json
craigbilner: it's the type of thing we're sorely lacking and trying to improve on here and in TPA - ideally we'd like to break down the journalistic language and make it something anyone could understand

If you could breakdown "puffing" that would be good

packages/puff/package.json
craigbilner: generally we add it to the lint script for the CI to run

chrishutchinson: Okay great, good to know

packages/puff/package.json
chrishutchinson: Sure can

packages/puff/__tests__/shared.js
chrishutchinson: I guess the API can’t provide `undefined` but someone else could just not pass the prop, right? Do we just ignore cases like that and go with what makes most sense for us?

packages/puff/__tests__/android/puff.test.js
chrishutchinson: Hm, odd! I’ll have a look and update as appropriate

packages/article-summary/article-summary.js
chrishutchinson: Okay cool, this makes more sense! I initially think I had this logic in the child but something else made me think that wasn’t a good idea, but now I forget what! Probably just a byproduct of using existing components for new purposes!

.eslintignore
chrishutchinson: Okay, so I should update the CLI to output the appropriate flow changes for all new components?

craigbilner: a separate PR updating all the ignores would be ideal thanks

packages/puff/__tests__/shared.js
craigbilner: through proptypes/flow types/GraphQL I'd expect the data flow from TPA output to components to be predictable and well typed

packages/puff/puff.flow.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
+export type PuffProps = {|
+  headline: string,
+  label: ?string,
+  text: *, // @TODO: What is the type of "markup"?
```
colincclark: Looks like an Array of objects.

```
[{ name: <string>, attributes: <Object>}]
```

packages/puff/puff.web.js
```diff
@@ -0,0 +1,67 @@
...
+import ArticleSummary from "@times-components/article-summary";
+import Card from "@times-components/card";
+import Link from "@times-components/link";
+import { IconChevronRight } from "@times-components/icons";
+import withResponsiveStyles from "@times-components/responsive-styles";
+import styles from "./styles";
+
+import type { PuffProps } from "./puff.flow";
+
+const CardContainer = withResponsiveStyles(View, {
```
colincclark: can these be moved into the `styles` directory into a file called `responsive.web.js`?

packages/puff/styles/index.base.js
```diff
@@ -0,0 +1,40 @@
...
+  },
+  link: {
+    color: "#CD0000",
+    fontFamily: "GillSansMTStd-Medium",
+    fontSize: 13,
+    lineHeight: 13,
+    textDecorationColor: "transparent"
+  },
+  ...createSectionStyles({
+    News: "#13354E",
```
colincclark: can we have a `@todo` here to move the colours to the style guide when it is done?

chrishutchinson:
Editorial deadlines have stopped me being able to finish this up. I'll be back on it next week and will re-open then

PR #587, opened 2018-02-01
Feat: REPLAT-1388 - Date Publication refactor

jimhunty:
In related articles, we don't want to show a publication, we just want to show the date. This refactor removes the default publication and allows this to be blank. Seemed this simplest way to solve this.


packages/date-publication/date-publication.js
```diff
@@ -32,13 +32,16 @@ class DatePublication extends React.Component {
   }
   render() {
     const { date, publication, showPublication } = this.props;
```
colincclark: you have an unused `showPublication` prop here and in PropTypes below

packages/date-publication/date-publication.js
```diff
@@ -32,13 +32,16 @@ class DatePublication extends React.Component {
...
+
     const datetimeUTC = getUTCTime(date);
     const isDateBST = isBST(datetimeUTC);
     const offset = isDateBST ? 60 : 0;
     const datetimeLondonTimezone = addMinutes(datetimeUTC, offset);
-    const publicationString = showPublication
+
+    const publicationString = publication
```
colincclark: I thought the purpose of this PR was to `export` `Date` and `Publication` separately? @craigbilner

packages/date-publication/__tests__/date-publication.test.js
```diff
@@ -27,16 +27,20 @@ function gmtTests(userTimezone) {
...
-      .create(<DatePublication date="2017-01-01T14:32:00.000Z" />)
+      .create(
+        <DatePublication date="2017-01-01T14:32:00.000Z" publication="TIMES" />
+      )
       .toJSON();
     expect(tree).toMatchSnapshot();
   });
 
-  it("renders a DatePublication component with default publication and relevant BST date", () => {
+  it("renders a DatePublication component with a relevant BST date", () => {
```
colincclark: The `with a relevant BST date` and `with a relevant GMT date` part of the test names do not make sense when this function is called on line 54 `gmtTests("Europe/Kiev");`

colincclark:
Again, I think this branch - when approved - can fold straight into master - it should not be directly dependent on the feature branch

jimhunty:
@colincclark Its branched from the article summary refactor so it cant without breaking a lot. Best to maintain the cascade.

packages/date-publication/__tests__/date-publication.test.js
jimhunty: Changed the naming. I didnt write these tests.

packages/date-publication/date-publication.js
jimhunty: I didn't see the need to do this as this just seemed simpler.

aronshamash:
ok to have it on the `article-summary` refactor branch but it should be independent of our related articles with 2 article feature branch

jimhunty:
@aronshamash its related because of your refactor of slice. Unless this is in master, we should maintain with the original plan.

PR #588, opened 2018-02-01
Test: Update the Article Summary package to enable Jest Configurator

jimhunty:
As part of the Article Summary work, we wished to update the test suite Jest Configurator to make sure we are covering all platforms.


colincclark:
This should merge straight into master, it is not dependent on the feature branch and does not help deliver 2 related articles

packages/article-summary/__tests__/android/article-summary.android.test.js
```diff
@@ -0,0 +1,6 @@
+import ArticleSummary from "../../article-summary";
+import test from "../article-summary-helper";
```
colincclark: can you rename this `shared` as per other packages? `test` is a Jest method already

jimhunty: Done

jimhunty:
@colincclark We need to have your related articles work completed first so we can finish this refactor, otherwise it will break master. Your plan is to merge that into a feature branch so we have to follow suit.

PR #612, opened 2018-02-05
chore: Removing watch scripts

jimhunty:
Watch scripts are now considered to be unnecessary. As a result, I have removed them from all the packages.


jimhunty:
Please note Dextrose failing is not affected by this PR.

PR #613, opened 2018-02-05
fix: utils/graphql.js now imports transpiled file

nikhedonia:
`jest-configurator` doesn't work, if you have `@times-components/utils` as a transitive dependency.
This incompatibility was caused by mistakenly importing an untranspiled file from `react-apollo-temp`

This begs the question should we use `babel-jest` inside of jest-configurator?


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/986d4c17a1ea663a3a0214c65485bc663f6320fd/dextrose-test/index.html

PR #615, opened 2018-02-05
chore: flowtypes icons

nikhedonia:
Flowtypes `@times-components/icons`.
This PR does not include flowtypes for svg hence flow comment comes after import statements.

Furthermore I've been more explicit as required by flowtype. Eg.
`(props:Props): Element<*>`  could be written as `(props:Props): *` as FlowType correctly identifies stateless components. 
We could even be more explicit and state which type of component is used eg.: `Element<SVG>`
What does the styleguide have to say?


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/52b9d539ca8984a6d84e9829d5c14e84abe4d50f/dextrose-test/index.html

packages/icons/.flowconfig
```diff
@@ -0,0 +1,11 @@
+[ignore]
```
jimhunty: Is the config required if it is empty?

packages/icons/icons/diamond.js
```diff
@@ -1,20 +1,25 @@
 import React from "react";
-import PropTypes from "prop-types";
-
 import Svg, { G, Path } from "svgs";
 
-const IconDiamond = ({ width, height, fillColour }) => (
+// @flow
+import type { Element } from "react";
+
+export type DiamondProps = {
```
jimhunty: type Icon instead that is shared between Diamond and Twitter?

packages/icons/icons/twitter.js
```diff
@@ -1,8 +1,20 @@
 import React from "react";
-import PropTypes from "prop-types";
 import Svg, { Path } from "svgs";
 
-const IconTwitter = ({ width, height, fillColour }) => (
+// @flow
```
jimhunty: Prefer the flow comment to be at the top of the file

nikhedonia: In  that case we need to flowtype the svg package

packages/icons/icons/diamond.js
nikhedonia: props have colour mandatory wheras Twitter has colour optional. Not sure what we gain if we use a common type

packages/icons/.flowconfig
nikhedonia: This way flow knows where the project root is

packages/icons/icons/twitter.js
nikhedonia: We could just flowtype the used components instead of all though.

packages/icons/icons/diamond.js
craigbilner: could use an `interface` but perhaps it has limited value

packages/icons/icons/twitter.js
craigbilner: if it's not too much work would be good to get some flow-typed svg?

packages/icons/package.json
```diff
@@ -51,6 +52,7 @@
     "react-test-renderer": "16.2.0"
   },
   "dependencies": {
+    "flow-bin": "0.64.0",
```
craigbilner: need to make sure version are aligned

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/418339141ded4e6ed31ae90963b04eeb77af7a7e/dextrose-test/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/fac400b368ce322a99bd82e35a43256ac4ca0492/dextrose-test/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/71fa15b92c56cc2d35091ffe13d4aabfd6fd0590/dextrose-test/index.html

packages/icons/icons/twitter.js
```diff
@@ -16,12 +27,6 @@ const IconTwitter = ({ width, height, fillColour }) => (
...
 );
 
-IconTwitter.propTypes = {
-  fillColour: PropTypes.string,
-  height: PropTypes.number.isRequired,
-  width: PropTypes.number.isRequired
-};
-
 IconTwitter.defaultProps = {
   fillColour: "#006699"
```
colincclark: so are we going with flow and defaultProps? might be better to run with ES6 defaults and flow, otherwise it is going to be confusing...

packages/icons/icons/twitter.js
```diff
@@ -1,8 +1,19 @@
...
 import React from "react";
-import PropTypes from "prop-types";
 import Svg, { Path } from "svgs";
+import type { Element } from "react";
 
-const IconTwitter = ({ width, height, fillColour }) => (
+export type TwitterProps = {
+  width: number,
+  height: number,
+  fillColour?: string
```
colincclark: this could call a function with `isColorOptional` or some such value. Then an `IconProps type` could be shared across these (and future) components?

packages/icons/icons/twitter.js
nikhedonia: I've experimented a bit and here are some facts: 
If you don't use defaultProps then defaultProps won't show in the documentation.

If you use defaultProps but no es6 props then flow doen't know that it has a default value.
This means:
1) you specify the type of the prop as required
2) or you specify the type as optional and handle the undefined case.

In case 1) leads to wrong documentation and an eslint error
In case 2) leads to extra code

So for now I'd suggest a bit of duplication and use both: defaultProps and es6 defaulted arguments.  This way we have correct documentation and flow analysis

jimhunty: Considering Facebook author both React and Flow, there must be a standard approach?

jimhunty:
@nikhedonia Why was this closed?

PR #616, opened 2018-02-06
Feat/add two related articles

colincclark:
Final PR in a series. Adds two related articles to the mix. Slice now handles the template styling, and related articles sits in the articles package, handling its own internal styles.

Final feature branch will handle any small design tweaks after cross browser testing.


packages/slice/slice.js
```diff
@@ -4,12 +4,20 @@ import PropTypes from "prop-types";
...
 
 const Slice = ({ children, template }) => {
-  if (!children || children.length === 0) return null;
+  if (!children) return null;
+
+  const templateName = template.toLowerCase();
+
   const styles = StyleSheet.create({
-    template: templateStyles[template]
+    container: templateStyles[templateName].container,
```
nikhedonia: what happens if templateName does not exist in templateStyles?

colincclark: as per discussion after refinement - no defensive code required

packages/article/related-articles/config.js
```diff
@@ -0,0 +1,4 @@
+export default template => ({
+  domain: "relatedArticles",
+  type: template.charAt(0).toUpperCase() + template.slice(1).toLowerCase()
```
craigbilner: optional: could just take 0 index and use template string

craigbilner: I wouldn't expect this to be needed because this should come from TPA strongly typed and come through as some form of `enum` via prop-types/flow?

packages/article/related-articles/related-articles.web.js
```diff
@@ -0,0 +1,65 @@
...
+
+  const articleCount = articles.length;
+  const StyledRelatedArticleContainer = RelatedArticleContainer(articleCount);
+  const StyledImageContainer = ImageContainer(articleCount);
+  const StyledSummaryContainer = SummaryContainer(articleCount);
+
+  return (
+    <View style={styles.container}>
+      <RelatedArticlesHeading />
+      <Slice template={templateObject}>
```
craigbilner: when will `Slice` take on the template/layout responsibility?

PR #611, opened 2018-02-05
feat: Ad page level configs

nehasri89:
Each ad sends out some targetting config to Ad server so that it can receive the most appropriate Ad on the page. This PR adds the page level config for the ads on web. 
https://nidigitalsolutions.jira.com/wiki/spaces/PD/pages/154796078/Advertising+web+spec

The population of the config is dealt in render side in another PR https://github.com/newsuk/cps-content-render/pull/2176

PR #626, opened 2018-02-08
Feat/ad slot targeting config

nehasri89:
This PR deals with the slot level config for Ads. The responsibility of providing the actual key values lies with the client i.e render and apps. There is another render PR for that change.

This PR branch is taken from https://github.com/newsuk/times-components/pull/611 PR's branch which lays the foundation, so needs to go after the base PR is merged.

Also looked at the live article pages to find out the correct slot `pos` attributes that we support and refactored code for it.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/279e49ef2fc96d93fba7bc8b056987f44cfd7fac/dextrose-test/index.html

PR #628, opened 2018-02-08
fix: Removed fixtures from ad

tuncaulubilge:
Fixtures shouldn't be used outside tests so removed pageOptions and slotOptions from ad package


nehasri89:
you have to comment out the references in here too
https://github.com/newsuk/times-components/blob/master/packages/ad/ad-init.js#L14
https://github.com/newsuk/times-components/blob/master/packages/ad/ad-init.js#L63

Also, there are PRs currently open to replace this fixtures with real config, so depending on the urgency we can make this change

PR #632, opened 2018-02-08
chore: upgrade react-apollo

rjanjua:
Now that there is a beta version of react-apollo 2.1.0, I thought I'd upgrade it. We need it to get dextrose to play nicely with web. 

Unfortunately, as part of this PR I had to update storybook dependencies up to version 3.3.12 as I was getting multiple versions of storybook in my yarn.lock. And I needed storybook to test the changes.

I'm not 100% sure on why we needed react-apollo-temp, so please test it if you know exactly what needs testing. All I checked was that the Author Profile story "With an error getting author" still works correctly in the web.


rjanjua:
tests are failing due to the following https://github.com/apollographql/react-apollo/issues/1633

craigbilner:
will address the mockprovider issues

PR #633, opened 2018-02-09
feat: add example article summary API

colincclark:
- shows an example of byline being consumed as a function by article-summary

Although this moves the ternary to the consumer, we could simply write a HOC that takes the Component and its wrapped content, does the decoration if the Component is not null. The HOC would return null if the Component was null. Did not want to spend too much time on this as not aware of the business requirements around  https://nidigitalsolutions.jira.com/browse/REPLAT-1403 anymore (as discussed with @jimhunty). But, all this to avoid ternaries in the code? Not convinved it is worth the effort...

### DIFF
--- a/packages/slice/fixtures/single-related-article.json
+++ b/packages/slice/fixtures/single-related-article.json
@@ -29,28 +29,13 @@
           "https://www.thetimes.co.uk/article/bayeux-tapestry-now-for-a-new-battle-bringing-fragile-masterpiece-to-britain-safely-2k629tpvh",
         "byline": [
           {
-            "name": "author",
-            "attributes": {
-              "slug": "camilla-long"
-            },
-            "children": [
-              {
-                "name": "text",
-                "attributes": {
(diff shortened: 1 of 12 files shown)


packages/slice/fixtures/single-related-article.json
```diff
@@ -28,29 +28,14 @@
         "url":
           "https://www.thetimes.co.uk/article/bayeux-tapestry-now-for-a-new-battle-bringing-fragile-masterpiece-to-britain-safely-2k629tpvh",
         "byline": [
-          {
```
colincclark: replaced this part of the fixture because links within links are invalid

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/a78c6fce8a5534688deccbf131715869566432b3/dextrose-test/index.html

colincclark:
Closing as too generic, as discussed

PR #638, opened 2018-02-09
feat: Add jest-configurator to Tracking package

jimhunty:
As stated above


packages/tracking/__tests__/android/jest.config.js
```diff
@@ -0,0 +1,6 @@
+const jestConfigurator = require("@times-components/jest-configurator").default;
+
+module.exports = jestConfigurator("android", __dirname, [
+  "flow-typed",
```
craigbilner: perhaps we should move this to JC core ignore?

PR #646, opened 2018-02-13
feat: Allowing platform independence in Jest Configurator

jimhunty:
Some packages do not require platform specific environments for testing. These currently have their own Jest files. In order to use Jest projects as the top level we would want to run a command like `jest --projects */**/jest.config.js`. This means if we have consistent Jest Configs in all packages using the Jest Configurator, we should be able to reach our goal.

Thoughts?

### DIFF
--- a/packages/jest-configurator/src/jest-configurator.js
+++ b/packages/jest-configurator/src/jest-configurator.js
@@ -72,7 +72,9 @@ export default (
       coverageIgnoreGlobs
     ),
     testMatch: [`${module}/__tests__/${platformPath}*.test.js`],
-    testPathIgnorePatterns: [`${module}/__tests__/${platformPath}jest.config.js`],
+    testPathIgnorePatterns: [
+      `${module}/__tests__/${platformPath}jest.config.js`
+    ],
     snapshotSerializers: ["enzyme-to-json/serializer"],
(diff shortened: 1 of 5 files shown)


packages/utils/__tests__/jest.config.js
```diff
@@ -0,0 +1,3 @@
+const jestConfigurator = require("@times-components/jest-configurator").default;
+
+module.exports = jestConfigurator("", __dirname);
```
jimhunty: Empty string is a bit yucky.

craigbilner: `null` or `NO_PLAT`?

packages/jest-configurator/src/jest-configurator.js
```diff
@@ -24,6 +24,25 @@ const webSpecific = {
...
+
+const platformCode = platform => {
+  switch (platform) {
+    case "web":
+      return webSpecific;
+    case "ios":
+      return nativeSpecific("ios");
+    case "android":
+      return nativeSpecific("android");
+    case null:
```
craigbilner: is this branch needed?

jimhunty: I originally tried without it and had problems. Tried again this morning and all fine. Must have messed up on the `prepublish`. This has now changed.

PR #660, opened 2018-02-16
fix: make ids accurate

craigbilner:
Now we have debounced paging it's possible to have new ids for the existing content because they're artificially based on the current page. We do have unique ids for articles though, so this PR uses them instead. Index has also been added for complete defence against any rogue duplicate articles....


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/be6901dcb5f1ae031629fe3d86fe8d8eec7e847a/dextrose-test/index.html

PR #662, opened 2018-02-16
feat: REPLAT-1311 Label Colour by Section

jimhunty:
This includes:

- Adding Knobs to Storybook
- Refactoring of Article Label for better coverage
- Use of styleguide to apply section colours
- Swap function required for Storybook terribleness that they will fix later
- Improved Storybook package for easier exporting
- Updated article provider to include section
- Updated styleguide colours to be uppercase


packages/article/related-articles/related-article-item.js
```diff
@@ -31,7 +40,10 @@ const RelatedArticleItem = ({ article, onPress }) => {
           bylineProps={{ ast: byline }}
           datePublicationProps={{ date: publishedTime }}
           headline={() => <ArticleSummaryHeadline headline={headline} />}
-          labelProps={{ title: label, color: "#333333" }}
+          labelProps={{
+            title: label,
+            color: sectionColours[section] || "#333333"
```
craigbilner: should all colours be in the styleguide?

packages/storybook/swap.js
```diff
@@ -0,0 +1,7 @@
+const swap = items =>
```
craigbilner: should this be a util?
optional: object spread

packages/article-label/article-label.android.js
```diff
@@ -0,0 +1,28 @@
...
+  title
+    .toUpperCase()
+    .split("")
+    .join("\u200A");
+
+const ArticleLabel = ({ title, color }) => (
+  <Text style={[styles.title, { color }]}>{beautifyTitle(title)}</Text>
+);
+
+ArticleLabel.propTypes = {
```
colincclark: because the propTypes and the render and the imports are so similar to `article-label.js`, did you want to use a `base` file?

packages/article/related-articles/related-article-item.js
colincclark: default label can also be in the style guide like `sectionColours['defaultLabel']`?

packages/styleguide/colours/section.js
```diff
@@ -1,36 +1,36 @@
 const sectionColours = {
   artsfeatures: "#622956",
-  business: "#005b8d",
+  business: "#005B8D",
```
colincclark: 👍 for upperCase

packages/storybook/swap.js
aronshamash: could use `_.invert`

packages/article/related-articles/related-article-item.js
aronshamash: I guess default label can be in styleguide if it is the default for all use cases (not just for related articles). There is still a case for having some colours local to the component if they are not shared.

jimhunty: Its an interesting one, I lean towards @aronshamash on this one. If we see greater consistency it can then be moved.

@craigbilner Eventually we should have all colours in the styleguide. I'll make it a tech debt task to do when I run out of tickets.

packages/storybook/swap.js
jimhunty: @craigbilner its a temporary util until Storybook has its next major release. Unless it should go elsewhere?

@aronshamash I guess you are suggesting lodash? Which is essentially doing the same job. Why add another library when its already written?

packages/article-label/article-label.android.js
jimhunty: I started refactoring it in this way, but in the end there was a lot more code with a base than this approach. So I went with efficiency.

packages/article/related-articles/related-article-item.js
colincclark: This colour is reused al over the place, so should definitely be a default colour. Also, it is the default for other labels... https://github.com/newsuk/times-components/blob/master/packages/author-profile/author-profile-item.web.js#L70

craigbilner: I would say the styleguide should be authoritative so colour changes can happen in one place rather than having to do route around for where a colour is coming from?

packages/storybook/swap.js
craigbilner: `_.invert` is tiny and already battle tested...

packages/article/related-articles/related-article-item.js
aronshamash: if it is default for other labels then should be in styleguide

aronshamash: Created new github issue https://github.com/newsuk/times-components/issues/665

packages/storybook/swap.js
aronshamash: it is very thoroughly tested

PR #666, opened 2018-02-19
Feat/related articles tracking

nikhedonia:
Adds render event tracking  for related articles.
https://nidigitalsolutions.jira.com/browse/REPLAT-1092

Currently the role of the article is determined by the template and by the position in the articles array.
I'm happy to define names or (use existing role names) and make the role-name more expressive. 
What do you think?  - I reached out to people who might have a list of role names.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/2815dde682787167654c3f31dd6c32e16abef162/dextrose-test/index.html

packages/article/related-articles/related-articles-tracking-context.js
```diff
@@ -0,0 +1,17 @@
+import get from "lodash.get";
```
jimhunty: Do we need the tracking context if this component is going to be part of an Article page which is already wrapped with a Tracking Context? I assumed this was just a page level thing. Perhaps @colincclark can help?

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/df99592b3e030197366d5deb56b81af1f1b93163/dextrose-test/index.html

packages/article/related-articles/related-articles-tracking-context.js
nikhedonia: TrackingContext has a double life.
One it provides a TrackingContext to it's descendants and then it also extracts analytics from the props.
We extract slightly different information from a related article then from an article.

packages/article/__tests__/android/__snapshots__/related-article.android.test.js.snap
```diff
@@ -1,8 +1,196 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Related articles test on android Related articles handles no related articles 1`] = `null`;
+exports[`Related articles test on android Related articles analytics handles no related articles 1`] = `
```
craigbilner: test names just need sorting out to reflect the testing of tracking props

packages/article/related-articles/related-articles-tracking-context.js
```diff
@@ -0,0 +1,21 @@
...
+import { withTrackingContext } from "@times-components/tracking";
+import role from "@times-components/slice/styles/role-map";
+
+export default Component =>
+  withTrackingContext(Component, {
+    trackingObjectName: "RelatedArticles",
+    isDataReady: ({ isLoading }) => !isLoading,
```
craigbilner: this seems a little redundant, if it passes isLoading does it not already know this information as a default?

nikhedonia: Who do you refer to when you say "it" ?
`withTrackingContext` invokes `isDataReady` to ensure that the data is ready.
When used in conjunction with a provider shouldn't we prevent to send data prematurely?
Ideally the `RelatedArticles` component should just work regardless of the context it is being used.

PR #672, opened 2018-02-20
chore: publish once

craigbilner:
The deploy script (making components.thetimes.co.uk) runs on every publish commit and every job. This moves the logic to the custom script where it will only publish for a single job when merged to master


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/c45a042ddc07b17489797931047bf2708fbc97df/dextrose-test/index.html

PR #674, opened 2018-02-20
fix: publish coverage for all jobs

craigbilner:
Accidentally broke the coveralls report for master by not reporting the coverage for all jobs, only bail out on the other three jobs before publishing


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/a09cedf711164f90837c859a09c0f58e37e174d2/dextrose-test/index.html

PR #675, opened 2018-02-21
feat: Slice API for default and lead

colincclark:
- only show Slice API logic in Slice stories
- separate logic for DEFAULT and LEAD_AND_TWO Slice APIs
- fix Slice tests to show better and clearer snapshots

### 1 Article
default-1

### 2 Articles
default-2

### 3 Articles
deafult-3

### 1 Article
lead-1

### 2 Articles
lead-2

### 3 Articles
lead-2-incminheight


packages/slice/__tests__/shared.js
```diff
@@ -1,51 +1,87 @@
...
-
-    const createItems = noOfItems =>
+    const createDefaultItems = noOfItems =>
       items
-        .map(item => {
-          const { key } = item;
-          if (key > noOfItems) return false;
-          return <ExampleChild key={key}>Test text {key}</ExampleChild>;
+        .map((item, index) => {
+          if (index >= noOfItems) return false;
```
fourlastor: is the change of logic here intentional? was key 1-based?

packages/slice/__tests__/shared.js
```diff
@@ -1,51 +1,87 @@
...
+        "5. LEAD_AND_TWO template renders two child elements"
       );
-      expect(wrapper).toMatchSnapshot();
     });
 
     it("renders three child elements", () => {
       const wrapper = shallow(
-        <Slice template={template}>{createItems(3)}</Slice>
+        <LeadSlice
+          lead={() => <ExampleChild />}
```
fourlastor: misaligned with the ones after

packages/slice/templates/default/default.web.js
```diff
@@ -0,0 +1,28 @@
...
+  ChildrenContainer.displayName = "ChildrenContainer";
+  return (
+    <SliceContainer>
+      <ChildrenContainer>
+        {children
+          .map(child => (
+            <ChildContainer key={child.key}>{child}</ChildContainer>
+          ))
+          .reduce((previous, current) => [
+            ...(previous.length > 0 ? previous : [previous]),
```
fourlastor: can you instead put a starting value as `[]` to concat them and remove the ternary?

packages/slice/templates/lead/responsive.web.js
```diff
@@ -0,0 +1,128 @@
...
+    flex-direction: column;
+    height: auto;
+  `
+});
+ChildrenContainer.displayName = "ChildrenContainer";
+
+export const getChildContainer = (isSecondChild, hasTwoChildren) =>
+  withResponsiveStyles(View, {
+    base: () => {
+      const firstChildStyle = hasTwoChildren
```
fourlastor: can we extract the 2 cases in methods? it's a bit hard to follow with all the backticks 👅

packages/slice/__tests__/shared.js
colincclark: yes it was

packages/slice/templates/default/default.web.js
colincclark: no, the separator will then be the first item in the resulting array

packages/slice/templates/lead/responsive.web.js
colincclark: i _could_ add an if statement, but not sure how much value this abstraction will add...will take a look

packages/slice/__tests__/shared.js
colincclark: all of these files were formatted using `yarn fmt`

packages/slice/templates/default/default.web.js
fourlastor: Ah I see

packages/slice/templates/lead/responsive.web.js
fourlastor: I meant something like:

```javascript
const firstChildStyle = hasTwoChildren ? twoChildrenFirstStyle() : oneChildFirstStyle()
```

colincclark: but yeah, totally agree in the cold light of day it looks a bit confusing...will try and fix it up a bit

colincclark: done

packages/responsive-styles/responsive-styles.js
```diff
@@ -1,6 +1,10 @@
 import styled from "styled-components";
 
 export { keyframes } from "styled-components";
+export const config = {
+  mediumBpWidth: "84%",
```
aronshamash: nice

packages/slice/templates/lead/lead.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import { View } from "react-native";
+import { propTypes, defaultProps } from "./proptypes";
+import styles from "../styles";
+
+const Slice = ({ lead, child1, child2 }) => {
+  const hasChildren = child1();
```
craigbilner: minor: naming is a little iffy? `hasChild` which could be either?

packages/slice/templates/lead/lead.web.js
```diff
@@ -0,0 +1,44 @@
...
+  ChildrenContainer,
+  Separator
+} from "./responsive";
+
+const LeadSlice = ({ lead, child1, child2 }) => {
+  const hasChildren = child1();
+  const hasTwoChildren = child2();
+  const Container = getContainer(!!hasChildren);
+  Container.displayName = "Container";
+  const LeadContainer = getLeadContainer(!!hasChildren);
```
craigbilner: would generally steer clear of boolean APIs

aronshamash: Can you give an example of how you would see it?

colincclark: done

packages/slice/templates/lead/lead.js
```diff
@@ -0,0 +1,31 @@
...
+  const hasTwoChildren = child2();
+  return (
+    <View style={styles.container}>
+      <View style={styles.childContainer}>
+        <View style={[styles.child, { paddingTop: 0 }]}>{lead()}</View>
+      </View>
+      {hasChildren ? (
+        <View style={styles.childContainer}>
+          <View style={styles.child}>{child1()}</View>
```
nikhedonia: is hasChildren not the resoult of child1() ?

packages/slice/templates/lead/lead.web.js
```diff
@@ -0,0 +1,44 @@
...
+  LeadContainer.displayName = "LeadContainer";
+  const ChildContainer1 = getChildContainer(false, !!hasTwoChildren);
```
nikhedonia: that looks awful.

how about the following approach:

```js
const wrapped = decorateChildren(children); // or instead of children let's call it supporting or simmilar 
return (
    <SliceContainer>
      <Container>
        <LeadContainer>{lead()}</LeadContainer>
        {wrapped.length? <Separator /> : null}
        {wrapped}
//....
)
```

packages/slice/templates/lead/responsive.web.js
```diff
@@ -0,0 +1,130 @@
...
+export const ChildrenContainer = withResponsiveStyles(View, {
+  base: () => `
+    flex: 1;
+    flex-direction: column;
+    height: auto;
+  `
+});
+ChildrenContainer.displayName = "ChildrenContainer";
+
+export const getChildContainer = (isSecondChild, hasTwoChildren) =>
```
nikhedonia: I would put each case into a seperate function and then create a decorator that consumes an array of cards.
This way would be imo easier to read (especially in `lead/lead.*.js`) as then we would have less conditionals in the jsx

Issue #677, opened 2018-02-21
Add accessibility checks to storybook 


aronshamash:
There are accessibility checks on components in storybook.

PR #678, opened 2018-02-22
Chore/a11y Add accessibility checks to storybook

aronshamash:
Adds an accessibility addon to storybook which highlights accessibility issues in the component stories. https://github.com/newsuk/times-components/issues/677

### DIFF
--- a/.storybook/addons.js
+++ b/.storybook/addons.js
@@ -1,3 +1,4 @@
 import '@storybook/addon-knobs/register'
 import "@storybook/addon-actions/register";
 import '@storybook/addon-options/register';
+import '@storybook/addon-a11y/register';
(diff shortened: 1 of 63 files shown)


jimhunty:
Can the decorator be added at the config so it will be applied to all stories? Do we want it on all stories? Does this work with React Native as well?

aronshamash:
@jimhunty Move to storybook config. I think they should be on any component with a UI. Its doesn't work on native.

aronshamash:
Found to not work with React Native

craigbilner:
@aronshamash how does the `checkA11y` work now?

aronshamash:
@craigbilner it didn't work and I have had to add the decorator in each story we would like a11y checks.

jimhunty:
This should be added to the CLI?

aronshamash:
@jimhunty Added to cli

PR #680, opened 2018-02-23
WIP: Refactor/card

colincclark:


packages/card/package.json
```diff
@@ -31,8 +31,8 @@
   },
   "homepage": "https://github.com/newsuk/times-components#readme",
   "devDependencies": {
+    "@storybook/addon-knobs": "3.3.13",
```
nikhedonia: we are at 3.3.14 now

packages/card/__tests__/web/__snapshots__/card.web.test.js.snap
```diff
@@ -136,76 +113,49 @@ exports[`Card test on web renders 1`] = `
...
-  -webkit-flex-shrink: 1 !important;
-  -ms-flex-negative: 1 !important;
-  flex-shrink: 1 !important;
-}
-
+exports[`Card tests on web renders loading state 1`] = `
 .c0 {
+  -webkit-flex: 1;
+  -ms-flex: 1;
+  flex: 1;
```
nikhedonia: This seems like a recurring pattern in our codebase.
I think we should use an auto prefixing solution or at least refactor this into utility functions...

However not the scope of this PR

packages/card/package.json
colincclark: done

packages/card/loading.js
```diff
@@ -0,0 +1,34 @@
...
+    <View style={styles.imageContainer}>
+      <Image aspectRatio={aspectRatio} />
+    </View>
+  );
+
+  return (
+    <View>
+      {showImage && imageComponent}
+      <View>
+        <Gradient style={styles.headerContainer} degrees={264} />
```
nikhedonia: 264 or 267 ?

PR #682, opened 2018-02-26
chore: upgrading dextrose version

kaliabadi:


packages/ad/package.json
```diff
@@ -45,7 +45,7 @@
     "@storybook/react-native": "3.3.14",
     "@times-components/eslint-config-thetimes": "0.1.3",
     "@times-components/utils": "0.5.1",
-    "dextrose": "^1.4.6",
+    "dextrose": "^1.5.0",
```
nikhedonia: Should we prefer exact version numbers?

craigbilner: we do, in the npmrc

PR #683, opened 2018-02-26
refactor: refactor the Slice logic

colincclark:
@nikhedonia 

- refactored Slice layout logic - lead and two now becomes an array called `supports` for better a functional API
- DRYed up Separator styled component


packages/slice/templates/lead/lead.js
```diff
@@ -3,27 +3,18 @@ import { View } from "react-native";
...
-        </View>
-      ) : null}
+const Slice = ({ lead, sections }) => (
+  <View style={styles.container}>
+    <View style={styles.childContainer}>
+      <View style={[styles.child, { paddingTop: 0 }]}>{lead()}</View>
     </View>
-  );
-};
+    {sections.map(section => (
```
aronshamash: Think we should try and avoid `sections` as it it used elsewhere with a different meaning.

colincclark: now called supports

aronshamash:
LGTM

packages/slice/templates/shared.responsive.web.js
```diff
@@ -23,16 +24,12 @@ export const Separator = withResponsiveStyles(View, {
...
     border-bottom: none;  
     border-right-style: solid;
     border-right-width: 1px;
     border-right-color: #dbdbdb;
     flex: 0 !important;
-    margin-bottom: 0;
-    margin-left: 10px;
-    margin-right: 10px;
-    margin-top: 0;
+    margin: ${hasLeftRightMargin ? "0 10px" : "0px"};
```
nikhedonia: small nitpick: 0 vs 0px

packages/slice/templates/lead/responsive.web.js
```diff
@@ -3,65 +3,38 @@ import withResponsiveStyles, {
...
-      padding-left: 10px;
-      padding-right: 10px;
-      ${isSecondChild ? secondChildStyle : firstChildStyle}
-    `;
+        flex: 1;
+        flex-wrap: wrap;
+        min-height: auto;
+        padding-left: 10px;
+        padding-right: 10px;
+        ${index === 0 ? firstChildStyle : secondChildStyle}
```
nikhedonia: You could get rid of this ternary if you either return an array of containers. The enduser would then decorate / wrap it's cards with those containers:
```js
_.zip( getContainers(), [card1, card2] ).map( ([Container, Card]) => <Container> <Card/> </Container>));
// or 
const cards = [card1, card2];
getContainers().map( (Container, i) => <Container>{cards[i]}</Container> )
```

packages/slice/templates/lead/index.web.js
```diff
@@ -0,0 +1,49 @@
...
+  const supports = [support1, support2].filter(support => support !== null);
+  const hasSupports = supports.length > 0;
+  const Container = getContainer({ hasSupports });
+  const LeadContainer = getLeadContainer({ hasSupports });
+  const Separator = getSeparator({ hasLeftRightMargin: false });
+
+  // for tests
+  Container.displayName = "Container";
+  LeadContainer.displayName = "LeadContainer";
+  Separator.displayName = "Separator";
```
nikhedonia: This belongs to the (Lead)Container/Seperator factory.

packages/slice/__tests__/android/jest.config.js
```diff
@@ -1,3 +1,8 @@
 const jestConfigurator = require("@times-components/jest-configurator").default;
 
-module.exports = jestConfigurator("android", __dirname, ["responsive.js"]);
+// @TODO: remove "**/styles" and "slice.js" when related articles has been refactored and styles removed
```
jimhunty: Rather than the comment why not remove them now. Unless the coverage drops so much you cant merge?

packages/slice/templates/shared.responsive.web.js
colincclark: yep, nice, done

packages/slice/templates/lead/responsive.web.js
colincclark: do not feel this is neccessary, and the way we have done it here follows patterns used elsewhere in the codebase

packages/slice/templates/lead/index.web.js
colincclark: this is where the component is first assigned to a Seperator variable. I feel it belongs here

packages/slice/__tests__/android/jest.config.js
colincclark: the coverage dropped so much we could not merge. Removing them would break related articles which will consume Slice in a separate PR

jimhunty: 👍

packages/slice/__tests__/ios/jest.config.js
```diff
@@ -1,3 +1,8 @@
 const jestConfigurator = require("@times-components/jest-configurator").default;
 
-module.exports = jestConfigurator("ios", __dirname, ["responsive.js"]);
+// @TODO: remove "**/styles" when related articles has been refactored and styles removed
```
jimhunty: Remove the to do and create a ticket. No need for them to go into Master.

packages/slice/templates/default/index.web.js
```diff
@@ -1,11 +1,16 @@
 import React from "react";
 import propTypes from "./proptypes";
-import { Separator, SliceContainer } from "../shared.responsive";
+import { getSeparator, SliceContainer } from "../shared.responsive";
```
jimhunty: Why cant this be Separator? Its just a component that takes a prop.

packages/slice/templates/lead/index.web.js
```diff
@@ -0,0 +1,49 @@
...
+  getSupportContainer,
+  SupportsContainer
+} from "./responsive";
+
+const LeadSlice = ({ lead, support1, support2 }) => {
+  const supports = [support1, support2].filter(support => support !== null);
+  const hasSupports = supports.length > 0;
+  const Container = getContainer({ hasSupports });
+  const LeadContainer = getLeadContainer({ hasSupports });
+  const Separator = getSeparator({ hasLeftRightMargin: false });
```
jimhunty: The same as getSeparator, Container and LeadContainer can just be `<LeadContainer hasSupports={hasSupports} />`

packages/slice/templates/default/index.web.js
colincclark: it is not a component, it is a function that returns a component - think this change was driven by comments from @craigbilner

packages/slice/__tests__/ios/jest.config.js
colincclark: I disagree that @todo comments should not go into master. It makes the code easier to find, on a PR which is just around the corner

packages/slice/templates/lead/index.web.js
colincclark: see above, these are functions and @craigbilner asked for them to be named as such

jimhunty: It returns a Styled Component, so thats not the case.

packages/slice/__tests__/ios/jest.config.js
jimhunty: I would consider anything in Master as released (which is true because we publish the packages via npm). @todo comments have no place in released code in my opinion.

packages/slice/templates/lead/index.web.js
colincclark: will revisit

packages/slice/__tests__/ios/jest.config.js
colincclark: Here is an example of some todo comments in the React repo: https://github.com/facebook/react/search?utf8=%E2%9C%93&q=todo

colincclark: I think too many @todo comments that never go away are indicative of a bad technical debt policy. These @todo comments _have_ to go away to meet our previous sprint goal. In retrospect, creating a feature branch would have been the way to go, but that is easy to say with hindsight. Maybe we should _always_ create a feature branch to avoid commiting @todo comments into `master`? @todo comments might fall out of sync with development, so I understand your concerns with having them, but that is unlikely in this scenario.

colincclark: As an adendum to this - I have just finished the work (in a separate branch) to remove Slice after making Related Articles consume the new templates. It was much easier to remove the old Slice code with the @todo comments in place....

packages/slice/templates/lead/index.web.js
nikhedonia: `displayname` is a property of the `React.Component`.
If i want to use Your Seperator - say I call `getSeperator()` I would like to use it immediately without needing to configure it's displayname.  If the displayname is not set inside the factory then every user needs to set the displayname.

colincclark: this can be refactored another time if you have a better resolution. going to merge this for now as Slice coverage in particular seems to be blocking other work

packages/slice/__tests__/ios/jest.config.js
aronshamash: I'm with not have @todo s

colincclark: Disagree, but removed controversial @todo comments, mostly because of the disagreement, but partly because the PR has been complete that utilises these comments...

packages/slice/templates/lead/index.web.js
colincclark: You convinced me. Done.

packages/slice/templates/lead/index.web.js
colincclark: @jimhunty not sure how to do this differently. Can you expand / provide an example / re-review?

PR #684, opened 2018-02-26
Feat: Author Profile Design Changes

jimhunty:
Jake requested and has subsequently reviewed the following:
* Last line of the loading card to be about 40% width.
* No image card to align with the keyline.


packages/card/card-styles.web.js
```diff
@@ -11,6 +11,7 @@ export const ImageContainer = withResponsiveStyles(View, {
     flex-grow: 2;
     flex-basis: 0 !important;
     margin-bottom: 0;
+    margin-right: 15px;
```
aronshamash: What is this @jimhunty ?! Can we use `10px` or `20px`

aronshamash: ok we have agreed a base 5 spacing scheme.

PR #685, opened 2018-02-26
Feat/bigger placeholder

nikhedonia:
Replaces the old T placeholder with a bigger one as requested by Jake Cox.

OLD:
t-old

NEW:
t-new

ViewBox/ratios computed from the extracted paths of attached svg in: https://nidigitalsolutions.jira.com/browse/REPLAT-1312

ViewBox retrieved via: `document.querySelector('path').getBBox()`
scaling factor computed via: 
 ```js
const scale = ( 
  document.querySelector('path').getBBox().width 
/ document.querySelector('svg').getBBox().width
)
// scale = 0.27411167512690354

```


jimhunty:
We should look to move all SVGs to the Icons package?

PR #686, opened 2018-02-26
Feat: Fonts in the Styleguide

jimhunty:
Putting all the references for the fonts into the styleguide.

Font names confirmed with Jake

Font sizes will be tackled in a separate PR and worked on with Jake to get more consistency

### DIFF
--- a/packages/styleguide/styleguide.stories.js
+++ b/packages/styleguide/styleguide.stories.js
@@ -30,6 +30,16 @@ const styles = {
     backgroundColor: "#CCCCCC",
     margin: 50,
     padding: 100
+  },
+  showoffFontsContainer: {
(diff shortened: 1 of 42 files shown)


packages/styleguide/styleguide.stories.js
```diff
@@ -83,4 +55,57 @@ storiesOf("Helpers/Styleguide", module)
         <Text style={styles.text}>Fade In</Text>
       </View>
     </Animations.FadeIn>
-  ));
+  ))
+
+  .add("Fonts", () => {
+    const phrase = "The Quick Brown Fox Jumped Over the Lazy Dog";
```
colincclark: nice pangram!

PR #687, opened 2018-02-27
Refactor/card package

colincclark:
- [x] refactor Card to be inline with current codebase patterns and standards
- [x]  add knobs for stories
- [x]  allow card to take a flex property to control the responsive layouts, in preparation for using Card in Related Articles
- [x]  fix up snapshots so they are readable


packages/author-profile/author-profile-item.web.js
```diff
@@ -56,6 +56,7 @@ const AuthorProfileItem = item => {
     return (
       <View style={[styles.container, style]}>
         <Card
+          childRatio={2.7}
```
jimhunty: 2.7 seems more like the default ratio that 1? This means we can reduce the number of occurances of this prop?

packages/card/__tests__/web/card.web.test.js
```diff
@@ -15,101 +14,97 @@ const cardProps = {
...
-      .create(
-        <Card {...cardProps}>
-          <span>A card</span>
-        </Card>
-      )
-      .toJSON();
-
-    expect(tree).toMatchSnapshot();
+  it("should render", () => {
+    const component = shallow(
```
jimhunty: Interested to know why we are doing shallow rendering? Is it because we want to make the snapshots smaller and as a composed component, we expect those child components to be fully tested?

packages/card/loading.js
```diff
@@ -0,0 +1,30 @@
+import React from "react";
```
jimhunty: Why are we removing the `card` prefix? This doesnt match the pattern used in other packages.

packages/card/proptypes.js
```diff
@@ -0,0 +1,31 @@
+import PropTypes from "prop-types";
```
jimhunty: Peronsal preference but the grouping of two different components proptypes and default props here doesnt feel right.

packages/author-profile/author-profile-item.web.js
colincclark: 1 is the default for mobile. The `childRatio` prop is only really used for responsiveness on web. 2.7 is specific to authorprofile tablet+ content atm

packages/card/__tests__/web/card.web.test.js
colincclark: We want the snapshots to be useful and readable, not just smaller. This can be seen in the snapshots on this PR and the snapshots in `Slice`

packages/card/loading.js
colincclark: I thought it was a better pattern ..will put it back for now

packages/card/proptypes.js
colincclark: My preference is in one file, but will move to separate named files to match elsewhere

packages/author-profile/author-profile-item.web.js
colincclark: this is a flex ratio, not an aspect ratio of an image...needs to be a number

colincclark: plus it is the specific style used in this one place only...

packages/card/card-loading.web.js
```diff
@@ -1,57 +1,48 @@
...
   return (
     <CardContainer>
-      {showImage && imageComponent}
-      <SummaryContainer>
-        <Gradient style={[gradientStyles.headerContainer]} degrees={264} />
-        <Gradient style={[gradientStyles.textContainer]} degrees={267} />
-        <Gradient style={[gradientStyles.textContainer]} degrees={267} />
+      {showImage && (
+        <ImageContainer>
+          <Image uri="" aspectRatio={aspectRatio} />
```
craigbilner: is the empty uri still needed?

colincclark: I am going with no, and removing it

packages/card/card-proptypes.js
```diff
@@ -0,0 +1,21 @@
+import PropTypes from "prop-types";
+
+export const cardPropTypes = {
+  image: PropTypes.shape({ uri: PropTypes.string }),
```
craigbilner: think image has been iterated on the most out of any component but....is there value if an object with a `uri` prop here?

colincclark: refactoring this would break the lazy loading stuff and involve a lot of digging / investigation into how to make it work again...felt that refactor was out of scope

craigbilner: :+1:

PR #690, opened 2018-02-28
Chore/depend cleanup & integration

nikhedonia:
minor changes:
- improves docs
- loads lerna.json from cwd if `-l` or `-e` are not set
- adds lint and fix scripts to root package.json
- list is now sorted
- fixes one test


packages/depend/README.md
```diff
@@ -91,7 +91,7 @@ Consider following repo:
 }
 ```
 
-`depend -g` outputs all dependencies in `.dot` format:
+`depend -l . -g` outputs all dependencies in `.dot` format:
```
craigbilner: this doesn't need to be documented now you've made the other change?

nikhedonia: true, but I prefer to be explicit

PR #694, opened 2018-03-01
Refactor: Functional colours part 2

aronshamash:
Updated:

- `article`
- `article-byline`
- `article-flag`
- `article-summary`

also:
- Agreed with Jake that we should use a consistent colour for the keylines hence the change in article package

screen shot 2018-03-02 at 12 07 57


packages/styleguide/colours/functional.js
```diff
@@ -1,20 +1,17 @@
...
-  gold: "#8F7F75",
-  grey: "#666666",
   greyDark: "#333333",
-  greyLight: "#D0CECE",
   greyMid: "#696969",
+  gunmetalGrey: "#4D4D4D",
   keyline: "#DBDBDB",
+  muletaRed: "#C51D24",
   offWhite: "#F9F9F9",
+  pictonBlue: "#3C81BE",
```
colincclark: can we get annotations on the designs with these names to make it easier to get the right one?

colincclark: can we use the website that you pointed out to ensure better naming consistency? for example, `"#1D1D1B"` is not _really_ black...

aronshamash: I'll speak with Jake, but in the case of `#1D1D1B` we are using it as *our* black

aronshamash: I agree with annotations though

packages/article/styles/article-header/shared.js
```diff
@@ -1,18 +1,18 @@
...
+    color: colours.functional.timesBlack,
     marginBottom: 7,
     fontFamily: fonts.headline
   },
   standFirst: {
     fontSize: 20,
     lineHeight: 26,
     fontFamily: fonts.headlineRegular,
-    color: "#333333",
+    color: colours.functional.mineShaftGrey,
```
colincclark: ⛏

PR #696, opened 2018-03-02
chore: adds tests and for roleMap in slice

nikhedonia:
Lets increase coverage for slice!

### DIFF
--- a/packages/slice/__tests__/shared.js
+++ b/packages/slice/__tests__/shared.js
@@ -92,11 +92,11 @@ module.exports = () => {
     });
 
     it("should throw if child does not exist", () => {
-      expect(() => roleMap("DEFAULT", 3)).toThrow();
+      expect(roleMap("DEFAULT", 3)).toBe(null);
     });
 
     it("should throw if template does not exist", () => {
-      expect(() => roleMap("NonExistantTemplateName", 0)).toThrow();
+      expect(roleMap("NonExistantTemplateName", 0)).toBe(null);
     });
   });
 };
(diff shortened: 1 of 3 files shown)


packages/slice/__tests__/shared.js
```diff
@@ -84,4 +85,18 @@ module.exports = () => {
...
     });
   });
+
+  context("RoleMaps", () => {
+    it("returns the name if template and index exists", () => {
+      expect(roleMap("DEFAULT", 0)).toEqual("default-1");
+    });
+
+    it("should throw if child does not exist", () => {
+      expect(roleMap("DEFAULT", 3)).toBe(null);
```
colincclark: this does not throw, it returns null....test name is incorrect

packages/slice/styles/role-map.js
```diff
@@ -9,5 +9,17 @@ export default (template, index) => {
     ]
   };
 
-  return templateObject[template][index];
+  const roles = templateObject[template];
```
colincclark: will `return templateObject[template][index] || null;` work just as well?

nikhedonia: would throw if template does not exists.

nikhedonia: `templateName[index]` would return `undefined` instead of `null` if index does not exist.
What does our styleguide say regarding `undefined` vs `null` ?

colincclark:
```
const templateName = templateObject[template] || null;
return templateName ? templateName[index] || null : null;
```

PR #701, opened 2018-03-03
Refactor: Change article as a standalone prop

nehasri89:
This PR exports article as an independent prop which can be used for validation in the consumers i.e. render

### DIFF
--- a/packages/article/article-prototype.js
+++ b/packages/article/article-prototype.js
@@ -1,0 +1,12 @@
+import PropTypes from "prop-types";
+import ArticleHeader from "./article-header/article-header";
+import ArticleMeta from "./article-meta/article-meta";
+
+const ArticlePrototype = {
+  article: PropTypes.shape({
+    ...ArticleHeader.propTypes,
+    ...ArticleMeta.propTypes
+  })
+};
+
+export default ArticlePrototype;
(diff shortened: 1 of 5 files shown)


packages/article/article-prototype.js
```diff
@@ -0,0 +1,12 @@
+import PropTypes from "prop-types";
+import ArticleHeader from "./article-header/article-header";
+import ArticleMeta from "./article-meta/article-meta";
+
+const ArticlePrototype = {
```
colincclark: Can the just be called `articlePropTypes` to match other packages?

packages/article/article.web.js
```diff
@@ -5,6 +5,7 @@ import Ad, { AdComposer } from "@times-components/ad";
 
 import ArticleError from "./article-error";
 import ArticleLoading from "./article-loading";
+import { ArticlePrototype, ArticleDefaultProps } from "./article-prototype";
```
colincclark: Can this be called `articleDefaultProps` instead of `ArticleDefaultProps`? Also, it does not exist in `"./article-prototype"`...should it?

packages/article/article-prototype.js
nehasri89: 👍

packages/article/article.web.js
nehasri89: yeah, somehow I failed to commit changes to article-proptype default vars. done!

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/e09f85690f0546807ed6f0d2be7f476abf115874/dextrose-test/index.html

PR #695, opened 2018-03-02
Feat/video icon

nikhedonia:
Needed for https://nidigitalsolutions.jira.com/secure/RapidBoard.jspa?rapidView=736&modal=detail&selectedIssue=REPLAT-1314

PR #703, opened 2018-03-05
Feat/video label

nikhedonia:
Based on the work of: https://github.com/newsuk/times-components/pull/695
Adds videoLabel story.
videolabel

Furthermore it makes width in Icons optional and computes widths from the aspect ratio and height

Design needs confirmation from jake


aronshamash:
@nikhedonia Could we not have extended the logic in the `article-label` component ?

nikhedonia:
@aronshamash sure, but that would make article-label package more complex as it would add an extra dependency (times/icons). I think packages should be minimal.

PR #706, opened 2018-03-06
feat: enable header bidding for ads

BernieSumption:
This PR ads the prebidding functionality to Ads. We have a number of bidders on the site.

- AppNexus
- Rubicon
- Amazon
- Criteo
- Pubmatic
- IndexExchange
We send the bids along with the `GPT` request for `DFP` to decide the winning bid. We then display the ad from the winning bidder on the slot

On the web, we send the `GPT` request once we have got the bids.
On native, we do not do any bidding

It also refactors and simplifies `DOMContext`, removing `dom-context-harness`.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/73efc88d7fe4fb84b5e1b424b2867c00b8ce0e7f/dextrose-test/index.html

packages/ad/ad-init.js
```diff
@@ -1,91 +1,368 @@
...
+          el.style.justifyContent = "center";
+          el.style.margin = "0 auto";
+          /* eslint-enable no-param-reassign */
+
+          const gptMapping = window.googletag.sizeMapping();
+          mappings.forEach(size =>
+            gptMapping.addSize([size.width, size.height], size.sizes)
+          );
+          slot.defineSizeMapping(gptMapping.build());
+          Object.entries(slotTargeting || {}).forEach(entry =>
```
nehasri89: Object.entries failed on the android emulator when I tested, can switch to Object.keys. It may not be available on all native devices

nikhedonia: If `Object.entries` does not exists maybe we want to `babel-polyfill` it ?

packages/ad/prebid-config.js
```diff
@@ -0,0 +1,146 @@
+import { getAdSizes } from "./generate-config";
+
+const timeout = 3000;
+const minPrice = 0.01;
+const maxBid = 15;
+const bucketSize = 0.25;
+
+const biddersConfig = {
```
nehasri89: As the config in this file would be environment specific i.e. different values for dev and prod, it should be provided by the consumer package which is render in this case?

packages/ad/package.json
```diff
@@ -50,7 +50,8 @@
     "enzyme-adapter-react-16": "1.1.0",
     "eslint": "4.9.0",
     "jest": "21.2.1",
-    "jsdom": "9.12.0",
+    "jsdom": "^9.12.0",
```
craigbilner: set-exact

packages/ad/webview-event-callback-setup.js
```diff
@@ -1,13 +1,12 @@
...
-// See the warning about the implications of this in dom-context-harness.js
 
 /* eslint-env browser */
 
 const webviewEventCallbackSetup = options => {
   const { window } = options;
   window.eventCallback = (type, detail) => {
     // delay until next frame as React Native message bridge is not set up immediately
-    window.requestAnimationFrame(() => {
+    window.setTimeout(() => {
```
craigbilner: is this a polyfill issue?

killanaca: no, it's related to the way that React Native message bridge works at the moment. We should take an eye on this for future improv.

PR #708, opened 2018-03-07
Feat/article with video label

nikhedonia:


packages/article/get-lead-asset.js
```diff
@@ -0,0 +1,11 @@
+export default function getLeadAsset({ leadAsset: asset }) {
+  if (!asset) return { isVideo: false, leadAsset: null };
+
+  const isVideo = !!asset.posterImage;
```
aronshamash: Is this a guarantee? will it always be populated?

packages/video-label/style/shared.js
```diff
@@ -0,0 +1,17 @@
...
+  title: {
+    fontFamily: fonts.supporting,
+    fontSize: 12,
+    lineHeight: 12,
+    fontWeight: "400",
+    letterSpacing: 1.2,
+    padding: 0,
+    margin: 0,
+    position: "relative",
+    top: 2
```
aronshamash: Can we remove this and increase the `lineHeight`?

packages/article/get-lead-asset.js
nikhedonia: Yes, It comes directly from the graphql-endpoint.

packages/video-label/style/shared.js
nikhedonia: Is it desired to remove the lineHeight in `article-label` also?

nikhedonia: I'm unable to center the text vertically by doing so.

packages/article/get-lead-asset.js
craigbilner: this isn't how you should determine a video though, you have a type system?

packages/article/article-header/article-header-label.js
```diff
@@ -1,24 +1,28 @@
 import React from "react";
 import PropTypes from "prop-types";
 import ArticleLabel from "@times-components/article-label";
+import VideoLabel from "@times-components/video-label";
```
craigbilner: the native and web versions are basically the same, would be tempted to create a base version with render prop?

packages/article/fixtures/article-with-video-asset.json
```diff
@@ -0,0 +1,113 @@
...
+              "attributes": {
+                "value":
+                  "Tony Hartley, who lives 50 yards away from the site, said he had found a man buried beneath the rubble. “Me and a friend lifted up a steel girder with about five other blokes and removed a man"
+              },
+              "children": []
+            }
+          ]
+        }
+      ],
+      "leadAsset": {
```
nehasri89: A video fixture should be something along the lines of http://cps-api-tnl-prod.elb.tnl-prod.ntch.co.uk/v0.8/document/sir-ken-dodd-dies-aged-90-two-days-after-marriage-32h7t6p2s

nehasri89: We have a story in the backlog for video integration, just wondering if there is any specific reason / requirement for doing just the video label or is the video integration already done if I may have missed it

nehasri89: 👍

packages/article/get-lead-asset.js
```diff
@@ -0,0 +1,11 @@
+export default function getLeadAsset({ type, leadAsset: asset }) {
+  if (!asset) return { isVideo: false, leadAsset: null };
+
+  const isVideo = asset.type === 'Video';
```
aronshamash: better

packages/article/article-header/article-header-label.js
aronshamash: This makes sense @nikhedonia

nikhedonia: Let's tackle this in a seperate PR

nikhedonia: create issue #723 to document this as a todo.

PR #710, opened 2018-03-08
feat: create a template system for Slice

colincclark:
- This is _the one PR that binds them all_
- Card no longer has breakpoints
- Author Profile consumes Card correctly
- Slice passes config to it's consumers (Related Articles) to determine layout options
- Related Article consumes Card
- Slice configures Related Article which in turn passes these config options to Card
- Web is configured using classNames
- Native is configured using properties


packages/slice/templates/lead/index.js
```diff
@@ -2,21 +2,24 @@ import React from "react";
...
+      <View style={[styles.item, { paddingTop: 0 }]}>{lead(leadConfig)}</View>
     </View>
-    {[support1, support2].filter(support => support !== null).map(support => (
-      <View key={support.key} style={styles.childContainer}>
-        <View style={styles.child}>{support}</View>
-      </View>
-    ))}
+    {[support1(supportConfig), support2(supportConfig)]
+      .filter(support => support !== null)
+      .map(support => (
```
craigbilner: optional: can destructure

packages/article/related-articles/related-articles.js
```diff
@@ -12,29 +12,63 @@ import withTrackingContext from "./related-articles-tracking-context";
...
+  const renderArticleItems = (articleItems, config) => {
+    const {
+      contentContainerClass = "",
+      headlineClass = "",
+      imageContainerClass = "",
+      summaryClass = "",
+      showImage = true,
+      showSummary = true
+    } = config;
+    return articleItems.map(articleItem => (
```
craigbilner: would move to the consumer i.e. default and call `renderArticleItem`

packages/article/related-articles/related-articles.js
```diff
@@ -12,29 +12,63 @@ import withTrackingContext from "./related-articles-tracking-context";
...
+              const article = articles[2];
+              return article ? renderArticleItems([article], config) : null;
+            }}
+          />
+        );
+    }
+  };
 
   return (
     <ScrollView style={{ marginTop: 10 }}>
```
colincclark: Will move this ScrollView to the stories

packages/slice/templates/lead/responsive.web.js
```diff
@@ -108,9 +77,40 @@ export const getLeadContainer = ({ hasSupports, supportCount }) => {
...
-      flex-grow: ${supportCount === 1 ? "1.75" : "1.25"};;
+      flex-grow: ${supportCount === 1 ? "2.75" : "1.5"};;
     `
   });
   LeadContainer.displayName = "LeadContainer";
   return LeadContainer;
 };
+
+export const getSupportContainer = ({ index }) => {
+  const SupportContainer = withResponsiveStyles(View, {
```
colincclark: Will ammend these Components to be semantic elements as they are only for web

packages/slice/templates/default/index.web.js
```diff
@@ -2,27 +2,29 @@ import React from "react";
...
-            <Separator key={`separator-${current.key}`} />,
-            current
-          ])}
-      </ChildrenContainer>
-    </SliceContainer>
+    <ConfigWrapper>
+      <SliceContainer>
+        <ChildrenContainer>
+          {renderItems(config)
+            .map(item => <ChildContainer key={item.key}>{item}</ChildContainer>)
```
craigbilner: consumer should take the item `ID` rather than use a `key` which is a React rendering concept which also fits into the other refactor to not use map

packages/article/related-articles/related-articles.js
colincclark: done

PR #712, opened 2018-03-08
fix: get dextrose web working

rjanjua:
Dextrose hasn't really worked on the web. This PR aims to fix that. 

I have updated the fructose version to fix an issue relating do dependecy of dependencies weirdness.

I have updated react-apollo to the latest release candidate to fix the 'getDataFromTree' issue we see in dextrose - which was a known bug in react-apollo a few versions ago.

I have also removed the dextrose clean and generate stories commands from the following packages as they use the storybook addon knobs - dextrose doesn't work with knobs:
  
- video-label
- slice


package.json
```diff
@@ -110,7 +110,8 @@
     "@storybook/react": "3.3.14",
     "@storybook/react-native": "3.3.14",
     "@times-components/depend": "*",
-    "@times-components/fructose": "3.1.14",
+    "@times-components/fructose": "3.2.2",
+    "babel-loader": "*",
```
nikhedonia: can we lock it down to `7.1.4` ?

packages/depend/package.json
```diff
@@ -25,7 +25,7 @@
       "!**/packages/depend/get-packages.js"
     ],
     "transformIgnorePatterns": [
-      "node_modules/(?!(react-native|react-apollo-temp|@times-components)/)"
+      "node_modules/(?!(react-native|react-apollo|@times-components)/)"
```
nikhedonia: would it still work if we remove react-apollo from the list?
We used to depend on a untranspiled file but I recall that I removed that dependency in one of my PRs

packages/provider/connect.js
```diff
@@ -30,14 +30,13 @@ export const makeGraphqlOptions = (
...
-    data: { error, loading, refetch, retry, ...result },
+    data: { error, loading, refetch, ...result },
     children,
     ...props
   }) =>
     children({
       error,
       refetch: () => {
-        retry(); // FIXME: remove this after react-apollo fixes https://github.com/apollographql/apollo-client/issues/2513
         refetch();
```
nikhedonia: this is redundant. Use a shorthand like in case of error

package.json
rjanjua: oh yeah, my bad

packages/provider/connect.js
rjanjua: i'm not sure what you mean

rjanjua: oh, yeah, don't worry - I understand now

packages/depend/package.json
rjanjua: no idea, but will try

packages/utils/graphql.js
```diff
@@ -2,8 +2,8 @@
 
 import React, { Component } from "react";
 import ApolloClient from "apollo-client";
-import { ApolloProvider } from "react-apollo-temp";
-import { MockLink } from "react-apollo-temp/lib/test-utils";
```
nikhedonia: Good spot! I wonder how that sneaked back into our codebase

craigbilner: it was never removed cos of the usual umd issues

packages/provider/connect.js
```diff
@@ -36,7 +36,7 @@ const connectGraphql = (query, propsToVariables) => {
   }) =>
     children({
       error,
-      refetch,
+      refetch: () => refetch(), // using shorthand causes a rect-native error
```
rjanjua: so I tested this in storybook, and the shorthand version results in a strange error: 

screen shot 2018-03-09 at 17 10 12

packages/utils/graphql.js
nikhedonia: You are right, was thinking about my older PR: 
https://github.com/newsuk/times-components/pull/613/files

PR #714, opened 2018-03-09
Feat: Styleguide spacing

aronshamash:
- First attempt to define and restrict the spacing and bring it all into the styleguide. 
- I only tackled a few packages in this PR but did a full audit of the codebase to condense the spacing to well defined multiples of our agreed design system base of `5`.
- Some feedback would be useful before rolling it out to the whole codebase.

```
const base = 5;

const spacing = {
  base: 1 * base,
  default: 2 * base,
  s: 3 * base,
  m: 4 * base,
  l: 5 * base,
  xl: 6 * base,
  xxl: 12 * base
};
```

### DIFF
--- a/packages/styleguide/spacing/index.js
+++ b/packages/styleguide/spacing/index.js
@@ -1,0 +1,15 @@
+const base = 5;
+
+const spacing = {
+  base: 1 * base,
+  standard: 2 * base,
+  headline: 3 * base,
+  insetStandard: 4 * base,
+  insetLarge: 6 * base,
+  stackSmall: 4 * base,
+  stackStandard: 5 * base,
+  stackMedium: 6 * base,
+  stackLarge: 12 * base
+};
+
+export default spacing;
--- a/packages/styleguide/spacing/index.js
+++ b/packages/styleguide/spacing/index.js
@@ -2,14 +2,12 @@ const base = 5;
 
 const spacing = {
   base: 1 * base,
-  standard: 2 * base,
-  headline: 3 * base,
-  insetStandard: 4 * base,
-  insetLarge: 6 * base,
-  stackSmall: 4 * base,
-  stackStandard: 5 * base,
-  stackMedium: 6 * base,
-  stackLarge: 12 * base
+  default: 2 * base,
+  s: 3 * base,
+  m: 4 * base,
+  l: 5 * base,
+  xl: 6 * base,
+  xxl: 12 * base
 };
 
 export default spacing;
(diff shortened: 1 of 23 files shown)


packages/styleguide/spacing/index.js
```diff
@@ -0,0 +1,13 @@
+const base = 5;
+
+const spacing = {
+  base: 1 * base,
+  default: 2 * base,
+  s: 3 * base,
```
colincclark: please do not abbreviate these to meanngless letters....`small`, `large`, etc is better. Also, I think these defaults are a layer of abstraction too far - I don't want to have to refer to this file everytime I want to check a property. Maybe just using the multiplication in place is enough?

aronshamash: I thought of just using `$spacing  *  2 ` but it would still allow a new developer to add spacing which is outside of the spacing guidelines. It would be too flexible imo.

The other extreme is to try and name spacing vars to match the component where they are used. Unfortunately there is such a wide variety of uses of `5px`, `10px`, `15px` and `20px` spacing both vertical and horizontal that there would either be huge repetition or spacing var names would need to be so generic as to cover all the use cases.

In the end we decided this was better then nothing as at least it will be in one place if we do decide to change in the future. 

The longer we wait the more 'non standard' spacing will get added.

aronshamash: reason to use `s` instead of `small` etc is that we could more easily add new larger spacing eg `xxxl`

trevorah:
👎 I personally wouldnt go down this route as it obscures simple values and becomes annoying when more edge-cases pop up. Just tell people that "spacing is *usually* a multiple of 5" and fix up by eye any mismatched components.

packages/styleguide/spacing/index.js
colincclark: IMO this does not really enforce using the `spacing` properties that you have defined, either. This approach does not cover enforecement. It does obfuscate the intended values by abstracting them away into a separate file. What is the intended goal of this work? Is it simply consistency, or enforcement, or both? Did you consider using Spacing components?

aronshamash: I think we want enforcement and consistency.  
Looks like we have a number of different opinions on this. Lets talk about it in the next refinement before any more work is done in this area.

jimhunty:
If designed according to the system, you could argue that no edge cases should pop up.

The purpose of this being to offer consistent pattern for spacing that is documented and used. This is then clearly defined so it can be used in UX, Design and Development.

aronshamash:
Agreed that we should take the route suggested by @trevorah. I will close this PR and open a new PR.

Issue #716, opened 2018-03-09
Storybooks not working due to multiple versions of storybook packaged


rjanjua:
**Do you want to request a _feature_ or report a _bug_?**
BUG

**What is the current behaviour?**
Running `yarn storybook` results in an empty storybook. 
If you `yarn install` on master (after removing the lock file - this seems to be the way that most people update the yarn.lock files) you will end up with multiple versions of storybook dependencies, this breaks storybooks. This is the second time this is happening in a month.

**If the current behaviour is a bug, please provide the steps to reproduce.**
checkout master
remove the yarn lock file *
yarn *
Run storybook.
*this is how people tend to update the yarn lock files in new branches.

**What is the expected behaviour?**
A storybook with lots of stories
**If this is a feature request, what is the motivation or use case for changing
the behaviour?**

**Please mention other relevant information such as the browser version, Node.js
version and Operating System.**

PR #717, opened 2018-03-09
chore: update storybook version to 3.3.15

rjanjua:
Updating storybook to 3.3.15 because it broke. Fixes issue #716 
I tested it against web and ios, It seems fine from what I can tell. But it will be worth someone with more familiarity with the stories to check it over.


aronshamash:
I've pulled checked locally and all seems good.

PR #719, opened 2018-03-12
Feat/responsive ads

killanaca:
Add support for responsive ads (different ads size for different breakpoint), no different bidding.
Note: branch off  from`feat/prebid-rebase`


packages/ad/ad.stories.js
```diff
@@ -96,7 +96,7 @@ storiesOf("Advertisement", module)
           Orci varius natoque penatibus et magnis dis parturient montes,
           nascetur ridiculus mus.
         </Text>
-        <Ad pos="header" section="news" contextUrl={articleUrl} />
+        <Ad pos="intervention" section="news" contextUrl={articleUrl} />
```
BernieSumption: "intervention" - not a valid pos? It's not in slot-options.json anyway. I think this is supposed to be "inline-ad". Other stories have it too, e.g. "render one ad - intervention".

nehasri89: @BernieSumption  it is a valid pos, I have checked in render. We should update the fixture for it, it has the size mapping for it and acts as a default ad size category

nehasri89: ok, I checked render again. `intervention` is only ever use as a default size mapping, and you are right it is not a valid pos @BernieSumption  👍

killanaca:
superseded by https://github.com/newsuk/times-components/pull/733

PR #722, opened 2018-03-13
feat: add support for article section

killanaca:

### DIFF
--- a/packages/article/article-body/article-body.web.js
+++ b/packages/article/article-body/article-body.web.js
@@ -3,11 +3,19 @@ import PropTypes from "prop-types";
 import ArticleRow from "./article-body-row";
 
 const ArticleBody = props => {
-  const contentArray = props.content.map((data, index) => ({
-    data,
-    index
-  }));
-
+  const { section, content: bodyContent } = props;
+  const contentArray = bodyContent.map((data, index) => {
+    const item = {
+      data,
+      index
+    };
+    if (data.name === "ad") {
+      item.data.attributes = Object.assign({}, item.data.attributes, {
+        section
+      });
+    }
+    return item;
+  });
   const BodyView = contentArray.map(content => (
     <ArticleRow key={`cont-${content.index}`} content={content} />
   ));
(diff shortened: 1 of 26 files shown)


packages/article/article-body/article-body.web.js
```diff
@@ -3,11 +3,19 @@ import PropTypes from "prop-types";
...
-
+  const { section, content: bodyContent } = props;
+  const contentArray = bodyContent.map((data, index) => {
+    const item = {
+      data,
+      index
+    };
+    if (data.name === "ad") {
+      item.data.attributes = Object.assign({}, item.data.attributes, {
+        section
```
fourlastor: should we add the section if it's not present? (it's not a required prop)

or should it be required?

packages/article/data-helper.js
```diff
@@ -17,11 +17,19 @@ const prepareDataForListView = articleData => {
...
   ].concat(
-    articleData.content.map((i, index) => ({
-      type: "articleBodyRow",
-      data: i,
-      index
-    }))
+    articleData.content.map((i, index) => {
+      const item = {
+        type: "articleBodyRow",
+        data: i,
```
fourlastor: maybe we can rename `i` to `data` (or to whatever `i` is currently)

packages/article/article-body/article-body.web.js
killanaca: if it not present, we're gonna use the default value from the ads component

packages/article/article-body/article-body.web.js
```diff
@@ -3,11 +3,19 @@ import PropTypes from "prop-types";
...
-  }));
-
+  const { section, content: bodyContent } = props;
+  const contentArray = bodyContent.map((data, index) => {
+    const item = {
+      data,
+      index
+    };
+    if (data.name === "ad") {
+      item.data.attributes = Object.assign({}, item.data.attributes, {
```
craigbilner: while it's not an eslint rule, perhaps object spread would be more consistent?

PR #725, opened 2018-03-13
Related links in article (Native only)

tuncaulubilge:
Added related articles to data-helper and to renderRow for articles. 
Fixed the onPress event for related article item press.

Paired with: @craigbilner on data-helper


packages/article/article-content.js
```diff
@@ -2,12 +2,12 @@ import React from "react";
 import PropTypes from "prop-types";
 import { FlatList } from "react-native";
 
-const ArticleContent = ({ data, renderRow }) => (
+const ArticleContent = ({ data, renderRow, onRelatedArticlePress }) => (
```
tuncaulubilge: Is there a better way to do this?

packages/article/data-helper.js
```diff
@@ -36,6 +42,10 @@ const prepareDataForListView = articleData => {
     })
   );
 
+  if (relatedArticlesData) {
+    data = data.concat({ type: "relatedArticles", data: relatedArticlesData });
```
craigbilner: reassigning to data is a little odd? Could just move to the bottom and return the concatted data. To make it more functional it would probably be nice to prefer destructuring over that splice too...

packages/article/article-content.js
craigbilner: as in having to pass around "event functions"? For native to control navigation it'll have to bubble up so the alternative would be more React `context` :smirk:

packages/article/data-helper.js
tuncaulubilge: Used push instead of concat, so we kept data as const 6159eb9

craigbilner:
@tuncaulubilge an alternative: https://gist.github.com/craigbilner/bf7d8d9eb1134c52c4702e31114edc22

packages/article/data-helper.js
```diff
@@ -1,5 +1,21 @@
 import getLeadAsset from "./get-lead-asset";
 
+const prepend = ({ type, data }, list) => {
```
aronshamash: could we unit test these 2 helpers?

tuncaulubilge: Working on it 👍

craigbilner: they should be covered with "behavioural" tests?

tuncaulubilge: they are, we have separate fixtures for these cases

PR #728, opened 2018-03-13
refactor: extract label logic

craigbilner:
We often have native and web code that is very similar but due to the inter mingled differences between web and native and/or web having to provide a "wrapper" the inversion of control isn't always obvious. This is a suggested approach by providing a "render function" which is essentially a callback to allow the consumer to decide how to render the label


jimhunty:
Why is Video Label a package? Surely this is an extension of the Article Label package? This removes the need for this code in Article. Also it keep everything label related in one place.

PR #730, opened 2018-03-14
refactor: remove test ignore hacks and increase coverage

colincclark:
https://nidigitalsolutions.jira.com/browse/REPLAT-1586

- remove most coverage ignore hacks in Slice
- rename files and assign default exports to variables for increased and appropriate code coverage and to follow convention
- fixes a rebase issue where roles where left in incorrectly named directories


packages/slice/templates/leadandtwo/index.js
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import { View } from "react-native";
 import { propTypes, defaultProps } from "./proptypes";
-import styles from "../styles";
+import styles from "../shared.native";
```
jimhunty: I dont see the need for the name change here, this goes against conventions in other packages. I would suggest creating a `styles` folder under templates and having `index.js` and `responsive.web.js`. This then falls in line better.

colincclark: Nice. Done

PR #750, opened 2018-03-15
Feat/Headerbidding slots props

nehasri89:
This PR moves the information about the slot that needs bidding to be provided by the page which is consuming the Ads


nehasri89:
https://github.com/newsuk/times-components/pull/753

PR #751, opened 2018-03-15
feat: refactor slot targeting and add timestestgroup

killanaca:
Refactor slot targeting to share common properties and add timestestgroup logic


nehasri89:
LGTM

PR #754, opened 2018-03-16
fix: Put back the missing summary for Author Profiles with no images

jimhunty:
This is to fix REPLAT-1579 where on Author Profile pages with no images, we lose the summary text.

### DIFF
--- a/packages/author-profile/__tests__/author-profile-helper.js
+++ b/packages/author-profile/__tests__/author-profile-helper.js
@@ -201,6 +201,8 @@ export default AuthorProfileContent => {
   it("renders profile content item component with no image", () => {
     const item = cloneDeep(pagedResult(0, 1).data.author.articles.list[0]);
     set(item, "leadAsset.crop.url", null);
+    set(item, "shortSummary", item.summary)
+    set(item, "longSummary", item.summary)
     const component = renderer.create(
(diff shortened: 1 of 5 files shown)


colincclark:
Can you add a test that would pick this up in future?

jimhunty:
@colincclark I have had a look and for some reason its not working as expected. I think the Author Profile tests need refactoring and making shallow. Ok to look at this in another ticket?

colincclark:
As long as github issue is raised, that's fine

packages/author-profile/__tests__/author-profile-helper.js
```diff
@@ -201,6 +201,8 @@ export default AuthorProfileContent => {
   it("renders profile content item component with no image", () => {
     const item = cloneDeep(pagedResult(0, 1).data.author.articles.list[0]);
     set(item, "leadAsset.crop.url", null);
+    set(item, "shortSummary", item.summary);
```
colincclark: should this be in a separate test? it does not relate to image _per se_..?

jimhunty: If you look at the logic, it does actually directly relate to not having an image.

colincclark: ok

PR #755, opened 2018-03-16
Feat: Dynamic Summary lengths

aronshamash:
https://nidigitalsolutions.jira.com/browse/REPLAT-1315

Agreement by the team to use a length of `125` for native and `small` breakpoint for all templates.

# Web
- outputs different summary lengths based on the slice template and the number of items in the slice.
- shows/hides summaries based on breakpoint in `templates/{template_name}/config.web.js`.

## Chrome

dynamice

# Native
- update to `provider` package to slit `article` and `article.web` as for native we only need `summary125` but on web we need various summary lengths.

## iOS
screen shot 2018-03-16 at 14 21 05


packages/slice/templates/leadandtwo/index.web.js
```diff
@@ -23,11 +28,16 @@ const LeadAndTwoSlice = ({ lead, support1, support2 }) => {
   });
   const Separator = getSeparator({ hasLeftRightMargin: false });
 
+  const config = {
+    ...leadAndTwoConfig,
+    summaryConfig: summaryConfig[supportCount + 1]
```
colincclark: this is a bit weird...maybe named properties for this file might be more understandable?

packages/article/related-articles/related-article-item.js
```diff
@@ -53,7 +46,15 @@ const RelatedArticleItem = ({
           bylineProps={{ ast: byline }}
           content={() =>
             showSummary && (
-              <ArticleSummaryContent className={summaryClass} ast={summary} />
+              <View>
+                {summaryConfig.map(item => (
+                  <ArticleSummaryContent
+                    className={`summaryHidden summary${item}Class`}
```
colincclark: this still does not differentiate between `lead` and `support` articles...

packages/slice/templates/leadandtwo/index.web.js
aronshamash: like: 
```
{
 itemCount1: [125],
 itemCount2: [125],
 itemCount3: [124, 145]
}
``` 
?

packages/article/related-articles/related-article-item.js
aronshamash: Why does it need to differentiate? We do not have different rules on summary length for `lead` and `support` articles.

colincclark: we have different rules for displaying them at all. `support` articles should remain hidden at al bps

packages/slice/templates/leadandtwo/index.web.js
colincclark: yes, something like that might be better

packages/article/related-articles/related-article-item.js
colincclark: this is because you are attaching `summaryConfig` to the lead config only, which is a bit confusing. Maybe add more classNames for summary, like `leadSummary125Class` and supportSummary125Class` would be moe intuitive. Then the classNames can be used as a pattern, rather than not passing certain styles at all...?

PR #759, opened 2018-03-19
refactor: rename related articles template fixtures and stories

colincclark:
- renames fixtures and stories with Standard and Lead and Two templates (in related articles)
- consolidates and DRYs up the related articles tracking tests


packages/article/__tests__/related-articles-shared.js
```diff
@@ -42,150 +42,119 @@ export default () => {
...
         }
       };
       const tree = renderer
-        .create(<RelatedArticles {...createRelatedArticlesProps(data)} />)
+        .create(
+          <RelatedArticles {...createRelatedArticlesProps(data, events)} />
+        )
         .toJSON();
       expect(tree).toMatchSnapshot();
+      expect(events.mock.calls).toMatchSnapshot();
```
jimhunty: This second expectation creates a new snapshot with the same name? It would be good to assign them to different names to make the snapshots more readable.

colincclark: Done

PR #760, opened 2018-03-19
fix: export article default props

nehasri89:
This PR exposes article props which can be used for validation by the clients


nehasri89:
so, actually the real purpose of this PR was so that the consumer of this package i.e. render etc are able to use the props validation just for the article, for eg. the parallel PR in render is
https://github.com/newsuk/cps-content-render/pull/2203/files

PR #779, opened 2018-03-22
Article modal image

tuncaulubilge:
Started using modal-image instead of image in article lead assets and inline images. 

Only supported in ios/Android and falls back to image in web.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/2df724a839879f16ceffc5595b77379f2d3a9bf8/index.html

PR #783, opened 2018-03-26
[WIP] - Brightcove event cleanup

tuncaulubilge:
Brightcove android native was firing too many events and play/pause events were fired on wrong callbacks. 
- Replaced "PLAY" event with "DID_PLAY", and same for pause and seek events. 
- Removed the callback on progress, because it's fired every 10ms or so. 
- Deleted .android file. We don't really need a android specific logic in js. 
- Default style for fullscreen.
- Replace contextWrapper with the activity context, so Brightcove's FullScreenController functions properly


packages/brightcove-video/android/src/main/java/uk/co/news/rntbrightcovevideo/RNTBrightcoveView.java
```diff
@@ -105,24 +108,22 @@ private Activity getActivity() {
...
+    public void emitState(final boolean isPlaying, final int progress, boolean isFullScreen) {
         WritableMap event = Arguments.createMap();
 
-        if (isPlaying != null) {
-            Integer duration = playerView.getDuration();
-
-            event.putBoolean("isPlaying", isPlaying);
-            event.putBoolean("isFullscreen", playerView.getIsFullscreen());
-            event.putDouble("progress", progress);
+        Integer duration = playerView.getDuration();
```
fourlastor: probably best to have this as an `int`

packages/brightcove-video/android/src/main/java/uk/co/news/rntbrightcovevideo/RNTBrightcoveView.java
```diff
@@ -84,8 +84,11 @@ private void initPlayerView() {
             boolean hideFullScreenButton = this.hideFullScreenButton != null ? this.hideFullScreenButton : false;
             int theme = hideFullScreenButton ? R.style.FullScreenButtonDisabled : R.style.FullScreenButtonEnabled;
 
-            playerView = new BrightcovePlayerView(new ContextThemeWrapper(getActivity(), theme));
-
+           if(hideFullScreenButton) {
+               playerView = new BrightcovePlayerView(new ContextThemeWrapper(getActivity(), theme));
```
fourlastor: we can remove `getActivity()` in favor of `getContext()`

packages/brightcove-video/android/src/main/java/uk/co/news/rntbrightcovevideo/RNTBrightcoveView.java
```diff
@@ -84,8 +84,11 @@ private void initPlayerView() {
...
+           } else {
+               playerView = new BrightcovePlayerView(getActivity());
```
fourlastor: discussed offline: it's good to know this works, but right now we're depending on the default theme of the user of this lib to have fullscreen enabled by default (which isn't an assumption we can make, e.g. the android times client has it disabled by default)

For now we'll go with using the context wrapper and the correct theme (and accept it's not working for the time being).

packages/brightcove-video/android/src/main/java/uk/co/news/rntbrightcovevideo/RNTBrightcoveView.java
```diff
@@ -105,24 +108,22 @@ private Activity getActivity() {
...
-            }
+        event.putBoolean("isPlaying", isPlaying);
+        event.putBoolean("isFullscreen", isFullScreen);
+        event.putDouble("progress", progress);
 
-            event.putBoolean("isFinished", duration == progress);
-            ReactContext reactContext = (ReactContext) getContext();
-            reactContext.getJSModule(RCTEventEmitter.class).receiveEvent(getId(), "topChange", event);
+        if (duration > 0) {
+            event.putDouble("duration", duration);
```
fourlastor: this can be `putInt`

packages/brightcove-video/brightcove-player.android.js
```diff
@@ -1,111 +0,0 @@
-import React, { Component } from "react";
```
fourlastor: 🙏

PR #790, opened 2018-03-28
Rename test file

fourlastor:
This is because a file named *.stories.*.js will be treated as a story by storybook on native, and then it won't be found when trying to run storybook, causing a compilation error.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/d3f6021e1ed5113100b4f0d18915dbaccc4b156f/index.html

PR #787, opened 2018-03-28
feat: Update related article field denoting the main article

aronshamash:
Use a common field name to denote the main article in a related article layout.

Related TPA PR : https://github.com/newsuk/times-public-api/pull/332

PR #782, opened 2018-03-23
Feat/layout opinion related articles

colincclark:
https://nidigitalsolutions.jira.com/browse/REPLAT-1615

- adds images to opinion template for related articles (there needs to be a future story to fix these images on the storybook for native as they currently do not display properly) - https://github.com/newsuk/times-components/issues/791)
- styles opinion related articles
- adds opinion byline to article summary for new opinion byline UI
- extend DatePublication component to show or hide the day within the date (a requirement for all related articles)
- fixes numerous layout issues across devices (tested on Firefox, Safari, Chrome, IE and native)
- changes the fixed article width at tablet and desktop breakpoints as directed by Jake Cox

## Chrome
opinion 2

PR #794, opened 2018-03-29
fix: Main Id in related article is now required

tuncaulubilge:
Converted mainId in related articles to be required, as it is required for leadAndTwo. Created a shared function so storybook, native and web can use the same ruleset.

Also added mainId to native and web articles.

### DIFF
--- a/packages/article/related-articles/get-main-related-article-id.js
+++ b/packages/article/related-articles/get-main-related-article-id.js
@@ -1,8 +1,0 @@
-export default function getMainRelatedArticleId(relatedArticlesLayout) {
-  return (
-    relatedArticlesLayout.lead ||
-    relatedArticlesLayout.main ||
-    relatedArticlesLayout.opinion ||
-    ""
-  );
-}
--- a/packages/article/related-articles/get-main-related-article-id.js
+++ b/packages/article/related-articles/get-main-related-article-id.js
@@ -1,0 +1,6 @@
+export default function getMainRelatedArticleId(relatedArticlesLayout) {
+  return (
+    relatedArticlesLayout.main ||
+    ""
+  );
+}
--- a/packages/article/related-articles/get-main-related-article-id.js
+++ b/packages/article/related-articles/get-main-related-article-id.js
@@ -1,6 +1,8 @@
 export default function getMainRelatedArticleId(relatedArticlesLayout) {
   return (
+    relatedArticlesLayout.lead ||
     relatedArticlesLayout.main ||
+    relatedArticlesLayout.opinion ||
     ""
   );
 }
--- a/packages/article/related-articles/related-articles-proptypes.js
+++ b/packages/article/related-articles/related-articles-proptypes.js
@@ -3,7 +3,7 @@ import { relatedArticleItemPropTypes } from "./related-article-item-proptypes";
 
 export const relatedArticlesPropTypes = {
   articles: PropTypes.arrayOf(relatedArticleItemPropTypes.article),
-  mainId: PropTypes.string.isRequired,
+  mainId: PropTypes.string,
   onPress: PropTypes.func.isRequired,
   template: PropTypes.string
 };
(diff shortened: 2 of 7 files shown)


packages/article/related-articles/get-main-related-article-id.js
```diff
@@ -0,0 +1,8 @@
+export default function getMainRelatedArticleId(relatedArticlesLayout) {
+  return (
+    relatedArticlesLayout.lead ||
```
aronshamash: I have removed `lead` and `opinion` fields in this PR https://github.com/newsuk/times-components/pull/787

tuncaulubilge: Great :+1: I'll remove the function when that pr is merged. Or you can just add mainId to articles.js and articles.web.js, and we destroy this pr altogether

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/65fa0673bab15603129ba25233a5a58420f6d556/index.html

packages/article/related-articles/get-main-related-article-id.js
colincclark: Work on this was done in https://github.com/newsuk/times-components/pull/782

tuncaulubilge: This is breaking the mobile apps right now, so if these all depend on the TPA release and your PR, that might take some time and I'd rather merge this in first, so the apps are unblocked.

tuncaulubilge: @colincclark We also need to pass in mainId in articles.js so we can use leadAndTwo inside mobile apps. Also I think mainId should be required, as leadAndTwo just crashes without it

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/fa262e9930be73b8c58c34b008b5cd07fcf2845e/index.html

packages/article/related-articles/related-articles-proptypes.js
```diff
@@ -3,7 +3,7 @@ import { relatedArticleItemPropTypes } from "./related-article-item-proptypes";
 
 export const relatedArticlesPropTypes = {
   articles: PropTypes.arrayOf(relatedArticleItemPropTypes.article),
-  mainId: PropTypes.string,
+  mainId: PropTypes.string.isRequired,
```
colincclark: the only problem with this, is the `StandardSlice` does not take a `mainId`, so should technically not be required.... maybe just error handling around `LeadAndTwoSlice` and `OpinionAndTwoSlice` which do require `mainId`?

packages/article/related-articles/get-main-related-article-id.js
```diff
@@ -0,0 +1,8 @@
+export default function getMainRelatedArticleId(relatedArticlesLayout) {
```
colincclark: this needs a simple unit test

packages/article/related-articles/get-main-related-article-id.js
colincclark: As discussed, we should treat the first article as the `main` one if there is no `mainId`. `mainId` should not be required because there won't be one for the `StandardSlice` template

packages/article/related-articles/related-articles-proptypes.js
tuncaulubilge: Done in 40de868

packages/article/related-articles/get-main-related-article-id.js
tuncaulubilge: Using this function in tests in b3f29f1 . That should be enough to cover it imo. Do we really need a separate unit test?

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/b3f29f1d4ed92b47fd2a8621be3e9b60bc01ef3c/index.html

packages/article/related-articles/get-main-related-article-id.js
colincclark: this will suffice

packages/article/related-articles/get-main-related-article-id.js
colincclark: you should remove `.lead` and `.opinion` now though...these should never exist anymore

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/3200c8d521beb9c03d04de9bec63662735e7975d/index.html

PR #799, opened 2018-04-03
fix: proptypes for related articles proptypes for native

colincclark:
- native only has the summary125

@tuncaulubilge


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/1be8c955a53c0b32509c17123ee87c5ed8796b1e/index.html

PR #801, opened 2018-04-03
Add filter to depend

fourlastor:
This PR adds an `--only` or `-on` (I went for `-on` instead of `-o` cause `-o` is usually used for output, could also just remove it) flag to restrict the packages on which `depends` act on.

The rationale behind that is that sometimes we're not interested in the whole deps graph, sometimes we are.

Talking to @nikhedonia it seems like in web implementations, for example, we want all the packages to be aligned because some CSS classes will be otherwise referring to different versions (so different styles), in native implementations, we're mostly interested in libraries that have a native library to be included on the project (e.g. `react-native-linear-gradient`), because there's only one entry point for these. The rest is ok because it'll simply be included with different versions in different packages.

### DIFF
--- a/packages/depend/__tests__/__snapshots__/cli.test.js.snap
+++ b/packages/depend/__tests__/__snapshots__/cli.test.js.snap
@@ -79,3 +79,18 @@ Array [
   ],
 ]
 `;
+
+exports[`depend cli tests restricts to only requested packages 1`] = `
+Array [
+  Array [
+    "1",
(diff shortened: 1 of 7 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/656df088fb86f15470104f5953c8e9f4835aac2c/index.html

PR #802, opened 2018-04-03
[BREAKING CHANGE] Change to render player component immediately on all platforms 

jackharvey1:
Fixes two tap mobile web bug

Code is non-viable in terms of a styled native component.

Functionality confirmed using storybook.


packages/brightcove-video/src/brightcove-video.js
```diff
@@ -87,26 +83,15 @@ class BrightcoveVideo extends Component {
...
-      );
-    }
-
     return (
-      <TouchableWithoutFeedback onPress={this.play}>
-        <View style={{ width: this.props.width, height: this.props.height }}>
-          <Splash {...this.props} />
-        </View>
-      </TouchableWithoutFeedback>
+      <Player
```
fourlastor: why this change if the bug is only on web?

packages/brightcove-video/src/brightcove-player.web.js
```diff
@@ -28,6 +56,14 @@ class BrightcoveVideo extends Component {
     document.body.appendChild(s);
   }
 
+  static attachStyles() {
+    const styleTag = document.createElement('style');
```
fourlastor: I think this shouldn't touch directly the document, instead you should do something similar to https://github.com/newsuk/times-components/blob/master/packages/article/src/styles/article-body/responsive.web.js

Then you can assign the style with the `style` attribute

jackharvey1: Correct me if I'm wrong but I don't think that would work. In this case we are styling classes which are attached to nodes that we do not have control over but are instead attached by the Brightcove script.

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/69501ba7172ead37c5a55684cb0f27e60a6f6bc3/index.html

packages/brightcove-video/src/brightcove-player.web.js
BernieSumption: @jackharvey1 The concern here is server-side rendering - at a minimum, you'd need to check whether `document` exists before using it to avoid breaking SSR.

However, do you have to create the style element programmatically? I'd have thought that you could just emit a style tag and CSS string as HTML, and this would work in SSR as well?

BernieSumption:
@jackharvey1 A general comment - This component had a major version bump recently, as we changed its internal HTML structure. Previously, 100% sizing didn't work, and so Render had to use CSS to [reach into the player DOM and fix things](https://github.com/newsuk/cps-content-render/blob/master/src/public/styles/_components/_media.scss#L271-L286)

It should be possible to remove some of the styles in the link above, certainly the bit inside `.Media-videoElement > div > div`. It's possible that other styles may need updating, ping me if you hit issues.

packages/brightcove-video/src/brightcove-video.js
jackharvey1: Craig has said that the A-team are carrying out an overhaul to the Brightcove player so this is acceptable for resolving the issue in the meantime.

packages/brightcove-video/src/brightcove-player.web.js
jackharvey1: @BernieSumption I wouldn't know as I'm not too experienced with React but I'm happy to carry out the change

jackharvey1:
@BernieSumption Yeah Toby told me about that before so I've already made that change to render

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/88f884f8f80b499309d01de68866ae6130813495/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/88f884f8f80b499309d01de68866ae6130813495/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/654897bd502b5ee9de887027e5678889d4f917c1/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/cca4b187996f5970a0b9e63800700cc1d7bfa098/index.html

packages/brightcove-video/src/brightcove-player.web.js
craigbilner: @BernieSumption this is in `componentDidMount` so not for SSR

packages/brightcove-video/src/brightcove-video.js
craigbilner: @fourlastor more that if the A-team could take any further changes related to this fix so @jackharvey1 can focus on editions. @jackharvey1 this could just be a `.web` change though?

jackharvey1: How would you suggest isolating the changes to `.web.js`?

packages/brightcove-video/src/brightcove-player.web.js
BernieSumption: Ah yes, of course. I take it all back it's fine :)

packages/brightcove-video/src/brightcove-video.js
craigbilner: @jackharvey1 if it's easier for the A-team you can cut a web version of that core player file and make the change just there

jackharvey1: Oh I see, a _new_ `.web` file. Yeah I can try that out.

jackharvey1: @craigbilner just spoken to @BernieSumption and as they are now going to be moving on to the video component work and as the component isn't currently used by any apps they're happy to take this in the state it's currently in and resolve the platform specific bits. 

With that in mind would you be happy to approve?

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/bc5f12cf2d0aaaf76accc32c4a9be466389251d5/index.html

PR #807, opened 2018-04-04
fix: Use modal image in lead asset image

tuncaulubilge:
Replaced image with modal image for lead assets. Modal image is only supported in native and falls back to image


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/bbb7a52ccb77c5f5c312750ba8a878f005eb8932/index.html

PR #812, opened 2018-04-04
Removes postinstall transpilation

jamesseanwright:
Removes `postinstall` transpilation of each package, as our dependents should be using the provided, pre-generated transpiled code. We originally included this behaviour to support specific dev setups, but in retrospect this is an edge case.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/3c683a06b91082b816f2898294c07542326c713a/index.html

PR #813, opened 2018-04-05
fix: fixed article lead asset provider

BernieSumption:
Fixed a mismatch between fixtures and graphql data that resulted in lead asset video not working.

Also added a development utility: <StorybookProvider> is a component to enable stories to be switched between mock data or a graphql endpoint supplied via a knob. Builds on @craigbilner's STORYBOOK_ENDPOINT feature.


packages/provider/src/article.js
```diff
@@ -21,6 +21,10 @@ export const query = gql`
           posterImage {
             ...imageProps
           }
+          brightcoveAccountId
+          brightcoveVideoId
+          brightcovePlayerId
```
fourlastor: we don't need this field on mobile

packages/storybook/storybook-provider.js
```diff
@@ -0,0 +1,49 @@
...
+        {props.children}
+      </ApolloProvider>
+    );
+  }
+
+  return <MockedProvider mocks={props.mocks}>{props.children}</MockedProvider>;
+};
+
+StorybookProvider.propTypes = {
+  mocks: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
```
fourlastor: an array of empty objects?

BernieSumption: `PropTypes.shape({})` is the only way of specifying arbitrary objects in PropTypes, since PropTypes.any is disabled. It's used in quite a few places.

packages/provider/src/article.web.js
```diff
@@ -21,6 +21,10 @@ export const query = gql`
           posterImage {
             ...imageProps
           }
+          brightcoveAccountId
+          brightcoveVideoId
+          brightcovePlayerId
+          brightcovePolicyKey
         }
```
nehasri89: there is another field `paidonly`

packages/storybook/storybook-provider.js
fourlastor: I'd honestly disable the linting is that the result we want to achieve. This feels like a workaround which isn't really better. Not for this PR tho

packages/provider/src/article.web.js
BernieSumption: Yes, but it's currently unused in the client.

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/515fd7a80b617372e15cef6abfee68840b59b06e/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/8c919ad99a7e58c2abcf7b6a1dd043ea94674c2d/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/0b81e118e729eab0abf80d579a41d0f2bda7ba9e/index.html

PR #815, opened 2018-04-05
Removes provider fixtures from .npmignore

jamesseanwright:
Removes `fixtures` from the provider package's .npmignore files, so that consumers can use the exposed `fixtureGenerator` member.


craigbilner:
will fix in render

PR #825, opened 2018-04-06
Feat/rewrite the story

craigbilner:
Storybook doesn't always play nice with our tooling and plus we may move away from it in the future. This PR introduces the concept of a showcase which is lib agnostic for the developer who just writes plain ol' JavaScript.

The component function is given `knobs` and `actions` respectively to be consumed by the component which can then be mocked out by tooling for example. Other than that it's a very straightforward "storybook builder" that iterates over the showcase structure.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/19cdb2d33684100bdcf18088c726111d87101e10/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/a83c10d770d306074fa10953e70cf4e247d6599a/index.html

colincclark:
> Storybook doesn't always play nice with our tooling

What specific issues have we faced that this PR resolves?

packages/caption/caption.showcase.js
```diff
@@ -0,0 +1,55 @@
...
+      name: "Without credits",
+      component: () => <Caption text={captionText} />
+    },
+    {
+      type: "story",
+      name: "With credits",
+      component: () => <Caption text={captionText} credits={credits} />
+    },
+    {
+      type: "story",
```
jimhunty: Is a `type` required?

packages/icons/icons.showcase.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+});
+
+export default {
+  name: "Primitives/Icons",
+  children: [
+    {
+      type: "story",
+      name: "Icons",
+      component: ({ color }) => (
```
jimhunty: This level of nesting is difficult to decipher. Can this be extracted?

nikhedonia: We could make it slightly better by returning a map:
```
export default {
  "Primitives/Icon": [
    {...},
    {...}
  ]
}
```
This would also enable us to return multiple story groups.

packages/caption/caption.showcase.js
nikhedonia: I don't think it is required we could just check define the following behavior:
1) apply decorator if defined
2) add story if component is defined

L0wry: type will be really useful for dextrose when it comes to filtering

craigbilner: indeed, I would see types as useful for other tools that want to use/ignore certain "elements"

packages/icons/icons.showcase.js
craigbilner: yeah, unfortunately the knobs have to be provided at the component/runtime level

craigbilner:
@colincclark the various tooling we want to use doesn't always play well with the various plugins, largely because they're only written and tested in a specific (likely web) use case. This allows us to be both storybook agnostic and also do more interesting things around Expo and Dextrose for example

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/985b0eacd4fd926b0c0ae3baa82cc90e4952a155/index.html

PR #848, opened 2018-04-11
fix: do not publish on npm settings files

killanaca:
like we already do inside some packages


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/917415daf3a986342fa3ed22f9e66b6e98f30cde/index.html

PR #857, opened 2018-04-12
feat: make card use a placeholder image if no image is found

colincclark:
https://nidigitalsolutions.jira.com/browse/REPLAT-1670

Card will now always show a times placeholder image unless showImage prop is explicitly set to false
Added tests and stories for this eventually

### DIFF
--- a/packages/card/card.stories.js
+++ b/packages/card/card.stories.js
@@ -63,6 +63,19 @@ storiesOf("Composed/Card", module)
       </Card>
     </CardWrapper>
   ))
+  .add("Default - with placeholder", () => (
+    <CardWrapper>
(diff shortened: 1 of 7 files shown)


packages/card/card.stories.js
```diff
@@ -88,6 +101,20 @@ storiesOf("Composed/Card", module)
       </Card>
     </ReversedCardWrapper>
   ))
+  .add("Default - reversed with placeholder", () => (
```
aronshamash: Do we need 2 reversed card stories?

colincclark: this one is for the placeholder...

PR #858, opened 2018-04-12
chore: bump react to version 16.3.1

BernieSumption:


packages/storybook/src/knobs.js
```diff
@@ -0,0 +1,9 @@
+import * as nativeKnobs from "@storybook/addon-knobs/react";
```
craigbilner: what makes them `nativeKnobs`?

BernieSumption: Built-in knobs would have been more appropriate.

PR #864, opened 2018-04-13
feat: adds a11y to storybook

nikhedonia:
...and also fixes addon info


colincclark:
Having pulled the branch, the accessibility tab does not seem to provide much value. Every single story just say

> Element has insufficient color contrast of 3.84 (foreground color: #ffffff, background color: #2288cc, font size: 9.0pt, font weight: normal). Expected contrast ratio of 4.5:1

craigbilner:
agree with @colincclark we should only add addons with value

nikhedonia:
Bummer...wonder if this is expected or a bug.

PR #865, opened 2018-04-13
fix: Update video attribute

nehasri89:


aronshamash:
I think there should be some snapshots to update and maybe fixtures too.

colincclark:
what does this do?

nikhedonia:
Apparently the schema changed:
https://github.com/newsuk/times-public-api/pull/354/commits/b9d560464786bab4c58bb246220a3f4f79890c3e

Is anybody relying on the old schema with `paidonly` ?
If so we need to deprecate it first before we can rename it.

nehasri89:
@nikhedonia - no, only TC is consuming it atm

nehasri89:
@colincclark - This is to align it to the latest tpa changes for video. @aronshamash - fixtures updated

PR #867, opened 2018-04-13
Feat/REPLAT-1761/488 - article byline styling

jamesseanwright:
Covered by [REPLAT-1761](https://nidigitalsolutions.jira.com/browse/REPLAT-1761) and [REPLAT-488](https://nidigitalsolutions.jira.com/browse/REPLAT-488).


packages/article-byline/article-byline.stories.js
```diff
@@ -2,6 +2,7 @@ import React from "react";
 import { Text } from "react-native";
 import { storiesOf } from "@storybook/react-native";
 import { fonts, colours, fontSizes } from "@times-components/styleguide";
+import { boolean, selectV2, withKnobs } from "@storybook/addon-knobs";
```
colincclark: can we be consistent and use `import { select } from "@times-components/storybook";` as used elsewhere?

packages/article-byline/article-byline.stories.js
```diff
@@ -29,53 +30,52 @@ const bylineLinkStyles = {
   }
 };
 
+const sections = Object.keys(colours.section);
+
+const ArticleBylineWithCommonProps = props => (
```
colincclark: Can you create a default props object and pass that to ArticleByline instead? This could be in a func if you did not want to repeat the <Component syntax... this would be consistent with the other packages, and is less confusing than with the weird Component name

colincclark:
```
const createArticleByline = props => (
   <Article...etc
```

packages/article/fixtures/article-with-byline-author.json
```diff
@@ -0,0 +1,183 @@
+{
+  "data": {
+    "article": {
+      "id": "17e53696-3dd7-11e8-b8eb-a2b1e54e4e88",
+      "headline": "The Queen may be a child of the Prophet Muhammad",
```
colincclark: lol @ headline

packages/article-byline/article-byline.stories.js
jamesseanwright: Done!

packages/article-byline/article-byline.stories.js
jamesseanwright: I had tried this, but it didn't work with the Storybook knobs. However, I was able to work around this by writing a function for getting the common props, so that said knobs are reinvoked on each render.

packages/article/fixtures/article-with-byline-author.json
jamesseanwright: I saw it in production and just had to use it.

packages/article-byline/src/article-byline.js
```diff
@@ -16,22 +16,49 @@ const linkStyles = StyleSheet.create({
...
+const ArticleByline = ({
+  ast,
+  isCurrentEdition,
+  isPastSixDays,
+  section,
+  style,
+  onAuthorPress
+}) => {
+  const sectionStyle = {
+    color: colours.section[section] || colours.default
```
tuncaulubilge: In the native world, we are still using the old api, so I suspect the section names are not the same as tpa. Also, section colouring is done in the native apps. So for the native apps, it would be better to pass in the colour code directly, instead of a section name

jamesseanwright: Good to know, I'll take a look.

tuncaulubilge: Had a short chat with Craig on the issue. We are going to use the section value that we receive from tpa for the article, so this bit is okay. I thought we are going to pass it through the native apps but it turns out that's not the case. :+1:

packages/article-byline/src/article-byline.js
```diff
@@ -16,22 +16,49 @@ const linkStyles = StyleSheet.create({
   }
 });
 
-const ArticleByline = ({ ast, style, onAuthorPress }) =>
-  renderTrees(ast, {
+const ArticleByline = ({
+  ast,
+  isCurrentEdition,
```
aronshamash: I was thinking that the parent component would just pass a prop like `useSectionColour` and the logic for determining this prop's value would be done in the parent. I don't think having business logic in such a simple component is correct. 

The article label also needs to follow the same business rules around colour so doing it in the lower level components `articleByline` and `articleLabel` would have to be replicated.

craigbilner: yah, a byline component wouldn't have any idea about editions

packages/article-byline/src/article-byline.js
craigbilner: yes the section colour will come from TPA so this logic should be stripped

craigbilner: https://github.com/newsuk/times-public-api/pull/366/files#diff-afbcd088e21fff8c5fcca6cac0a73182R6

packages/article-byline/src/article-byline.js
jamesseanwright: @craigbilner @aronshamash I've updated it to simply take a `sectionColour` property.

jamesseanwright:
@craigbilner I've dumbified the component. Please can you take a look when you have a chance?

packages/article-byline/__tests__/web/__snapshots__/article-byline.web.test.js.snap
```diff
@@ -115,9 +142,11 @@ exports[`Article Byline test on web renders correctly with styles 1`] = `
 
 exports[`Article Byline test on web renders correctly with the author at the end 1`] = `
 Array [
-  <span>
+  <div
```
craigbilner: is `span` to a `div` expected?

jamesseanwright: Oddly, this `inline` AST entry is still rendered within a `span` in Storybook. Perhaps it's an issue with the serialiser.

jamesseanwright: Looks like it's caused by this marvellous line in `react-native-web`:

```js
const component = isInAParentText ? 'span' : 'div';
```

I'll investigate how my changes are causing this tomorrow morning.

packages/article-byline/article-byline.stories.js
```diff
@@ -29,44 +31,54 @@ const bylineLinkStyles = {
   }
 };
 
+const sectionColours = Object.assign(
```
colincclark: This loses some of the section colors. Assigning keys like this overwrites some of the keys, so the duplicates fall out of the assigned object. For example, `artsfeatures` has a value of "#622956", but so does `"#622956"`, so `artsfeatures` does not end up in the final list. This needs a rethink, and should probably then be a tested utils function

jamesseanwright: That's a good spot. The reason I'm computing this object is because our own `select` abstraction reverses the keys and the values of the object that's passed as a parameter; it worked OK for me with Storybook's own `selectV2`; could we make an exception in this case and use it directly here?

colincclark: Yes, maybe add the reason to the README if appropriate 👍

jamesseanwright: I'll add it as a comment.

jamesseanwright: I've done this, but there's still a quirk where selecting a shared colour will change it, but will then select the first occurrence of said colour in the list e.g. selecting `times2` will successfully change the colour but the rendered `select` will display `artsfeatures`. Looks like this is the behaviour of `selectV2`, so I can't work around it, but at least the missing colours are now present.

packages/article-byline/__tests__/web/__snapshots__/article-byline.web.test.js.snap
craigbilner: Looks like it uses `display:inline` if not wrapped in a `Text` element so will let it go

PR #868, opened 2018-04-13
build: add clean up to dist folder when transpile is run

colincclark:


jamesseanwright:
The `cleanup-dist` script needs to be added to the CLI.

PR #875, opened 2018-04-16
refactor: tealium package

colincclark:
- BREAKING CHANGE: new API, tealium function now takes the scheduler Class
- amended `tealium-utils` package to use the new API
- fixed, refactored and extended the tests


packages/tealium/__tests__/web/tealium.js
```diff
@@ -0,0 +1,43 @@
+import createTealiumReport from "../../src";
+import TealiumSendScheduler from "../../src/tealium-send-scheduler";
+
+jest.mock("../../src/tealium-send-scheduler");
```
craigbilner: this isn't ideal, and the index/default export thing is a little odd, could we just move the code into index and allow it ot take a scheduler?

colincclark: done

PR #876, opened 2018-04-16
feat: display secondary video

nehasri89:
This PR adds the capability to display a secondary video as a '16:9 '


aronshamash:
Do we not need the capability to display video with display `primary` and `secondary`?

nehasri89:
@aronshamash we do, now on current render we display all videos as 'primary', so a similar behaviour on TC, we ignore the display type, and display all videos as 16:9

PR #877, opened 2018-04-17
Fix: read Jest intra-monorepo dependencies from source

jamesseanwright:
Currently, when Jest runs a test from one of our monorepo packages, any other packages upon which it is dependent will be loaded from the transpiled code. This means that if we were to change the `ArticleByline` component, then the `ArticleSummary` package would be unaware of any updates when we run Jest, thus we'd have to transpile the former manually.

This PR introduces a custom transformer. It's based upon `babel-jest`, as is the `react-native` Jest preset, but any requests for `@times-components` monorepo dependencies will be read from the `src` directory instead of `dist`. This means that local changes are immediate between test re-runs.


jamesseanwright:
Had a brainwave: will reopen shortly.

nikhedonia:
Cool stuff! maybe you want to move it into a seperate package?

jamesseanwright:
@nikhedonia I was on-the-fence about this, but ultimately decided to keep it in the `jest-configurator` package as it won't be required anywhere else. Do you have any particular advantages to separating it in mind?

packages/jest-configurator/__tests__/source-loader.test.js
```diff
@@ -0,0 +1,100 @@
+import * as babelJest from "babel-jest";
+import * as fs from "fs";
+import { getCacheKey, process as transform } from "../src/source-loader";
+
+jest.mock("babel-jest");
+jest.mock("fs", () => jest.genMockFromModule("fs"));
+
+describe("source loader", () => {
+  describe("getCacheKey", () => {
```
colincclark: top tip: context is better for nesting in jest - optional

packages/jest-configurator/src/source-loader.js
```diff
@@ -0,0 +1,53 @@
...
+
+const pointToSource = (filename: string): boolean =>
+  filename.replace("dist", "src");
+
+/* Based upon the babel-jest impl, but only
+ * changes if raw source code has been updated.
+ */
+const getPackageCacheKey = (filename: string): string =>
+  crypto
+    .createHash("md5")
```
colincclark: this is kicking out a warning: `DeprecationWarning: crypto.createCredentials is deprecated. Use tls.createSecureContext instead.` weird, as `createHash()` still seems legit?

jamesseanwright: I saw this, but I thought it was a red herring thrown from another package. I'll have another look.

jamesseanwright: So it does only occur in my branch, but as you say I'm not using the deprecated method. I imagine it's being used under-the-hood in our current Node.js version. I'll dig deeper.

packages/jest-configurator/__tests__/source-loader.test.js
jamesseanwright: Are we using `context` elsewhere in the project? I can't seem to find any references to it.

nikhedonia: I think I removed `context` a while ago as after one of our updates it stopped working.
Not sure if it would still cause an issue.

colincclark: It is still in use. https://github.com/newsuk/times-components/blob/master/packages/related-articles/__tests__/shared.js#L1

PR #886, opened 2018-04-19
Feat/shavingbar icons

nikhedonia:
* Icons added for Shaving bar
* height prop made optional as SVG to enable automatic scaling when height&width not set
* title now passed to svg (for accessibility)
* stroke colour added to all icons to be able to invert colours.
* cleaned up tests


packages/icons/src/icons/facebook.js
```diff
@@ -0,0 +1,30 @@
...
+    >
+      <Path d="M20.8754687,30.3972581 L20.8754687,21.0930804 L23.9984274,21.0930804 L24.4660462,17.4670773 L20.8754687,17.4670773 L20.8754687,15.1521034 C20.8754687,14.1022918 21.1669846,13.3868799 22.6723889,13.3868799 L24.5924599,13.3860409 L24.5924599,10.1429146 C24.2603909,10.0987257 23.1206168,10 21.7945782,10 C19.0262491,10 17.1309761,11.6898042 17.1309761,14.7929991 L17.1309761,17.4670773 L14,17.4670773 L14,21.0930804 L17.1309761,21.0930804 L17.1309761,30.3972581 L20.8754687,30.3972581 Z" />
+    </G>
+  </Svg>
+);
+
+IconFacebook.propTypes = iconPropTypes;
+
+IconFacebook.defaultProps = {
+  fillColour: colours.functional.brandColour
```
colincclark: I think `defaultProps` would be better colocated with the props themselves

colincclark: Also, some of the unrequired props are missing defaults

packages/icons/src/icons/proptypes.js
```diff
@@ -1,7 +1,9 @@
 import PropTypes from "prop-types";
 
 export default {
+  title: PropTypes.string,
```
colincclark: pls change the file name to `prop-types.js`

packages/icons/src/icons/facebook.js
nikhedonia: comment in prop-types.js

PR #887, opened 2018-04-20
fix: storybook native compilation error

BernieSumption:
fixed a bug around tealium-utils. It was lacking a cleanup-dist step in transpile, and this allowed a compilation bug to creep in that was hidden by stale files in the dist folder.

### DIFF
--- a/packages/tealium-utils/package.json
+++ b/packages/tealium-utils/package.json
@@ -9,7 +9,8 @@
     "prettier:diff": "prettier --list-different '**/*.*'",
     "lint": "eslint . && yarn prettier:diff",
     "prepublishOnly": "yarn transpile",
-    "transpile": "babel src --ignore node_modules,__tests__ -d dist"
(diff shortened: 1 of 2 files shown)


packages/tealium-utils/package.json
```diff
@@ -9,7 +9,8 @@
     "prettier:diff": "prettier --list-different '**/*.*'",
     "lint": "eslint . && yarn prettier:diff",
     "prepublishOnly": "yarn transpile",
-    "transpile": "babel src --ignore node_modules,__tests__ -d dist"
+    "cleanup-dist": "rm -rf dist",
+    "transpile": "yarn cleanup-dist && babel src --ignore node_modules,__tests__ -d dist"
```
nikhedonia: do we still need the `--ignore` ?

BernieSumption: Not sure... I defaulted to not touching it.

PR #889, opened 2018-04-20
feat: paidOnly videos and new video package

BernieSumption:
This PR adds a web-only feature to disable the play button on a video marked as paid-only.

This PR contains a major refactor of the BrightcoveVideo component, which removes all native functionality. This native functionality is used by external contributors, so instead of making a breaking change to `@times-components/brightcove-video`, a new package video has been created called `@times-components/video`.

Highlights:

* removed native implementations of inline player - new player uses a callback handled by the host app
* removed numerous unused features such as the ability to externally play and pause the video and respond to progress events
* Rewrote stories for new functionality, and ported to use showcase
* Added {platform: "web|native|ios|android"} option to showcase to allow platform-specific stories.


packages/video/src/video.js
```diff
@@ -0,0 +1,73 @@
...
+const Video = ({
+  poster,
+  width,
+  height,
+  onVideoPress,
+  brightcoveAccountId,
+  brightcovePolicyKey,
+  brightcoveVideoId
+}) => (
+  <TouchableOpacity
```
tuncaulubilge: Can we use Link component instead of this? Link package uses TouchableNativeFeedback on android, instead of the TouchableOpacity, which is better suited for android ux

BernieSumption: Sure, I'll update it now.

BernieSumption: @tuncaulubilge I couldn't use Link because it's designed to take an URL and complains if you don't provide one. So I just updated Video to use TouchableNativeFeedback instead of TouchableOpacity on Android. If we get another use case of a clickable element that doesn't have a URL then we can split the Touchable element I created into a new package.

packages/video/src/video.proptypes.js
```diff
@@ -0,0 +1,22 @@
+import PropTypes from "prop-types";
+
+const numberOrString = PropTypes.oneOfType([
+  PropTypes.string,
+  PropTypes.number
+]);
+
+export const propTypes = {
+  accountId: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
BernieSumption: This naming is copied from the old component which this is mostly a drop-in replacement for. brightcovePolicyKey is converted to policyKey at a higher level in the article component tree.

PR #892, opened 2018-04-23
Native inline videos

tuncaulubilge:
Inline video click callback added and inline video splash aspect ratio fixed for native


packages/article/src/article-body/article-body-row.js
```diff
@@ -85,18 +84,22 @@ const ArticleRow = ({ content: { data, index }, onLinkPress }) =>
         caption
       }
     ) {
+      const aspectRatio = 16 / 9;
```
aronshamash: I think we have this aspectRatio in various components. Would it be worth sharing it?

tuncaulubilge: There is an AspectRatioContainer component with takes strings as props: https://github.com/newsuk/times-components/blob/master/packages/article/src/article-body/media-aspect-ratio.js#L6

This seems to be the only place where we use 16/9 as a hardcoded value, other than tests. We could have a shared set of constants for widely used aspect ratios throughout the project

aronshamash: Also in related articles https://github.com/newsuk/times-components/blob/master/packages/related-articles/src/related-article-item.js#L39

tuncaulubilge: https://github.com/newsuk/times-components/issues/895

aronshamash: cool

PR #894, opened 2018-04-24
fix: warnings in font storybook

colincclark:
- fix warnings about invalid style and missing keys

### DIFF
--- a/packages/styleguide/styleguide.stories.js
+++ b/packages/styleguide/styleguide.stories.js
@@ -21,7 +21,7 @@ ColourBox.propTypes = {
 
 const fontDisplayer = (fontFamily, phrase) =>
   Object.keys(fontSizes).map(fontSize => (
-    <Fragment>
+    <Fragment key={fontSize}>
       <Text style={styles.subHeadline}>{fontSize}</Text>
       <Text
         style={[
--- a/packages/styleguide/styleguide.stories.styles.js
+++ b/packages/styleguide/styleguide.stories.styles.js
@@ -36,8 +36,7 @@ const styles = {
   headline: {
     fontSize: fontSizes.body,
     fontWeight: "bold",
-    marginBottom: 10,
-    textDecoration: "underline"
+    marginBottom: 10
   },
   subHeadline: {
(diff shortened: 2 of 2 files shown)


packages/styleguide/styleguide.stories.js
```diff
@@ -21,7 +21,7 @@ ColourBox.propTypes = {
 
 const fontDisplayer = (fontFamily, phrase) =>
   Object.keys(fontSizes).map(fontSize => (
-    <Fragment>
+    <Fragment key={fontSize}>
```
jimhunty: Should protect against font sizes not being unique.

colincclark: why? React gives warnings when keys are matched, and the fontSizes are supposed to be different. If someone made them the same, it would probably not be correct. Also, this follows the pattern elsewhere in the file: https://github.com/newsuk/times-components/blob/master/packages/styleguide/styleguide.stories.js#L43. Can you suggest a way to protect uniqueness in this case?

jimhunty: Add the index from the map

colincclark: That would not pass linting

craigbilner: a combination would, plus it's just a story where nothing changes

colincclark: @jimhunty 
```
const fontDisplayer = (fontFamily, phrase) =>
  Object.keys(fontSizes).map((fontSize, index) => (
    <Fragment key={fontSize + index}>
      <Text style={styles.subHeadline}>{fontSize}</Text>
      <Text
        style={[
          styles.showoffFonts,
          { fontFamily, fontSize: fontSizes[fontSize] }
        ]}
      >
        {phrase}
      </Text>
    </Fragment>
  ));
```

gets me:

`error  Do not use Array index in keys  react/no-array-index-key`

definitely does _not_ pass linting.

colincclark: I think this might be a non-issue - this at least fixes the multiple warnings in storybook which is an improvement?

PR #896, opened 2018-04-24
Feat/article snapshots

nehasri89:
This PR cleans up the article snapshots. It mocks the components outside of article package, to make the snapshot more relevant to the functionality done inside of the article package.

The components that the article are using should be tested in their own packages


packages/article/__tests__/android/__snapshots__/article.android.test.js.snap
```diff
@@ -1429,23 +1429,10 @@ exports[`Article tests on android should render a full article 1`] = `
...
-                "fontSize": 12,
-                "fontWeight": "400",
-                "letterSpacing": 1.2,
-                "marginBottom": 0,
-              }
-            }
-          >
-            H U R R I C A N E   I R M A
-          </Text>
+          <MockArticleLabel
```
colincclark: rather than calling them <MockBlah.../> can we just call them by the actual Component names?

packages/article/__tests__/shared-tracking.js
```diff
@@ -0,0 +1,81 @@
+import "react-native";
+import React from "react";
+import renderer from "react-test-renderer";
+import Article from "../src/article";
+
+import fullArticleFixture from "../fixtures/full-article.json";
+
+const adConfig = {
```
colincclark: `adConfig` is also used in `shared.js` - can we DRY it up a bit?

packages/article/__tests__/web/article.web.test.js
```diff
@@ -7,53 +7,75 @@ import ArticleRow from "../../src/article-body/article-body-row";
...
+jest.mock("@times-components/topics", () => "MockTopics");
+jest.mock("@times-components/watermark", () => "MockWatermark");
+jest.mock("@times-components/tracking", () => {
+  const mockTracking = component => component;
+  return {
+    withTrackingContext: mockTracking
   };
+});
 
+const adConfig = {
```
colincclark: and here...

PR #897, opened 2018-04-25
remove criteo and pubmatic

killanaca:


nehasri89:
tests need to be updated too

PR #898, opened 2018-04-25
fix: video package code review updates

BernieSumption:
Removed video from error-view package stories
Added ErrorView around Video component
More user friendly error message c/o Jake

Plus changes required after 1st attempt to integrate component into Render:

Exported IsPaidSubscriber.Provider
Fixed play button styling on newer versions of VideoJS


packages/video/__tests__/web/__snapshots__/video.web.test.js.snap
```diff
@@ -95,40 +95,54 @@ exports[`Video on web renders a paidOnly video correctly for unpaid users 1`] =
...
       style={
         Object {
-          "display": "block",
-          "position": "absolute",
-          "width": "100%",
-          "zIndex": 1,
+          "display": "table",
+          "height": 0,
+          "overflow": "hidden",
+          "paddingBottom": "NaN%",
```
colincclark: weird

packages/video/__tests__/web/inline-video-player.web.test.js
```diff
@@ -84,10 +84,12 @@ describe("InlineVideoPlayer", () => {
     expect(dispose).toHaveBeenCalled();
   });
 
-  it("renders correctly if the script fails to load", () => {
+  it("throws an exception from render if the script fails to load", () => {
+    jest.spyOn(console, "error").mockImplementation();
```
colincclark: should this `spy` be here?

packages/video/src/video-error.web.js
```diff
@@ -1,19 +1,27 @@
...
     textAlign: "center",
     color: "white",
-    width: "auto !important",
-    height: "auto !important",
+    width: "auto",
+    height: "auto",
     marginBottom: spacing(2)
   },
+  body: {
+    fontFamily: "TimesDigitalW04",
```
colincclark: can the styleguide package be utilised here?

packages/video/src/video.web.js
```diff
@@ -1,29 +1,72 @@
...
   paidOnlyValue === "false" ? false : !!paidOnlyValue;
 
+const styles = StyleSheet.create({
+  noSubscriptionWrapper: {
+    position: "absolute",
+    top: "50%",
+    left: 0,
+    right: 0,
+    height: 65,
+    marginTop: -32
```
colincclark: `-32` does not seem to fit with the vertical rhythm introduced in the `spacing` property - or is this a special case?

packages/video/src/video.web.js
```diff
@@ -1,29 +1,72 @@
...
+  },
+  noSubscriptionMessage: {
+    marginLeft: "auto",
+    marginRight: "auto",
+    width: 300,
+    maxWidth: "80%",
+    backgroundColor: "rgba(0, 0, 0, 0.7)",
+    color: "#FFFFFF",
```
colincclark: styleguide?

packages/video/src/video.web.js
```diff
@@ -1,29 +1,72 @@
...
+    fontFamily: fonts.body,
+    padding: spacing(2),
+    textAlign: "center"
+  }
+});
+
 const Video = props => {
```
colincclark: you can destruct here if you fancied a quick refactor

packages/video/src/video-error.web.js
```diff
@@ -1,19 +1,27 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { View, Text, StyleSheet } from "react-native";
-import withResponsiveStyles from "@times-components/responsive-styles";
 import { spacing } from "@times-components/styleguide";
 
-const errorStyling = StyleSheet.create({
+const styles = StyleSheet.create({
```
colincclark: the video files seem to have a lot of styles within them. Can we split those into a styles file in a styles folder please?

packages/video/__tests__/web/inline-video-player.web.test.js
BernieSumption: Yes, the behaviour we're testing involves a render method throwing and a componentDidCatch catching, which causes React to emit console errors. This silences them in a way that gets reset in the afterEach.

packages/video/src/video.web.js
BernieSumption: It's half the height, to vertically centre the panel in its container.

packages/video/__tests__/web/inline-video-player.web.test.js
colincclark: why do we need to silence console errors in a test?

BernieSumption: Because they're noisy. When you run tests and you get a single real failure, that failure can be drowned in the noise of 10 passing tests that emit console errors because the test is testing the handling of a failure condition. It's bad enough in development when you know what tests you're working on and so which ones are likely to be the real failures, and worse when investigating a arbitrary test failure in CI.

BernieSumption:
@colincclark made [some updates](https://github.com/newsuk/times-components/pull/898/commits/da71bf6e0af07e8790138e4fff2fb7f2de3791bc) driven by your review can you take a peek?

packages/video/src/styles/video-error.styles.web.js
```diff
@@ -0,0 +1,29 @@
...
+    color: "white",
+    width: "auto",
+    height: "auto",
+    marginBottom: spacing(2)
+  },
+  body: {
+    fontFamily: fonts.body,
+    fontSize: "14px",
+    textAlign: "center",
+    color: "rgba(255, 255, 255, 0.8)",
```
nehasri89: should take from styleguide

BernieSumption: Talked over this with Colin and decided against - this style is only used in one place.

PR #902, opened 2018-04-26
refactor: destruct Component in all the places

colincclark:
- destruct Component
- move eslint rules to appropriate jest mock folder


craigbilner:
@colincclark  we should really have an eslint rule for this if you want it as a practice, otherwise there's no value in the change?

colincclark:
@craigbilner the value without the rule is consistency, but it is definitely a good idea to enforce that with a rule too

PR #911, opened 2018-05-01
REPLAT-2210/Use React ART for gradient

fourlastor:
This PR switches `react-native-linear-gradient` with react native ART implementation.

On Android, this works out of the box, on iOS, we need to include the `node_modules/react-native/Libraries/ART` project and add the artifact as per [documentation](http://facebook.github.io/react-native/docs/linking-libraries-ios.html). React native link won't work for it so that's a manual process (and a tradeoff that we need to consider).

Currently the story without a style attribute won't show anything, I'm investigating what's up with it.


packages/icons/icons.stories.js
```diff
@@ -1,4 +0,0 @@
-import { showcaseConverter } from "@times-components/storybook";
```
tuncaulubilge: Don't we need these stories?

fourlastor: 😱 that wasn't supposed to happen

L0wry:
dextrose should be running again. Would be good to see it run on this PR!

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/d9323b064d304586f914bc720c94c245a8a6b118/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/01ea134ff9c1d60e13aad8264216c20649a899d0/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/7f447d25ba79e38d0ff308a4ba40f3ad076a88c3/index.html

PR #919, opened 2018-05-03
Fix/new dextrose

kaliabadi:
This breaks the testing of only changed components.

Implementing new fructose/dextrose to accomodate showcase format.

I have had to make one compormise in my implementation by abstracting the image showcase out into separate web and rest of the showcase files. This is because of it's import to react-dom.

We filter out stories based on platform within the application, so for web this works perfectly but for native if we have imports withing the showcase that are not in the module map then the bundle generation will fail.

The alternative solution I have though of was to use haul, however I could need further investigation into this as I can forsee some potential limitations for different users of fructose/dextrose. This could be a potential avenue for investitgation in the future, however this is a quick fix to get dextrose back up and running.


dextrose/dextrose.ios.js
```diff
@@ -14,7 +14,9 @@ const config = {
     "TrackingEvent",
     "Error",
     "error",
-    "AuthorProfileLoading"
+    "AuthorProfileLoading",
+    "Video",
+    "Tracking"
```
L0wry: add to android too!
Or export a common array to share

packages/pagination/pagination.showcase.js
```diff
@@ -5,7 +5,7 @@ import { LateralSpacingDecorator } from "@times-components/storybook";
 import Pagination, { withPageState } from "./src/pagination";
 import { PreviousPageIcon, NextPageIcon } from "./src/pagination-icons";
 
-export const pagination = {
+const pagination = {
```
L0wry: might be worth running storybook and testing if this still works after changing export

dextrose/dextrose.ios.js
kaliabadi: done

packages/pagination/pagination.showcase.js
kaliabadi: fixed 👍

PR #924, opened 2018-05-04
feat: updating to latest tc

nehasri89:
This PR is a result of changes required for integration with Render. 
We need to make sure that the version of libraries that TC and Render are using should be same, otherwise it causes a lot of integration issues i.e. styles broken on Render etc.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/a74006e9331fd45c59d804f52e62715e4685b8ad/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/41a63e45fb6f6a1d3a4626419cb92d0789dc43f4/index.html

craigbilner:
The tests are choking on the `ActivityIndicator` which uses this CSS:

`@media all {@-webkit-keyframes rn-anim-1309axw{0%{-webkit-transform:rotate(0deg);transform:rotate(0deg)}100%{-webkit-transform:rotate(360deg);transform:rotate(360deg)}}}` fails the CSSOM parser.

Either we need to allow nested keyframes in media queries on CSSOM or change RNW. As it appears to be valid CSS? the change would have to be in CSSOM which hasn't published for one year. As the loader is very temporary for the article and we'll probably move to skeletons when we move to a SPA? we can just remove the test

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/b34616a953cf440e7f5653ab4e3837556b6167c9/index.html

PR #926, opened 2018-05-08
feat: Topic provider

jimhunty:
Adds the Topic provider which will hook up to the new query on Times Public API.

This follows what is used on Author Profile as the Topic page looks to mimic this.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/6f87e746e376f5a83633d263395f3c6b1d92c54a/index.html

packages/provider/src/topic.js
```diff
@@ -0,0 +1,43 @@
+import gql from "graphql-tag";
+import connectGraphql from "./connect";
+
+export const query = gql`
+  query TopicQuery($slug: Slug!, $first: Int, $skip: Int, $imageRatio: Ratio!) {
```
tuncaulubilge: Don't we want this data in two separate queries, just like in author-profile? One for the head (including article count) and one for the actual article list, which is re-used in pagination. Otherwise, we'll end up fetching topic name and description for each page even if we don't need them.

jimhunty: Correct. A copy and paste fail. Now it is focussed just for the articles.

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/326c12c19528697eddb13673e117bda982cc4cdf/index.html

packages/provider/src/provider.js
```diff
@@ -19,5 +20,7 @@ export {
   articleQuery,
   authorProfileQuery,
   articleListNoImagesQuery,
-  articleListWithImagesQuery
+  articleListWithImagesQuery,
+  TopicProvider,
```
tuncaulubilge: TopicArticlesProvider and topicArticlesQuery?

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/bfe7853c516b010b938969f41c35ead32d33c64e/index.html

PR #927, opened 2018-05-08
Fix: Add storybook to devDependencies in CLI

aronshamash:
- Add storybook to devDependencies
- Remove extraneous colon in test template

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,8 @@ requires [Xcode](https://developer.apple.com/xcode),
 
 You can try without these requirements, but you'd be on your own.
 
-
 ## Viewing Our Components
+
 If you use Expo you are able to scan the QR code below
 
 <img src="https://s3.eu-west-1.amazonaws.com/nu-tools-expo/MasterExpo.png">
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -20,8 +20,8 @@ requires [Xcode](https://developer.apple.com/xcode),
 
 You can try without these requirements, but you'd be on your own.
 
-
 ## Viewing Our Components
+
```
colincclark: why is the README doing this?

lib/cli/templates.js
```diff
@@ -59,14 +59,14 @@ const testEsLintConfigFile = handlebars.compile(`{
 const testFileNative = platform =>
   handlebars.compile(`import shared from "../shared";
 
-describe("{{{component}}} tests on ${platform}: ", () => {
+describe("{{{component}}} tests on ${platform}", () => {
```
colincclark: 👍

README.md
craigbilner: it's not obvious why this has happened, will try and find some time to look into it but it should have failed linting if it was wrong previously....

PR #939, opened 2018-05-14
Infinite Scroll on Article List

tuncaulubilge:
Article list fetch more on scroll feature (infinite scroll)
- Used Apollo's fetchMore functionality, together with FlatList's onEndReached callback
- Added a loading state as footer
- Added a retry button as a footer in case of an error
- Bunch of tests
- Updated React Apollo as fetch more is buggy in older version. 

Paired with: @jackharvey1 

infinite-scroll-with-error


packages/article-list/__tests__/shared-native.js
```diff
@@ -107,47 +107,121 @@ export default () => {
...
+    return expect(
+      wrapper
+        .dive()
+        .dive()
+        .dive()
+        .dive()
+        .find("ListFooterComponent")
+        .dive()
+        .debug()
+    ).toEqual("<ArticleListRetryButton refetch={[Function: refetch]} />");
```
colincclark: you can make this a snapshot

packages/article-list/src/styles/index.js
```diff
@@ -27,6 +27,10 @@ const styles = StyleSheet.create({
   listErrorContainer: {
     flex: 1,
     margin: spacing(3)
+  },
+  loadingContainer: {
+    paddingTop: spacing(3),
+    paddingBottom: spacing(3)
```
colincclark: `paddingVertical`?

packages/article-list/__tests__/shared-native.js
jackharvey1: We actually pulled it out of a snapshot as I thought a snapshot was overkill, I think it's nicer to be able to see the expected output in the actual test where possible. Not too pushed though.

packages/article-list/src/styles/index.js
tuncaulubilge: 6989d7e

packages/article-list/__tests__/shared-native.js
tuncaulubilge: Let's move this to snapshots, and aim to make every snapshot as clear as this one :)

tuncaulubilge:
@colincclark , I'll add a test with article count, it's a good idea :+1:

packages/article-list/src/article-list.js
```diff
@@ -105,16 +78,28 @@ class ArticleList extends Component {
...
+        return null;
+      }
+
+      return fetchMore({
+        variables: {
+          skip: data.length
+        },
+        updateQuery
+      }).catch(err => this.setState({ loadMoreError: err }));
+    };
```
jackharvey1:
```
const fetchMoreOnEndReached = () =>
  this.state.loadMoreError
    ? null
    : fetchMore({
        variables: {
          skip: data.length
        },
        updateQuery
      }).catch(err => this.setState({ loadMoreError: err }));
```
I like ternaries

tuncaulubilge: d4ae324

packages/article-list/__tests__/shared-native.js
tuncaulubilge: 30d8aac

packages/article-list/__tests__/ios/__snapshots__/article-list.ios.test.js.snap
```diff
@@ -1,30 +1,19 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`ArticleList tests on ios should display error when fetch more fails 1`] = `
+"<View>
+  <ArticleListItemSeparator />
+  <View style={{...}}>
+    <ArticleListRetryButton style={{...}} refetch={[Function: refetch]} />
```
colincclark: should this be rendering an error, as the test name suggests, or just the retry button?

packages/article-list/src/styles/index.js
```diff
@@ -27,6 +27,16 @@ const styles = StyleSheet.create({
   listErrorContainer: {
     flex: 1,
     margin: spacing(3)
+  },
+  loadingContainer: {
+    paddingVertical: spacing(5)
+  },
+  showMoreRetryContainer: {
+    paddingHorizontal: spacing(5),
+    paddingVertical: spacing(5)
```
colincclark: this can just be `padding: spacing(5)`

packages/article-list/src/styles/shared.js
```diff
@@ -47,7 +47,7 @@ const styles = {
     backgroundColor: colours.functional.action,
     flex: 1,
     justifyContent: "center",
-    height: 45
+    height: 36
```
colincclark: Jake was pretty insistent on `45`, was this meant to change?

packages/author-profile/__tests__/shared-native.js
```diff
@@ -0,0 +1,22 @@
...
+      <MockedProvider mocks={mockArticles}>
+        <AuthorProfile {...props} />
+      </MockedProvider>
+    );
+
+    const flatList = tree.root.findByType(FlatList);
+
+    expect(flatList.props.data.length).toBe(pageSize);
+    await flatList.instance.props.onEndReached();
+    return expect(flatList.props.data.length).toBe(pageSize * 2);
```
colincclark: why the `return`?

packages/article-list/__tests__/ios/__snapshots__/article-list.ios.test.js.snap
tuncaulubilge: We just renders an retry button when there is an error. We didn't want to re-use the text+button layout as this will be displayed inside the list itself, shouldn't be that big

packages/article-list/src/styles/shared.js
tuncaulubilge: We got a new design on the retry button for the infinite scroll, it's better to keep it consistent so I also changed it here

times-tools:

times-tools:

times-tools:

packages/article-list/__tests__/shared-native.js
```diff
@@ -107,47 +107,123 @@ export default () => {
...
+
+    expect(
+      wrapper
+        .dive()
+        .dive()
+        .dive()
+        .dive()
+        .find("ListFooterComponent")
+        .dive()
+        .debug()
```
colincclark: not sure `debug()` is intended for usage in passing tests. use the `html()` method instead

packages/article-list/__tests__/shared-native.js
```diff
@@ -107,47 +107,123 @@ export default () => {
...
+      />
+    ).dive();
+    wrapper.setState({ loadMoreError: null });
+
+    await wrapper
+      .find("FlatList")
+      .props()
+      .onEndReached();
+
+    await wrapper
```
colincclark: seems duplicated

packages/article-list/src/article-list-prop-types-base.js
```diff
@@ -0,0 +1,18 @@
...
+  articleListHeader: PropTypes.element,
+  articles: PropTypes.arrayOf(PropTypes.shape()),
+  articlesLoading: PropTypes.bool,
+  isLoading: PropTypes.bool,
+  onArticlePress: PropTypes.func,
+  onViewed: PropTypes.func,
+  receiveChildList: PropTypes.func
+};
+
+export const defaultProps = {
```
colincclark: loading states need defaults. should be set to `true`

packages/article-list/src/styles/index.js
```diff
@@ -27,6 +27,15 @@ const styles = StyleSheet.create({
...
     margin: spacing(3)
+  },
+  loadingContainer: {
+    paddingVertical: spacing(5)
+  },
+  showMoreRetryContainer: {
+    padding: spacing(5)
+  },
+  showMoreRetryButton: {
+    elevation: 10
```
colincclark: this has android support only. will this mean inconsistencies between android and ios?

packages/article-list/src/article-list-prop-types-base.js
```diff
@@ -0,0 +1,18 @@
+import PropTypes from "prop-types";
+
+export const propTypes = {
+  articleListHeader: PropTypes.element,
+  articles: PropTypes.arrayOf(PropTypes.shape()),
```
colincclark: should pull in propTypes from somewhere to fill out the `shape()`

packages/article-list/src/article-list-prop-types-base.js
```diff
@@ -0,0 +1,18 @@
...
+  articlesLoading: PropTypes.bool,
+  isLoading: PropTypes.bool,
+  onArticlePress: PropTypes.func,
```
colincclark: shouldn't `onArticlePress` be required?

packages/article-list/__tests__/shared-native.js
jackharvey1: `html` is all one line isn't it? Means any difference will be for entire snapshot. I do hate how it's called `debug` though.

colincclark: Actually I tried the `html()` method and it did nort work at all! @tuncaulubilge Just lose the `debug()` method and snapshot the component with the style objects, etc. I am strongly opposed to using `debug()` in a test

packages/article-list/__tests__/shared-native.js
tuncaulubilge: It's intentional. Calling onEndReached twice and expecting it to be called only once, testing if functionality is blocked after an error. We could rename the test if it's not clear enough

packages/article-list/src/styles/index.js
tuncaulubilge: Only android designs have a shadow in zeplin, but I'll double check this with Jake

packages/article-list/__tests__/shared-native.js
tuncaulubilge: We could use the react-test-renderer instead but we'd end up with a huge snapshot, which is quite unreadable. How else can we snapshot the component?

packages/article-list/src/article-list-prop-types-base.js
tuncaulubilge: dae0a0a

times-tools:

times-tools:

packages/article-list/__tests__/shared-native.js
```diff
@@ -107,47 +107,122 @@ export default () => {
...
       />
-    ).root;
-    comp
-      .findAllByType(Pagination)[0]
-      .findAllByType(Link)[0]
-      .props.onPress();
-    expect(onPrev).toHaveBeenCalled();
+    );
+
+    tree.root.findByType(FlatList).props.onEndReached();
```
craigbilner: @tuncaulubilge does this render new data on the next tick? Might be a nicer test than checking the implementation?

packages/article-list/__tests__/shared-native.js
```diff
@@ -107,47 +107,122 @@ export default () => {
...
+    ).dive();
+    wrapper.setState({ loadMoreError: null });
+
+    await wrapper
+      .find("FlatList")
+      .props()
+      .onEndReached();
+
+    expect(
+      wrapper
```
craigbilner: :boat:

packages/article-list/__tests__/shared-web.js
```diff
@@ -0,0 +1,48 @@
...
+      <ArticleList
+        {...articleListProps}
+        articles={results.articles.list}
+        onArticlePress={onArticlePressMock}
+        page={1}
+        pageSize={pageSize}
+      />
+    );
+
+    wrapper
```
craigbilner: Don't you only need to dive if the whole tree isn't rendered which is the issue on Native but on web you can deeply render and then just find the element of interest?

packages/article-list/src/article-list.js
```diff
@@ -105,22 +78,33 @@ class ArticleList extends Component {
...
     if (!articlesLoading) this.props.receiveChildList(data);
 
+    const fetchMoreOnEndReached = () =>
+      this.state.loadMoreError
+        ? null
+        : fetchMore({
+            variables: {
+              skip: data.length
+            },
+            updateQuery
```
craigbilner: a little odd having this passed in to invoke with a passed in func. Perhaps refactor to just have `fetchMore` take `data` and then the consumer can do whatever it likes with that.

packages/article-list/__tests__/shared-native.js
tuncaulubilge: I'll add a new test with a mock provider in here

tuncaulubilge: Tried this, but we need a proper ArticleListProvider to be able to test this inside article-list package. I think we should re-discuss this issue after refactoring author-profile and topics to be "dumb" components.

times-tools:

tuncaulubilge:
We removed pagination.showcase.js and only using pagination.showcase.web.js 
We no longer have a native use for the pagination package so it only need a showcase.web. Somehow, expo is picking up the .web showcase in a native build

L0wry:
is it possible that pagination is still being used somewhere else a different component that is being picked up by expo?

times-tools:

craigbilner:
going to close until the review

times-tools:

packages/article-list/__tests__/shared-native.js
```diff
@@ -107,47 +107,122 @@ export default () => {
...
+        .dive()
+        .dive()
+        .dive()
+        .dive()
+        .find("articleListFooter")
+        .dive()
+    ).toMatchSnapshot();
+  });
+
+  it("should re-fetch more when retry button clicked", () => {
```
craigbilner: could you not simulate click on this one?

craigbilner: does the title of this test reflect what happens? It actually tests there's no error in the state?

tuncaulubilge: Renamed it to include error state, and also checking if fetchMore is called :+1:

tuncaulubilge: Turns out I can, if I dive deep enough. :) updated

times-tools:

PR #954, opened 2018-05-21
fix: Get Topic "Viewed" analytics working

jimhunty:
REPLAT-2154

Analytics matches that of Author Profile other than:

1. Articles Count has been removed as it seemed unnecessary and wouldn't work correctly. If this is wanted, then I suggest we should put it back in when integrating the Name and Description provider as this will then closer mimic Author Profile. (Awkward in-between phase)
2. Passing in `isLoading: false` to get around the fact that we don't have the name and description provider yet,

### DIFF
--- a/packages/topic/__tests__/topic-functional.js
+++ b/packages/topic/__tests__/topic-functional.js
@@ -82,17 +82,9 @@ export default () => {
       </MockedProvider>
     );
 
-    expect(reporter).toHaveBeenCalledWith(
-      expect.objectContaining({
-        object: "Topic",
-        component: "Page",
-        action: "Viewed",
-        attrs: expect.objectContaining({
-          topicName: "Chelsea",
-          page: 1,
-          pageSize
-        })
-      })
-    );
+    const call = reporter.mock.calls[0][0];
+    const callWithoutEventTime = call.attrs && call.attrs.eventTime ? [{ ...call, attrs: {...call.attrs, eventTime: null}}] : call;
+
+    expect(callWithoutEventTime).toMatchSnapshot("3. Send analytics when rendering a topics page (with null event time)");
   });
 };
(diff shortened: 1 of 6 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/00c7d56118b5204db8ee5bcb4d6b2e28a7a5df23/index.html

times-tools:

packages/topic/__tests__/topic-functional.js
```diff
@@ -35,19 +36,48 @@ export default () => {
...
+          {...topicProps}
+          page={1}
+          pageSize={pageSize}
+          analyticsStream={reporter}
+        />
+      </MockedProvider>
+    );
+
+    expect(reporter).toHaveBeenCalledWith(
+      expect.objectContaining({
```
craigbilner: for consistency this should be a snapshot?

jimhunty: I can change this, but this was a copy from Author Profile. Want me to update this and AP?

jimhunty: As this is checking a call on the analytics, i'm not sure how this could relate to a snapshot?

craigbilner: snapshotting is just a visual way to make an assertion

packages/topic/__tests__/topic-functional.js
```diff
@@ -73,17 +73,9 @@ export default () => {
...
-        component: "Page",
-        action: "Viewed",
-        attrs: expect.objectContaining({
-          topicName: "Chelsea",
-          page: 1,
-          pageSize
-        })
-      })
-    );
+    const call = reporter.mock.calls[0][0];
```
colincclark: does `expect(reporter.mock.calls[0]).toMatchSnapshot();` work just as well without the line below?

colincclark: Oh....you are removing eventTime so the snapshot is not affected. 👍

PR #956, opened 2018-05-21
Feat: REPLAT-2167 Topic name query

jimhunty:
This takes the work @jackharvey1 has done on TPA to expose Name and Description in a separate request and wraps the Topic component in the same way that Author Profile does. This means our Name and Description are data-driven.


times-tools:

L0wry:
rebasing should fix dextrose

packages/topic/__tests__/topic-functional.js
```diff
@@ -4,10 +4,12 @@ import { fixtureGenerator } from "@times-components/provider-test-tools";
 import { delay, MockedProvider } from "@times-components/utils";
 import Topic from "../src/topic";
 
-jest.mock("@times-components/article-list", () => "ArticleList");
+// eslint-disable-next-line global-require
+jest.mock("@times-components/article-list", () => require("./articleListMock"));
```
jackharvey1: does `jest.mock('./blah', () => ({ default: "ArticleList", ArticleListPageError: "ArticleListPageError" }));` not work?

jackharvey1:
Should we be updating the provider rather than adding another? Domain knowledge on this is still fuzzy but with this pattern aren't we going to require two providers feeding into the `TopicPage` component? Additionally this is two API calls when it could be one.

times-tools:

jimhunty:
@jackharvey1 Possibly. At the moment, I have just been copying the example set out in Author profile. If this is the case then we can look to refactor later.

jackharvey1:
@jimhunty should we not just do it now?

packages/provider/src/provider.js
```diff
@@ -8,6 +8,7 @@ import AuthorArticlesWithImagesProvider, {
   query as articleListWithImagesQuery
 } from "./author-articles-with-images";
 import ArticleProvider, { query as articleQuery } from "./article";
+import TopicProvider, { query as topicQuery } from "./topic";
```
jackharvey1: should we call this `topic-head`? In my head topic-head and topic-articles are siblings whose parent is both topic.

packages/topic/__tests__/topic-functional.js
colincclark: 👍

jackharvey1:
```
jest.mock("@times-components/article-list", () => ({
  default: "ArticleList",
  ArticleListPageError: "ArticleListPageError"
}));
```
works for me

packages/topic/__tests__/topic-functional.js
```diff
@@ -55,12 +57,18 @@ export default () => {
   });
 
   it("should render the loading state", () => {
+    const tree = renderer.create(<Topic {...props} isLoading />);
+
+    expect(tree).toMatchSnapshot("2. Render an topics page loading state");
```
colincclark: render _a_ topics page

packages/provider/src/provider.js
jimhunty: Again this was following the naming conventions set in Author Profile. If we change this here, we should change it there as well. Same with the provider as mentioned above. Something we should tackle in a separate ticket

packages/topic/__tests__/topic-functional.js
jimhunty: Still doesn't work for me locally

times-tools:

PR #958, opened 2018-05-21
Feat/rnw bundle

nikhedonia:
Creates a production ready react-native-web bundle for each package ready for consumption (in-render).

The bundle is placed in the top-level instead of in dist.
This allows you to import the RNW-bundle using
`import Article from "@times-components/article/rnw"` instead of  `import Article from "@times-components/article/dist/rnw"`


times-tools:

times-tools:

package.json
```diff
@@ -6,7 +6,7 @@
   "main": "index.js",
   "engines": {
     "node": ">=8.9.0",
-    "yarn": "1.6.0"
+    "yarn": ">=1.6.0"
```
colincclark: would it not be better to pin the `yarn` version here and in the CI rather than being _more_ permissive?

nikhedonia: AFAIK thinks work quite well with 1.6.0 and 1.7.0.
It just adds friction to artificially limit the number of compatible tools for this project.

craigbilner: `yarn` previously was a pain between versions and there's no real value in manually checking whether there were "allowable" breaking changes between versions?

packages/webpack-configurator/index.js
```diff
@@ -0,0 +1,70 @@
+const { resolve } = require("path");
+const { readFileSync, existsSync } = require("fs");
```
craigbilner: optional: async/await gives best of both worlds

packages/webpack-configurator/index.js
```diff
@@ -0,0 +1,70 @@
...
+
+const parseJson = path => {
+  return existsSync(path)
+    ? JSON.parse(readFileSync(path, "utf8").toString())
+    : {};
```
craigbilner: need lint and prettier in this package, empty object literal is a bit strange to represent the absence of something?

packages/webpack-configurator/index.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const babelConfig = dir => {
+  const babelrcPath = resolve(dir,'.babelrc');
+  const babelrc = parseJson(babelrcPath);
+  return {
```
craigbilner: :cake:

packages/webpack-configurator/index.js
```diff
@@ -0,0 +1,70 @@
...
+    ...babelrc,
+    plugins: [...babelrc.plugins, "react-native-web"]
+  };
+};
+
+function getEntry(dir, entry) {
```
craigbilner: perhaps abstract and test these?

packages/webpack-configurator/index.js
nikhedonia: unfortunately this would force us to return a promise but i don't think webpack supports promises of config files.

times-tools:

times-tools:

times-tools:

PR #960, opened 2018-05-22
chore: fix unhoisted modules

craigbilner:
There's currently a handful of packages that have their own `node_modules` because their versions aren't compatible for hoisting to the root. Have bumped the versions of these packages (`styled-components` for `hoist-non-react-statics`)


times-tools:

PR #961, opened 2018-05-23
Disable delta bundling on android for haul

tuncaulubilge:
Haul does not support delta bundles, so it had to be turned off manually. Found a way to turn it off automatically for android

### DIFF
--- a/android/app/src/main/java/com/storybooknative/MainActivity.java
+++ b/android/app/src/main/java/com/storybooknative/MainActivity.java
@@ -1,6 +1,9 @@
 package com.storybooknative;
 
+import android.os.Bundle;
+
 import com.facebook.react.ReactActivity;
+import com.facebook.react.devsupport.DevInternalSettings;
 
 public class MainActivity extends ReactActivity {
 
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/storybooknative/MainActivity.java
```diff
@@ -12,4 +15,13 @@
     protected String getMainComponentName() {
         return "storybooknative";
     }
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        DevInternalSettings settings = (DevInternalSettings) getReactNativeHost().getReactInstanceManager().getDevSupportManager().getDevSettings();
```
fourlastor: what happens on non dev builds? does this count as a no-op?

tuncaulubilge: This is times-components, so we only use it for dev builds. We will need some checks if we decide to do the same thing in the native app

fourlastor: I'm blind 😄

PR #969, opened 2018-05-24
chore: stop coverage collection of provider package showcase file

colincclark:
Coverage was being collected by `provider.showcase.js`, which it should ignore.

- remove `jest` config in provider package.json, as it should be using jest-configurator only
- remove unused eslint file
- moved `mount`ed debounce test to `web` folder to give it an environment from `jest-configurator`


times-tools:

PR #974, opened 2018-05-29
Topic article query pagination fix

tuncaulubilge:
Fixes pagination issue on topic articles


tuncaulubilge:
Better to fix this in TPA and update query along with the schema

PR #984, opened 2018-06-01
fix: article list static image

colincclark:
- fix article list error page image on native by making it base64

This is so the image is present to allow us to style the layout of the page properly across viewport sizes as part of https://nidigitalsolutions.jira.com/browse/REPLAT-2571


packages/article-list/src/article-list-page-error.js
```diff
@@ -9,8 +8,14 @@ const ArticleListPageError = ({ refetch }) => (
   <View style={styles.pageErrorContainer}>
     <View style={styles.pageErrorImageContainer}>
       <Image
-        aspectRatio={700 / 770}
-        uri="https://www.thetimes.co.uk/d/img/internal-error-c45d0e8347.png"
+        accessible
+        accessibilityLabel="Server error image"
```
aronshamash: Think `Server error` makes more sense here as for a screen reader the fact that it is an image is irrelevant.

Issue #912, opened 2018-05-01
BUG: Image package does not handle query strings well


colincclark:
**What is the current behaviour?**

Snaoshots prove that the image package mishandles query strings, and sometimes outputs `http://example.com/image.jpg&resize=1440` instead of 
`http://example.com/image.jpg?resize=1440` for example

**What is the expected behaviour?**

Query string should not be malformed, and should use `?` for the first parameter and `&` for subsequent parameters

colincclark:
Done in https://github.com/newsuk/times-components/pull/985

PR #985, opened 2018-06-01
fix: image component resize query string params

colincclark:
https://github.com/newsuk/times-components/issues/912

- fix query string formatting for images
- fix data url images - should not have query strings

### DIFF
--- a/packages/image/src/image.js
+++ b/packages/image/src/image.js
@@ -35,7 +35,10 @@ class TimesImage extends Component {
       onLoad: this.handleLoad
     };
 
-    if (cleanUri && width > 0) {
+    const regExImageDataUrl = new RegExp("data:", "gi");
+    const isDataImageUrl = regExImageDataUrl.test(uri);
+
+    if (cleanUri && width > 0 && !isDataImageUrl) {
       const regExResize = new RegExp("([?&])resize", "gi");
       const hasResizeParameter = regExResize.test(cleanUri);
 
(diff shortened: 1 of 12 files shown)


packages/video/__tests__/android/__snapshots__/video.android.test.js.snap
```diff
@@ -65,7 +65,7 @@ exports[`Video on native renders correctly 1`] = `
         onLoad={[Function]}
         source={
           Object {
-            "uri": "data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7&resize=1440",
+            "uri": "data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7?resize=1440",
```
tuncaulubilge: Is this a desired result?

packages/image/src/image.js
```diff
@@ -36,11 +36,18 @@ class TimesImage extends Component {
     };
 
     if (cleanUri && width > 0) {
-      const resizedUri =
-        cleanUri.indexOf("&resize") === -1
-          ? `${cleanUri}&resize=${width}`
-          : cleanUri;
-      props.source = { uri: resizedUri };
+      const regExResize = new RegExp("([?&])resize", "gi");
```
craigbilner: it's a crying shame we support IE

packages/video/__tests__/android/__snapshots__/video.android.test.js.snap
colincclark: I think this might be a separate bug @craigbilner should the base64 urls have resize query strings?

PR #986, opened 2018-06-01
Feat/replat 2658 update article query

jimhunty:
This PR covers:

1. Updating the Article Query to match the work done here: https://github.com/newsuk/times-public-api/pull/434
2. Add the ability to click on an `ArticleTopic` and be taken to the correct topic cross platform.
3. As a result of 2, for the `onPress` prop as required.


packages/article-topics/__tests__/shared.js
```diff
@@ -59,9 +65,16 @@ module.exports = () => {
...
     };
 
-    shallow(<ArticleTopic id={topicData[0].id} name={topicData[0].name} />, {
-      context
-    }).simulate("press", "events");
+    shallow(
+      <ArticleTopic
+        name={topicData[0].name}
+        slug={topicData[0].slug}
+        onPress={() => {}}
```
craigbilner: `onPress` as a noop for an `onPress` test?

PR #999, opened 2018-06-07
chore: set or fix node version to less than 10

colincclark:
- fix node version to less than 10, as t-c does not work on 10 yet.

Note: this will only throw a warning not an error if a node version greater than 10 is used

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "description": "A collection of presentational components for The Times and Sunday Times",
   "main": "index.js",
   "engines": {
-    "node": ">=8.9.0",
+    "node": "<10",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "description": "A collection of presentational components for The Times and Sunday Times",
   "main": "index.js",
   "engines": {
-    "node": ">=8.9.0",
+    "node": "<10",
```
craigbilner: perhaps want a greater than too?

aronshamash: greater than 8.9.0

colincclark: done

PR #1003, opened 2018-06-08
fix: Add bundle step to button package

aronshamash:
Package created before cli was updated to include bundling

### DIFF
--- a/packages/button/package.json
+++ b/packages/button/package.json
@@ -7,14 +7,15 @@
   "scripts": {
     "fmt": "prettier --write '**/*.*'",
     "prettier:diff": "prettier --list-different '**/*.*'",
-    "lint": "eslint . && npm run prettier:diff",
+    "lint": "eslint . && yarn prettier:diff",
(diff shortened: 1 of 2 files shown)


packages/button/package.json
```diff
@@ -7,14 +7,15 @@
...
+    "prepublishOnly": "yarn transpile && yarn bundle",
     "test:android": "jest --config='./__tests__/android/jest.config.js'",
     "test:ios": "jest --config='./__tests__/ios/jest.config.js'",
     "test:web": "jest --config='./__tests__/web/jest.config.js'",
     "test:all": "yarn test:android --coverage & yarn test:ios --coverage & yarn test:web --coverage",
-    "prepublishOnly": "npm run transpile",
     "cleanup-dist": "rm -rf dist",
-    "transpile": "yarn cleanup-dist && babel src -d dist"
+    "transpile": "yarn cleanup-dist && babel src -d dist",
+    "bundle": "NODE_ENV=production webpack -p"
```
colincclark: this should be added to the cli if it is required in every package

PR #1005, opened 2018-06-11
refactor: code coverage for button package to 100%

colincclark:
- Button onPress event is required


craigbilner:
@colincclark is this not just working around the actual problem that the default value is not being tested for? If rolled out it would also add a lot more null checks?

PR #1009, opened 2018-06-12
REPLAT-2285/Use ART for SVGs

fourlastor:
This PR introduces an ART implementation for SVGs, also switches to pre-rendered images for the watermark svg on mobile devices.

This is due to a missing feature in ART (the `fillType` property can't be set to `evenodd`)

Web falls back to `svgs`.

There are some packages that aren't still using `@times-components/svgs`, the plan is to move them once this is stable: `brightcove-video`, `jest-serializer`, `pagination`, `video`, `watermark`.


packages/svgs/src/g.js
```diff
@@ -0,0 +1,50 @@
+import React from "react";
+import { ART } from "react-native";
+import PropTypes from "prop-types";
+
+const { Group } = ART;
+
+const G = ({ fill, stroke, strokeWidth, opacity, children }) => {
+  const childrenWithProps = React.Children.map(children, child => {
```
craigbilner: could split this out to make more readable

packages/svgs/src/path.js
```diff
@@ -0,0 +1,38 @@
...
+  d: PropTypes.oneOfType([PropTypes.string, PropTypes.instanceOf(ARTPath)])
+    .isRequired,
+  fill: PropTypes.string,
+  stroke: PropTypes.string,
+  strokeWidth: PropTypes.string,
+  opacity: PropTypes.string
+};
+
+Path.defaultProps = {
+  fill: undefined,
```
craigbilner: I don't think humans should use `undefined`

packages/svgs/src/svg.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+
+import * as React from "react";
+import { ART } from "react-native";
+import PropTypes from "prop-types";
+
+const { Surface, Group, Transform } = ART;
+
+const Svg = ({ width, viewBox, children }) => {
+  const [shapeX, shapeY, shapeWidth, shapeHeight] = viewBox
```
craigbilner: perhaps need some more robust error/edge case handling here?

fourlastor: would a regex work? also should we do error handling here or in proptypes?

packages/svgs/src/path.js
fourlastor: 66882ea

packages/svgs/src/svg.js
fourlastor: 1b8fcb6

packages/svgs/src/g.js
fourlastor: c208d7d

PR #1010, opened 2018-06-12
fix: topic design review

colincclark:
- add section colour handling for `ArticleLabel` in `ArticleList`
- split out styles in `AuthorProfile` to unrequire `Platform` and added platform specific padding ammends around `AuthorProfileHeader` components
- fixed `Button` text layout in web
- fixed padding in `Topic` header for native

Note: this is not a comprehensive fix. Other PRs to follow


packages/topic/__tests__/android/__snapshots__/topic-head-styling.test.js.snap
```diff
@@ -10,9 +10,8 @@ exports[`TopicHead styling tests on android should render styling correctly 1`]
       "borderBottomWidth": 0.5,
       "flexDirection": "column",
       "minHeight": 90,
-      "paddingBottom": 20,
       "paddingHorizontal": 10,
-      "paddingTop": 35,
+      "paddingVertical": 35,
```
tuncaulubilge: Requirement was to increase the top padding, but looks like we increased the bottom padding

colincclark: The layout now matches the design

tuncaulubilge: I'll take a look in storybook and the app and let you know :+1:

packages/article-list/src/article-list-item.js
```diff
@@ -22,6 +22,7 @@ const ArticleListItem = props => {
     onPress,
     publicationName,
     publishedTime,
+    section,
```
craigbilner: should article list "know" about `section` rather than some sort of colour "theme"?

colincclark: Think section is OK. Adding theming probably out of scope for these design fixes?

packages/topic/__tests__/android/__snapshots__/topic-head-styling.test.js.snap
tuncaulubilge: Had a quick chat with Jake, let's go with 50 top padding, and 20 bottom. (apparently, we got some updated designs in zeplin for native)

packages/article-list/src/article-list-item.js
aronshamash: I think article should not know about `section` too. If we merge like this then raise a github issue.

packages/topic/__tests__/android/__snapshots__/topic-head-styling.test.js.snap
colincclark: done

packages/article-list/src/article-list-item.js
colincclark: There is no requirement to theme pages at the moment, with anything other than `section`. Therefore the concepts of themes are totally out of scope of this PR, and do not relate to these design changes. Theming is not even a requirement so should not even be a GH issue IMHO. Happy to discuss external to the PR

PR #1013, opened 2018-06-12
fix: lazy loading on safari

colincclark:
- fix lazy loading of `ArticleList` on safari. Safari does not use `IntersectionObserver`. The registration of images was happening _after_ the images were being fetched from the Component Class `images.map()`. Set the image width directly rather than going through the map as it was unnecessary

### DIFF
--- a/packages/article-list/__tests__/web/__snapshots__/article-list.web.test.js.snap
+++ b/packages/article-list/__tests__/web/__snapshots__/article-list.web.test.js.snap
@@ -66,7 +66,7 @@ exports[`ArticleList tests on web should render an article list 1`] = `
                     >
                       <img
                         alt=""
-                        src="//www.thetimes.co.uk/imageserver/image/%2Fmethode%2Ftimes%2Fprod%2Fweb%2Fbin%2F1b5afe88-cb0d-11e7-9ee9-e45ae7e1cdd4.jpg?crop=4252%2C2835%2C0%2C0&resize=100"
+                        src="//www.thetimes.co.uk/imageserver/image/%2Fmethode%2Ftimes%2Fprod%2Fweb%2Fbin%2F1b5afe88-cb0d-11e7-9ee9-e45ae7e1cdd4.jpg?crop=4252%2C2835%2C0%2C0&resize=1440"
                         style={
                           Object {
                             "display": "block",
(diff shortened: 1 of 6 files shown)


packages/article-list/__tests__/web/__snapshots__/helper.web.test.js.snap
```diff
@@ -82,7 +82,7 @@ exports[`Lazy loading and pagination tests on web Lazy loading tests should rend
 Array [
   <TimesImage
     aspectRatio={1.5}
-    uri="//www.thetimes.co.uk/imageserver/image/%2Fmethode%2Ftimes%2Fprod%2Fweb%2Fbin%2F1b5afe88-cb0d-11e7-9ee9-e45ae7e1cdd4.jpg?crop=4252%2C2835%2C0%2C0&resize=660"
+    uri="//www.thetimes.co.uk/imageserver/image/%2Fmethode%2Ftimes%2Fprod%2Fweb%2Fbin%2F1b5afe88-cb0d-11e7-9ee9-e45ae7e1cdd4.jpg?crop=4252%2C2835%2C0%2C0&resize=1440"
```
craigbilner: ideally you need to fix this test which confirms it correctly uses the `clientWidth`, perhaps an equally unfortunate global mock

packages/article-list/src/article-list.web.js
```diff
@@ -90,15 +90,11 @@ class ArticleList extends Component {
...
   registerNode(node) {
-    if (!node) {
+    if (!node || !this.observer) {
       return;
     }
 
-    if (this.observer) {
-      this.observer.observe(node);
-    } else {
-      this.images.set(node.id, normaliseWidth(node.clientWidth));
```
craigbilner: you can now remove `images` from the instance

packages/article-list/__tests__/web/__snapshots__/article-list.web.test.js.snap
```diff
@@ -66,7 +66,7 @@ exports[`ArticleList tests on web should render an article list 1`] = `
                     >
                       <img
                         alt=""
-                        src="//www.thetimes.co.uk/imageserver/image/%2Fmethode%2Ftimes%2Fprod%2Fweb%2Fbin%2F1b5afe88-cb0d-11e7-9ee9-e45ae7e1cdd4.jpg?crop=4252%2C2835%2C0%2C0&resize=100"
+                        src="//www.thetimes.co.uk/imageserver/image/%2Fmethode%2Ftimes%2Fprod%2Fweb%2Fbin%2F1b5afe88-cb0d-11e7-9ee9-e45ae7e1cdd4.jpg?crop=4252%2C2835%2C0%2C0&resize=1440"
```
craigbilner: this implies that rendering with an observer no longer works....although it does in storybook?

packages/article-list/src/article-list.web.js
colincclark: done

packages/article-list/__tests__/web/__snapshots__/helper.web.test.js.snap
colincclark: fixed

packages/article-list/__tests__/web/helper.web.test.js
```diff
@@ -144,6 +144,9 @@ describe("Lazy loading and pagination tests on web", () => {
     });
 
     it("should render good quality images if there is no IntersectionObserver", () => {
+      Object.defineProperty(window, "clientWidth", {
```
craigbilner: ideally this would be torn down too

PR #1019, opened 2018-06-14
test: move ad package tests to jest-configurator

craigbilner:


packages/ad/__tests__/android/android.test.js
```diff
@@ -0,0 +1,5 @@
+import shared from "../share.native";
+
+describe("On android", () => {
```
colincclark: "Ad tests on android" is the convention

PR #1021, opened 2018-06-14
Fix: Ad placeholder layout in IE

aronshamash:
## Before
screen shot 2018-06-14 at 12 38 55

## After
screen shot 2018-06-14 at 12 39 07


tuncaulubilge:
That watermark svg is quite a huge one and might perform poorly. We had discussed it with Jake and Daniele and decided to replace it with an image, but haven't done that yet. That might be a good task before we go live with topics & ads.

PR #1022, opened 2018-06-14
feat: article list ads

colincclark:
- add an ad to article list, every fifth article. Added storybook and test

screen shot 2018-06-14 at 13 34 14

### DIFF
--- a/packages/article-list/__tests__/shared-web.js
+++ b/packages/article-list/__tests__/shared-web.js
@@ -1,7 +1,7 @@
 import "react-native";
 import React from "react";
 import renderer from "react-test-renderer";
-import { mount, shallow } from "enzyme";
+import { mount } from "enzyme";
 import ArticleListPagination from "../src/article-list-pagination";
 import ArticleList from "./../src/article-list";
 import ArticleListItem from "./../src/article-list-item";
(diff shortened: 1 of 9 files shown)


packages/article-list/article-list.showcase.js
```diff
@@ -14,7 +14,56 @@ const preventDefaultedAction = decorateAction =>
...
+    section: "news",
+    zone: "current_edition",
+    slot: "news"
+  },
+  biddersConfig: {
+    timeout: 3000,
+    minPrice: 0.01,
+    maxBid: 15,
+    bucketSize: 0.25,
+    bidders: {
```
aronshamash: I thought there was no requirement for header bidding?

packages/article-list/__tests__/shared-web.js
```diff
@@ -40,4 +84,20 @@ export default () => {
...
+      <ArticleList
+        {...articleListProps}
+        adConfig={adConfig}
+        articles={results.articles.list}
+        page={1}
+        pageSize={pageSize}
+      />
+    );
+
+    expect(tree).toMatchSnapshot();
```
tuncaulubilge: Can we assert that the view is there, with a shallow renderer? I don't think we need a snapshot here

colincclark: I would argue that for now, a snapshot is best, to see the advert after the fifth article, until we get around to fixing up the tests in the article-list package in totality

packages/article-list/article-list.showcase.js
colincclark: I do not have a solid understanding of these properties. I figured they would be sorted with the next sub-task

aronshamash: yes we can remove in the next PR if needed.

PR #1025, opened 2018-06-14
fix: topic and author profile stories

colincclark:
- fixes ArticleList ads by passing the adConfig down through the relevant consumer packages, and limiting ads to only showing when the loading of articles is completed

### DIFF
--- a/packages/article-list/src/article-list.web.js
+++ b/packages/article-list/src/article-list.web.js
@@ -199,7 +199,9 @@ class ArticleList extends Component {
                       }
                     </ErrorView>
                   </div>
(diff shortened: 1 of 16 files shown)


packages/article-list/src/article-list.web.js
```diff
@@ -199,7 +199,9 @@ class ArticleList extends Component {
                       }
                     </ErrorView>
                   </div>
-                  {index === 4 && Object.keys(adConfig).length !== 0 ? (
+                  {!articlesLoading &&
```
aronshamash: Could put in a  function `shouldShowAd()` ?

PR #1028, opened 2018-06-15
fix: ads in article list for pagination and loading states

colincclark:
This fixes the lack of ads on second page and loading state ArticleLists. The ad slots need to be unique, and refreshed for it to work.

1. ArticleList is created in a loading state
2. Articles and Header load, and a new ad is shown
3. Clicking "next page" also renders a new ad


packages/article-list/src/article-list.web.js
```diff
@@ -172,19 +175,28 @@ class ArticleList extends Component {
...
                 index > 0 ? <ArticleListItemSeparator /> : null;
 
-              const renderArticle = () => {
-                if (
-                  articlesLoading ||
-                  index !== 4 ||
-                  Object.keys(adConfig).length === 0
-                ) {
+              const renderAd = () => {
+                if (index !== 4 || Object.keys(adConfig).length === 0) {
```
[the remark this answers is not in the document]
colincclark: done

packages/ad/src/dom-context.web.js
```diff
@@ -25,6 +26,10 @@ export default class DOMContext extends React.PureComponent {
     this.processEventQueue();
   }
 
+  componentWillUnmount() {
+    this.processEvent = () => false;
```
craigbilner: perhaps look at an alternative to overwriting the class method entirely. Flush the queue and guard on unmount?

packages/article-list/src/article-list.web.js
```diff
@@ -172,19 +176,28 @@ class ArticleList extends Component {
...
+                  if (articlesLoading) {
+                    return "inline-articles-loading";
+                  }
+                  return "inline-ad";
+                };
+
+                const AdComponent = (
                   <AdComposer adConfig={adConfig}>
-                    <Ad pos="inline-ad" />
+                    <Ad counter={this.counter} pos={renderPosTag()} />
```
craigbilner: could we change this to `slotPrefix` or something similar? As an API Ad shouldn't particularly care about it being a `counter` per se?

craigbilner: `pos` could do with a more appropriate name

packages/article-list/src/article-list.web.js
```diff
@@ -172,19 +176,28 @@ class ArticleList extends Component {
...
                 index > 0 ? <ArticleListItemSeparator /> : null;
 
-              const renderArticle = () => {
-                if (
-                  articlesLoading ||
-                  index !== 4 ||
-                  Object.keys(adConfig).length === 0
-                ) {
+              const renderAd = () => {
+                if (index !== this.advertPosition || Object.keys(adConfig).length === 0) {
```
craigbilner: minor: potentially don't need to check the adConfig every render, just when it changes?

packages/article-list/src/article-list.web.js
aronshamash: It looks like that is a name the ad guys have used for targeting so the prop name is just following that.

colincclark: done

colincclark: `slot.setTargeting("pos", containerID);` - but I can still change the JS variables

colincclark: done

packages/article-list/src/article-list.web.js
colincclark: moved outside of the loop

PR #1029, opened 2018-06-15
compose serializers

craigbilner:
This is the start of being able to curate snapshots in conjunction with `jest-lint`. `jest-serializer` has now been split into separate functionality and applied to each package at a minimum in order to preserve the current snapshots.

The next steps would be to iterate on `jest-lint` and the various `serializers` in order to make "good" snapshots as easy as possible. The first obvious feature being, blitzing external components to that involve integration headaches

The PR is rather large because it adds serializers to every package but you can focus on any changes to existing snapshots which is the important part. For example `ad` now has an example where `rnw` styles are printed in the snapshot where the current snapshots are fairly meaningless

For the more interested `jest-serializer` has the actual functionality change where we have a traverser, printers and transformer to allow us to curate the snapshots on a case by case basis

breaking change: add options object to jest-configurator
feat: hoist dev specific rnw styles in snapshots a la styled components
feat: remove empty values from snapshots e.g. stickyHeaderIndices
feat: web retains className even if empty and native does not
chore: remove coverage flags from test scripts
chore: bump enzyme-to-json


packages/article-label/package.json
```diff
@@ -53,7 +54,7 @@
     "babel-plugin-styled-components": "1.5.1",
     "babel-preset-react-native": "4.0.0",
     "enzyme": "3.3.0",
-    "enzyme-to-json": "3.3.0",
+    "enzyme-to-json": "3.3.4",
```
nikhedonia: to we use `enzyme-to-json` directly?

packages/article-list/__tests__/android/serializers.js
```diff
@@ -0,0 +1,12 @@
+const {
+  compose,
+  minimalNativeTransform,
+  flattenStyleTransform,
+  print
+} = require("@times-components/jest-serializer");
+const { createSerializer } = require("enzyme-to-json");
+
+expect.addSnapshotSerializer(createSerializer({ mode: "deep" }));
```
nikhedonia: I expected this to be inside compose. Maybe it makes sense to export this serializer from the `jest-serialize` package?

nikhedonia:
Jest-serializer would benefit from a readme that describes the various serializes and how they compose.

packages/ad/__tests__/web/__snapshots__/placeholder.test.js.snap
```diff
@@ -1,8 +1,19 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Placeholder renders a large placeholder 1`] = `
+<style>
+{
+  "S1": {
```
nikhedonia: would be prettier if it actually would look like css.

packages/card/__tests__/web/__snapshots__/card.web.test.js.snap
```diff
@@ -14,6 +14,7 @@ exports[`1. Renders default layout 1`] = `
     }
   >
     <View
+      className=""
```
nikhedonia: maybe get rid of those empty classNames

packages/jest-serializer/__tests__/web/__snapshots__/flatten-style.test.js.snap
```diff
@@ -0,0 +1,16 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`The flatten style serializer should for web flatten styles 1`] = `
```
nikhedonia: maybe "The flatten style serializer for web should flatten styles 1" ?

packages/ad/__tests__/web/__snapshots__/placeholder.test.js.snap
craigbilner: Have raised #1031

packages/card/__tests__/web/__snapshots__/card.web.test.js.snap
craigbilner: will remove in the future, this "proves" that the snapshots aren't correct

packages/jest-serializer/__tests__/web/__snapshots__/flatten-style.test.js.snap
craigbilner: will pick up

PR #1037, opened 2018-06-19
bc: Update provider parameter types

jimhunty:
As a result of https://github.com/newsuk/times-public-api/pull/472, we now have new types for the parameters for the list queries in the providers.

Once TPA goes live, then the providers will error. This will fi that.


craigbilner:
pending TPA changes

PR #1042, opened 2018-06-20
fix: article list ads

colincclark:
This makes ads more stable in general, especailly when they are mounted twice+

DEPENDENT ON https://github.com/newsuk/times-components/pull/1041


craigbilner:
pending ads working

PR #1043, opened 2018-06-20
test: make article flag pass linting

craigbilner:
This PR has three elements to it:

### A proposal for snapshot testing

Using the `Article Flag` package as an example, the idea is to get the snapshots to pass `jest-lint` and to also cover the type of things we would like to test.

The tests have been split into the article flag "flavours" (do they pass the correct props?) and then separate detailed snapshots to verify the styling. None of the snapshots care about the implementation of `IconDiamond` which is a dependency with it's own tests.

You may also note the mixture of `react-test-renderer` (RTR) and `enzyme`. Currently `enzyme` is much better at creating snapshots for web because that is what it's designed for and an RN adapter is outstanding. RTR is therefore much better at RN but doesn't give the nice "host" abstractions that `enzyme` provides for web. Therefore if you want to render the "whole" tree and be able to replace "host" elements, you'll need to use a mixture of the two across native and web.

### New features in `jest-serializer`

To get `jest-lint` to pass and to make it easier for the developer to make "good snapshots" two new features have been added to `jest-serializer`, `replace` and `hoist-style`. `replace` means no more `jest.mock` which was misused to create better snapshots but is meant for the tests themselves and can have unintended side effects/consequences. `hoist-style` obviates the need to satisfy `jest-lint`s `maxAttr` error by allow the style block to have as many attributes as it wants yet keeping the JSX as clean as possible. If styles change the diff should be obvious and just in the `style` block!

Further props were removed under the `minimal` transforms which is an iterative process to find the right balance between testing and bare bones snapshots.

### Snapshot updates

You can safely ignore these changes if the removed props seem sensible, from a test perspective they just provide implementation noise we don't care about or need to worry about at a PR level.

### Summary

feat: add replace serializer
feat: add hoist style
feat: remove props (`accessible`, `allowFontScaling`, `ellipsizeMode`, `dir`)
refactor: `jest-serializer` code to make more manageable/extensible
chore: rename `rnwPrinter` to `stylePrinter`
chore: remove redundant Enzyme code


packages/article-flag/__tests__/web/article-flag-content.web.test.js
```diff
@@ -0,0 +1,48 @@
...
+  enzymeTreeSerializer,
+  hoistStyleTransform,
+  justChildren,
+  meltNative,
+  minimalWebTransform,
+  propsNoChildren,
+  replaceTransform,
+  stylePrinter,
+  rnwTransform
+} from "@times-components/jest-serializer";
```
colincclark: will these be required for all web tests?

packages/article-flag/__tests__/android/article-flag-content.android.test.js
```diff
@@ -0,0 +1,5 @@
+import shared from "../article-flag-content.native";
+
+describe("Article Flag on android", () => {
```
colincclark: Can we differentiate between the content tests, and the non-content tests in the test runner, and in the snapshots?

packages/article-flag/__tests__/shared.js
```diff
@@ -1,37 +1,43 @@
...
-    expect(tree).toMatchSnapshot();
-  });
+      expect(wrapper).toMatchSnapshot("2. New flag");
+    });
 
-  it("renders Exclusive flag correctly", () => {
-    const tree = renderer.create(<ExclusiveArticleFlag />).toJSON();
-    expect(tree).toMatchSnapshot();
-  });
+    it("Updated flag correctly", () => {
```
colincclark: can we ditch the "correctly"

packages/article-flag/__tests__/web/article-flag-content.web.test.js
craigbilner: the current approach is to create a "raw snapshot" then tailor it with our "serialization toolbelt". If patterns crop up such as this one we could export them much like a babel preset for example. I have an example of this called `minimalRnw`.

PR #1045, opened 2018-06-21
Fix: Defensive check for window object

aronshamash:
- Fixes `article-list` on the server.

### DIFF
--- a/packages/article-list/src/article-list.web.js
+++ b/packages/article-list/src/article-list.web.js
@@ -54,6 +54,10 @@ class ArticleList extends Component {
   }
 
   getImageSize(nodeId) {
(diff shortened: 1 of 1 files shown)


packages/article-list/src/article-list.web.js
```diff
@@ -54,6 +54,10 @@ class ArticleList extends Component {
   }
 
   getImageSize(nodeId) {
+    if (typeof window === "undefined") {
```
colincclark: don't think this method should return 0 as a default. Maybe 100?

aronshamash: ok `100` seems ok.

PR #1046, opened 2018-06-21
fix: article list ad re-render issues

colincclark:
- ensure ad DOM component does not try to execute code after it has unmounted
- do not re-render the article list if the page prop is changed - we should only rerender when the loading states or data changes
- destroy ad slot when the ad component is unmounted
- moves the instantiation of the ad component out of a loop

Note: the ad will not appear when the article list is loading. This will be fixed in a subsequent PR when the ad placeholder is available to export.

### DIFF
--- a/packages/ad/src/ad-init.js
+++ b/packages/ad/src/ad-init.js
@@ -4,7 +4,7 @@
 // defined outside the function, so that it can be passed into a WebView.
 
 const adInit = args => {
-  const { slotSuffix, el, data, platform, eventCallback, window } = args;
+  const { el, data, platform, eventCallback, window } = args;
   const { document, setTimeout, Promise } = window;
 
   const scriptsInserted = {};
@@ -100,19 +100,12 @@ const adInit = args => {
         this.scheduleAction(() => {
           const adUnitPath = `/${networkId}/${adUnit}/${section}`;
           const { slotName, sizes, mappings } = slotConfig;
-          const containerID = slotSuffix
-            ? `${slotName}-${slotSuffix}`
-            : slotName;
 
-          const slot = window.googletag.defineSlot(
-            adUnitPath,
-            sizes,
-            containerID
-          );
+          const slot = window.googletag.defineSlot(adUnitPath, sizes, slotName);
 
           if (!slot) {
             throw new Error(
-              `Ad slot ${containerID} ${
+              `Ad slot ${slotName} ${
                 adUnitPath
               } could not be defined, probably it was already defined`
             );
@@ -120,7 +113,7 @@ const adInit = args => {
           slot.addService(window.googletag.pubads());
           /* eslint-disable no-param-reassign */
           el.id = `wrapper-${slotName}`;
-          el.innerHTML = `<div id="${containerID}"></div>`;
+          el.innerHTML = `<div id="${slotName}"></div>`;
           el.style.display = "flex";
           el.style.alignItems = "center";
           el.style.justifyContent = "center";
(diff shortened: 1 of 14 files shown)


packages/article-list/__tests__/web/__snapshots__/article-list.web.test.js.snap
```diff
@@ -1,5 +1,9 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`ArticleList tests on web should not rerender if the page prop changes 1`] = `1`;
```
craigbilner: not sure these are in keeping with our snapshot drive?

packages/article-list/src/article-list.web.js
```diff
@@ -156,6 +160,15 @@ class ArticleList extends Component {
       </ListContentContainer>
     );
 
+    const AdComponent = (
+      <AdComposer adConfig={adConfig}>
```
craigbilner: should this wrap around the whole list?

packages/article-list/src/article-list.web.js
```diff
@@ -183,29 +196,18 @@ class ArticleList extends Component {
...
-                    <AdComposer adConfig={adConfig}>
-                      <Ad
-                        slotSuffix={this.advertPositionCounter.toString()}
-                        slotName={renderSlotName()}
-                      />
-                    </AdComposer>
-                  );
-                  this.advertPositionCounter = this.advertPositionCounter + 1;
+                  if (this.pending.size === 0) {
+                    this.advertPositionCounter = this.advertPositionCounter + 1;
```
craigbilner: would it be neater to move into state and use `getDerivedStateFromProps`?

packages/article-list/src/article-list.web.js
colincclark: not sure why we would want to do that? it made more sense to keep the ad stuff together to me

packages/article-list/src/article-list.web.js
colincclark: the class property is succinct and does not need the added lifecycle event management so is neater

packages/article-list/__tests__/web/__snapshots__/article-list.web.test.js.snap
colincclark: this package defintiely needs improved snapshots, but you can't get more succinct than one line of code? what would you image this test to look like? ditch the snapshot altogether?

colincclark: done

packages/article-list/src/article-list.web.js
colincclark: will make this a function, but in a later PR

packages/article-list/src/article-list.web.js
colincclark: will be removed, as it was causing ads to fail intermittently

packages/ad/src/ad-init.js
```diff
@@ -370,6 +367,10 @@ const adInit = args => {
       }
     },
 
+    destroySlots() {
+      this.gpt.destroySlots();
```
aronshamash: Can you explain why we need to do `window.googletag.destroySlots();` and `this.gpt.destroySlots();`

colincclark: The `gpt` object is the only one that interacts with gpt itself. The "public" method above is for use by the consumer, and the "private" `gpt` object is for internal use by the init code. It will probably be refactored and renamed shortly, but this follows the pattern of other methods in the meantime

PR #1047, opened 2018-06-21
test: article byline

craigbilner:
The next stop on the snapshot journey, `Article Byline`

Not only should snapshots be as small as possible and specific to the test they should also limit the surface area for change.

This PR splits out testing styles from testing structure of the component. In addition it puts back in missing styles on the web test which were hidden in the React Native Web classes. This should mean when a style changes only the 3 expected snapshots change or example.

`href` has also been removed from native because it provides no value which should be covered by interactive tests

### tl;dr

* refactor: article byline snaps to only show styles when needed
* chore: remove href from native


packages/article-byline/__tests__/android/__snapshots__/article-byline-with-links.android.test.js.snap
```diff
@@ -1,206 +1,61 @@
...
-    }
   >
     Mike Atherton
   </Text>
 </View>
 `;
 
+exports[`4. Render null when AST is empty 1`] = `null`;
+
 exports[`6. Render multiple authors separated by commas 1`] = `
```
colincclark: numbers out of whack

packages/article-byline/__tests__/shared-with-styles.web.js
```diff
@@ -0,0 +1,29 @@
+import {
```
colincclark: should this just be in the `web` directory?

PR #1054, opened 2018-06-26
chore: Remove accessibilityLabel from date publication

jimhunty:
When VoiceOver reads the article summary as a link, it reads "Date Publication" rather than the actual date and publication.

### DIFF
--- a/packages/article-summary/__tests__/ios/jest.config.js
+++ b/packages/article-summary/__tests__/ios/jest.config.js
@@ -1,6 +1,6 @@
 const jestConfigurator = require("@times-components/jest-configurator").default;
 const path = require("path");
 
-module.exports = jestConfigurator("android", __dirname, {
+module.exports = jestConfigurator("ios", __dirname, {
   setupTestFrameworkScriptFile: path.join(__dirname, "./serializers")
 });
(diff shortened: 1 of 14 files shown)


packages/article-summary/__tests__/ios/jest.config.js
```diff
@@ -1,6 +1,6 @@
 const jestConfigurator = require("@times-components/jest-configurator").default;
 const path = require("path");
 
-module.exports = jestConfigurator("android", __dirname, {
+module.exports = jestConfigurator("ios", __dirname, {
```
colincclark: 👍

packages/article-summary/src/article-summary.js
```diff
@@ -71,11 +71,7 @@ const ArticleSummary = props => {
       {headline()}
       {content()}
       {datePublicationProps ? (
-        <Text
-          accessibilityLabel="datePublication"
-          style={styles.metaText}
-          testID="datePublication"
-        >
+        <Text style={styles.metaText} testID="datePublication">
```
colincclark: could the aria-label be set to the date rather than removed altogether?

jimhunty: It's not necessary as the screen reader will pick up the content. It's already accessible.

PR #1061, opened 2018-06-27
Show image loading state instead of hiding it

tuncaulubilge:
Images are flickering on native apps while re-rendering because we are removing the view in loading state, instead of showing the placeholder. Removed the null checks for uri's as the image component itself has a placeholder state

### DIFF
--- a/packages/author-profile/src/author-profile-head.js
+++ b/packages/author-profile/src/author-profile-head.js
@@ -24,7 +24,7 @@ const AuthorProfileHead = ({
     );
   };
 
-  const renderImage = () => <AuthorProfileHeadImage uri={uri} />;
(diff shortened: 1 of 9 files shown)


packages/author-profile/src/author-profile-head.js
```diff
@@ -24,10 +24,7 @@ const AuthorProfileHead = ({
     );
   };
 
-  const renderImage = () => {
-    if (!uri) return null;
-    return <AuthorProfileHeadImage uri={uri} />;
-  };
+  const renderImage = () => <AuthorProfileHeadImage uri={uri} />;
```
craigbilner: remove func?

tuncaulubilge: Done :+1:

PR #1063, opened 2018-06-28
chore: Update types to match TPA

jimhunty:
The types have been approved in TPA: https://github.com/newsuk/times-public-api/pull/472

This updates these types in times-components. 

**These two PRs should ship together otherwise there will be errors.**


craigbilner:
potentially need to version TPA

PR #1064, opened 2018-06-28
Fix/article list rerender

tuncaulubilge:
We were rendering separate ArticleLists in loading states and loaded states, which was causing a complete re-render of the page. This is causing glitchy ux in the mobile apps.

- Now we are using the same article list with a loading state.
- Added a isLoading prop to MockedProvider that doesn't return anything by using an empty ApolloLink

### DIFF
--- a/packages/author-profile/__tests__/ios/__snapshots__/author-profile.test.js.snap
+++ b/packages/author-profile/__tests__/ios/__snapshots__/author-profile.test.js.snap
@@ -4353,29 +4353,7 @@ exports[`3. Render an author profile page loading state 1`] = `
         </View>
       </View>
     </View>
-    <View>
-      <View>
-        <View
-          style={
-            Object {
-              "backgroundColor": "#DBDBDB",
-              "height": 1,
-            }
-          }
-        />
-        <ActivityIndicator
-          animating={true}
-          color="#DBDBDB"
-          hidesWhenStopped={true}
-          size="large"
-          style={
-            Object {
-              "paddingVertical": 25,
-            }
-          }
-        />
-      </View>
-    </View>
+    <View />
   </View>
 </RCTScrollView>
 `;
--- a/packages/topic/src/topic.js
+++ b/packages/topic/src/topic.js
@@ -30,10 +30,12 @@ const Topic = ({
     return <ArticleListPageError refetch={refetch} />;
   }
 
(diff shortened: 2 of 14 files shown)


aronshamash:
Verified by using this version of components in render and SSR is working as before. @tuncaulubilge

packages/author-profile/__tests__/ios/__snapshots__/author-profile.test.js.snap
```diff
@@ -4353,29 +4353,7 @@ exports[`3. Render an author profile page loading state 1`] = `
...
-          size="large"
-          style={
-            Object {
-              "paddingVertical": 25,
-            }
-          }
-        />
-      </View>
-    </View>
+    <View />
```
craigbilner: now you've "fixed" loading state, there are only 2 activity indicators in the snapshots that are for "renders a page", is this correct?

packages/utils/src/graphql.js
```diff
@@ -29,8 +31,12 @@ class MockedProvider extends Component {
   constructor(props, context) {
     super(props, context);
 
+    const link = this.props.isLoading
```
craigbilner: As discussed, looks cool, perhaps need to take our "fixes" here and put them back into Apollo

packages/topic/src/topic.js
```diff
@@ -30,27 +30,12 @@ const Topic = ({
...
-        isLoading
-        pageSize={initPageSize}
-        refetch={() => {}}
-        showImages
-      />
-    );
-  }
-
-  const { name, description } = topic;
+  const { name, description } = isLoading
```
colincclark: we map `isLoading` for article to `articlesLoading` (line 53 of this branch). can we map `isLoading` for the description/name to `headerLoading` or similar?

packages/author-profile/__tests__/ios/__snapshots__/author-profile.test.js.snap
tuncaulubilge: That last activity indicator was the one for infinite scroll, which wasn't supposed to be shown on loading state

packages/topic/src/topic.js
tuncaulubilge: Good idea, fixed in 04995be

PR #1075, opened 2018-07-02
test: refactor card snapshots

craigbilner:
* split out styles on web and render full stylesheet
* trim native down further


packages/card/__tests__/shared.base.js
```diff
@@ -0,0 +1,212 @@
...
+  image: {
+    uri: "https://img.io/img"
+  },
+  imageRatio: 2 / 3,
+  imageSize: 360,
+  showImage: true
+};
+
+export default renderMethod => {
+  // magic to stop the React Native Animated library from dying as each test kicks of another animation that uses timing
```
colincclark: "kicks off"

PR #1076, opened 2018-07-02
fix: handle ad blocker

colincclark:
https://nidigitalsolutions.jira.com/browse/REPLAT-2861

Note: ad component testing is limited atm, I believe it should be done with the `jest-serializer` package on a separate PR

### DIFF
--- a/packages/ad/src/ad-init.js
+++ b/packages/ad/src/ad-init.js
@@ -9,6 +9,7 @@ const adInit = args => {
 
   const scriptsInserted = {};
   let initCalled = false;
+  let hasAdBlocker = false;
 
   const initialiser = {
     utils: {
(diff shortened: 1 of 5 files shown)


packages/ad/src/ad-init.js
```diff
@@ -24,6 +25,7 @@ const adInit = args => {
               resolve();
             },
             () => {
+              hasAdBlocker = true;
```
aronshamash: I would prefer a more generic name such as `isScriptError`

colincclark: done

packages/ad/src/ad.js
```diff
@@ -105,8 +113,8 @@ class Ad extends Component {
 
     return (
       <View style={[style]}>
-        {isLoading ? null : AdComponent}
-        {isLoading || !this.state.adReady ? AdPlaceholderComponent : null}
+        {isLoading || hasError ? null : AdComponent}
```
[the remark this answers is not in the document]
colincclark: done

PR #1079, opened 2018-07-03
refactor: gradient snapshots

craigbilner:
split and focus tests on parts of the gradient


packages/gradient/__tests__/shared.base.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { Text } from "react-native";
+import Gradient from "../src/gradient";
+
+module.exports = renderMethod => {
```
colincclark: `export default`

PR #1080, opened 2018-07-04
chore: Max number of Article Topics to 5

jimhunty:
On an article, the maximum number of Article Topics shown can be 5. This enforces this.


craigbilner:
@jimhunty should this not be a provider/TPA change? Also it would be ideal if the diff didn't change (by adding the additional topic at the end)?

jimhunty:
@craigbilner Putting the max number of topics on TPA restricts the API for other consumers?

craigbilner:
@jimhunty didn't mean hard code it!
@aronshamash yes, the consumer should be able to request the topics they would like, the client shouldn't have to handle the data per se

jimhunty:
Why was this closed @craigbilner? Can you give more information on what you mean by `take` ?

jimhunty:
Alternative approach will be taking by paging topics from TPA as discussed with @craigbilner

PR #1084, opened 2018-07-05
chore: fix up CLI generation code

colincclark:


lib/cli/templates.js
```diff
@@ -6,15 +6,16 @@ const indexFile = handlebars.compile(`import React from "react";
 import { Text } from "react-native";
 import styles from "./styles";
 
-export default function {{{component}}}() {
+function {{{component}}}() {
```
craigbilner: this is a little old fashioned :smirk:

PR #1085, opened 2018-07-05
test: refactor markup snaps

craigbilner:
Has highlighted a bug where replaced nodes still print their styles, will fix separately


packages/markup/__tests__/shared.native.js
```diff
@@ -0,0 +1,30 @@
...
+
+  shared(TestRenderer.create);
+
+  it("renders multiple paragraphs with ads", () => {
+    const testInstance = TestRenderer.create(
+      <View>{renderTrees(multiParagraphWithAds)}</View>
+    );
+
+    expect(testInstance).toMatchSnapshot(
+      "14. renders multiple paragraphs with ads"
```
colincclark: this might end up being a little annoying to maintain, in terms of numbering. Can it be split out so files only have fixed numbering?

packages/markup/__tests__/shared.native.js
```diff
@@ -0,0 +1,30 @@
...
+import { View } from "react-native";
+import {
+  addSerializers,
+  minimalNative
+} from "@times-components/jest-serializer";
+import TestRenderer from "react-test-renderer";
+import { renderTrees } from "../src/markup";
+import shared from "./shared.base";
+
+const multiParagraphWithAds = require("../fixtures/multi-paragraph-with-ads.json");
```
colincclark: could this just be `import`ed?

craigbilner: without helpers I think this is the most straightforward way to import `json`?

PR #1088, opened 2018-07-06
Feat: Article ad config

aronshamash:
- Fix article ads

### DIFF
--- a/packages/article/fixtures/article-ad-config.json
+++ b/packages/article/fixtures/article-ad-config.json
@@ -1,0 +1,59 @@
+{
+  "networkId": "25436805",
+  "adUnit": "d.thetimes.co.uk",
+  "pageTargeting": {
+    "testmode": "null",
+    "log": "1",
+    "ra": "0",
+    "teaser": false,
+    "pw": "1",
+    "shared": "0",
+    "share_token": "null",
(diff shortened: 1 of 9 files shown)


packages/article/fixtures/article-ad-config.json
```diff
@@ -0,0 +1,59 @@
+{
```
colincclark: do we need to maintain parity between this file and https://github.com/newsuk/times-components/blob/master/packages/ad/fixtures/article-ad-config.json?

packages/markup/src/markup.web.js
```diff
@@ -35,7 +35,7 @@ const defaultRenderers = {
   },
   ad(key, attributes) {
     return (
-      <Ad key={key} slotName="inline-ad" style={styles.ad} {...attributes} />
+      <Ad key={key} slotName="ad-inline" style={styles.ad} {...attributes} />
```
colincclark: What should the one in `ArticleList` be called? Because atm that is called "inline-ad"...

packages/article/fixtures/article-ad-config.json
aronshamash: yes but I don't think think we should have a shared file

PR #1093, opened 2018-07-09
refactor: enforce and adopt alphabetical ordering of props

colincclark:
- add `eslint . --fix` to the npm scripts
- re-order props across all the packages

This will automatically be fixed by running the usual `yarn fmt` command. devs will be able to more easily identify props on Components without having to focus on manually maintaining ordering.


aronshamash:
beautiful

PR #1096, opened 2018-07-10
test: improve snapshots of watermark

colincclark:
- added README
- improved snapshots
- added knobs to story for watermark sizes


packages/watermark/README.md
```diff
@@ -0,0 +1,39 @@
+# Watermark
+
+A Times standard image (native) or svg (web) that acts as a placeholder image
+wherever required. The Watermark component can act as a placeholder for failed
+images, or as a loading state. It requires height and width props for layout.
```
craigbilner: we need to automate this along the lines of TypeScript -> prop types -> doc gen -> README

packages/watermark/README.md
```diff
@@ -0,0 +1,39 @@
...
+The code can be formatted and linted in accordance with the agreed standards.
+
+```
+yarn fmt
+yarn lint
+```
+
+## Testing
+
+This package uses [yarn](https://yarnpkg.com) (latest) to run unit tests on each
```
craigbilner: think you can use the `npm` client too?

packages/watermark/__tests__/android/__snapshots__/watermark.android.test.js.snap
```diff
@@ -1,9 +1,8 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Watermark tests on android renders correctly 1`] = `
+exports[`1. watermark 1`] = `
```
craigbilner: would be good to test the correct src is used?

packages/watermark/__tests__/shared.js
```diff
@@ -1,14 +1,31 @@
...
-      .create(<Watermark height={250} width={300} />)
-      .toJSON();
+export default () => {
+  addSerializers(
+    expect,
+    enzymeTreeSerializer(),
+    compose(
+      print,
+      minimaliseTransform(
+        (value, key) => key === "d" || key === "id" || key === "source"
```
craigbilner: I would test `id` and `d` myself. Would be good to see that rather odd id of "Page-1" and if `d` had changed in anyway?

packages/watermark/README.md
colincclark: removed `yarn` link - would just be more maintenance re: yarn version, etc

packages/watermark/__tests__/shared.js
colincclark: done and amended `id` to something more sensible

packages/watermark/__tests__/android/__snapshots__/watermark.android.test.js.snap
colincclark: done - mocked the asset, but this is not ideal so also created https://github.com/newsuk/times-components/issues/1102

PR #1098, opened 2018-07-10
Fix apollo-client import for storybooks

tuncaulubilge:
Storybooks are broken due to an import issue with MockedProviders.

### DIFF
--- a/packages/provider-test-tools/src/mocked-provider.js
+++ b/packages/provider-test-tools/src/mocked-provider.js
@@ -1,5 +1,5 @@
 import React, { Component } from "react";
-import ApolloClient from "apollo-client";
+import { ApolloClient } from "apollo-client";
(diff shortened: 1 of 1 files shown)


packages/provider-test-tools/src/mocked-provider.js
```diff
@@ -1,5 +1,5 @@
 import React, { Component } from "react";
-import ApolloClient from "apollo-client";
+import { ApolloClient } from "apollo-client";
```
colincclark: this wasn't picked up in a test?

tuncaulubilge: It only fails in the storybook, which is weird. ApolloClient has both named and default exports, so both imports should be correct. There might be an issue with transpilation process on storybooks

PR #1110, opened 2018-07-13
refactor: test utils hash helper

colincclark:
- make iterator a non-default export from test utils
- add hash test helper function for long keys in snapshots
- rolled out and replaced hashing in icon, pagination and watermark packages


packages/test-utils/package.json
```diff
@@ -50,6 +49,9 @@
     "webpack": "4.6.0",
     "webpack-cli": "2.1.4"
   },
+  "dependencies": {
```
craigbilner: should `test-utils` just provide the hashing ability rather than tying it into serialization?

PR #1113, opened 2018-07-16
test: refactor slice snaps

craigbilner:
* Split out role snaps
* Changed snaps to show rendered elements
* Followed related article naming conventions
* Print web `styled-component` styles


packages/slice/package.json
```diff
@@ -8,6 +8,7 @@
     "prettier:diff": "prettier --list-different '**/*.*'",
     "lint": "eslint . && npm run prettier:diff",
     "test:android": "jest --config='./__tests__/android/jest.config.js'",
+    "test": "jest roles.test.js --config='./__tests__/jest.config.js'",
```
colincclark: bit odd to be called in the script explicitly and be ignored in the coverage glob too. why not just use the `roles.[platform].test.js` pattern?

craigbilner: I think for "pure js" code we should test with an agnostic platform, there's no value in repeating it 3 times either?

PR #1125, opened 2018-07-19
Feat: Android ad config

aronshamash:
- Populate ad config for Android.
- Corresponding Android Smartphone PR: https://github.com/newsuk/nu-projectd-times-smartphone-android/pull/420

### DIFF
--- a/android-app/pages/article.js
+++ b/android-app/pages/article.js
@@ -15,20 +15,20 @@ const {
 } = NativeModules.ArticleEvents;
 const ArticlePageView = Article(config)(fetch);
 
-const getPlatformAdConfig = config => ({
+const getPlatformAdConfig = adConfig => ({
   adUnit: "d.thetimes.co.uk",
   networkId: "25436805",
   testMode: "",
-  sectionName: config.sectionName,
-  appVersion: config.appVersion,
+  sectionName: adConfig.sectionName,
+  appVersion: adConfig.appVersion,
   operatingSystem: "Android",
-  operatingSystemVersion: config.operatingSystemVersion,
-  environment: config.environment,
-  deviceId: config.deviceId,
-  cookieEid: config.cookieEid,
-  cookieAcsTnl: config.cookieAcsTnl,
-  cookieIamTgt: config.cookieIamTgt,
-  isLoggedIn: config.isLoggedIn,
+  operatingSystemVersion: adConfig.operatingSystemVersion,
+  environment: adConfig.environment,
+  deviceId: adConfig.deviceId,
+  cookieEid: adConfig.cookieEid,
+  cookieAcsTnl: adConfig.cookieAcsTnl,
+  cookieIamTgt: adConfig.cookieIamTgt,
+  isLoggedIn: adConfig.isLoggedIn,
   platform: "mobile"
 });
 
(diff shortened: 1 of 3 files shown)


android-app/pages/article.js
```diff
@@ -2,6 +2,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import { NativeModules } from "react-native";
 import { Article } from "@times-components/pages";
+import DeviceInfo from "react-native-device-info";
```
tuncaulubilge: We should be getting rid of this dependency eventually. I'll check if I can pass the device id from the native app.

tuncaulubilge: Created an issue to remove it completely: https://github.com/newsuk/times-components/issues/1126

aronshamash: We can get from native app.

android-app/pages/article.js
```diff
@@ -15,41 +16,69 @@ const {
...
+  cookieIamTgt: config.cookieIamTgt,
+  isLoggedIn: config.isLoggedIn,
   platform: "mobile"
-};
+});
+
+const ArticleView = ({
+  articleId,
+  sectionName,
+  appVersion,
```
tuncaulubilge: We should get this, and all the props below from `config`. I'm moving these inside config in the native app

android-app/pages/article.js
```diff
@@ -15,41 +16,69 @@ const {
...
-  deviceId: "",
-  deviceIdHash: "",
-  environment: "",
-  isLoggedIn: true,
+  sectionName: config.sectionName,
+  appVersion: config.appVersion,
+  operatingSystem: "Android",
+  operatingSystemVersion: config.operatingSystemVersion,
+  environment: config.environment,
+  deviceId: DeviceInfo.getUniqueID(),
```
tuncaulubilge: Should be available through `config.deviceId` soon

android-app/pages/article.js
```diff
@@ -15,41 +15,48 @@ const {
 } = NativeModules.ArticleEvents;
 const ArticlePageView = Article(config)(fetch);
 
-const platformAdConfig = {
+const getPlatformAdConfig = config => ({
```
tuncaulubilge: This needs to be renamed, as it clashes with the upper level config (` const config = NativeModules.ReactConfig;`)

android-app/pages/article.js
```diff
@@ -15,41 +15,41 @@ const {
...
-  cookieAcsTnl: "",
-  cookieIamTgt: "",
-  deviceId: "",
-  deviceIdHash: "",
-  environment: "",
-  isLoggedIn: true,
-  platform: "mobile"
-};
+const ArticleView = ({ articleId, sectionName }) => {
+  const adConfig = {
```
craigbilner: is there value in binding this on every function call?

packages/pages/src/client/ad-targeting-config.js
```diff
@@ -1,15 +1,13 @@
 export default (platformAdConfig, articleAdConfig) => ({
-  networkId: platformAdConfig.adUnit,
-  adUnit: platformAdConfig.networkId,
+  networkId: platformAdConfig.networkId,
```
craigbilner: shouldn't this have  broken a test?

PR #1128, opened 2018-07-20
test: refactor article testing

craigbilner:
This PR removes most of the snapshot testing and also minimises the snaps that are left achieving :100: coverage. There were a couple of breaking changes:

1) `onPress` wasn't needed for web and is't provided by render for article links. The other `onPress` events should probably be removed?
2) `ActivityLoadingIndicator` from RNW doesn't work with snapshots so has been replaced. This needs a full replacement with a skeleton anyway and will only be visible on deeplinked native articles.

It's probably best reviewed "as is" in your IDE rather than as a diff on GitHub due to the radical changes.

* file sizes are still quite large but there are a lot of styles
* split tracking, styles and structural tests
* related article tracking is still outstanding but has now been plugged in to the article stream


packages/article/__tests__/android/__snapshots__/article-with-style.android.test.js.snap
```diff
@@ -0,0 +1,949 @@
...
+          Some content
+        </Text>
+      </View>
+    </View>
+    <View />
+  </View>
+</RCTScrollView>
+`;
+
+exports[`9. an article with updated byline 2`] = `
```
colincclark: would it be better to just have a full article here for the styles tests, rather than testing the functionality?

PR #1131, opened 2018-07-23
feat: article keywords native ad brand protection

colincclark:
https://nidigitalsolutions.jira.com/browse/REPLAT-3133

- added keywords handler for articles, so we can pass the `kw` to adtech for brand protection on native
- fixed test names inline with TC convention

NOTE: this is somewhat dependent on tpa being deployed after https://github.com/newsuk/times-public-api/pull/543 but can be reviewed and merged in the meantime


packages/utils/src/schema.json
```diff
@@ -441,6 +441,18 @@
               "isDeprecated": false,
               "deprecationReason": null
             },
+            {
```
craigbilner: not sure how you generated this but looks like it _does_ have a description on TPA?

colincclark: did not realise these were generated - have used the npm script instead

colincclark: Although it seems like loads of stuff has changed...?

PR #1137, opened 2018-07-25
test: improve video snapshots

colincclark:


packages/video/.jestlint
```diff
@@ -0,0 +1,4 @@
+{
+  "maxAttributes": 9,
```
craigbilner: I think this could still be broken down? You don't need `className` outside of the style snaps and you can split out the testing of the `data-` attributes which only number 5?

packages/video/__tests__/video-shared-with-style.native.js
```diff
@@ -1,39 +1,48 @@
...
   flattenStyleTransform,
-  minimalNativeTransform,
+  minimaliseTransform,
   print
 } from "@times-components/jest-serializer";
 import { iterator } from "@times-components/test-utils";
 import Video from "../src/video";
 import defaultVideoProps from "./default-video-props";
 
+jest.mock("@times-components/image", () => "PlayIcon");
```
craigbilner: people may find this confusing?

packages/video/__tests__/video-shared.native.js
```diff
@@ -6,13 +6,15 @@ import {
...
+  print,
+  replacePropTransform
 } from "@times-components/jest-serializer";
-import { iterator } from "@times-components/test-utils";
+import { hash, iterator } from "@times-components/test-utils";
 import Video from "../src/video";
 import defaultVideoProps from "./default-video-props";
 
 jest.mock("@times-components/image", () => "Image");
+jest.mock("@times-components/image", () => "PlayIcon");
```
craigbilner: did you mean to double mock?

packages/video/__tests__/video-shared-with-style.native.js
colincclark: typo. fixed

packages/video/__tests__/video-shared.native.js
colincclark: removed duplicate

PR #1145, opened 2018-07-26
fix: Allow poster image to show with lead asset videos

jimhunty:
This PR fixes the issue mentioned in: https://nidigitalsolutions.jira.com/browse/REPLAT-2809

### DIFF
--- a/packages/utils/src/schema.json
+++ b/packages/utils/src/schema.json
@@ -438,31 +438,6 @@
               "deprecationReason": null
             },
             {
-              "name": "keywords",
-              "description":
-                "A field that is populated from the article headline, a string delineated with commas such as [\"this,is,a,headline\"]",
-              "args": [],
-              "type": {
-                "kind": "NON_NULL",
-                "name": null,
-                "ofType": {
-                  "kind": "LIST",
-                  "name": null,
-                  "ofType": {
-                    "kind": "NON_NULL",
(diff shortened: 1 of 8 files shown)


packages/utils/src/schema.json
```diff
@@ -437,31 +437,6 @@
               "isDeprecated": false,
               "deprecationReason": null
             },
-            {
-              "name": "keywords",
```
colincclark: keywords have been removed

jimhunty: As stated on slack this exists here: https://github.com/newsuk/times-components/blob/c3e60c899c33defc06de4783123dc63996d81a0b/packages/utils/src/schema.json#L491

Not sure why its in two places but still works.

Can also be seen working here on the branch deployment: https://deploy-preview-1145--xenodochial-benz-aa23b6.netlify.com/?knob-Size%20of%20ad%20placeholder%3A=default&selectedKind=Helpers%2FProvider&selectedStory=Article&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

colincclark: Oh! that is what you meant...! yes this is the correct one to remove. It must have slipped in on an earlier PR. It was created manually, but erroneously, so yes, good shout in removing it

PR #1146, opened 2018-07-26
Receive font size from android app and assign a listener for updates

tuncaulubilge:
Bridges the initial font size and adds listeners for continuous updates. Font size is passed into articles through the context package as the scale.


android-app/pages/article.js
```diff
@@ -31,26 +32,52 @@ const platformAdConfig = {
...
-      onVideoPress={onVideoPress}
-      onTopicPress={onTopicPress}
-      platformAdConfig={adConfig}
-    />
-  );
-};
+class ArticleView extends Component {
+  constructor(props) {
+    super(props);
+    this.subscription = null;
```
craigbilner: is there value in this?

android-app/pages/article.js
```diff
@@ -31,26 +32,51 @@ const platformAdConfig = {
...
-      onVideoPress={onVideoPress}
-      onTopicPress={onTopicPress}
-      platformAdConfig={adConfig}
-    />
-  );
-};
+class ArticleView extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { fontSize: this.props.fontSize };
```
colincclark: please destruct `fontSize` above this from `props`

android-app/pages/article.js
```diff
@@ -31,26 +32,51 @@ const platformAdConfig = {
...
+
+  render() {
+    const { articleId, sectionName } = this.props;
+    const adConfig = { ...platformAdConfig, sectionName };
+
+    return (
+      <ArticlePageView
+        articleId={articleId}
+        analyticsStream={track}
+        fontSize={this.state.fontSize}
```
colincclark: neater to destruct above

packages/pages/src/article.js
```diff
@@ -54,7 +56,12 @@ ArticleDetailsPage.propTypes = {
   onAuthorPress: PropTypes.func.isRequired,
   onVideoPress: PropTypes.func.isRequired,
   onLinkPress: PropTypes.func.isRequired,
-  onTopicPress: PropTypes.func.isRequired
+  onTopicPress: PropTypes.func.isRequired,
+  fontSize: PropTypes.string
```
colincclark: easier to read if these remain alphabetised (optional)

tuncaulubilge: Didn't we have a lint rule for this? Does it only work for props but not for prop-types?

colincclark: Correct. Only for props atm. We are intorducing these rules gently, lol

android-app/pages/article.js
```diff
@@ -31,26 +32,53 @@ const platformAdConfig = {
...
-};
+class ArticleView extends Component {
+  constructor(props) {
+    super(props);
+    const { fontSize } = this.props;
+    this.state = { fontSize };
+  }
+
+  componentDidMount() {
+    this.subscription = articleEventEmitter.addListener(
```
craigbilner: @tuncaulubilge did you say this can go now?

tuncaulubilge: If we remove the button from the toolbar, we can also remove these. Let's keep it in for now as it'll make dev easier

android-app/pages/article.js
```diff
@@ -31,26 +33,53 @@ const platformAdConfig = {
...
-      onVideoPress={onVideoPress}
-      onTopicPress={onTopicPress}
-      platformAdConfig={adConfig}
-    />
-  );
-};
+class ArticleView extends Component {
+  constructor(props) {
+    super(props);
+    const { scale } = this.props;
```
colincclark: can just do `const { scale } = props;` here

craigbilner:
will raise a different one

PR #1148, opened 2018-07-27
Version bump react-native to 0.55.4

tuncaulubilge:
React Context does not work with our current react-native version, so we had to bump it.

React-native bumped to 55.4


packages/article-flag/src/utils/beautify.android.js
```diff
@@ -1,8 +0,0 @@
-// apply transformations to add uppercase and letter spacing.
-// letterSpacing CSS prop does not work on android:
-// https://github.com/facebook/react-native/pull/13199
-export default title =>
-  title
-    .toLowerCase()
-    .split("")
-    .join("\u200A");
```
colincclark: 💯

packages/video-label/src/beautify-title.android.js
```diff
@@ -1,9 +0,0 @@
-// apply transformations to add uppercase and letter spacing.
```
colincclark: please amend the README which references this issue

tuncaulubilge: Done :+1:

packages/utils/src/strings.js
```diff
@@ -1,15 +1,5 @@
 /* eslint-disable import/prefer-default-export */
```
colincclark: this is unrequired now I think?

tuncaulubilge: Removed

tuncaulubilge:
I couldn't find anything on react-native release notes about View to Component change. Compared a few snapshots, and seems like only the shallow rendered ones have changed to Component.

PR #1150, opened 2018-07-31
docs: add article readme

colincclark:
- add a simple article package README
- adopt a standard when refering to ASTs in the docs

### DIFF
--- a/packages/article/README.md
+++ b/packages/article/README.md
@@ -2,19 +2,42 @@
 
 The article component is a composed collection of components and features which
 go to make up an article. This is distinct from the concept of an article page
-in the pages package, as that page deals with the data provider, whereas this
-package is intended to be a dumb component.
+in the pages package, as that page deals with the data provider, whereas the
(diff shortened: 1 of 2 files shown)


packages/article/README.md
```diff
@@ -0,0 +1,53 @@
+# Article
+
+Article essentially represents an article page and all of the components and
```
craigbilner: subtly different from `pages/article` though which is a composition of Provider/Context/Dumb Component. This is the dumb component part

packages/article/README.md
```diff
@@ -0,0 +1,53 @@
...
+features found therein. This package is comprised of numerous other components
+and packages, such as `ArticleHeader`, `ArticleTopics` and `RelatedArticles`,
+all of which are related to a specific article. Some of the components utilised
+in the package are self-contained with the article package itself, and some have
+been split out into separate packages. The article data which forms the article
```
craigbilner: can we explain the philosophy here for future travellers?

PR #1153, opened 2018-07-31
Introduce theme package with scale props

tuncaulubilge:
A simple theme package that uses the new Context API. Currently, we only have the font size scale property. It will be used to pass around font size changes by the native apps.

No stories at the moment since this package is not really meaningful on its own.


packages/theme/src/theme.js
```diff
@@ -0,0 +1,5 @@
+import React from "react";
+import articleThemeDefault from "./defaults/theme";
+
+export default React.createContext(articleThemeDefault);
```
colincclark: destruct `createContext` at the top

packages/theme/README.md
```diff
@@ -0,0 +1,46 @@
+# Theme
```
colincclark: should the font-size changes really be considered theming? i envisaged this package as `context` rather than as `theme`....

packages/theme/src/defaults/theme.js
```diff
@@ -0,0 +1,5 @@
+import scales from "./scales";
```
colincclark: change this file name to `index` to enable shorted import paths

packages/theme/src/defaults/scales.js
```diff
@@ -0,0 +1,5 @@
+export default Object.freeze({
```
colincclark: rather than `scales`, can we just call this file `font-sizes`?

packages/theme/src/theme.js
```diff
@@ -0,0 +1,5 @@
+import React from "react";
+import articleThemeDefault from "./defaults/theme";
```
colincclark: if the file was `./defaults/index`, then this could just be `import { articleThemeDefault } from "./defaults"`;

packages/theme/.babelrc
```diff
@@ -0,0 +1,7 @@
+{
+  "presets": ["react-native"],
+  "plugins": [
+    "styled-components",
+    "add-react-displayname"
```
colincclark: are these neccessary?

packages/theme/README.md
tuncaulubilge: This is more like a "scale change" instead of a font-size change, as line height and few paddings are also going to change. And we can even add section colours to the same theme context.

I've started the PR with a "context" package, but after discussing with James, it made more sense to go with a theme package instead, since context is just the implementation detail of it. What do you think?

packages/theme/src/defaults/scales.js
tuncaulubilge: It also changes line heights and paddings etc

packages/theme/README.md
colincclark: I am not convinced that changing the scale of the various elements falls under theming, but happy to go with the consensus. Would prefer this package to be called `context`

craigbilner: @kaliabadi actually has a case now where he needs to give a callback for the platform to construct an environmental specific URL which would naturally fit on `context` but not `theme`. Theme can live on Context though?

packages/theme/__tests__/shared.js
```diff
@@ -0,0 +1,26 @@
...
+import themeDefaults from "../src/defaults";
+
+export default () => {
+  it("with default values", () => {
+    const { scale: defaultScale } = themeDefaults;
+    const testInstance = TestRenderer.create(
+      <Theme.Consumer>{theme => theme.scale}</Theme.Consumer>
+    );
+
+    expect(testInstance.toJSON()).toBe(defaultScale);
```
craigbilner: FYI you often don't need `toJSON` because snapshotting performs the same `toString` action

tuncaulubilge: Getting this error when I remove toJSON:
`Comparing two different types of values. Expected string but received object.`

packages/context/__tests__/shared.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import TestRenderer from "react-test-renderer";
+import Context from "../src/context";
+import defaults from "../src/defaults";
+
+export default () => {
+  it("with default theme", () => {
```
colincclark: with default scale

packages/context/__tests__/shared.js
```diff
@@ -0,0 +1,26 @@
...
+    const { theme: { scale: defaultScale } } = defaults;
+    const testInstance = TestRenderer.create(
+      <Context.Consumer>{context => context.theme.scale}</Context.Consumer>
+    );
+
+    expect(testInstance.toJSON()).toBe(defaultScale);
+  });
+
+  it("with inline theme", () => {
```
colincclark: with inline scale

PR #1154, opened 2018-08-01
refactor: refactor article flag, add default showcase, add readme

colincclark:


packages/article-flag/README.md
```diff
@@ -0,0 +1,40 @@
+# Article Flag
+
+Artivle flags are attached to an article like a lable to indicate the article's
```
craigbilner: Article
label

packages/article-flag/README.md
```diff
@@ -0,0 +1,40 @@
...
+status. There are some common statuses that are exported as components from this
+package, such as "new", "exclusive" or "sponsored", but the default component
```
craigbilner: a bit odd to name 3/4?

craigbilner: perhaps we should also say that the editors/CMS determines the flag for an article and they currently "expire" server side

packages/article-flag/README.md
colincclark: fixed typos

packages/article-flag/README.md
colincclark: was my way of not tieing the docs to the code to tightly. Have added the other one

colincclark: Not keen on coupling the docs for TC with what happens server side, but have amended

PR #1159, opened 2018-08-01
docs: add article list readme

colincclark:

### DIFF
--- a/packages/article-list/README.md
+++ b/packages/article-list/README.md
@@ -1,0 +1,50 @@
+# Article List
+
+This represents a paginated list of articles. These articles typically appear on
+a page, associated to a particular author or topic. An article list is composed
+of many other packages such as card, article summary and link. These components
+are primarily used for the layout of the individual article items themselves.
+
+## Contributing
+
+Please read [CONTRIBUTING.md](./CONTRIBUTING.md) before contributing to this
+package
+
+## Running the code
+
+Please see our main [README.md](../README.md) to get the project running locally
+
+## Development
+
+The code can be formatted and linted in accordance with the agreed standards.
+
+```
+yarn fmt
+yarn lint
+```
+
+## Testing
+
+This package uses [yarn](https://yarnpkg.com) (latest) to run unit tests on each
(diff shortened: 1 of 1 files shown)


packages/article-list/README.md
```diff
@@ -0,0 +1,50 @@
+# Article List
+
+This represents a paginated list of articles. These articles typically appear on
+a page, associated to a particular author or topic. An article list is composed
+of many other packages such as card, article summary and link. These components
+are primarily used for the layout of the individual article items themselves.
```
craigbilner: think this could do with a lot more explanation around native/web, lazy loading images error handling, split providers etc. etc.?

colincclark: done

packages/article-list/README.md
```diff
@@ -0,0 +1,71 @@
...
+Whilst the web flavour of article list utilises the pagination package, the
+native version contains its own brand of pagination in the form of infinite
+scrolling. This utilises the react-native `FlatList` and the GraphQL `fetchMore`
+method to show more articles when a user has scrolled to the bottom of the
+article list.
+
+## Error Handling
+
+The error view package utilises the React `componentDidCatch` lifecycle event to
+set yup an
```
craigbilner: yup, any cards that fail simply don't show but the page count is unchanged

packages/article-list/README.md
```diff
@@ -0,0 +1,71 @@
...
+[storybook](http://components.thetimes.co.uk/?knob-Size%20of%20ad%20placeholder%3A=default&selectedKind=Composed%2FArticle%20List&selectedStory=Default%20with%20images&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)
+to see our available article list templates.
+
+## Future
+
+Inifinite scroll pagination is baked into this package for the native
+experience. We are planning to fish this out into the pagination package in the
+future.
+
+Lazy loading for the web is baked into this package at the moment. We intend to
```
craigbilner: also mobweb should move to something like react-virtualised which may or may not be provided by RNW

PR #1160, opened 2018-08-01
fix(depend): add missing param for --only

mcheshkov:
Without `<glob>` part commander interpret this option as boolean flag. Tests didn't catch this because commander is stubbed.

Without this change I get this:
```
$ npx @times-components/depend --only foo --list --lerna ./
TypeError: glob pattern string required
```


craigbilner:
thanks @mcheshkov

PR #1164, opened 2018-08-02
chore: add jest-lint

### LINKED ISSUE #589: Improving snapshots
Snapshot testing could use some rationalising. Snapshots are too large, insufficiently readable, and often change as a result of distant refactorings in other parts of the codebase. Some of this may be fixable by writing and enforcing guidelines for snapshot best practices. The rest can be fixed with some simple utility functions to control what goes into snapshots.

Suggestions:

1. By default, don't let implementation details of one component leak into snapshot tests in another (shallow rendering is supposed to fix this, but sometimes we use enzyme.mount() or react-test-renderer for technical reasons, and these do full depth renders).

2. Provide a way for tests to either blacklist parts of the snapshot that they know are not relevant, or  to whitelist the parts that they are interested in.

3. Make an upper limit on the size of individual snapshots, potentially overridable. article.web.test.js.snap is 1.3MB in size with 11 snapshots.

craigbilner:
A journey was started on 20 June 2018...and that journey is nearly at an end. This PR is the final boss.

A further optimisation needs to be made so we can run linting from the root which isn't supported by `jest-lint` yet but for now we use lerna to run our lint scripts per package.

Close #589


times-tools:
?

times-tools:
olo

times-tools:
help

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/test/index.html

packages/topic/src/topic-prop-types.base.js
```diff
@@ -0,0 +1,23 @@
...
+export const propTypes = {
+  error: PropTypes.object,
+  isLoading: topicHeadPropTypes.isLoading,
+  page: PropTypes.number,
+  pageSize: PropTypes.number,
+  refetch: PropTypes.func.isRequired,
+  slug: PropTypes.string.isRequired,
+  topic: PropTypes.shape({
+    name: topicHeadPropTypes.name,
+    description: topicHeadPropTypes.description
```
colincclark: a to z

PR #1167, opened 2018-08-03
Feat - Comments button in native article

tuncaulubilge:
Added comments button at the bottom of the native article page
Launching the comments page is done by the native apps via comment press callback.

Also includes a disabled state

Paired with @colincclark 

screen shot 2018-08-03 at 11 37 07


packages/link/src/text-link-prop-types.js
```diff
@@ -0,0 +1,13 @@
+import PropTypes from "prop-types";
+import { basePropTypes, baseDefaultProps } from "./text-link-prop-types-base";
+
+export const propTypes = {
+  ...basePropTypes,
+  url: PropTypes.string,
```
colincclark: these can exist in the `text-link-prop-types-base.js` file and just be overridden in the `web` one

colincclark: then propTypes will just = basePropTypes

packages/article/src/article-comments/article-comments.js
```diff
@@ -0,0 +1,57 @@
...
+  commentCount,
+  commentsEnabled,
+  onCommentsPress,
+  onCommentGuidelinesPress,
+  url
+}) =>
+  commentsEnabled ? (
+    <View style={styles.container}>
+      <Text style={styles.headline}>{`${commentCount} comment${
+        commentCount === 1 ? "" : "s"
```
colincclark: small one: this should go in a named function above, so it is more obvious what it is doing. `handleString` or `handleGrammar` or some such thing

packages/article/__tests__/shared.native.js
```diff
@@ -143,6 +147,32 @@ export default () => {
 
         expect(onLinkPress).toHaveBeenCalled();
       }
+    },
+    {
+      name: "a full article with disabled comments",
```
colincclark: and a test that _has_ comments would be good

tuncaulubilge: All other tests have a comment by default as it's part of the default fixture now. Do we still need a separate test for it?

colincclark: That is fine. It was done on an earlier PR which threw me. I was expecting snapshots here

packages/link/src/text-link-prop-types.js
tuncaulubilge: I tried to keep base file as the default values, and put platform specifics into platform files. Neither of these values are actually the default prop types. This way, the differences are clearer, imho.

colincclark: Heard, but we are supposed to be iOS-first. Optional

tuncaulubilge: :+1: will convert to the convention then

packages/article/src/article-comments/article-comments.js
```diff
@@ -0,0 +1,59 @@
...
+
+const ArticleComments = ({
+  articleId,
+  commentCount,
+  commentsEnabled,
+  onCommentsPress,
+  onCommentGuidelinesPress,
+  url
+}) => {
+  const handlePluralSuffix = count => (count === 1 ? "" : "s");
```
craigbilner: good thing we only support english....

packages/article/src/article-comments/article-comments.js
```diff
@@ -0,0 +1,59 @@
...
+  return commentsEnabled ? (
```
craigbilner: for me large render ternaries are a little hard to read, would be nicer to  break this down?

packages/article/src/article-comments/article-comments.js
tuncaulubilge: 😱

tuncaulubilge:
@craigbilner I mocked comments in article tests and created a new snapshot file for it

packages/article/src/article-comments/article-comments.js
colincclark: I have refactored this to be translatable

packages/article/src/article-comments/article-comments.js
```diff
@@ -0,0 +1,61 @@
...
+      </Text>
+      <Text style={styles.supporting}>
+        For more details, please see our {"\n"}
+        <TextLink onPress={onCommentGuidelinesPress} style={styles.link}>
+          community guidelines
+        </TextLink>
+      </Text>
+    </View>
+  );
+  return commentsEnabled ? CommentsView : CommentsDisabledView;
```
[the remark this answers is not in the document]
colincclark: I have refactored this so it only build up the comments/disbaled comments views as appropriate

packages/article/fixtures/full-article.js
```diff
@@ -1803,8 +1803,11 @@ const defaultTopics = [
 const defaultUrl =
   "https://www.thetimes.co.uk/edition/news/france-defies-may-over-russia-37b27qd2s";
 
+const defaultCommentsEnabled = true;
```
craigbilner: should  be alphabetised above

packages/article/src/article-comments/article-comments.js
```diff
@@ -0,0 +1,60 @@
...
+
+const ArticleComments = ({
+  articleId,
+  commentCount,
+  commentsEnabled,
+  onCommentsPress,
+  onCommentGuidelinesPress,
+  url
+}) => {
+  const renderComments = () => (
```
craigbilner: would it be neater to simply have these as functional components and the top and render to just say:

```
if commentsEnabled ? <Comments {...props} /> : <DisabledComments {...props} />
```

packages/article/__tests__/android/__snapshots__/article-comments.android.test.js.snap
```diff
@@ -0,0 +1,316 @@
...
+    <View>
+      <View>
+        <Text>
+          Some content
+        </Text>
+      </View>
+    </View>
+    <View>
+      <View>
+        <ArticleTopics
```
craigbilner: not massively blocking but would have been neater to do with no topics or related articles i.e. the minimum article with comments?

PR #1172, opened 2018-08-06
refactor: caption and add readme

colincclark:


packages/caption/README.md
```diff
@@ -0,0 +1,38 @@
+# Caption
+
+The caption component usually wraps an image and adds optional caption and/or
```
craigbilner: should really be the `caption-credit` package :smirk:

PR #1176, opened 2018-08-07
refactor: error view tests and readme

colincclark:
- add readme
- improve snapshots

### DIFF
--- a/packages/error-view/README.md
+++ b/packages/error-view/README.md
@@ -10,6 +10,15 @@ documentation itself:
 > during rendering, in lifecycle methods, and in constructors of the whole tree
 > below them.
 
+This package provides an `ErrorView` wrapper component that takes a render prop
+as a direct child of the component. This function returns three possible values
+which can be used to handle the enclosed child components:
+
+* `error` - the error that was returned
+* `hasError` - boolean, did the component error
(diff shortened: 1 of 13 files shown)


packages/error-view/README.md
```diff
@@ -0,0 +1,45 @@
+# Error View
+
+The error view package is a React
```
craigbilner: While it's a simple wrapper around the `ErrorBoundary` it does provide a `renderProp` interface so perhaps a "How to use" would be useful?

colincclark: done

packages/error-view/__tests__/shared.js
```diff
@@ -18,49 +19,54 @@ module.exports = () => {
...
+          <ErrorView>
+            {({ hasError, error }) =>
+              hasError ? <ErrorState error={error} /> : <ThrowsError />
+            }
+          </ErrorView>
+        );
 
-    expect(tree).toMatchSnapshot();
-  });
+        expect(testInstance.toJSON()).toMatchSnapshot();
```
craigbilner: FYI - `toJSON` is not required because the way it's coerced to a string is the same

colincclark: removed

PR #1177, opened 2018-08-07
docs: add eslint-config readme

colincclark:


packages/eslint-config-thetimes/README.md
```diff
@@ -0,0 +1,29 @@
+# Eslint Config
+
+This package contains the eslint configuration for Times Components (TC). New TC
+packages created with the CLI will automatically add a pointer to this package
+in their `.eslintrc.json` files.
+
+```
+{
```
craigbilner: As a rule, we should only abide by `prettier` and `Airbnb` but in extenuating circumstances rules can be overriden here should they not make sense for the project

colincclark: done

PR #1179, opened 2018-08-07
refactor: gestures tests and readme

colincclark:
- add readme
- improved tests and coverage
- removed unsafe lifecycle event

Issue raised: https://github.com/newsuk/times-components/issues/1180 due to commented test


packages/gestures/README.md
```diff
@@ -0,0 +1,39 @@
+# Gestures
+
+Gestures is a native wrapper that allows touch gestures on mobile of its child
+components. Gestures supports two finger zoom and rotate, in either direction,
```
craigbilner: perhaps mention that zoom currently only works at the centrepoint of the image

colincclark: done

PR #1188, opened 2018-08-08
Fix: Bylines coloured by section

aronshamash:
First part of https://nidigitalsolutions.jira.com/browse/REPLAT-2522.

- Fixes bylines that should be coloured by section.

## Before
before

## After
after


tuncaulubilge:
@craigbilner, context is a great fit to pass section from the native app, as in "current navigation context". But we are just using the section value from the api article data, so I'm not sure context is the right way for this.

What's the actual requirement? What do we do when the article is in two sections?

PR #1193, opened 2018-08-09
feat: add caption data to modal images

colincclark:
https://nidigitalsolutions.jira.com/browse/REPLAT-2477
https://nidigitalsolutions.jira.com/browse/REPLAT-2456 (android)
https://nidigitalsolutions.jira.com/browse/REPLAT-2989 (iOS)

- added caption/credits to the modal image
- ensure close button is positioned as per designs (right on Android and left on iOS)
- passed caption/credit data through article lead and inline images
- fixed up modal storybook to better showcase the modal (with knobs for caption data)

Future PR to come will fix other issues mentioned at the bottom of the JIRA ticket:
- https://nidigitalsolutions.jira.com/browse/REPLAT-3315
- https://nidigitalsolutions.jira.com/browse/REPLAT-3316


packages/image/src/modal-image.js
```diff
@@ -27,9 +30,11 @@ class ModalImage extends Component {
   }
 
   render() {
+    const { caption, credits } = this.props;
+
     const closeButton = (
       <Svg height={24} viewBox="0 0 24 24" width={24}>
-        <G fill={colours.functional.cancel}>
+        <G fill="#FFFFFF">
```
craigbilner: this should still come from the styleguide though?

yarn.lock
```diff
@@ -937,6 +937,13 @@
     fs-extra "6.0.0"
     glob "7.1.2"
 
+"@times-components/caption@2.1.1":
```
craigbilner: got dep issues

packages/article/src/article-lead-asset/article-lead-asset-image.js
```diff
@@ -1,18 +1,31 @@
...
-export const propTypes = {
+ArticleLeadAssetImage.propTypes = {
+  caption: PropTypes.string,
+  credits: PropTypes.string,
   crop: cropPropTypes.isRequired
 };
 
-ArticleLeadAssetImage.propTypes = propTypes;
+ArticleLeadAssetImage.defaultProps = {
+  caption: "",
```
craigbilner: I think if we use default to represent the absence of something `null` is favourable over empty string?

colincclark: that feels odd to me, but have amended to `null`

yarn.lock
colincclark: yep, fixed dep

packages/image/src/modal-image.js
colincclark: done

packages/styleguide/src/colours/functional.js
```diff
@@ -14,7 +14,8 @@ const functionalColours = {
   articleFlagExclusive: "#C51D24",
   articleFlagNew: "#E34605",
   articleFlagUpdated: "#3C81BE",
-  cancel: "rgba(255, 255, 255, 0.8)"
+  cancel: "rgba(255, 255, 255, 0.8)",
```
tuncaulubilge: Looks like we are not using this cancel colour anywhere else, we can ditch it

colincclark: removed

PR #1194, opened 2018-08-09
fix: article fixtures

colincclark:
- this was breaking iOS

### DIFF
--- a/packages/article/article.showcase.js
+++ b/packages/article/article.showcase.js
@@ -1,15 +1,12 @@
 /* eslint-disable react/prop-types */
 /* eslint-env browser */
 import React from "react";
-import { Text } from "react-native";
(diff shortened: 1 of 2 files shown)


packages/article/article.showcase.js
```diff
@@ -1,14 +1,12 @@
 /* eslint-disable react/prop-types */
 /* eslint-env browser */
-
 import React from "react";
 import { addTypenameToDocument } from "apollo-utilities";
-
-import { ArticleProvider, articleQuery } from "@times-components/provider";
+import { ArticleProvider } from "@times-components/provider";
+import { article as articleQuery } from "@times-components/provider-queries";
```
craigbilner: this should have failed linting :-(

colincclark: which lint rule?

Issue #1186, opened 2018-08-08
Iterator to take extra args for setup/teardown


colincclark:
**Do you want to request a _feature_ or report a _bug_?**

Feature

**What is the current behaviour?**

The `iterator` helper in `test-utils` currently only takes an array fo tests and sets up those tests, numerically. It would be beneficial if the API handled setup and teardown functions so we could do away with annoying empty `describe` methods elsewhere

ValGeorgiev:
This is not needed anymore, I will close it.

PR #1196, opened 2018-08-09
Refactor: Related Articles - Split-up snapshots

aronshamash:
- Snapshot file sizes were close to `jest-lint` size limit so split by number of articles.


packages/related-articles/__tests__/shared-util.js
```diff
@@ -59,6 +51,13 @@ export default ({
     mockDate.reset();
     global.Intl = realIntl;
   });
+}
+
+export const noArticlesTests = ({
```
aronshamash: Suggestions on how to split this or rename this needed.

packages/related-articles/__tests__/android/__snapshots__/la2-1article.android.test.js.snap
```diff
@@ -0,0 +1,90 @@
...
+                <Text>
+                  <Text>
+                    
+                    Summary 125
+                    ...
+                  </Text>
+                </Text>
+              </View>
+              <Text>
+                March 13 2015, 06:54pm
```
colincclark: is this how we do dates? looks weird

packages/related-articles/__tests__/shared-util.js
```diff
@@ -34,15 +34,7 @@ export const createRelatedArticlesProps = (
...
-export default ({
-  fixture1,
-  fixture2,
-  fixture3,
-  one,
-  template,
-  three,
-  two
-}) => renderComponent => {
+const beforeAndAfterEach = () => {
```
colincclark: my only comment would be we should move this stuff to the test-utils iterator function. But as discussed, on another PR _a la_ https://github.com/newsuk/times-components/issues/1186

packages/related-articles/__tests__/android/__snapshots__/la2-1article.android.test.js.snap
craigbilner: now we've bumped Jest I was going to do a tech debt task to use the new date matchers so we don't have to do the funny mocking stuff

Issue #456, opened 2017-12-11
Add script to fail CI on unhoisted packages


craigbilner:
**Do you want to request a _feature_ or report a _bug_?**

Feature

**What is the current behaviour?**

Yarn workspaces happily splits out packages when "hoisted" without notification

**If the current behaviour is a bug, please provide the steps to reproduce.**

Change versions of the same package between package deps

**What is the expected behaviour?**

CI to fail if packages aren't aligned

**If this is a feature request, what is the motivation or use case for changing
the behaviour?**

To save on yarn time

**Please mention other relevant information such as the browser version, Node.js
version and Operating System.**

craigbilner:
@thetimes/nmkeep will have this capability, need to add to linting on repo

craigbilner:
Blocked by storybook. Need to upgrade to v4 with webpack 4 support before packages will be completely hoisted

PR #1199, opened 2018-08-10
chore: refactor deps to ensure all are hoisted

craigbilner:
Currently we don't hoist all our packages because of out-of-date deps and moving to webpack 4 for our bundling. I tried to get to the alpha version of storybook which will remove the webpack 3 dependency but it was rather miserable. This PR cleans up the other deps so should at least make yarning a little faster, reduce I/O a little bit and just leave the storybook bump outstanding

After that we can address #456 

nb: the noise is all about babel 7 now which will be YAPR


package.json
```diff
@@ -6,7 +6,7 @@
   "main": "index.js",
   "engines": {
     "node": ">=8.10.0 <10",
-    "yarn": ">1.9.0 <2"
+    "yarn": "1.9.4"
```
colincclark: we are going for the fixed version of `yarn`? this will break if CI continues to bump automatically?

craigbilner: that's correct. Without fixed versions of yarn it leads to more problems with a team of people and integrating with render. Yarn can make breaking changes across versions.

colincclark: can we lock down our CI versions then and choose when to bump those to keep everything in sync? `yarn` do seem to use semver, although their changelog does not highlight breaking changes very well

PR #1200, opened 2018-08-10
fix: Font scaling design review

jimhunty:
Minor changes that Jake has requested. All approved in iOS and Android


packages/article/src/styles/article-body/shared.js
```diff
@@ -15,6 +15,15 @@ const sharedStyles = scale => ({
...
     paddingRight: spacing(2)
+  },
+  articleLink: {
+    color: colours.functional.action,
+    ...styleguide({ scale }).fontFactory({
+      font: "body",
+      fontSize: "bodyMobile"
+    }),
+    marginBottom: spacing(5),
+    marginTop: 0
```
colincclark: should this be `spacing(0)`?

Issue #599, opened 2018-02-05
Find a linter for styled components


aronshamash:

craigbilner:
Looks like this is the [official](https://github.com/styled-components/stylelint-processor-styled-components) one?

PR #1203, opened 2018-08-10
feat: add stylelint with example

craigbilner:
Proposal for #599 

With the way `withResponsiveStyles` works at the moment it's not possible for the linter to infer the css to lint. This refactor would allow us to move to responsive styles (and fix my bugbear with using native elements on web).


craigbilner:
to roll out

PR #1204, opened 2018-08-10
Feat: Colour Article byline, Article label and Key facts using context 

aronshamash:
- Article byline, article label and key facts coloured by section using `context`.

Notes:

- In the apps we need to cater for when it is a deeplinked article with no section in the navigation (no context). In this case the section is taken from TPA.


packages/article/src/article-header-label/article-header-label.base.js
```diff
@@ -15,6 +16,13 @@ export default render => ({ isVideo, label }) => {
       testID: "label",
       style: styles.articleLabel
     },
-    <Label color={colours.section.default} title={label} />
+    <Context.Consumer>
+      {({ theme: { sectionColour } }) => (
+        <Label
+          color={sectionColour || colours.section[section]}
```
craigbilner: still wouldn't expect it to know anything about section?

colincclark: does this need a whitebaord sesh? @aronshamash

aronshamash: This logic is due to this requirement:
`In the apps we need to cater for when it is a deeplinked article with no section in the navigation (no context). In this case the section is taken from TPA.`

craigbilner:
WIP

packages/pages/src/article.js
```diff
@@ -1,10 +1,16 @@
...
 import Article from "@times-components/article";
+import Context, { defaults } from "@times-components/context";
+import { ArticleProvider } from "@times-components/provider";
+import { colours } from "@times-components/styleguide";
 import withClient from "./client/with-client";
 import adTargetConfig from "./client/ad-targeting-config";
 
+const pickSectionColour = (sectionName, article) =>
+  colours.section[sectionName] ||
+  (article ? colours.section[article.section] : undefined);
```
craigbilner: can we use `null` instead?

packages/pages/src/article.js
```diff
@@ -1,10 +1,16 @@
...
 import PropTypes from "prop-types";
-import { ArticleProvider } from "@times-components/provider";
 import Article from "@times-components/article";
+import Context, { defaults } from "@times-components/context";
+import { ArticleProvider } from "@times-components/provider";
+import { colours } from "@times-components/styleguide";
 import withClient from "./client/with-client";
 import adTargetConfig from "./client/ad-targeting-config";
 
+const pickSectionColour = (sectionName, article) =>
```
craigbilner: minor: bit of an odd API to take `article`? Perhaps just `section`?

packages/key-facts/src/key-facts-title.js
```diff
@@ -1,16 +1,9 @@
...
 
-const KeyFactsTitle = ({ title }) => (
-  <Context.Consumer>
-    {({ theme: { scale } }) => {
-      const styles = styleFactory(scale);
-      return <Text style={styles.title}>{title.toUpperCase()}</Text>;
-    }}
-  </Context.Consumer>
+const KeyFactsTitle = ({ styles, title }) => (
+  <Text style={styles.title}>{title.toUpperCase()}</Text>
```
craigbilner: usually you wouldn't pass a prop called `styles` and assume it has a `title` prop. Either a component should just take the given styles, or better accept a whitelist of styles.

tuncaulubilge: Instead of breaking styles into smaller chunks and overcomplicating styles by splitting them by "themeable and non-themeable" props, I'd prefer using context and stylesheet factory on each component. What do you think?

PR #1208, opened 2018-08-13
Feat: Use shortHeadline in related articles with fallback to headline

aronshamash:
- Use `shortHeadline` with a fallback to `headline` in related articles.


packages/related-articles/src/related-articles-tracking-context.js
```diff
@@ -22,7 +22,7 @@ export default Component =>
           return {
             id,
             byline: get(byline, "[0].children[0].attributes.value", ""),
-            headline,
+            headline: shortHeadline || headline,
```
craigbilner: bit sad the logic is in two places :thinking:

PR #1213, opened 2018-08-14
feat: add semantic elements

### LINKED ISSUE #606: Use semantic tags in all packages

craigbilner:
The web articles weren't well structured in terms of "semantic HTML". This PR adds stronger semantics for SEO and a11y. There could be extra aria work as we get more rigorous with our a11y testing but this gives the article a sense of structure

Close #606

### DIFF
--- a/packages/article/src/article-topics/index.web.js
+++ b/packages/article/src/article-topics/index.web.js
@@ -9,22 +9,26 @@ const ShowTopics = ({ device, onPress, topics }) => {
     if (device === "DESKTOP") {
       return (
         <TopicsMetaContainer>
-          <ArticleTopics
-            onPress={onPress}
-            style={styles.topicsMetaContainer}
-            topics={topics}
-          />
+          <nav>
+            <ArticleTopics
+              onPress={onPress}
+              style={styles.topicsMetaContainer}
+              topics={topics}
+            />
+          </nav>
         </TopicsMetaContainer>
       );
     }
 
     return (
       <TopicsContainer>
-        <ArticleTopics
-          onPress={onPress}
-          style={styles.topicsContainer}
-          topics={topics}
-        />
+        <nav>
+          <ArticleTopics
+            onPress={onPress}
+            style={styles.topicsContainer}
+            topics={topics}
+          />
+        </nav>
       </TopicsContainer>
     );
   }
--- a/packages/article/src/article-topics/index.web.js
+++ b/packages/article/src/article-topics/index.web.js
@@ -9,7 +9,7 @@ const ShowTopics = ({ device, onPress, topics }) => {
     if (device === "DESKTOP") {
       return (
         <TopicsMetaContainer>
-          <nav>
+          <nav role="navigation">
             <ArticleTopics
               onPress={onPress}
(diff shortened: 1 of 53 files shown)


packages/article/src/article-topics/index.web.js
```diff
@@ -9,22 +9,26 @@ const ShowTopics = ({ device, onPress, topics }) => {
     if (device === "DESKTOP") {
       return (
         <TopicsMetaContainer>
-          <ArticleTopics
-            onPress={onPress}
-            style={styles.topicsMetaContainer}
-            topics={topics}
-          />
+          <nav>
```
aronshamash: could add `role="navigation"` here

packages/article/src/article.web.js
```diff
@@ -60,51 +60,57 @@ const renderArticle = (
...
-    <Fragment>
+    <article>
       <HeaderAdContainer key="headerAd">
-        <Ad
-          contextUrl={url}
-          section={section}
-          slotName="header"
-          style={adStyle}
-        />
+        <aside>
```
aronshamash: According to http://html5doctor.com/aside-revisited/  `<aside>` inside an `<article>`  the content show be related to the article (not the page). Here the ads are not related to the article.

packages/article/src/article.web.js
```diff
@@ -60,51 +60,57 @@ const renderArticle = (
...
           <ArticleBody content={content} contextUrl={url} section={section} />
         </BodyContainer>
       </MainContainer>
       <ArticleTopics onPress={onTopicPress} topics={topics} />
-      {displayRelatedArticles}
-      <Ad contextUrl={url} section={section} slotName="pixel" />
-      <Ad contextUrl={url} section={section} slotName="pixelteads" />
-      <Ad contextUrl={url} section={section} slotName="pixelskin" />
-    </Fragment>
+      <aside>
```
aronshamash: same as above

packages/article/src/article.web.js
aronshamash: I think I would remove the `<aside>`

aronshamash: Maybe the ad should have `aria-hidden="true"` ?

packages/article/src/article-topics/index.web.js
craigbilner: because it's implicit it fails linting

packages/related-articles/.jestlint
```diff
@@ -1,5 +1,5 @@
 {
   "maxAttributeStringLength": 44,
-  "maxFileSize": 11118,
-  "maxLines": 517
+  "maxFileSize": 11245,
```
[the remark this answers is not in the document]
craigbilner: I know :-(, not sure the best approach to take with them now

packages/pull-quote/src/pull-quote.web.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import PQ from "./pull-quote.base";
```
colincclark: `PullQuoteBase`

PR #1216, opened 2018-08-14
Feat: Topic description type update

aronshamash:
- Alongside https://github.com/newsuk/times-public-api/pull/584
- update Topic `description` to allow editorial to add basic style

### DIFF
--- a/packages/topic/src/topic-head-prop-types.js
+++ b/packages/topic/src/topic-head-prop-types.js
@@ -1,13 +1,14 @@
 import PropTypes from "prop-types";
+import { treePropType } from "@times-components/markup";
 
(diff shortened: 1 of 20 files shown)


packages/topic/src/topic-head-prop-types.js
```diff
@@ -1,13 +1,14 @@
 import PropTypes from "prop-types";
+import { treePropType } from "@times-components/markup";
 
 export const propTypes = {
   name: PropTypes.string,
-  description: PropTypes.string,
+  description: PropTypes.arrayOf(treePropType),
```
colincclark: this is causing lots of warnings in the test runner @aronshamash

PR #1219, opened 2018-08-15
docs: add provider readme

colincclark:


packages/provider/README.md
```diff
@@ -0,0 +1,49 @@
...
+yarn test:web
+```
+
+Visit the official
+[storybook](http://components.thetimes.co.uk/?selectedKind=Primitives%2FSlice&selectedStory=Default%20template%20with%20one%20item&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)
+to see our available provider templates.
+
+## Future
+
+We are working on improving the developer experience around fixture generation,
```
craigbilner: I think `provider-test-tools` will be mostly effected now that `provider-queries` has been split out

packages/provider/README.md
```diff
@@ -0,0 +1,49 @@
+# Provider
+
+The provider package manages the connections to [GraphQL](https://graphql.org/)
+(GQL). At a fundamental level, the package simply exports providers that take
+GQL queries (from the provider queries package), and use
+[React Apollo](https://github.com/apollographql/react-apollo) to fetch data from
```
craigbilner: we use the old version of [this](https://www.apollographql.com/docs/react/essentials/queries.html#basic) we should reference and update

packages/provider/README.md
```diff
@@ -0,0 +1,49 @@
...
+the GQL server. These providers can be configured to take a `debounceTimeMs`
+prop which adds debounce functionality to the provider calls to ensure better
+performance and user experience.
```
craigbilner: primarily gives us control over the API to our dumb components and is where we can put extra functionality/fixes/things that are missing from the Query HOC

packages/provider/README.md
colincclark: removed

packages/provider/README.md
colincclark: done

packages/provider/README.md
colincclark: done

PR #1220, opened 2018-08-15
fix: article accessibility

tuncaulubilge:
Couple of accessibility improvements
- Rename video splash accessibility label to "Video Splash" from "splash-component"
- Add button traits to article-topics to emphasise click-ability
- Remove accessibility label from article labels so it uses the actual label text


packages/article/src/article-header-label/article-header-label.base.js
```diff
@@ -12,7 +12,6 @@ export default render => ({ isVideo, label }) => {
 
   return render(
     {
-      accessibilityLabel: "label",
```
aronshamash: Is this not needed?

tuncaulubilge: Accessibility label is read by the screen readers. So it says "label" instead of reading the actual text of that label. We should avoid using accessibility label on TextViews.

We should only use accessibility labels when we want screen reader to read some extra info instead of the actual text, like we do in article flags. Article flags read "Updated Flag" instead of just reading "Updated".

Issue #1222, opened 2018-08-16
Squash styles


craigbilner:
**Do you want to request a _feature_ or report a _bug_?**

Feature

**What is the current behaviour?**

react-native-web styles are hoisted in a snapshot for comparison but are bloated by not being unique

**What is the expected behaviour?**

See [article-with-style.web.test.js.snap](https://github.com/newsuk/times-components/blob/master/packages/article/__tests__/web/__snapshots__/article-with-style.web.test.js.snap)

```
"S10": {
    "line-height": "15px",
    "margin-bottom": "0px",
    "margin-top": "0px",
    "padding-bottom": "0px",
    "padding-top": "0px",
  },
  "S11": {
    "margin-bottom": "0px",
    "margin-top": "0px",
    "padding-bottom": "5px",
    "padding-top": "5px",
  },
  "S12": {
    "line-height": "15px",
    "margin-bottom": "0px",
    "margin-top": "0px",
    "padding-bottom": "0px",
    "padding-top": "0px",
  },
```

S10 and S12 can be squashed

**If this is a feature request, what is the motivation or use case for changing
the behaviour?**

Smaller snapshots

PR #1226, opened 2018-08-16
refactor: remove resp package

### LINKED ISSUE #599: Find a linter for styled components

craigbilner:
This PR removes the responsive styles package and uses styled components straight up from each package that needs it. I've also made it a lot less DRY in order to simplify the dev experience of understanding what style something has. This also means everything can be explicitly linted with `stylelint` (found one bug)

*important*

We can't use web elements e.g. `styled.div` because the style system is largely based on RNW with all it's overrides.

When composing complex breakpoint logic, make sure the media queries are still ordered correctly otherwise you can get odd behaviour. e.g. do complex logic for medium breakpoint, then compose complex logic for wide breakpoint _even if_ there's overlap

When reviewing:
* `article-image` and `card` have some odd snapshots (from early experiments) that I'll change separately
* check I'm using the correct RN component and haven't messed up the media query
* particularly on `slice` I refactored the inline logic to compose styles, small downside are a couple of extra css rules, upside is better linting and readability
* check I haven't missed something in storybook
* sadly had to bump some snapshot rules, will address as part of #1222 

refactor: remove responsive styles package and use `styled` explicitly
feat: add `stylelint`

Close #599


packages/article-image/src/styles/responsive.web.js
```diff
@@ -1,19 +1,18 @@
...
-    margin-top: ${spacing(-2)};
-    padding-bottom: ${spacing(4)};
-    padding-left: ${spacing(2)};
-    padding-right: ${spacing(2)};
-    padding-top: 0;
-    vertical-align: top;
-    width: 50%;
-  `,
-  mediumUp: () => `
+export const InsetCaptionContainerStyle = styled(View)`
```
colincclark: in the spirit of no RN components in `web`, why not just have `styled.div` here?

craigbilner: see desc, would have to be a much bigger change, I tried it and backed out

packages/article-list/src/styles/responsive.web.js
```diff
@@ -46,92 +39,72 @@ export const ListItemWrapper = withResponsiveStyles(
...
-    base: () => `
-    margin-left: ${spacing(2)};
-    margin-right: ${spacing(2)};
-  `,
-    mediumUp: () => `
+  }
+`;
+
+export const ListItemSeparator = styled(View)`
+  margin-left: ${spacing(2)};
```
colincclark: these could be refactored to `marginHorizontal`

packages/article/src/article-topics/styles/responsive.web.js
```diff
@@ -1,22 +1,27 @@
...
-export const TopicsMetaContainer = withResponsiveStyles(
-  View,
-  {
-    base: () => `display: none;`,
-    wideUp: () => `display: block;`
-  },
-  "TopicsMetaContainer"
-);
+  @media (min-width: ${breakpoints.medium}px) {
+    width: 80.8%;
```
colincclark: think this `width` should remain a global setting as it is used in several places and has been changed a few times

PR #1233, opened 2018-08-17
feat: article error state

tuncaulubilge:
Add an error state for native apps with a retry button

screen shot 2018-08-17 at 15 12 17


packages/article/src/article-error.web.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+import { Text, View } from "react-native";
+
+const ArticleError = props => (
+  <View>
+    <Text>An error occurred</Text>
+    <Text>{JSON.stringify(props, null, 2)}</Text>
```
colincclark: can we fix this at the same time as doing android?

packages/article/src/styles/article-error/index.js
```diff
@@ -0,0 +1,41 @@
...
+  fontSizes,
+  spacing
+} from "@times-components/styleguide";
+
+export default StyleSheet.create({
+  errorContainer: {
+    height: "100%",
+    justifyContent: "space-between",
+    marginLeft: spacing(2),
+    marginRight: spacing(2),
```
colincclark: `marginHorizontal`

packages/article/src/styles/article-error/index.js
```diff
@@ -0,0 +1,41 @@
...
+    paddingTop: spacing(2),
+    paddingBottom: spacing(14)
+  },
+  errorImageContainer: {
+    alignSelf: "center",
+    marginBottom: spacing(4),
+    marginTop: spacing(4)
```
colincclark: `marginVertical`

packages/article/src/article-error.web.js
craigbilner: we should just remove?

tuncaulubilge: We are not really using this error state in web, so I just left the error message in here. 

I considered using the same view for web, but we have a local image asset for this view and web doesn't seem to play along with local images.

tuncaulubilge: We could also remove it completely, and just return null on error for web

colincclark: Yes, remove it completely please

tuncaulubilge: Removed in 4a91707

packages/article/src/article.web.js
```diff
@@ -2,7 +2,6 @@ import React from "react";
 import Ad, { AdComposer } from "@times-components/ad";
 import RelatedArticles from "@times-components/related-articles";
 import ArticleBody from "./article-body/article-body";
-import ArticleError from "./article-error";
```
colincclark: can the `.web` file be deleted now?

tuncaulubilge: article-error.web is deleted now

PR #1235, opened 2018-08-17
feat: squash styles

### LINKED ISSUE #1222: Squash styles

craigbilner:
We have a lot of duplicate RNW styles in our snapshots, this change squashes them by removing dupes and using a single hoisted `className`

Probably best reviewed by commit:

* Provide `AppRegistry`: it's hard for external users of `jest-serializer` to use without RNW, this abstracts it to a dev dep
* Use AppRegistry: effectively the chore to implement the above without major bumping everyone
* Squash styles: the actual feature, find an existing hoisted style, use that instead
* Update snaps: the change run across all packages
* Restrict lint: now we can reduce some of the `jestlint` cheating

Close #1222


packages/author-profile/__tests__/web/__snapshots__/author-profile-with-style.web.test.js.snap
```diff
@@ -79,7 +79,6 @@ exports[`1. an article list header 1`] = `
     "textDecorationLine": "none",
   },
   "S1": {},
-  "S10": {},
```
colincclark: could we go further and just remove the empties?

packages/jest-serializer/__tests__/web/rnw.test.js
```diff
@@ -78,8 +79,41 @@ describe("The React Native Web serializer should", () => {
...
+    const component = (
+      <View style={styles.parent}>
+        <Text style={styles.child1}>child 1</Text>
+        <Text style={styles.child2}>child 2</Text>
+        <View>
+          <Text style={styles.child3}>child 3</Text>
+        </View>
+      </View>
+    );
+    const tree = renderer.create(component);
```
colincclark: where is that `testRenderer` lint rule when you need it?

packages/jest-serializer/src/rnw.js
```diff
@@ -138,5 +162,5 @@ export const rnwTransform = includeStyleProps => (
   };
 };
 
-export default includeStyleProps =>
-  traverse(stylePrinter, rnwTransform(includeStyleProps));
+export default (AppRegistry, includeStyleProps) =>
```
colincclark: why is `AppRegistry` passed in as an arg to `rnwTransform ` and not just imported into this file?

craigbilner: RNW caches styles at the consumer and is not accessible by the serializer's instance. So it only works in times-components and nowhere else atm. By allowing the consumer to say, here's my RNW instance, other consumers can use the serializer.

PR #1238, opened 2018-08-21
Fix: check for array before array length

aronshamash:
- check for existence of `Array` before `Array` length


packages/topic/src/topic-head.js
```diff
@@ -7,7 +7,7 @@ import styles from "./styles";
 
 const TopicHead = ({ name, description, isLoading }) => {
   const showDescription = () =>
-    description.length > 0 ? (
+    Array.isArray(description) && description.length > 0 ? (
```
colincclark: you can just do `description && decription.length > 0`. The type-checking is done with `prop-types` atm

PR #1245, opened 2018-08-22
fix: video svg to use times svgs

tuncaulubilge:
We had a few places that still uses svgs instead of times-components/svgs and this was crashing articles for native (since svgs depends on react-native-svg)

Made few changes to fix it
- times svgs now supports opacity
- Video play icon re-implemented using svgs
- All other components that are depending on svgs moved to times svgs. Times SVGS exports svgs for web by default


packages/video/__tests__/android/__snapshots__/video.android.test.js.snap
```diff
@@ -10,24 +10,25 @@ exports[`1. video 1`] = `
...
         width={70}
       >
-        <rect
-          fill="rgba(0,0,0)"
+        <Rect
+          fill="#000000"
           fillOpacity="0.4"
           height="100"
-          stroke="rgb(255,255,255)"
-          strokeWidth="8"
```
tuncaulubilge: Needs to be passed to svgs

PR #1248, opened 2018-08-23
docs: add storybook readme

colincclark:


packages/storybook/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+```
+import { showcaseConverter } from "@times-components/storybook";
+import showcase from "./my-component.showcase";
+
+showcaseConverter(module, showcase);
+```
+
+This package introduces the concepts of showcases, rather than stories.
+Showcases provide a consistent and simple API for a developer to create stories.
```
craigbilner: perhaps for future travellers we should mention the motiviation, a la dextrose, decoupling ourselves from storybook removes addon/knobs issues cross platform?

packages/storybook/README.md
```diff
@@ -0,0 +1,113 @@
...
+## Showcase API
+
+* name - a "/" delineated string describing the location of the showcase in the
+  component library. Each "/" moves the showcase one level deeper in the
+  component library tree
+* children - an array of objects containing showcases
+
+### Showcase Object
+
+* type - whether the showcase object is a story or a [decorator](#decorators)
```
craigbilner: I think this would be a little clearer if it just showed the types e.g.

type: story | decorator
name: string
platform: ios | android | native | web

packages/storybook/README.md
colincclark: done

packages/storybook/README.md
colincclark: changed

PR #1249, opened 2018-08-23
docs: add styleguide readme

colincclark:


packages/styleguide/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+The styleguide package contains the shared styles and functionalities which are
+commonly used across Times Components packages and components. The following
+items can be exported as named exports:
+
+```
+import { colours, spacing } from "@times-components/styleguide";
+```
+
+Or, the default export is a method which takes a scale. Scale manages the
```
craigbilner: effectively it's a factory at this point where the constructor gives you back a styleguide built with the given options, `scale` happens to be the first one

PR #1250, opened 2018-08-23
Fix: byline line height

jimhunty:
This tackles both the issue with the random character causing more space and the line height of multiple line bylines across all three platforms.

### iOS
screen shot 2018-08-23 at 11 57 11

### Android
screen shot 2018-08-23 at 11 55 58

### Web
screen shot 2018-08-23 at 11 59 46


packages/article-byline/__tests__/web/__snapshots__/article-byline-style.web.test.js.snap
```diff
@@ -109,3 +109,80 @@ exports[`3. with given styles 1`] = `
   </div>
 </div>
 `;
+
+exports[`4. with a a very long byline 1`] = `
+<style>
+{
+  "IS1": {
```
craigbilner: ...need to squash these too :man_facepalming:

PR #1251, opened 2018-08-23
docs: add svgs readme

colincclark:


packages/svgs/README.md
```diff
@@ -0,0 +1,41 @@
+# SVGs
+
+This package ensures a consistent API for SVGs that are used across Times
+Components. The web version relies on [`svgs`](https://github.com/godaddy/svgs).
+The native version provides its own set of components and utilises `ART` from
+`react-native`. Components available include `G`, `Path`, `Polygon`, `Rect` and
+`Svg`.
+
```
tuncaulubilge: One bit of difference between APIs, that's worth mentioning.
"Opacity behaves different on native and web svgs. In web, fillOpacity and strokeOpacity are applied separately, but native only has one opacity prop that affects the opacity on both fill colours and stroke colours."

colincclark: added

PR #1259, opened 2018-08-23
Feat: UX scaling updates

jimhunty:
Scaling updates specifically to Comments Button and Article Topics
https://nidigitalsolutions.jira.com/browse/REPLAT-3322

### DIFF
--- a/packages/article-topics/package.json
+++ b/packages/article-topics/package.json
@@ -62,7 +62,7 @@
     "webpack-cli": "2.1.4"
   },
   "dependencies": {
-    "@times-components/context": "0.4.6",
+    "@times-components/context": "0.4.8",
     "@times-components/link": "3.0.17",
     "@times-components/styleguide": "3.2.9",
     "@times-components/tracking": "2.1.14",
--- a/packages/article-topics/package.json
+++ b/packages/article-topics/package.json
@@ -62,6 +62,7 @@
     "webpack-cli": "2.1.4"
   },
   "dependencies": {
+    "@times-components/context": "0.4.6",
     "@times-components/link": "3.0.17",
     "@times-components/styleguide": "3.2.9",
     "@times-components/tracking": "2.1.14",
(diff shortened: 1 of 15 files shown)


packages/article-topics/package.json
```diff
@@ -62,6 +62,7 @@
     "webpack-cli": "2.1.4"
   },
   "dependencies": {
+    "@times-components/context": "0.4.6",
```
craigbilner: need to bump

packages/button/src/button.js
```diff
@@ -5,9 +5,8 @@ import transformTitle from "./utils";
 import { propTypes, defaultProps } from "./button-prop-types";
 import styles from "./styles";
 
-const Button = ({ onPress, style, title }) => {
+const Button = ({ onPress, style, textStyle, title }) => {
```
craigbilner: ideally we shouldn't allow arbitrary styling of components and have a concrete api

packages/article-topics/src/article-topics.js
```diff
@@ -1,18 +1,44 @@
...
+import styleguide from "@times-components/styleguide";
 import ArticleTopic from "./article-topic";
 import styles from "./styles";
 import propTypes from "./article-topic-prop-types";
 
 const { style: ViewPropTypesStyle } = ViewPropTypes;
 
+const renderArticleTopics = (topics, onPress, fontStyle) =>
+  topics.map(({ name, slug }) => (
+    <ArticleTopic
```
craigbilner: same re:style api on `Button`

jimhunty: @craigbilner This is following the change in technical direction that was used in Key Facts. Do you have a suggestion? Im not a fan of the above approach.

craigbilner: :speaking_head: @tuncaulubilge ideally a component shouldn't be arbitrarily styled and instead look something like this if possible:

```
<Component
  fontWeight={400}
/>

// component.js

styles=[styles.componentBaseStyles, {fontWeight}]
```

jimhunty: So for the above example, we would have a separate prop for font size and line height?

PR #1260, opened 2018-08-23
feat: add new fields to article query

nehasri89:
This PR adds the new fields which will allow the render to construct the canonical url in render


craigbilner:
merged

PR #1271, opened 2018-08-28
feat: add markup support for `emphasis` and `strong`.

Belema:
This feature is an update to the markup package so that is can support the content elements,
- `emphasis`,
- `strong`,

that will soon be returned by the Times Public API, once Methode Parser is updated ([#15](https://github.com/newsuk/times-methode-parser/pull/15)).


L0wry:
I'm looking at the android dextrose build

PR #1273, opened 2018-08-28
feat: add ssr package

craigbilner:
* feat: made schema and fragment-matcher "live" to pick up on bad queries sooner. This will be even more important when the mocking PR drops to also fail tests and stories. This might be a bit disruptive (or hugely beneficial?) so might have to see "how it goes"

* feat: add SSR package. Not published, dev only feature to test our SSR. Also provides example code for a basic SSR implementation of the components and client-side bundling. Could be shared/published in the future for use by render/others.

* refactor: move schema and fragment matcher into their own package for maintainability and to optimise the bundle. We're currently shipping the whole of the `schema.json` in our client bundle but don't need it.

* refactor: move the timers to `test-utils` which didn't exist when they were created

* chore: upgrade `node-fetch`

* chore: eslint needs to treat `webpack` as a dev dep

* chore: add missing `webpack-configurator` dep


packages/schema/fetch-gql-schema.js
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable no-console */
+
+const { introspectionQuery } = require("graphql");
```
colincclark: I think this package deserves to have some tests

packages/ssr/author-profile.js
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable import/no-unresolved */
+
+const React = require("react");
```
colincclark: would this package still work with es6 imports as used elsewhere?

packages/topic/src/styles/responsive.web.js
```diff
@@ -8,24 +8,45 @@ import {
   spacing
 } from "@times-components/styleguide";
 
-export const getHeadContainer = ({ hasDescription }) => styled(View)`
+export const HeadContainerWithDescription = styled(View)`
```
colincclark: could use a github issue to ensure we are using styled components correctly across all packages? I know `slice` in particular uses a lot of functions which return styled components

PR #1280, opened 2018-08-29
fix: ssr publish

craigbilner:
For a package to be ignored by the lerna npm publish step, it needs to be set to `private`. The `publish.ignoreChanges` glob is to simply ignore that package for the purposes of change detection.

fix: revert testing change
chore: update out of date script flags

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -5,7 +5,7 @@
   "useWorkspaces": true,
   "command": {
     "publish": {
-      "ignoreChanges": ["ssr/"]
+      "ignoreChanges": ["packages/ssr/**"]
     }
   }
 }
--- a/lerna.json
+++ b/lerna.json
@@ -1,11 +1,11 @@
 {
-  "lerna": "3.0.0-rc.0",
+  "lerna": "3.2.1",
   "packages": ["packages/*", "android-app", "ios-app"],
   "version": "independent",
   "useWorkspaces": true,
   "command": {
     "publish": {
-      "ignoreChanges": ["packages/ssr"]
+      "ignoreChanges": ["ssr/"]
     }
   }
 }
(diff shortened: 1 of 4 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/022505c01ca4cb7c604b07c69c41b026be611d89/index.html

packages/ssr/package.json
```diff
@@ -52,5 +52,8 @@
     "react-native": "0.55.4",
     "react-native-web": "0.5.1",
     "styled-components": "3.4.0"
+  },
+  "publishConfig": {
+    "access": "private"
```
colincclark: this is not a valid value for `access` (https://docs.npmjs.com/misc/config#access). Can we just set `private: true` in here?

lerna.json
```diff
@@ -1,11 +1,11 @@
 {
-  "lerna": "3.0.0-rc.0",
+  "lerna": "3.2.1",
   "packages": ["packages/*", "android-app", "ios-app"],
   "version": "independent",
   "useWorkspaces": true,
   "command": {
     "publish": {
-      "ignoreChanges": ["packages/ssr"]
+      "ignoreChanges": ["ssr/"]
```
colincclark: lerna docs say the glob is always related to the location of the `lerna.json` file, so should this not remain as `packages/ssr`?

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/bce8be3fa514dc1041c2dfb7ca9b391ff4e9de83/index.html

L0wry:
web dextrose should pass on this re-run

PR #1284, opened 2018-08-31
Fix: Label title should not be required field

aronshamash:
- `Article-summary` label proptypes update to stop warnings in client.

### DIFF
--- a/packages/article-summary/src/article-summary.js
+++ b/packages/article-summary/src/article-summary.js
@@ -82,7 +82,7 @@ const ArticleSummary = props => {
 
 ArticleSummary.propTypes = {
   labelProps: PropTypes.shape({
-    title: PropTypes.string.isRequired,
+    title: PropTypes.string,
     color: PropTypes.string,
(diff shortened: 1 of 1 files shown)


packages/article-summary/src/article-summary.js
```diff
@@ -82,7 +82,7 @@ const ArticleSummary = props => {
 
 ArticleSummary.propTypes = {
   labelProps: PropTypes.shape({
-    title: PropTypes.string.isRequired,
+    title: PropTypes.string,
```
colincclark: now it will need a `defaultProp` of `null`

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/dd702d724a307f52a3f72a2a112ae37520d18c12/index.html

packages/article-summary/src/article-summary.js
aronshamash: labelProps is set to default to `null` already

PR #1286, opened 2018-08-31
feat: add server-side paging

craigbilner:
The current UX is a bit naff when you navigate to a list item and then hit back in the browser, we effectively broke the web experience :-(

This PR updates history state and provides a server-side implementation

chore: refactor prop-types and the way they were shared. We should always provide a no-js fallback for paging and sharing prop-types made this brittle

### DIFF
--- a/packages/article-list/src/article-list-prop-types.web.js
+++ b/packages/article-list/src/article-list-prop-types.web.js
@@ -1,14 +1,17 @@
 import PropTypes from "prop-types";
-import Pagination from "@times-components/pagination";
 import {
   propTypes as basePropTypes,
   defaultProps
 } from "./article-list-prop-types-base";
 
 export const propTypes = {
-  adConfig: PropTypes.shape({}).isRequired,
   ...basePropTypes,
-  ...Pagination.propTypes
+  adConfig: PropTypes.shape({}).isRequired,
+  count: PropTypes.number,
+  onNext: PropTypes.func,
+  onPrev: PropTypes.func,
+  page: PropTypes.number,
+  pageSize: PropTypes.number
 };
 
 export { defaultProps };
--- a/packages/related-articles/__tests__/web/__snapshots__/la2-with-style.web.test.js.snap
+++ b/packages/related-articles/__tests__/web/__snapshots__/la2-with-style.web.test.js.snap
@@ -166,13 +166,6 @@ exports[`default styles 1`] = `
 
 <style>
 {
-  "IS1": {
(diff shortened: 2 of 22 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/87e2e1915a774f7337fa234969ae2075b31a3f3c/index.html

packages/related-articles/__tests__/web/__snapshots__/la2-with-style.web.test.js.snap
```diff
@@ -166,13 +166,6 @@ exports[`default styles 1`] = `
```
jimhunty: Why are the snapshots for related articles changed?

craigbilner: There were some more SSR `StyleSheet` fixes for label

packages/ssr/package.json
```diff
@@ -7,9 +7,9 @@
     "bundle:dev": "webpack --config=webpack.config.dev.js",
     "bundle:prod": "NODE_ENV=production webpack --config=webpack.config.prod.js -p",
     "bundle:profile": "webpack --config=webpack.config.prod.js --profile --json > dist/stats.json",
-    "depcheck": "depcheck --ignores='eslint,prettier,webpack*' --skip-missing",
+    "depcheck": "depcheck --ignores='depcheck,eslint,prettier,webpack*' --skip-missing",
     "fmt": "eslint . --fix && prettier --write '**/*.*'",
-    "lint": "eslint . && yarn prettier:diff && yarn depcheck && jest-lint",
```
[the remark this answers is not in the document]
craigbilner: I think we need linting for SSR still?

craigbilner:
@jimhunty atm SSR is just for dev purposes so in itself is a "showcase" and doesn't need testing per se. However, should it become "new render", then we can backfill it's responsibility and start publishing it as such?

packages/article-list/src/article-list-prop-types.web.js
```diff
@@ -1,14 +1,17 @@
...
 import {
   propTypes as basePropTypes,
   defaultProps
 } from "./article-list-prop-types-base";
 
 export const propTypes = {
-  adConfig: PropTypes.shape({}).isRequired,
   ...basePropTypes,
-  ...Pagination.propTypes
+  adConfig: PropTypes.shape({}).isRequired,
```
colincclark: `PropTypes.object.isRequired`?

PR #1287, opened 2018-08-31
Add slicename to relatedArticleSlice query

tuncaulubilge:
SliceName was missing in the query. Added it in so related articles can be rendered when using the provider.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/0a4d34fdeb7d79a12397ec1d76560c387c39ecb6/index.html

PR #1292, opened 2018-09-04
chore: update markup > tag-mixture in storybook

Belema:
I noticed that I forgot to put an emphasis tag in the storybook. This is to fix this oversight.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/e86d7ba37496c311f03f400e31ebcbfe7106bb46/index.html

PR #1299, opened 2018-09-06
BREAKING CHANGE: split out high and low res images

### LINKED ISSUE #1173: Card should not be involved with image url manipulation

Refactor card so that the consumer of any images passed into card manage the image url, and card does not.

craigbilner:
While relatively large, essentially this PR changes our component APIs to allow more fine grained control over how we handle images and push the image optimisations into the image components. The consumers of the `Image` component still need to decide what `high resolution` and `low resolution` images mean to them.

In our current working example, `ArticleList` decides to load every image with a low resolution size of `100` and only the width of the page for list items that are in the viewport if the browser supports the `IntersectionObserver`. This allows those browsers to provide a faster experience on poor networks/devices. For browsers without an `IntersectionObserver`, the low/high resolution images are still rendered, but for the whole page rather than what is in the viewport.

A future enhancement is also needed to make the `highRes` decision above even more intelligent rather than just asking for a massive image when in reality the list has a small image left-aligned.

This PR highlights areas which are currently not providing an optimal image experience such as `Article`, `Related Articles` and `Native`. Future PRs should address these points.

In addition some UX improvements have been made to make the list pages "less flashy".

## tl;dr

* make high and low res images first class (make it obvious article/related articles/native don't respect it etc.) (edited)
* allow web images to use desired size to provide retina sizes etc in the future
* smooth transition of text and images with no "flashes"
* remove url logic from Card
* standardardise uri (edited)
* fix browsers with no IntersectionObserver

## changes

BREAKING CHANGE: split out high and low res images
refactor: split out fade-in effect
chore: fix meta
fix: no observer SSR

close #1173


packages/card/__tests__/android/__snapshots__/card-with-style.android.test.js.snap
```diff
@@ -23,7 +23,10 @@ exports[`1. card default state 1`] = `
   >
     <Image
       aspectRatio={0.6666666666666666}
-      uri="https://img.io/img&resize=360"
+      fadeImageIn={false}
```
colincclark: strangely, card is not showing coverage of tests in the CLI so cannot see what it is on this branch

packages/image/__tests__/web/__snapshots__/image.web.test.js.snap
```diff
@@ -1,69 +1,299 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`1. default layout 1`] = `
+.c0 {
```
colincclark: odd, cannot see image coverage either

packages/article-list/src/article-list.web.js
```diff
@@ -43,6 +44,19 @@ class ArticleList extends Component {
     );
   }
 
+  componentDidMount() {
+    const newState = {
+      fadeImageIn: true
+    };
+
+    if (typeof window !== "undefined" && !this.observer) {
+      newState.fallback = normaliseWidth(window.clientWidth);
```
colincclark: presumably, the idea is to split out lazy loading to another package at a later date still?

packages/related-articles/src/related-article-item.js
```diff
@@ -59,11 +59,12 @@ const RelatedArticleItem = ({
     <Link onPress={e => onPress(e, { url: article.url })} url={url}>
       <Card
         contentContainerClass={contentContainerClass}
-        image={{ uri: imageUri }}
         imageContainerClass={imageContainerClass}
         imageRatio={imageRatio}
         imageStyle={imageStyle}
+        imageUri={imageUri}
         isReversed={isReversed}
+        lowResSize={100}
```
colincclark: because related articles does not lazy load, the images are low res and blurry on the article page :(

colincclark:
Native seems OK, but I am not sure what a related article image would look like at the bottom of an article page.

packages/article-list/src/article-list.web.js
craigbilner: yep, will pick this up after mocking :+1:

packages/card/__tests__/android/__snapshots__/card-with-style.android.test.js.snap
craigbilner: I run `yarn test:android --coverage`? You have to use the `--no-cache` flag if you have previous jest caches for that package

packages/related-articles/src/related-article-item.js
craigbilner: yep, was done on purpose to make sure this gets picked up with the article work. We should be fetching high res images on mobweb as they'll just be the placeholders otherwise and/or waste bandwidth if never look at

PR #1308, opened 2018-09-08
chore: bump react-native-web to 0.8.9

craigbilner:
Button is currently broken in production after the last RNW upgrade. This upgrade appears to fix it, we need to bring in a "clicky test" to prevent this in the future.

## tl;dr

* `react-art` is a peer dep of RNW
* colours have changed to `rgba`
* some `!importants` were removed :-)
* the `babel-plugin-react-native-web` isn't really needed at each package level but should still be used to make the bundle smaller by the "tree shaking faking"
* a new `data-focusable` attr is dumped on everything which has been removed from the snaps for clarity
* the RNW aliasing changed to `react-native$`
* RNW has moved to es modules, need to specify `commonjs` in babel config for it to work with SSR

## future

In the same vain as RNW moving to esm we should also "es module everywhere" to get the benefits of tree shaking. It's still relatively painful with webpack and the way we do `externals` to keep our bundles "shallow" so going to look at rollup instead.


colincclark:
Closing this one until Q2 as per Adam Francis - will revisit in a few weeks

PR #1311, opened 2018-09-10
fix: replace list button with text link

colincclark:
- this was due to RNW not working with the `TouchableHighlight` component in our `Button` package. We opted for the small design tweak rather than implementing a RNW version bump.
- we moved the link package mock to the native tests which needed it to allow the web tests that used a `TextLink` component from the same package to work


tuncaulubilge:
I think we should keep using the button package. What about creating a `button.web` which is basically the same component without a wrapping touchableHighlight?

colincclark:
@tuncaulubilge @jimhunty @aronshamash I made an alternative approach here: https://github.com/newsuk/times-components/pull/1312

PR #1319, opened 2018-09-12
fix: android ad id

tuncaulubilge:
- Use mobile ad units on apps.
- Use networkId passed by the app, based on the build flavour


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/test/index.html

PR #1320, opened 2018-09-13
feat: subscript and superscript markup

tuncaulubilge:
This PR adds support for subscript and superscript

- For web, it uses `<sub>` and `<sup>` tags
- For native, we use special unicode characters for supported characters. If any of the characters in the markup is not supported, we fallback to a textview with a smaller font. 

For supported unicode characters: https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/613d553dc5fc0a5d610e0adbb8e846af1b64802c/index.html

PR #1325, opened 2018-09-14
fix: Remove unused dependency from TimesReactIOS podspec

mentrena:
React as imported by the iOS project must have changed and fails to satisfy this dependency, so we cannot install any cocoa pods until we fix this podspec.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/a4363070d99a558b2c5d5f0861ada9b29cd8a9eb/index.html

PR #1326, opened 2018-09-14
fix: Use whatwg-url for creating urls in android

tuncaulubilge:
React-native Android is overwriting URL class, which causes an error in our image component. 

Applied the following fix to polyfill URL using whatwg-url: https://github.com/facebook/react-native/issues/16434#issuecomment-392147974


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/c8d7eddadd4463af1696918c70398e740b859f73/index.html

PR #1335, opened 2018-09-17
Use url-polyfill instead of whatwg-url

tuncaulubilge:
The url polyfill we used for Android seem to only work when js debug is turned on, but crashing on release builds. Replaced with "url-polyfill" instead. Also added the same polyfill to storybook


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/a1a604c96e338a7ebb89f1ac4abaf9d4b9f5528b/index.html

PR #1337, opened 2018-09-18
feat: Style video error screen

aronshamash:
- Use poster image when avaliable
- Retry button just refreshes the page

### Before

 
screen shot 2018-09-18 at 10 44 46

### After

screen shot 2018-09-18 at 10 45 02

### DIFF
--- a/packages/button/src/styles/index.web.js
+++ b/packages/button/src/styles/index.web.js
@@ -6,19 +6,19 @@ const styles = {
   button: {
     alignItems: "center",
     backgroundColor: colours.functional.action,
-    borderRadius: 2,
     borderColor: colours.functional.action,
+    borderRadius: 2,
     color: colours.functional.white,
     cursor: "pointer",
-    height: 45,
-    justifyContent: "center",
-    minWidth: 100,
     ...fontFactory({
       font: "supporting",
       fontSize: "button"
     }),
-    paddingTop: 4,
+    height: 45,
+    justifyContent: "center",
     lineHeight: 0,
+    minWidth: 100,
+    paddingTop: 4,
     width: 160
   }
 };
--- a/packages/button/src/styles/index.web.js
+++ b/packages/button/src/styles/index.web.js
@@ -7,6 +7,7 @@ const styles = {
     alignItems: "center",
     backgroundColor: colours.functional.action,
     borderRadius: 2,
+    borderColor: colours.functional.action,
     color: colours.functional.white,
     cursor: "pointer",
     height: 45,
--- a/packages/video/src/styles/index.web.js
+++ b/packages/video/src/styles/index.web.js
@@ -1,33 +1,69 @@
 import { StyleSheet } from "react-native";
-import { fonts, spacing } from "@times-components/styleguide";
+import {
+  colours,
+  fonts,
+  fontFactory,
+  spacing
+} from "@times-components/styleguide";
 import sharedStyles from "./shared";
 
 const styles = StyleSheet.create({
   ...sharedStyles,
+  heading: {
+    color: colours.functional.contrast,
+    ...fontFactory({
+      font: "headline",
+      fontSize: "infoTitle"
+    }),
+    textAlign: "center",
+    width: "auto",
+    height: "auto",
+    marginBottom: spacing(2)
+  },
   body: {
     fontFamily: fonts.body,
-    fontSize: "14px",
     textAlign: "center",
-    color: "rgba(255, 255, 255, 0.8)",
-    width: 410,
+    color: colours.functional.contrast,
+    ...fontFactory({
+      font: "supporting",
+      fontSize: "secondary"
+    }),
+    lineHeight: 25,
+    opacity: 0.7,
     height: "auto",
-    maxWidth: "80%"
+    maxWidth: 285
   },
   background: {
     display: "flex",
     alignItems: "center",
     justifyContent: "center",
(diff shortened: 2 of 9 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/07136266b09ae83ecb859fc3d6949d8183f96001/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/888c4e1f530574c75ed18c14a16310ca92e8da5d/index.html

packages/video/src/video-error.web.js
```diff
@@ -1,21 +1,53 @@
...
+          title="Retry"
+        />
+      </View>
+    </View>
+  );
+};
 
 VideoError.propTypes = {
   height: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
+  poster: PropTypes.shape({ uri: PropTypes.string.isRequired }),
```
jimhunty: Is it likely other properties might be added?

packages/video/src/video-error.web.js
```diff
@@ -1,21 +1,53 @@
...
   width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired
 };
 
+VideoError.defaultProps = {
+  poster: ""
```
jimhunty: Doesnt match the prop type

packages/video/src/video-error.web.js
```diff
@@ -1,21 +1,53 @@
...
+    );
+  };
+
+  return (
+    <View style={[styles.background, { width, height }]}>
+      <View style={styles.posterContainer}>{renderPosterImage()}</View>
+      <View style={styles.textContainer}>
+        <Text style={styles.heading}>Something&apos;s gone wrong</Text>
+        <Text style={styles.body}>
+          Please check your network connection and retry to try again
```
jimhunty: Sentence needs work.

packages/video/src/video-error.web.js
```diff
@@ -1,21 +1,53 @@
...
+        </Text>
+        <Button
+          onPress={() => {
+            if (typeof window !== "undefined") {
```
jimhunty: Should this wrap the button and we change the message based on this?

packages/video/src/video-error.web.js
aronshamash: Changed to just be a string. Originally was an object to be consistent with `video-prop-types.js`

packages/video/src/video-error.web.js
aronshamash: So this is a compromise situation. This screen could display if it is an network error or an invalid id which refreshing will not help. We would need to do more work to be able to differentiate between different types of error. We need some wording which makes sense either for most common errors or for all error types. Suggestions as to better copy?

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/05e75c4b7751bde68884122b4eecbdefb3589447/index.html

packages/video/src/styles/index.web.js
```diff
@@ -1,33 +1,69 @@
...
-    color: "white",
-    width: "auto",
-    height: "auto",
-    marginBottom: spacing(2)
+  textContainer: {
+    position: "absolute",
+    maxWidth: "80%"
+  },
+  posterContainer: {
+    filter: "blur(8px)",
```
craigbilner: :cake:

packages/video/src/video-error.web.js
colincclark: > Please check your network connection

with a retry button should suffice?

packages/button/src/styles/index.web.js
```diff
@@ -7,6 +7,7 @@ const styles = {
     alignItems: "center",
     backgroundColor: colours.functional.action,
     borderRadius: 2,
+    borderColor: colours.functional.action,
```
colincclark: alphabetty spaghetti

packages/video/src/video-error.web.js
colincclark: it seems to be an object though? https://github.com/newsuk/times-components/pull/1337/files#diff-7f35ad8aff234a89efbd14e42fd7171eR16

packages/video/src/video-error.web.js
```diff
@@ -1,21 +1,58 @@
...
+      />
+    );
+  };
+
+  const renderRetryButton = () => {
+    if (typeof window === "undefined") return null;
+    return (
+      <Button
+        onPress={() => {
+          window.location.reload();
```
colincclark: why doesn't this tie in with the API call and re-try that instead?

packages/video/src/video-error.web.js
```diff
@@ -1,21 +1,58 @@
...
-const VideoError = ({ height, width }) => (
-  <View style={[styles.background, { width, height }]}>
-    <Text style={styles.heading}>Video unable to play</Text>
-    <Text style={styles.body}>
-      Please check your network connection and try refreshing the page. If that
-      doesn&apos;t work, please try again later
-    </Text>
-  </View>
-);
+const VideoError = ({ height, width, poster }) => {
```
colincclark: there are no changes to the tests or the showcase that passes in this new `poster` prop

PR #1341, opened 2018-09-19
fix: article list inline ad with adblocker keyline bug

aronshamash:
- Fixes Article list inline ad with adblocker keyline bug

### Before
before

### After
after

### DIFF
--- a/packages/article-list/src/styles/index.web.js
+++ b/packages/article-list/src/styles/index.web.js
@@ -20,7 +20,8 @@ const styles = StyleSheet.create({
     borderColor: colours.functional.keyline,
     borderTopWidth: 1,
     borderBottomWidth: 1,
-    paddingVertical: spacing(2)
+    paddingVertical: spacing(2),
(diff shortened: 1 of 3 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/82dcd13435cb164084334266e994bfb716941ad3/index.html

packages/article-list/src/styles/index.web.js
```diff
@@ -20,7 +20,8 @@ const styles = StyleSheet.create({
     borderColor: colours.functional.keyline,
     borderTopWidth: 1,
     borderBottomWidth: 1,
-    paddingVertical: spacing(2)
+    paddingVertical: spacing(2),
+    top: 1
```
jimhunty: Feels a bit hacky?
Also should it be `StyleSheet.hairlineWidth` ?

PR #1343, opened 2018-09-20
fix: Split article view event from track event

tuncaulubilge:
Part of the android analytics fix (REPLAT-1706). Splitted article page view event from track event so the native app can treat these events separately.

### DIFF
--- a/android-app/pages/article.js
+++ b/android-app/pages/article.js
@@ -5,7 +5,7 @@ import { Article } from "@times-components/pages";
 
 const config = NativeModules.ReactConfig;
 const { fetch } = NativeModules.NativeFetch;
-const { onArticleView, track } = NativeModules.ReactAnalytics;
+const { track } = NativeModules.ReactAnalytics;
 const {
   onArticlePress,
   onArticleLoaded,
@@ -45,7 +45,7 @@ const ArticleView = ({
   return (
     <ArticlePageView
       articleId={articleId}
-      analyticsStream={(event) => {
+      analyticsStream={event => {
         if (event.object === "Article" && event.action === "Viewed") {
           onArticleLoaded(event.attrs.articleId, event);
         } else {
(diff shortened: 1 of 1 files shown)


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/1b3d047f72f95fb9ce5dd8226332dacd7cccad74/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/c4325ebb314ca80dcf35f6f33717c5a6d33d9c3f/index.html

tuncaulubilge:
Will reopen when native counterparts are ready

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/7e6c406ae997755f9c4d89d6d436035db9048e95/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/7e6c406ae997755f9c4d89d6d436035db9048e95/index.html

android-app/pages/article.js
```diff
@@ -44,7 +45,13 @@ const ArticleView = ({
   return (
     <ArticlePageView
       articleId={articleId}
-      analyticsStream={track}
+      analyticsStream={event => {
+        if (event.object === "Article" && event.action === "Viewed") {
```
craigbilner: everything seems on the `event` already so why does it need the separate callback?

tuncaulubilge: Native doesn't know what's inside event and doesn't need to know other than this particular case. It'd be better to keep knowledge on times-components about the contents of the event object.

tuncaulubilge: Also, these two functions come from separate native modules, which makes our life a bit easier on native-land

craigbilner: I see, ta

PR #1346, opened 2018-09-20
fix: Change object name so that it works

jimhunty:
Object naming for Interactives in TC compared to TPA didn't match. This fixes the integration of Interactives.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/41d30994afdd0380c282490703c2e2768b36e93e/index.html

PR #1348, opened 2018-09-24
fix: storybooks

tuncaulubilge:
Native storybooks are broken because of two issues, this PR fixes both:

- Graphql transpilation problems. Added mjs to list of extensions: (see https://github.com/graphql/graphql-js/issues/1272)
- Lazy load stories are converted to web only

### DIFF
--- a/.storybook.native/webpack.haul.storybook.js
+++ b/.storybook.native/webpack.haul.storybook.js
@@ -24,7 +24,7 @@ module.exports = ({ platform }, { module, resolve, plugins }) => ({
   },
   resolve: {
     ...resolve,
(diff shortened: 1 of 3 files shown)


.storybook.native/webpack.haul.storybook.js
```diff
@@ -15,11 +15,16 @@ module.exports = ({ platform }, { module, resolve, plugins }) => ({
...
+        include: /node_modules/,
+        test: /\.mjs$/,
+        type: "javascript/auto",
+      },
     ]
   },
   resolve: {
     ...resolve,
-    extensions: [`.${platform}.js`, ".native.js", ".js"],
+    extensions: [`.${platform}.js`, ".native.js", ".js", ".webpack.js", ".web.js", ".mjs", ".js"],
```
craigbilner: .web?

PR #1349, opened 2018-09-24
fix: Ads not loading - Enqueue window post messages until it's ready

tuncaulubilge:
window.postMessage doesn't work until the webView is loaded. Our previous fix for this was to add a timeout to postMessages, but that's not 100% reliable. Implemented a separate approach where the window enqueues messages until post message is ready:

Based on: https://github.com/facebook/react-native/issues/11594#issuecomment-274689549

### DIFF
--- a/packages/ad/src/utils/webview-event-callback-setup.js
+++ b/packages/ad/src/utils/webview-event-callback-setup.js
@@ -34,7 +34,6 @@ const webviewEventCallbackSetup = options => {
   }
 
   window.eventCallback = (type, detail) => {
-    // delay until next frame as React Native message bridge is not set up immediately
     window.postMessage(
       JSON.stringify({
         isTngMessage: true,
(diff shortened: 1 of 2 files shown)


packages/ad/src/utils/webview-event-callback-setup.js
```diff
@@ -5,19 +5,43 @@
...
+      set(fn) {
+        currentPostMessageFn = fn;
+        isReactNativePostMessageReady = true;
+        window.setTimeout(sendQueue, 0);
+      }
+    });
+  }
+
   window.eventCallback = (type, detail) => {
     // delay until next frame as React Native message bridge is not set up immediately
```
craigbilner: can remove the comment

packages/ad/src/utils/webview-event-callback-setup.js
```diff
@@ -5,19 +5,43 @@
...
-    window.setTimeout(() => {
-      const method = window.reactBridgePostMessage || window.postMessage;
-      method(
-        JSON.stringify({
-          isTngMessage: true,
-          type,
-          detail
-        }),
-        "*"
```
craigbilner: what was magic asterix for?

tuncaulubilge: RN overrides the method and adds a second argument called "origin". Now we are just using the default postMessage function :+1:

PR #1350, opened 2018-09-24
feat: Make article text selectable

tuncaulubilge:
Adds selectable prop to article text to allow selections and copy-paste

copy-paste


packages/article/__tests__/images-with-style.web.js
```diff
@@ -24,7 +24,8 @@ const styles = [
   "marginBottom",
   "marginTop",
   "paddingBottom",
-  "paddingTop"
+  "paddingTop",
+  "selectable"
```
craigbilner: sweet hack....can you remove the minimalise serializer and customise it separately instead?

tuncaulubilge: Looks like I added this here by mistake. It's already being filtered by the minimalise

PR #1356, opened 2018-09-25
chore: bump react native web to 0.9.0

craigbilner:
`haul` needs `react` `16.5.2` which means we need to go to `react-native-web` `0.9.0` and latest `enzyme`. They needed to be upgraded anyway....

### highlights

* don't need `babel-plugin-react-native-web` in every package
* colours have moved to rgba with a as a float
* some of the other styles are tweaked and less important
* the `moduleNameMapper` key changed slightly
* the react-native-web plugin isn't needed for SSR - rnw has moved to esm and now need to force into cjs. Will address this with the bundling stories as well where we can probably get smaller bundles by using esm throughout for client bundle and cjs for SSR (and IE11)
* enzyme doesn't render `React.Fragment` anymore

tl;dr 

chore: update snaps
chore: remove prop
chore: fix commonjs


packages/article-flag/package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "prettier:diff": "prettier --list-different '**/*.*'",
     "fmt": "eslint . --fix && prettier --write '**/*.*'",
-    "depcheck": "depcheck --ignores='@thetimes/jest-lint,babel-*,depcheck,eslint,jest,prettier,webpack*' --ignore-bin-package=false --skip-missing",
+    "depcheck": "depcheck --ignores='@thetimes/jest-lint,babel-*,depcheck,eslint,jest,prettier,react-art,webpack*' --ignore-bin-package=false --skip-missing",
```
colincclark: react-art does not seem to be in the deps?

packages/ssr/webpack.config.dev.js
```diff
@@ -1,7 +1,7 @@
 const path = require("path");
 const webpack = require("webpack");
 
-const alias = { "react-native": "react-native-web" };
+const alias = { "react-native$": "react-native-web" };
```
colincclark: what does this do? and what is the difference between `react-native$` and `^react-native$`?

packages/article-flag/package.json
craigbilner: RNW should have been a dev dep, added, ta

packages/article-summary/__tests__/web/__snapshots__/article-summary-with-style.web.test.js.snap
```diff
@@ -18,23 +18,27 @@ exports[`1. article summary component with a single paragraph 1`] = `
     "color": "inherit",
     "font-family": "inherit",
     "font-size": "inherit",
+    "font-weight": "inherit",
+    "line-height": "inherit",
```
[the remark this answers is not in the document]
craigbilner: Raised: https://github.com/newsuk/times-components/issues/1367

packages/ssr/webpack.config.dev.js
craigbilner: best reference I could find was [this](https://github.com/necolas/react-native-web/issues/961#issuecomment-391164614)

It's just a regex and I'm guessing the `moduleNameMapper` and `alias` are slightly different without looking into it

PR #1359, opened 2018-09-26
fix: Remove hairline width

jimhunty:
Jake has requested that we remove hairline width and go with `1px` widths.


packages/article/src/article-comments/styles/shared.js
```diff
@@ -1,12 +1,11 @@
-import { StyleSheet } from "react-native";
 import styleguide from "@times-components/styleguide";
 
 const { colours, fontFactory, spacing } = styleguide();
 const styles = {
   container: {
     alignItems: "center",
     borderTopColor: colours.functional.keyline,
-    borderTopWidth: StyleSheet.hairlineWidth,
+    borderTopWidth: 1,
```
nehasri89: minor, should we do spacing(1) instead?

jimhunty: That would make it 5px. Could do `spacing(0.2)` but I think we should leave it.

PR #1362, opened 2018-09-26
WIP: Feature/Environment specific article canonical urls on listing pages

nehasri89:
This PR introduces the concept of environment specific URLs from the listing pages. 
TC will now not rely on absolute URL's `url` field from TPA but will construct its own env specific canonical URL based on the fields `slug` and `shortIdentifier`.


packages/article-list/__tests__/android/__snapshots__/article-list.android.test.js.snap
```diff
@@ -278,6 +279,7 @@ Array [
             "name": "paragraph",
           },
         ],
+        "slug": "This is slug 1",
```
Belema: This does not look like a slug

packages/article-list/src/article-list-item.web.js
```diff
@@ -34,7 +34,7 @@ const ArticleListItem = props => {
     shortSummary,
     showImage,
     summary,
-    url
+    canonicalUrl
```
colincclark: can these remain alphabetised and therefore easier to read?

packages/article-list/src/article-list-prop-types.web.js
```diff
@@ -11,7 +11,8 @@ export const propTypes = {
   onNext: PropTypes.func,
   onPrev: PropTypes.func,
   page: PropTypes.number,
-  pageSize: PropTypes.number
+  pageSize: PropTypes.number,
+  makeUrl: PropTypes.func
```
colincclark: a->z

packages/article-list/src/article-list.web.js
```diff
@@ -124,6 +125,11 @@ class ArticleList extends Component {
             data.map((article, index) => {
               const { elementId } = article;
 
+              const canonicalUrl = makeUrl(
```
colincclark: can this API be an object instead of seperate arguments, it is easier to use....?

packages/article-list/__tests__/utils.js
```diff
@@ -19,3 +19,18 @@ export const omitNative = new Set([
...
 export const omitWeb = new Set(["className", "data-testid", "style"]);
+
+const config = {
+  get() {
+    return "www.thetimes.co.uk";
+  }
+};
+
+export const makeUrl = (slug, shortIdentifier) => {
+  if (process.env.NODE_ENV === "local" && !process.env.IS_E2E_CI) {
```
colincclark: what is `process.env.IS_E2E_CI`? where is it set?

nehasri89: this is set for CI

colincclark: does CI set the `process.env.NODE_ENV` to `local`? If not a simple `isLocal` check could be added here, and used in a template literal instead

packages/article-list/src/article-list.web.js
colincclark:
```
const { elementId, slug, shortIdentifier } = article;

makeUrl({slug, shortIdentifier});
```

packages/article-list/__tests__/utils.js
nehasri89: Please refer to this https://github.com/newsuk/cps-content-render/blob/f2e2802640484d1d624f75dadb153e491613e7f4/src/test/e2e/wdio.conf.js#L264

colincclark: so can we add an `isDevEnvironment` variable that is used in a template literal to either add the port or not instead of the if statement?

packages/article-list/__tests__/utils.js
```diff
@@ -19,3 +19,18 @@ export const omitNative = new Set([
...
 
 export const omitWeb = new Set(["className", "data-testid", "style"]);
+
+const config = {
+  get() {
+    return "www.thetimes.co.uk";
+  }
+};
+
+export const makeUrl = ({ slug, shortIdentifier }) => {
```
craigbilner: this should probably just return `/foo/bar/${slug}/${shortIdentifier}`? The envars aren't set and it doesn't need to know about render?

packages/article-list/src/article-list-prop-types.web.js
```diff
@@ -8,6 +8,7 @@ export const propTypes = {
   ...basePropTypes,
   adConfig: PropTypes.shape({}).isRequired,
   count: PropTypes.number,
+  makeUrl: PropTypes.func,
```
craigbilner: required?

packages/article-list/src/article-list-prop-types-base.js
```diff
@@ -19,5 +19,6 @@ export const defaultProps = {
   error: null,
   isLoading: true,
   pageSize: 10,
-  receiveChildList: () => {}
+  receiveChildList: () => {},
+  makeUrl: () => {}
```
colincclark: if it is required, no default is needed

packages/article-list/__tests__/utils.js
```diff
@@ -19,3 +19,5 @@ export const omitNative = new Set([
 ]);
 
 export const omitWeb = new Set(["className", "data-testid", "style"]);
+
+export const makeUrl = ({ slug, shortIdentifier }) => `www.thetimes.co.uk/article/${slug}-${shortIdentifier}`;
```
colincclark: does this need a protocol?

packages/article-list/__tests__/android/__snapshots__/article-list.android.test.js.snap
craigbilner: [this is not a pipe](https://en.wikipedia.org/wiki/The_Treachery_of_Images#/media/File:MagrittePipe.jpg)

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/6736cdc93cf7a16ca3ab98da027378fbacf73163/index.html

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/cbf2d990cb7a9f3a7d3203e064e64def56d66b5a/index.html

craigbilner:
WIP

PR #1366, opened 2018-09-27
fix: article flag layout

colincclark:
Part of https://nidigitalsolutions.jira.com/browse/REPLAT-3472

NB: margins here are for layout _vis a vis_ font, rather than vertical rhythm, hence, did not use `spacing`

### DIFF
--- a/packages/article-flag/src/style/shared.js
+++ b/packages/article-flag/src/style/shared.js
@@ -7,8 +7,8 @@ const styles = {
     alignItems: "center"
   },
   diamond: {
-    marginRight: 5,
-    marginBottom: 1
+    marginBottom: -1,
+    marginRight: 5
   },
   title: {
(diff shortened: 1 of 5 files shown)


packages/article-flag/src/style/shared.js
```diff
@@ -7,8 +7,8 @@ const styles = {
     alignItems: "center"
   },
   diamond: {
-    marginRight: 5,
-    marginBottom: 1
+    marginBottom: -1,
```
craigbilner: look a bit hacky, is this an issue with the `svg`?

colincclark: put it back to `0` - seems ok this way, but it is the line-height that is screwing us over, and perhaps the weird font issues due to lack of a standard working font

PR #1371, opened 2018-09-27
fix: Ad padding

jimhunty:
Fixes the ad padding across all three platforms


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/da86d1c67d9dcc6b1a04e79ebff0bf43a7f3848a/index.html

Issue #1375, opened 2018-10-01
fixtures: move to `gql` driven


craigbilner:
**Do you want to request a _feature_ or report a _bug_?**

feature

**What is the current behaviour?**

There are static `json` fixtures which are often wrong and/or become stale with time

**What is the expected behaviour?**

showcases/tests are driven by the underlying `gql` query

**If this is a feature request, what is the motivation or use case for changing
the behaviour?**

Fixtures become "live" and continually represent the actual API that times-components integrates with

nehasri89:
Tools team are piking this up

PR #1376, opened 2018-10-01
refactor: port article fixtures

craigbilner:
Move articles to the new fixture pattern where they're driven by the `gql` query itself. This removes a lot of fixture boilerplate and allows the developer to see showcase/test changes by just changing the query itself.

`article.js` has been added to `provider-test-tools` which provides the magic schema resolvers based on types to abstract away (the rather ugly) index/knowledge of article structure hacks. A consumer can just tap into `makeArticle`, `makeRelatedArticle`, `chooseMedia` or provide it's own Apollo `error`.

The many stories have been aggregated where it made sense. For the storybook to "magically" update when changing the storybook `booleans` a helper had to be created to force a re-render. This is because the new props come in and by having different mock params, React isn't very concerned with re-rendering anything. `ArticleConfigurator` abstract this configuration and uses state to force a re-render.

Removed the mock from the `storybook` `provider` because it's confusing and doesn't fit in with the general pattern of here's mocked showcases here's a real one

There are now a handful of packages left that have their own fixtures, ideally with time they should move over to be driven by pieces of mocks derived from queries too. #1375


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/53e611fabd91bbec1a2d7e173760819ab87d0dbc/index.html

colincclark:
would be good if this PR fixed this story: http://components.thetimes.co.uk/?knob-Size%20of%20ad%20placeholder%3A=default&selectedKind=Helpers%2FProvider&selectedStory=Article&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

packages/provider/provider.showcase.js
```diff
@@ -1,10 +1,10 @@
 import React from "react";
 import { Text } from "react-native";
-import articleFixture from "@times-components/provider-test-tools/fixtures/article";
 import { addTypenameToDocument } from "apollo-utilities";
 import gql from "graphql-tag";
 import {
   authorProfile as makeAuthorParams,
+  fixtures,
```
jimhunty: Why does this differ to the author profile and topic approach?

packages/article/showcase-helper.js
```diff
@@ -0,0 +1,137 @@
...
+  MockedProvider,
+  schemaToMocks
+} from "@times-components/provider-test-tools";
+
+const FLAGS = 1;
+const LABEL = 2;
+const LEAD_ASSET = 4;
+const LINKED_BYLINE = 8;
+const STANDFIRST = 16;
+const VIDEO = 32;
```
jimhunty: This seems a bit odd?

jimhunty:
Similar to @colincclark, It would be good to see all the stories fixed.

packages/article/showcase-helper.js
craigbilner: I've tried to go for something that's clear and nice to extend. If you want to add more toggles to an article you can just extend the flags. It means you don't need to worry about `componentDidUpdate` or `shouldComponentUpdate` for any complex data structures

craigbilner:
@colincclark @jimhunty fixed the story :+1:

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/a0e3914ad3dacb6a7fe4be960e868dca216c8bc6/index.html

PR #1378, opened 2018-10-02
Feature/article canonical urls

nehasri89:
This PR allows Author Profile and Topics articles to have env agnostic URLs. Also RA will have env agnostic URLs too.

There are issues with Context API and enzyme shallow rendering https://github.com/airbnb/enzyme/issues/1647
https://github.com/airbnb/enzyme/pull/1513

So therefore we need to rethink about tests that are failing in Related Articles where we use Context API and Shallow render.


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/1cde2d39aa29f31a7b53323d89008df63127259b/index.html

packages/article-list/__tests__/web/article-list-lazy.web.test.js
```diff
@@ -1,14 +1,17 @@
...
 import {
   addSerializers,
   enzymeRenderedSerializer,
   minimalise
 } from "@times-components/jest-serializer";
+import Context from "@times-components/context";
 import articleListFixture from "../../fixtures/articles.json";
 import adConfig from "../../fixtures/article-ad-config.json";
 import ArticleList from "../../src/article-list";
+import { makeUrl } from "./../utils";
```
tuncaulubilge: could be just "../utils"

packages/article-list/__tests__/shared-states.web.js
```diff
@@ -40,12 +41,14 @@ export default () => {
       test() {
         const apolloError = new ApolloError("Test");
         const testInstance = TestRenderer.create(
-          <ArticleList
-            adConfig={adConfig}
-            emptyStateMessage="Empty State"
-            error={apolloError}
-            refetch={() => {}}
-          />
+          <Context.Provider value={{ makeUrl }}>
```
tuncaulubilge: Why don't we add this as a default value instead?

packages/author-profile/author-profile.showcase.js
```diff
@@ -15,6 +16,9 @@ import {
 import storybookReporter from "@times-components/tealium-utils";
 import AuthorProfile from "./src/author-profile";
 
+const makeUrl = ({ slug, shortIdentifier }) =>
+  `https://www.thetimes.co.uk/article/${slug}-${shortIdentifier}`;
```
tuncaulubilge: I'd prefer a default value in context package so we can dry these up

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/de3b2f9b584f308da52d3d3b1b7bd310cb51746b/index.html

packages/article-list/__tests__/shared-states.web.js
nehasri89: The reason I did not put it in the default is that then it becomes the `default` way to do it, whereas a noop is just saying this is just a default and client's responsible for providing it

packages/author-profile/author-profile.showcase.js
colincclark: at least move the function to the `utils` package

packages/related-articles/related-articles.showcase.js
```diff
@@ -17,6 +18,9 @@ import opinionAndTwo1RelatedArticleFixture from "./fixtures/opinionandtwo/1-arti
 import opinionAndTwo2RelatedArticlesFixture from "./fixtures/opinionandtwo/2-articles.js";
 import opinionAndTwo3RelatedArticlesFixture from "./fixtures/opinionandtwo/3-articles.js";
 
+const makeUrl = ({ slug, shortIdentifier }) =>
+  `https://www.thetimes.co.uk/article/${slug}-${shortIdentifier}`;
```
colincclark: move to `utils ` package

packages/related-articles/src/related-article-item.web.js
```diff
@@ -0,0 +1,138 @@
+import React from "react";
```
colincclark: this whole file is just a duplicate of `related-article-item.js`. Please create a `related-article-item-base.js` file and compose what you need out of that to keep the code DRY

packages/topic/topic.showcase.js
```diff
@@ -6,11 +6,15 @@ import {
...
   topic as makeParams
 } from "@times-components/provider-test-tools";
+import Context from "@times-components/context";
 import storybookReporter from "@times-components/tealium-utils";
 import Topic from "./src/topic";
 import TopicProvider from "../provider/src/topic";
 import adConfig from "./fixtures/topic-ad-config.json";
 
+const makeUrl = ({ slug, shortIdentifier }) =>
+  `https://www.thetimes.co.uk/article/${slug}-${shortIdentifier}`;
```
colincclark: `utils` package

packages/related-articles/src/related-article-item.web.js
nehasri89: The plan is to do the same change for the native side pretty soon, so the web file lives only until then. Therefore I would be tempted to make minimal changes for now

colincclark: I disagree. we either maintain device parity, or keep the codebase DRY and maintainable. Ideally both? @craigbilner to weigh-in?

times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/31f4ca618a2807c9aa7038bdf805406296066575/index.html

craigbilner:
WIP

PR #1380, opened 2018-10-03
feat: expose tracking events to native storybook

nathggns:
Call the action logger from the native storybook `tealium-util`. Looks like it was just an oversight that it wasn't added.

`if (!global.storiesOf)` has been removed which was a previous hack when `fructose` first started

### DIFF
--- a/packages/tealium-utils/src/storybook.js
+++ b/packages/tealium-utils/src/storybook.js
@@ -1,1 +1,4 @@
-export default () => {};
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { action } from "@storybook/addon-actions";
+
+export default e => action("analytics-event")(e);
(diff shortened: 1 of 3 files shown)


packages/tealium-utils/src/storybook.js
```diff
@@ -1 +1,4 @@
-export default () => {};
+// eslint-disable-next-line import/no-extraneous-dependencies
```
colincclark: why the eslint disable? this package is declared as a dependency in the `package.json`

packages/tealium-utils/src/storybook.web.js
```diff
@@ -22,5 +22,5 @@ const reporter = tealiumReporter(tealiumSendScheduler);
 export default e => {
   if (reporter) reporter.analytics(e);
 
-  if (!global.storiesOf) action("analytics-event")(e);
+  return action("analytics-event")(e);
```
colincclark: the PR title says on native, but this is a web file. Can you add a contextual description please?

craigbilner: think this was a tools hack for early versions of fructose/dextrose

L0wry:
can confirm `storiesOf` was an early hack that is no longer used

PR #1383, opened 2018-10-04
fix: ios article storybook has window and DOM references

colincclark:
The article story now matches the ad one which also uses the `window.top.location` value as a link


times-tools:
Please find visual snapshots of your changed components here: https://s3-eu-west-1.amazonaws.com/times-components-snaps/5ace6cc6c98b26a0a8817ddc0ea4d1c3745b8eaf/index.html

PR #1387, opened 2018-10-08
feat: add lazy load to article

craigbilner:
We implemented the `lazy-load` package on list pages to improve the UX for people who scroll on a poor network. This PR implements it on the article page before we refactor for templates.

n.b. there's an additional improvement we could make by not requesting related article images if they're not visible.

feat: add size to native lead asset and use both resolutions for web
feat: add `registerNode` and `observed` to article body images
feat: add `id` to related articles container to determine when they're visible
feat: minimise `SSR` output for better perf testing
feat: pass `AUTH_TOKEN` envar to TPA for unteased articles
fix: if you provide both high and low res sizes to image it didn't render the high res one
refactor: remove redundant article row for web and use `renderTrees` instead. This gives us better "indexes for keys"
refactor: split out `Article` prop-types (still need some love around xxxpress events)
refactor: cache `styled-components` in `slice` constructors to avoid dismounting when props change
refactor: move card fade in to consumer (only used on web lists atm)
refactor: remove keys/indexes from child components (only needed for top level in lists)
chore: add stories around images
chore: add `lowResSize` to `article-image`


packages/article-image/article-image.showcase.js
```diff
@@ -16,12 +16,30 @@ export default {
...
-      component: () => (
-        <ArticleImage
-          captionOptions={primaryImage.captionOptions}
-          imageOptions={primaryImage.imageOptions}
-        />
-      )
+      // eslint-disable-next-line react/prop-types
+      component: ({ boolean }) => {
+        const withHighRes = boolean("As high resolution");
+        const withLowRes = boolean("As low resolution");
```
colincclark: knobs are a bit confusing. Surely just one checkbox is needed for high res? default would be low res? if both are checked it is odd

packages/article-list/src/article-list-item.web.js
```diff
@@ -110,7 +110,7 @@ const ArticleListItem = props => {
           lowResSize={lowResSize}
           showImage={showImage}
         >
-          {children}
+          <Animations.FadeIn>{children}</Animations.FadeIn>
```
colincclark: doesn't `Card` do the animation already?

packages/article/__tests__/web/article-lazy.web.test.js
```diff
@@ -0,0 +1,220 @@
...
+          onLinkPress={() => {}}
+          onRelatedArticlePress={() => {}}
+          onTopicPress={() => {}}
+          onTwitterLinkPress={() => {}}
+          onVideoPress={() => {}}
+        />
+      );
+
+      const [
+        ,
```
colincclark: not sure what is happening here

packages/card/__tests__/web/__snapshots__/card.web.test.js.snap
```diff
@@ -499,12 +443,10 @@ exports[`15. card should re-render when loading state changes 1`] = `
...
-      <div>
-        <div />
-        <div />
-        <div />
-        <div />
-      </div>
+      <div />
+      <div />
+      <div />
+      <div />
```
colincclark: this seems weird

packages/article-list/src/article-list-item.web.js
craigbilner: :speaking_head: this is specifically the content of the card

packages/article/__tests__/web/article-lazy.web.test.js
craigbilner: destructuring the second item :-)

packages/article-image/article-image.showcase.js
craigbilner: np :+1:

packages/card/__tests__/web/__snapshots__/card.web.test.js.snap
craigbilner: looks like the genuine structure? I've made a note to modernise these tests though

PR #1390, opened 2018-10-09
feat: Drop Cap

tuncaulubilge:
This PR introduces the drop cap package, with only Android and Web support. 

On Android, we are using a 3rd party library (https://github.com/aMarCruz/react-native-text-size) to measure text sizes and split the text into 3 textboxes. The latest version of this library hasn't been released yet, which includes the changes we need, so for now the library is added to storybook android code for simplicity. iOS support can be added once the library adds support for iOS on "lineEnd" calculation.

On web, it uses CSS floats in order to implement the feature. 

- RenderTreeAsText is added to markup forest as it'll be used to convert the first paragraph to a pure text, removing any markup. Further improvement can be done to support different markups inside the paragraph with drop caps in the future.
- Supports multiple text scales and section colours
- Supports multiple chars as the drop cap

Paired with @nehasri89 & @nathggns 

Default:
screenshot_1539094217

Multiple characters:
screenshot_1539188780

Xlarge scale:
screenshot_1539094254

With Layout bounds displayed:
screenshot_1539094323


packages/drop-cap/README.md
```diff
@@ -0,0 +1,53 @@
+# DropCap (Android Only)
+
+This package provides a paragraph starting with a drop cap letter. In native,
+this is done by measuring the length of the text and splitting it into 3
+textboxes.
+
+Currently, this component only supports text-only paragraphs.
+
+Uses 3rd party library to measure text sizes on native:
+https://github.com/aMarCruz/react-native-text-size
```
craigbilner: can tidy up with a markdown link

packages/drop-cap/README.md
```diff
@@ -0,0 +1,53 @@
+# DropCap (Android Only)
+
+This package provides a paragraph starting with a drop cap letter. In native,
+this is done by measuring the length of the text and splitting it into 3
+textboxes.
+
+Currently, this component only supports text-only paragraphs.
```
craigbilner: perhaps elaborate on what "text-only" means?

packages/drop-cap/README.md
```diff
@@ -0,0 +1,53 @@
...
+yarn lint
+```
+
+## Testing
+
+Testing can be done on each platform individually
+
+```
+yarn test:android
+yarn test:ios
```
craigbilner: don't think these are relevant?

packages/drop-cap/__tests__/android/__snapshots__/drop-cap.test.js.snap
```diff
@@ -0,0 +1,86 @@
...
+    Object {
+      "flexDirection": "row",
+      "flexWrap": "wrap",
+      "paddingLeft": 10,
+      "paddingRight": 10,
+    }
+  }
+>
+  <Text
+    accessible={true}
```
craigbilner: can use `minimaliseNative` to dump these

packages/drop-cap/__tests__/android/__snapshots__/drop-cap.test.js.snap
```diff
@@ -0,0 +1,86 @@
...
+    allowFontScaling={true}
+    ellipsizeMode="tail"
+    selectable={true}
+    style={
+      Array [
```
craigbilner: `flattenStyles` will also tidy this

packages/drop-cap/__tests__/android/__snapshots__/drop-cap.test.js.snap
```diff
@@ -0,0 +1,86 @@
...
+          "lineHeight": 26,
+          "marginBottom": 20,
+        },
+        Object {
+          "width": 730,
+        },
+      ]
+    }
+  >
+    wars, he said later that he had learnt “lessons which I have never forgotten. If, in some respects, they may have left too deep an impression on my mind, the gain was greater than the loss.” The gain was a lifelong conviction that the central aim of domestic policy must be to avoid the horror of mass unemployment. Forget ideological posturing; the job of a responsible Conservative government was to keep people in work.
```
craigbilner: for snapshotting I'd keep the "cruft" to a minimum

packages/drop-cap/__tests__/shared.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import TestRenderer from "react-test-renderer";
+import "./mock-text-measure-module";
+import DropCap from "../src/drop-cap";
+
+export default () => {
+  it("paragraph with single drop cap", async () => {
+    const testInstance = await TestRenderer.create(
```
craigbilner: this isn't async?

packages/drop-cap/__tests__/shared.js
```diff
@@ -0,0 +1,17 @@
...
+      <DropCap
+        dropCap="I"
+        text="n 1924 Harold Macmillan became MP for Stockton-on-Tees. Witnessing brutal poverty there between the wars, he said later that he had learnt “lessons which I have never forgotten. If, in some respects, they may have left too deep an impression on my mind, the gain was greater than the loss.” The gain was a lifelong conviction that the central aim of domestic policy must be to avoid the horror of mass unemployment. Forget ideological posturing; the job of a responsible Conservative government was to keep people in work."
+      />
+    );
+
+    expect(testInstance.toJSON()).toMatchSnapshot();
```
craigbilner: probably don't need the `toJSON`

packages/drop-cap/README.md
```diff
@@ -0,0 +1,53 @@
+# DropCap (Android Only)
```
craigbilner: `Drop Cap`?

packages/markup-forest/__tests__/__snapshots__/markup-forest.test.js.snap
```diff
@@ -153,3 +153,5 @@ exports[`9. multiple children 1`] = `
   </p>
 </div>
 `;
+
+exports[`10. nested tags with a text only renderer 1`] = `"Some text is here  and there is some  more text  and in here text of a special element inline text of a special element  after it too  and here"`;
```
craigbilner: some dodgy spacing?

packages/markup-forest/src/markup-forest.js
```diff
@@ -20,6 +20,18 @@ export const renderTree = (tree, renderers, key = "0", indx = 0) => {
...
+export const renderTreeAsText = (tree, key = "0") => {
+  const { attributes, children } = tree;
+
+  if (attributes && attributes.value) return attributes.value;
+
+  const renderedChildren = children
+    .map((child, index) => renderTreeAsText(child, `${key}.${index}`))
+    .join(" ");
+
+  return renderedChildren;
```
craigbilner: can tidy up and return the `join`

packages/drop-cap/src/drop-cap.js
```diff
@@ -0,0 +1,129 @@
...
+    const { measuredForScale } = this.state;
+
+    return (
+      <Context.Consumer>
+        {({ theme: { scale } }) => {
+          const stylesScaled = styleFactory(scale);
+          if (measuredForScale === scale) {
+            return this.renderParagraph(colour, stylesScaled, dropCap, text);
+          }
+          this.measureTextBoxes(stylesScaled, dropCap, text, scale);
```
craigbilner: can't `setState` in `render`

tuncaulubilge: Seems to work as we only set the state if it wasn't set before. What would be the better way to achieve this? We need to re-measure the widths when scale changes

packages/drop-cap/__tests__/shared.js
tuncaulubilge: It was going to be sync but then went for a sync mock for NativeModules, and just forgot these in. In real use case, rendering is async due to async nature of native bridges

packages/drop-cap/src/drop-cap.js
tuncaulubilge: Worked with @nathggns on this one :+1:

packages/drop-cap/__tests__/android/__snapshots__/drop-cap.test.js.snap
tuncaulubilge: It might be worth raising a github issue to add default serializers to our component templates

nathggns:
To further explain the problems discussed in https://github.com/newsuk/times-components/pull/1390#discussion_r223803540, there are two issues with how this was implemented here

1. *Render, generally, cannot set state*. That's because setState itself calls render. This only works here by virtue of the condition included, and also because the setState is done asynchronously. Otherwise, you'd see an infinite loop blocking the main thread, where render calls setState and setState calls render. 

2. *Render should always be a pure function*, and as efficient as possible. That's because in future versions of React, and I believe even in some circumstances here, render can be called multiple times during any rendering cycle. This becomes especially important when it comes to the `time slicing` and `async rendering` coming down the line. We cannot rely on render being called at predictable times.

To fix this, we did a number of things.

1. We made the `DropCapParagraph` component more dumb. This allows us to more easily use React's lifecycle methods to recalculate rendering information, as the scale is passed as a prop and is therefore known in every lifecycle method, rather than just as part of a render prop from a child component within render.

2. Rather than calculating within render, we calculated initially within `componentDidMount` and then within `componentDidUpdate` if the scale prop had changed. These lifecycle methods are far more predictable in when they are called than render is, and are expected to be used for side effects. **It is important that componentDidUpdate only sets state when something has actually changed, as setState causes a render, which causes a componentDidUpdate, providing yet another risk for infinite loops* 

There were other alternative ways of we thought about, but that we ruled out for various reasons

1. *Always* calling the calculation functions within the render method, but memoizing the result. This was ruled out because the calculation functions are asynchronous, and we didn't want to go down the rabbit hole of asynchronous rendering where it could be avoided.

2. Using `getDerivedStateFromProps`, which always you to calculate state based on your passed props (which is more or less what we're trying to do here). We ruled this out because again, this is not meant to have side effects and you can't actually do anything asynchronous in this method. I also get the sense that the people maintaining React don't really want you to use this but provided it as an escape hatch for some existing behaviour after they deprecated `componentWillReceiveProps`.

Hope this clears some of our thoughts up.

craigbilner:
wip

PR #1391, opened 2018-10-09
Feat: Lead asset image crop fallbacks

aronshamash:
- Fallback lead asset image crop as per https://nidigitalsolutions.jira.com/browse/REPLAT-3984
- For the standard article template the order of crops should be as follows:

`16:9`,`3:2`, `1.25:1`, `1:1`, `4:5`, `2:3`
- This also fixes an issue with poster image crops.
- This _does not_ deal with fallback to the poster image when there is no available crop. This will come in a future PR.


aronshamash:
I didn't think it was worth added to storybook for this PR. Would demonstrate the fallback to placeholder though. @craigbilner ?

PR #1402, opened 2018-10-12
fix: remove date leading zero

colincclark:
https://nidigitalsolutions.jira.com/browse/REPLAT-3387

- added a morning and afternoon test example and also in storybook for morning vs afternoon and BST vs GMT


packages/date-publication/src/date-time.js
```diff
@@ -28,8 +28,8 @@ class DatePublication extends Component {
     const offset = isDateBST ? 60 : 0;
     const datetimeLondonTimezone = addMinutes(datetimeUTC, offset);
     const formatString = showDay
-      ? "dddd MMMM DD YYYY, hh:mma"
-      : "MMMM DD YYYY, hh:mma";
+      ? "dddd MMMM DD YYYY, h:mma"
+      : "MMMM DD YYYY, h:mma";
```
aronshamash: Can DRY this up as `MMMM DD YYYY, h:mma` is common to both

colincclark: done

PR #1403, opened 2018-10-12
feat: adding canonical Urls for articles in topic and author-profile

craigbilner:
While it looks like a large change...it's actually just adding a new `Provider` to `article-list` and `article` so that the platform determines what the `url` of an article is, this includes list items and related articles. Philosophically `makeArticleUrl` should be provided by the platform and given an `article` object to allow it to determine how to "route" to that article.

feat: add `makeArticleUrl` provider to `Article` and `ArticleList`
feat: add `makeArticleUrl` stub to `test-utils`
feat: add `makeArticleUrl` to SSR package for navigation across Author Profiles, Topics and Articles

refactor: use `onPress` for native and `url` for web
refactor: make `isLoading` and `article` explicit in `article-list`
refator: pass `article` into list items for ease of use
refactor: split `related-article-item` for web and native `Links`
refactor: move `onPress` tests into native for `related-articles`

test: add `shortIdentifier` and `slug` to fixtures


packages/article-list/src/article-list-item.js
```diff
@@ -13,26 +13,24 @@ import { getImageUri, getHeadline } from "./utils";
...
     shortSummary,
-    showImage,
     summary,
     url
-  } = props;
+  } =
+    props.article || {};
 
-  const imageUri = getImageUri(props);
+  const imageUri = getImageUri(props.article);
```
nehasri89: should we get rid of url from the link then?

craigbilner: awkward linting error, will address separately

PR #1405, opened 2018-10-15
fix: Platform specific post message

tuncaulubilge:
iOS is not happy when we overwrite `window.postMessage`. This pr uses the workaround by checking the current platform and using reactBridgePostMessage instead when on iOS


packages/ad/__tests__/utils/webview-event-callback-setup.js
```diff
@@ -101,10 +102,18 @@ export default () => {
...
     jest.runAllTimers();
-    expect(window.reactBridgePostMessageDecoded).toHaveBeenCalledWith({
+    expect(window.postMessageDecoded).toHaveBeenCalledWith({
       detail: "a\nb\nc",
       isTngMessage: true,
       type: "error"
     });
   });
+
+  it("posts a message to reactBridgePostMessage instead of postMessage on ios", () => {
```
craigbilner: wont' this run as part of the Android tests?

tuncaulubilge: Webview setup is only ran in web tests. We would need to jump through couple of hoops to get webview tests running in native, so I went for a simpler approach.

tuncaulubilge: We should have some integration tests running on a simulator for proper ad testing

PR #1407, opened 2018-10-15
fix: prop type warnings in article

colincclark:
- ad config
- refetch method

prop type warnings fixed in story book and test runner. Ad config is on the `ArticlePage` level, not `Article` anymore


colincclark:
@craigbilner the error state needs a refetch in native. it is only required in native?

PR #1408, opened 2018-10-16
chore: use tile override for related articles

Belema:
This should conclude the listing asset saga.

See [JIRA](https://nidigitalsolutions.jira.com/browse/REPLAT-1920) ticket.


packages/provider-queries/src/article.js
```diff
@@ -150,8 +136,34 @@ export default addTypenameToDocument(gql`
         }
       }
     }
-    ...articleProps
-    ...summaries
+    article {
+      leadAsset {
+        ... on Image {
+          crop169: crop(ratio: "16:9") {
```
craigbilner: does this not need the same fallbacks?

packages/related-articles/src/related-article-item.js
```diff
@@ -15,6 +15,18 @@ import {
 import styles from "./styles";
 import getHeadline from "./utils";
 
+const getImageUri = (leadAsset, leadAssetOverride, cropSize) => {
```
craigbilner: perhaps a separate PR....but the philosophy and code seems slightly conflated

`leadAssetOverride` is a `slice` concern but we've put the logic in the consumer of the `slice`. Our `slice` package is actually about responsively laying out a `slice` on the various platforms and not about it's "filling". There is actually nothing special about a `RelatedArticleItem` so this should probably move to the `slice` package as `SliceFilling` (name tbc :smirk:  )? Then this logic is correctly at the `slice` level and is shared?

packages/provider-queries/src/article.js
Belema: This is, in some way, the unchanged part of the code.

packages/related-articles/__tests__/android/__snapshots__/oa2-3articles-leadassetoverride.android.test.js.snap
```diff
@@ -0,0 +1,209 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`1. one opinion and two support related articles 1`] = `
```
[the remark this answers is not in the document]
Belema: Oops, I meant to rename this test.

(not sure what you mean by separate snapshots. All tests in this package follow the pattern used here.)

packages/related-articles/src/related-article-item.js
Belema: Yes, I sadly agree.

packages/provider-queries/src/article.js
craigbilner: @aronshamash are you picking this up?

aronshamash: The fallbacks we have in place are for the article lead asset. this would display the placeholder if there is no `16:9` crop. I think falling back to anything other crop is not a good idea in slices for design reasons.

craigbilner: my only concern is it blowing up, so it should fallback to the T which still needs a ticket?

aronshamash: @craigbilner I created a story in storybook for the `image` to show the fallback to a 'T' : http://components.thetimes.co.uk/?knob-Size%20of%20ad%20placeholder%3A=default&selectedKind=Primitives%2FImage&selectedStory=Falls%20back%20to%20display%20the%20placeholder&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

PR #1411, opened 2018-10-16
fix: Remove div in related articles and use view instead

tuncaulubilge:
The div in related-articles came back through another refactor pr. So this pr removes it once again, same as https://github.com/newsuk/times-components/pull/1406


tuncaulubilge:
Storybooks blow up and snapshots have divs in it, which we missed while merging. Dextrose should be able to catch these as most stories just crash due to this error

PR #1413, opened 2018-10-17
feat: article scroll tracking

colincclark:
https://nidigitalsolutions.jira.com/browse/REPLAT-741

- native tracks all components in the article AST
- web merely fulfils the JIRA ticket and tracks the related articles scroll depth

Ignore `viewabilityConfigCallbackPairs` and mock the scroll tracking package in article (this matches what we do in article list for scroll tracking)


packages/tracking/__tests__/android/__snapshots__/tracking.android.test.js.snap
```diff
@@ -87,6 +87,7 @@ Array [
       "attrs": Object {
         "scrollDepth": Object {
           "itemNumber": 0,
+          "name": undefined,
```
jimhunty: This looks unexpected?

colincclark: The test describes "unseen children" and because the `itemNumber` is zero I assumed this was actually expected. Can you see anything specifically that I have missed?

colincclark: @craigbilner any idea what this test actually does and why?

craigbilner: we don't want to spam the reporter as you scroll around so it's pretending something has been "viewed twice" and that only one event resulted. You need to change the test to provide `name` as a prop.

packages/article/src/article.js
```diff
@@ -150,8 +153,20 @@ class ArticlePage extends Component {
...
   }
 
+  onViewableItemsChanged(info) {
+    if (!info.changed.length) return [];
+
+    return info.changed
+      .filter(viewableItem => viewableItem.isViewable)
+      .map(
+        viewableItem =>
+          this.props.onViewed &&
```
craigbilner: should we not provide this to `ArticleList` if there's nothing to do with it (no prop) and remove this check?

packages/tracking/__tests__/android/__snapshots__/tracking.android.test.js.snap
colincclark: Makes sense. Done. Thanks

packages/article/src/article.js
colincclark: Have refactored. Had to remove a test that did not make sense or cover anything. Maybe if any more changes this can be moved into a separate PR...

packages/article-list/__tests__/android/__snapshots__/article-list-states.android.test.js.snap
```diff
@@ -52,7 +52,9 @@ exports[`3. an empty list 1`] = `
 `;
 
 exports[`4. loading state 1`] = `
-<RCTScrollView>
+<RCTScrollView
+  onViewableItemsChanged={null}
```
craigbilner: do we need this in the snaps?

colincclark: omitted

PR #1414, opened 2018-10-17
refactor: remove unrequired adConfig prop type

colincclark:
- remove `adConfig` prop type - Ad config is on the ArticlePage level, not Article anymore


aronshamash:
@colincclark has it been added as a prop to `ArticlePage` ?

colincclark:
> @colincclark has it been added as a prop to `ArticlePage` ?

https://github.com/newsuk/times-components/blob/master/packages/article/src/article-page-prop-types.js#L6

PR #1418, opened 2018-10-17
refactor: strip flow out of tracking

colincclark:
Remove flow from tracking to avoid others thinking it was a good idea.

- we do not use flow in any other package or file
- there was an agreement to move to typescript so flow's presence is confusing
- the flow stuff is more difficult to debug for those unfamiliar to the syntax
- less guff to maintain and wade through when debugging / refactoring this tracking code


colincclark:
@craigbilner disagree. strongly. we are not using flow anywhere else, so this is just sending mixed signals especially to newcomers

colincclark:
closed: team would prefer a github issue and discussion approach first

PR #1420, opened 2018-10-17
chore: output CSS in react native style serialisers. Fixes #1031

### LINKED ISSUE #1031: rnw serializer css

feature

An object is printed in a `style` block

**If the current behaviour is a bug, please provide the steps to reproduce.**

Would look better as a "css stylesheet"

nathggns:
- Modifies `hoist-style.js` to make use of `react-native-web` internals & dependencies to make outputted inline styles more inline with outputted style sheets. (Unfortunately there was not a single method that did everything we needed so had to use a couple)
- Modifies the style printer to output styles as CSS in web snapshots rather than as JSS objects 
- Updates all existing snapshots to use CSS rather than JSS in web snapshots

### DIFF
--- a/packages/ad/__tests__/web/__snapshots__/ad-placeholder-with-style.web.test.js.snap
+++ b/packages/ad/__tests__/web/__snapshots__/ad-placeholder-with-style.web.test.js.snap
@@ -2,114 +2,122 @@
 
 exports[`advert placeholder 1`] = `
 <style>
-{
-  "IS1": {
-    "left": 0,
-    "position": "absolute",
-    "top": 0,
-  },
-  "IS2": {
-    "backgroundColor": "#F9F9F9",
-    "borderColor": "#DBDBDB",
-    "borderStyle": "solid",
-    "borderWidth": 1,
-    "color": "#696969",
-    "fontFamily": "TimesDigitalW04",
-    "fontSize": 11,
-    "letterSpacing": 1.5,
-    "lineHeight": 15,
-    "paddingBottom": "5px",
-    "paddingLeft": "10px",
-    "paddingRight": "10px",
-    "paddingTop": "5px",
-    "zIndex": 1,
-  },
-  "IS3": {
-    "height": "300px",
-    "width": "970px",
-  },
-  "IS4": {
-    "alignItems": "center",
-    "backgroundColor": "#F9F9F9",
-    "borderColor": "#DBDBDB",
-    "borderStyle": "solid",
-    "borderWidth": 1,
-    "height": 300,
-    "justifyContent": "center",
-    "overflow": "hidden",
-    "width": 970,
-  },
-  "IS5": {
-    "backgroundColor": "rgba(255,0,0,1.00)",
-  },
-  "IS6": {
-    "alignItems": "center",
-    "backgroundColor": "red",
-    "flex": 1,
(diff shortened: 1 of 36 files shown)


packages/jest-serializer/.eslintignore
```diff
@@ -2,3 +2,5 @@
 **/coverage/*
 dist
 rnw.js
+hoist-style.js
```
colincclark: why are these files not subject to linting?

nathggns: `printers.js` uses the same `css` package as `rnw.js` (which is also ignored for this reason, it seems) that is hoisted as a top level dependency and not explicitly a dependency in `jest-serializer`. Lint complains about this.

`hoist-style.js` uses a dependency of `react-native-web`, `hyphenate-style-name`, that also isn't listed as a dependency but is hoisted as a top level dependency. Lint also complains about this.

I considered adding both of these packages to the `jest-serializer` dependencies, but my yarn kept hanging when installing dependencies at a package level, and I couldn't figure out why, so I went down the path of least resistance.

If you can help me sort that issue, happy to undo this.

colincclark: `import hyphenateStyleName from "hyphenate-style-name"; // eslint-disable-line import/no-extraneous-dependencies`

packages/ad/__tests__/web/__snapshots__/ad-placeholder-with-style.web.test.js.snap
```diff
@@ -2,114 +2,157 @@
...
+}
+
+.IS1 {
+  left: 0px;
+  position: absolute;
+  top: 0px;
+}
+
+.IS2 {
+  background-color: rgba(249,249,249,1.00);
```
craigbilner: styles have exploded, ideally we want to keep snapshots as small as possible

packages/jest-serializer/src/hoist-style.js
```diff
@@ -1,6 +1,22 @@
+import hyphenateStyleName from "hyphenate-style-name";
+import normalizeValue from "react-native-web/dist/cjs/exports/StyleSheet/normalizeValue";
```
craigbilner: these need removing and a GitHub issue raising instead for a sensible solution

packages/ad/__tests__/web/__snapshots__/ad-placeholder-with-style.web.test.js.snap
nathggns: Inline styles have sure, but only to the point where they're consistent with `StyleSheet` styles (`classNames`). Is consistency not important enough?

If we want to find a way to make the styles smaller, that's a worthwhile conversation have, but we should do that in a consistent way.

packages/jest-serializer/src/hoist-style.js
nathggns: So to clarify – I should remove the bit that makes inline styles consistent with `StyleSheet` styles?

craigbilner:
going to investigate exploding styles

nathggns:
Asked RNW to implement an API to make this nicer. The API would allow us to selectively apply some of these transforms

https://github.com/necolas/react-native-web/issues/1144

PR #1425, opened 2018-10-19
chore: output CSS in react native style serialisers. Fixes #1031

### LINKED ISSUE #1031: rnw serializer css

nathggns:
This fixes #1031 

- Modifies `hoist-style.js` to make use of `hyphenate-style-name` to make inline styles look more like CSS
- Modifies the style printer to output styles as CSS in web snapshots rather than as JSS objects 
- Updates all existing snapshots to use CSS rather than JSS in web snapshots 
- Updates eslint and fixes some linting mistakes in jest-serializer 

**Note**: For inline styles, the outputted "CSS" in snapshots will still have react-native properties & unitless values. This is because there is no public API for arbitrarily applying these the transformations applied to StyleSheets by React Native Web. https://github.com/necolas/react-native-web/issues/1144

### DIFF
--- a/lib/cli/package-json.js
+++ b/lib/cli/package-json.js
@@ -44,7 +44,7 @@ module.exports = variables => ({
     "@times-components/storybook": "*",
     "@times-components/webpack-configurator": "*",
     "babel-cli": "6.26.0",
-    eslint: "4.9.0",
+    eslint: "4.19.1",
     jest: "23.3.0",
     prettier: "1.8.2",
     react: "16.5.2",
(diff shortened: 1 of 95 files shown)


nathggns:
We should find a way to condense the StyleSheet styles in the snapshots

- Remove the prefixes? 
- Merge CSS properties that can be merged?

It may well be we avoid RNW's transformations for StyleSheet styles just as we have for inline styles, and create consistency that way.

Not sure we should hold this back for now though.

PR #1428, opened 2018-10-23
feat: Article Loading Package

jimhunty:
In order to make the Article Refactor PR a lot smaller, getting this in as a separate PR.

This package in this PR is not used by any other as it would only be used by `article` which is being drastically changed.


jimhunty:
We have collectively decided against doing a loading package.

PR #1434, opened 2018-10-25
BREAKING CHANGE: Use Apollo's new Query component. Fixes #1225

### LINKED ISSUE #1225: Use the new `Query` component from React-Apollo
We currently use an older version of React-Apollo and do the GQL connections in an older fashion. We should update to using the new Query component found here: https://www.apollographql.com/docs/react/essentials/queries.html#basic

nathggns:
Fixes #1225 

- Adds a `Debounce` component with a renderProp
- Use the Debounce component rather than the higher order component inside connect
- Abstract most of the innards of connect to a `QueryProvider` component
- Writes a few more tests for `QueryProvider`

`Debounce` uses `debounceRender` rather than `children` for its renderProp because `children` is a legitimate prop that Debounce might pass through / debounce. 

*I think this is a breaking change* because it changes the structure that `connectGraphql` and `withDebounce` renders, which might break snapshots by consumers downstream.


packages/provider/__tests__/connect.test.js
```diff
@@ -1,25 +1,45 @@
...
 
-    expect(options(props)).toEqual({ variables: { a: 10, b: 20 } });
-  });
+const propsToVariables = () => ({});
 
-  it("transforms props with the propsToVariables function", () => {
-    const propsToVariables = ({ a, b }) => ({ A: a + 10, B: b + 10 });
-    const options = makeGraphqlOptions(["A", "B"], propsToVariables);
-    const props = { a: 1, b: 2, c: 3 };
+describe("connectGraphql", () => {
```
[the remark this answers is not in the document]
nathggns: Pardon? Every test file should have a top level describe, surely?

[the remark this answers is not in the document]
colincclark: Incorrect. The `describe` methods make the snapshot names look weird. `QueryProvider 1.` and `QueryProvider 2.` for example. The file name tells us what is being tested, the `it` or `test` methods tell us, in human readable language what each test does. The pattern to drop `desribe` methods was implemented some months ago when the snapshots started looking weird like they do in several places here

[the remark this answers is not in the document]
colincclark: That being said, this issue should not block the PR, so have dismissed my review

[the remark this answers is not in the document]
nathggns: I actually thought jest *required* a top level describe. Interesting that it doesn't, I didn't know this.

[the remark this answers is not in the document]
nathggns: In this case, I think the snapshot name looks perfectly fine though.

`connectGraphql renders the correct QueryProvider component 1`

packages/provider/__tests__/__snapshots__/connect.test.js.snap
```diff
@@ -0,0 +1,11 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`connectGraphql renders the correct QueryProvider component 1`] = `
+<QueryProvider
+  debounceTimeMs={1000}
+  foo="baz"
+  propsToVariables={[Function]}
```
craigbilner: can remove

packages/provider/__tests__/connect.test.js
```diff
@@ -1,25 +1,45 @@
...
+    const ConnectedComponent = connectGraphql(query, propsToVariables);
+    const component = shallow(
+      <ConnectedComponent debounceTimeMs={1000} foo="baz">
+        {() => null}
+      </ConnectedComponent>
+    );
 
-    expect(options(props)).toEqual({ variables: { A: 11, B: 12 } });
+    expect(component).toMatchSnapshot();
+    expect(component.find(QueryProvider).props()).toEqual(
```
craigbilner: why would you do a snapshot and this?

packages/provider/src/connect.js
```diff
@@ -1,56 +1,8 @@
...
-    debouncedProps: PropTypes.shape({}).isRequired
-  };
-
-  const GraphQlComponent = graphql(query, {
-    options: makeGraphqlOptions(variableNames, propsToVariables)
-  })(Wrapper);
-
-  return withDebounce(GraphQlComponent);
-};
+const connectGraphql = (query, propsToVariables) => props => (
```
craigbilner: not obvious why this moved?

packages/provider/src/debounce.js
```diff
@@ -1,84 +1,102 @@
...
   }
+}
 
+Debounce.propTypes = {
+  debounceRender: PropTypes.func.isRequired,
+  debounceTimeMs: PropTypes.number.isRequired
+};
+
+const withDebounce = WrappedComponent => {
+  const WithDebounce = props => (
```
craigbilner: strangely this has no test coverage?

packages/provider/__tests__/__snapshots__/debounce.test.js.snap
```diff
@@ -1,82 +1,89 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Debounce handles different debounce delays 1`] = `
+exports[`Debounce 3. immediately updates props, but delays update of debouncedprops 1`] = `
```
craigbilner: not sure why the tests needed to change but looks like the odd use of `iterator` has given them number starting from 3

packages/provider/__tests__/connect.test.js
nathggns: To check the correct propToVariables and query are passed through, as the snapshot doesn't check that.

nathggns: Will separate into multiple tests

packages/provider/src/debounce.js
nathggns: Does now.

packages/provider/__tests__/debounce.test.js
```diff
@@ -26,11 +32,17 @@ describe("Debounce", () => {
   });
 
   it("does not change debouncing status if component recieves same props", () => {
-    const Outer = withDebounce(Inner);
-    const component = shallow(<Outer debounceTimeMs={1000} foo="initialFoo" />);
+    const component = shallow(
```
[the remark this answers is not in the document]
nathggns: 👍

packages/provider/__tests__/connect.test.js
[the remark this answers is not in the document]
nathggns: Removed the top level describe.

packages/provider/__tests__/__snapshots__/connect.test.js.snap
nathggns: 👍

packages/provider/src/connect.js
nathggns: Moved it back.

packages/provider/src/debounce.js
nathggns: Does now

packages/provider/__tests__/__snapshots__/debounce.test.js.snap
nathggns: Fixed this

packages/provider/__tests__/query-provider.test.js
```diff
@@ -0,0 +1,76 @@
...
+const queryWithVariable = gql`
+  query ArticleQuery($id: ID!) {
+    article(id: $id) {
+      content
+    }
+  }
+`;
+
+describe("QueryProvider", () => {
+  it("1 when passing no props to variables, it should call its children with the correct props", () => {
```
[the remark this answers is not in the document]
nathggns: 👍

PR #1442, opened 2018-10-29
chore: add cypress

craigbilner:
Much to @colincclark's chagrin we haven't had tests in SSR for they were awaiting this little PR. This is the seed to something hopefully much bigger, where e2e testing is brought into the repo and we move closer to TDD with our editorial content.

See the SSR README for more details

BREAKING CHANGE: change make-mocks to expose a schema instead of GraphQL and put on the consumer
feat: add mock TPA which runs on port 4000
fix: article mock generator wasn't stateless so couldn't be used as a server
fix: get underlying ref for related article image width
test: add cypress test for related article hi res images
fix: change topic showcase to use correct import


packages/provider-test-tools/src/article.js
```diff
@@ -14,16 +14,23 @@ const convertRatio = ratio => {
   return "100/100";
 };
 
+const getMediaUrl = (obj, ratio) => {
+  const crop = obj[`crop${ratio.replace(":", "")}`];
+
+  return {
+    url: crop ? crop.url : `https://placeimg.com/${convertRatio(ratio)}/tech`
```
L0wry: could there be a case where `crop` exists but `crop.url` does not?

packages/ssr/cypress/integration/article.js
```diff
@@ -0,0 +1,26 @@
+describe("Article", () => {
+  beforeEach(() => {
+    cy.visit("/article/8763d1a0-ca57-11e8-bde6-fae32479843d");
+  });
+
+  it("loads hi-res images for related articles", () =>
+    cy
+      .get("#related-articles")
```
L0wry: should probably start implementing some sort of page model design pattern, or exporting selectors direct from the code and consuming them here otherwise having selectors like this will get really messy down the line

packages/ssr/cypress/integration/article.js
```diff
@@ -0,0 +1,26 @@
...
+      .scrollIntoView()
+      .then(() => {
+        // wait for the image to transition and be removed (unfortunately Cypress doesn't auto wait for this)
+        cy.wait(2000);
```
L0wry: does cypress support explicit waits?

L0wry:
we should also have a task to update bitrise to run these tests, if you haven't already got one

packages/ssr/cypress/integration/article.js
L0wry: cy.wait('@getProfile')

packages/ssr/README.md
```diff
@@ -51,6 +51,23 @@ command such as `npx webpack-bundle-analyzer stats.json` in the `dist` folder to
...
+```
+
+This starts an [Apollo sever](https://www.apollographql.com/docs/apollo-server/)
+with a mock implementation using the fixture generator from `provider-test-tools`.
+This should be expanded on to provide different use cases for particular queries
+and used for testing with Cypress so that we have a closely integration
+"fixture server" for robust e2e testing.
+
+```bash
+yarn start:test
```
L0wry: `Error: Cannot find module '@times-components/provider/rnw'`

packages/ssr/cypress/integration/article.js
craigbilner: sadly there's no API to wait for the DOM element to be removed :-(

packages/provider-test-tools/src/article.js
craigbilner: not currently

packages/ssr/cypress/integration/article.js
craigbilner: will pick up next :+1:

PR #1443, opened 2018-10-29
feat: split article comments

tuncaulubilge:
This PR moves native-only article-comments out of the article package and into its own package. 

This way, we can wrap comments with its own provider (ArticleCommentsProvider) so comments are only requested when it's scrolled to. With this PR, even if we supply article data offline, the page will still make a network call to fetch the latest comment count, making sure the counts are always up-to-date.

Also added a error state with a retry button.

screenshot_1540834343
screenshot_1540834341
screenshot_1540834336


packages/provider-test-tools/src/article-comments.js
```diff
@@ -0,0 +1,29 @@
+import { articleComments as articleCommentsQuery } from "@times-components/provider-queries";
+
+export default ({
+  error = () => {},
```
craigbilner: alphabetical? :smirk:

packages/article-comments/__tests__/single-comment.js
```diff
@@ -0,0 +1,35 @@
...
+    const testInstance = TestRenderer.create(
+      <ArticleComments
+        articleId="dummy-article-id"
+        onCommentGuidelinesPress={() => {}}
+        onCommentsPress={() => {}}
+        url="dummy-article-url"
+      />
+    );
+
+    const comments = findViewByText(testInstance, "1 comment");
```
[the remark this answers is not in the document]
tuncaulubilge: There is special wording for 1 comment and 0 comment cases, which is hard to see inside a snapshot, so I'm checking if that wording is correct directly

packages/article/package.json
```diff
@@ -112,4 +113,4 @@
   "publishConfig": {
     "access": "public"
   }
-}
+}
```
colincclark: add newline

packages/article-comments/__tests__/single-comment.js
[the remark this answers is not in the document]
craigbilner: If you see snapshots as ways of visually snapshotting then you would find and snapshot. Making specific snapshots is the correct way to snapshot imo

packages/article/package.json
tuncaulubilge: Fixed my vscode settings :+1:

colincclark:
Approved, but agree with Craig's change requests above

PR #1449, opened 2018-10-31
BREAKING CHANGE fix: remove makeArticleUrl & usages of test-utils from showcases

nathggns:
<s>This is intended to fix dextrose, by removing a test-util that depends on being in node from the default export of the `test-utils` package, which is currently used by showcases (mostly for makeArticleUrl)

It may be that the right thing to do is to move `makeArticleUrl` to `utils`, and make `test-utils` an explicitly node-only package. But I think that requires wider consideration and so I present this as an alternative solution.

Discussed this with @L0wry and @nehasri89 which is how I ended up with this approach. 

*This may not fix dextrose if there are other issues. This just fixes the one that I've come across so far*. 
</s>

This fixes dextrose by removing usages of the node-only package `test-utils` within showcases.

**To do this, I removed the `makeUrlArticle` method from the `test-utils` package. That makes this a breaking change.**

### DIFF
--- a/packages/topic/topic.showcase.js
+++ b/packages/topic/topic.showcase.js
@@ -6,7 +6,6 @@ import {
   MockFixture,
   topic as makeParams
 } from "@times-components/provider-test-tools";
-import { makeArticleUrl } from "@times-components/test-utils";
 import Context from "@times-components/context";
 import storybookReporter from "@times-components/tealium-utils";
 import Topic from "./src/topic";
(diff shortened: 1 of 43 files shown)


nathggns:
Going to change approach and remove `makeArticleUrl` instead.

packages/topic/topic.showcase.js
```diff
@@ -30,7 +29,12 @@ const getProps = decorateAction => ({
 const articleImageRatio = "3:2";
 const name = "Chelsea";
 const pageSize = 20;
-const slug = "chelsea";
+const articleSlug = "chelsea";
```
craigbilner: `topicSlug`?

nathggns: Oh good catch 🤦‍♀️

PR #1451, opened 2018-11-02
feat: Zoom at centre of the two touches during pinch to zoom

nathggns:
React Native doesn't support transform origin, so we have to use translations to emulate this. https://github.com/facebook/react-native/issues/1964

nov-02-2018 12-17-57

First we apply a translation to reset the origin of the transformation to the top left of the gesture view, and then we apply another translation to move the origin to the offset from that to the centre of the two touches. Then, we apply the zoom transformation, and then undo the translations. Then the rotation is applied.

### DIFF
--- a/packages/gestures/__tests__/shared.native.js
+++ b/packages/gestures/__tests__/shared.native.js
@@ -24,12 +24,12 @@ const makeTouchEvent = (active, history = []) => ({
 });
 
 const renderComponent = () => {
-  const component = TestRenderer.create(
+  const testRenderer = TestRenderer.create(
     <Gesture>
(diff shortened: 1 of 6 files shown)


packages/gestures/__tests__/shared.native.js
```diff
@@ -23,12 +23,24 @@ const makeTouchEvent = (active, history = []) => ({
   }
 });
 
-const renderComponent = () =>
-  TestRenderer.create(
+const renderComponent = () => {
+  const component = TestRenderer.create(
```
colincclark: the convention is `testRenderer` not `component`

PR #1452, opened 2018-11-02
fix: Use screen width if highRes is not provided in images

tuncaulubilge:
Image resize was broken on native apps since highResSize is not passed at the moment. This pr uses screenWidth if highResSize is not available

### DIFF
--- a/packages/image/src/image.js
+++ b/packages/image/src/image.js
@@ -1,6 +1,10 @@
 import React, { Component } from "react";
 import { Image, View } from "react-native";
-import { addMissingProtocol } from "@times-components/utils";
+import {
+  addMissingProtocol,
+  normaliseWidth,
+  screenWidthInPixels
+} from "@times-components/utils";
 import appendSize from "./utils";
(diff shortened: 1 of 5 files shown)


packages/image/src/image.js
```diff
@@ -22,13 +27,13 @@ class TimesImage extends Component {
...
     const { aspectRatio, highResSize, style, uri } = this.props;
-    const { isLoaded } = this.state;
+    const { isLoaded, width } = this.state;
 
     const isDataImageUri = uri && uri.indexOf("data:") > -1;
 
     const srcUri = isDataImageUri
       ? uri
-      : addMissingProtocol(appendSize(uri, "resize", highResSize));
+      : addMissingProtocol(appendSize(uri, "resize", highResSize || width));
```
colincclark: this is missing a test

tuncaulubilge:
@craigbilner in our current state, the images we have are all full width, with 1 or 2 exceptions, so deciding on package level is alright. For the exception cases, we can pass the "highResSize" prop when needed

PR #1464, opened 2018-11-08
feat: Filter interactives from article ast

tuncaulubilge:
Adds ability to enable/disable interactives on the native apps. We do the filtering by modifying the article ast.

It's added as a prop to article and set to false by default. This feature flag should be removed completely once we want to release interactive elements.


packages/pages/src/article/filter-interactives.js
```diff
@@ -0,0 +1,9 @@
+export default article =>
+  article
+    ? {
+        ...article,
+        content: [
+          ...article.content.filter(node => node.name !== "interactive")
```
nathggns: Why the filter + the spread? The filter makes a new array anyway.

packages/pages/src/article/article-prop-types.js
```diff
@@ -7,7 +7,8 @@ export const propTypes = {
   omitErrors: PropTypes.bool,
   refetch: PropTypes.func,
   scale: PropTypes.string,
-  sectionName: PropTypes.string
+  sectionName: PropTypes.string,
+  showInteractives: PropTypes.bool
```
nathggns: Where does `showInteractives` come from?

tuncaulubilge: Native apps pass these while creating an article. It'll be a setting in the native app debug screen

nathggns: And render passes true in always?

nathggns: So how does this showInteractives prop get passed on the website?

tuncaulubilge: This is just filtering it for the native apps, it's unfiltered for web

nathggns: But the filter is in the base article? https://github.com/newsuk/times-components/pull/1464/files#diff-e6e55958a0d63c66307aa9d108def32b

tuncaulubilge: It's not a base in the sense of native or web articles (it's still in pages package, which is native only). It's called "base" because it's shared between Article with Provider and without provider scenarios (inside article/index.js)

nathggns: Ok

PR #1465, opened 2018-11-08
fix: Use short headline in Related Articles on web

nathggns:
The short headline isn't displaying in Related Articles slices on web, because its not included in the GraphQL query on web.


packages/provider-queries/src/article.web.js
```diff
@@ -106,6 +106,7 @@ export default addTypenameToDocument(gql`
     publishedTime
     section
     shortIdentifier
+    shortHeadline
```
aronshamash: good spot

PR #1476, opened 2018-11-12
chore: SSR cypress tidy-up

Belema:
As the name suggests, this is to tidy up the current cypress tests/configuration
- rename folder `cypress` to `__tests__`
- add `cypress.json` configuration file
- fix/tidy up `package.json` scripts
  - wait-on is now used/working in the intended way
  - we only terminate the processes that we start
  - scripts names are more descriptive


packages/ssr/package.json
```diff
@@ -6,18 +6,16 @@
     "bundle:dev": "yarn cleanup-dist && webpack --config=webpack.config.js",
     "bundle:prod": "yarn cleanup-dist && NODE_ENV=production webpack --config=webpack.config.js -p",
     "bundle:profile": "yarn cleanup-dist && NODE_ENV=production webpack --config=webpack.config.js --profile --json > dist/stats.json",
-    "cleanup:servers": "pkill node",
```
tuncaulubilge: 💥 lol

PR #1479, opened 2018-11-13
fix: Add id to images inside author and topic articles queries

tuncaulubilge:
Apollo normalised cache has an issue which fails the network request if an object has an id in one query and doesn't have it in another. Added id to all image objects inside topic and author articles

Here is the error we are receiving at the moment: 
```
Store error: the application attempted to write an object with no provided id but the store already contains an id of Image:d3494166-0c9b-4971-86c0-602892d5b11c for this object. The selectionSet that was trying to be written is:
```
Here is the related github issue for more details: https://github.com/apollographql/apollo-client/issues/2510

### DIFF
--- a/packages/provider-queries/src/topic-articles.js
+++ b/packages/provider-queries/src/topic-articles.js
@@ -21,18 +21,19 @@ export default addTypenameToDocument(gql`
           shortIdentifier
           leadAsset {
             ... on Image {
-              id
-              title
               crop(ratio: $imageRatio) {
                 url
               }
+              id
+              title
             }
             ... on Video {
               posterImage {
-                title
                 crop(ratio: $imageRatio) {
                   url
                 }
+                id
+                title
               }
             }
           }
--- a/packages/provider-test-tools/src/author-profile.js
+++ b/packages/provider-test-tools/src/author-profile.js
@@ -118,7 +118,10 @@ export default ({
             imageIndex += 1;
             return {
               __typename: "Image",
-              title: `Some title ${imageIndex}`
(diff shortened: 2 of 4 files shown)


packages/provider-test-tools/src/author-profile.js
```diff
@@ -118,7 +118,10 @@ export default ({
             imageIndex += 1;
             return {
               __typename: "Image",
-              title: `Some title ${imageIndex}`
+              title: `Some title ${imageIndex}`,
+              id: `e98c2${imageIndex
+                .toString()
+                .padStart(2)}c-cb16-11e7-b529-95e3fc05f40f`,
```
colincclark: this feels a bit weird

packages/provider-queries/src/topic-articles.js
```diff
@@ -21,18 +21,19 @@ export default addTypenameToDocument(gql`
           shortIdentifier
           leadAsset {
             ... on Image {
-              id
-              title
               crop(ratio: $imageRatio) {
                 url
               }
+              id
```
colincclark: not uuid? weird

packages/provider-test-tools/src/author-profile.js
tuncaulubilge: Saw this in other providers. I agree it's a bit weird but at least it ensures a certain level of uniqueness

packages/provider-queries/src/topic-articles.js
tuncaulubilge: Its type is "uuid" but in apollo it's called "id" by default

PR #1480, opened 2018-11-13
test: article main tests

colincclark:
- fixes the `main-article-standard` web snapshots
- fixes issue with jest-lint in the same package

### DIFF
--- a/packages/article-main-standard/__tests__/android/__snapshots__/article-lead-asset.android.test.js.snap
+++ b/packages/article-main-standard/__tests__/android/__snapshots__/article-lead-asset.android.test.js.snap
@@ -2,11 +2,6 @@
 
 exports[`1. a full article with an image as the lead asset 1`] = `
 <RCTScrollView
-  ListHeaderComponent={
-    <bound renderHeader
-      width={1440}
-    />
-  }
   initialNumToRender={10}
   maxToRenderPerBatch={10}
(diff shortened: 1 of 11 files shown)


packages/article-main-standard/__tests__/android/__snapshots__/article-lead-asset.android.test.js.snap
```diff
@@ -2,11 +2,6 @@
 
 exports[`1. a full article with an image as the lead asset 1`] = `
 <RCTScrollView
-  ListHeaderComponent={
```
colincclark: jest-lint had a problem with this. Have added an issue https://github.com/newsuk/jest-lint/issues/26 but these fixes need to get in in order to allow me to bump various package dependencies

packages/article-main-standard/__tests__/mocks.web.js
```diff
@@ -1,6 +1,5 @@
 // eslint-disable-next-line global-require
 jest.mock("@times-components/ad", () => require("./ad-mock"));
-jest.mock("@times-components/article", () => "Article");
```
tuncaulubilge: Why are we removing this mock?

colincclark: so that the snapshots can show the article properly

PR #1484, opened 2018-11-14
Feat/fixture generator

L0wry:
And so begins our journey into creating type-safe fixtures for articles

This PR creates a new package called fixture gen and migrates away from the provider test tool package.
The fixture gen package has a responsibility of providing correct fixtures to any client (such as showcases, cypress) 
It uses schema introspection and graphqlgen to create types based on the schema.

This means that whenever the schema is updated the types will break and be very obvious as to why.

As part of this, we have refactored an existing article cypress test.
unfortunately, this has broken the other cypress tests that @nehasri89  created. In the interim, I have switched them off and tools have another ticket in this sprint to pick up the work to refactor.


packages/fixture-generator/index.ts
```diff
@@ -0,0 +1,3 @@
+import MockArticle from "./src/mock-article";
+
+export default { MockArticle };
```
nathggns: Have you tried importing this anywhere?

My feeling is if you tried to import it like

```
import { MockArticle } from "@times-components/fixture-generator"
```

It wouldn't work. I might be wrong but that's how I remember default/named exports working.

I think if you do the following instead it will work. 

```
export { default as MockArticle } from "./src/mock-article";
```

nathggns: That's because 

```
import { MockArticle } from "@times-components/fixture-generator"
```

is synonymous with

```
const { MockArticle } = require("@times-components/fixture-generator");
```

but 

```
export default { MockArticle };
```

is synonymous with

```
module.exports.default = { MockArticle };
```

nathggns: Whereas 

```
export { default as MockArticle } from "./src/mock-article";
```

is synonymous with

```
module.exports.MockArticle = require("./src/mock-article").default;
```

packages/fixture-generator/src/mock-article.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  withSundayTimes() {
+    this.article.publicationName = getPublicationName(
+      PublicationName.SUNDAYTIMES
+    );
+    return this;
+  }
+
+  fetch() {
```
nathggns: Can you explain why this is called `fetch` instead of like, `generate` or `create` or something that doesn't suggest asynchrony?

nathggns: Also I was thinking, might be cool for generators to have an interface of some kind?

packages/fixture-generator/src/resolvers/publication-name.ts
```diff
@@ -0,0 +1,5 @@
+import { PublicationName } from "../types";
+
+export default (
```
nathggns: I believe if you give this a name like

```
const publicNameResolver = ....;

export default publicNameResolver;
```

then tools like WebStorm will be able to find it for you, and stack traces may be slightly easier to follow.

packages/fixture-generator/src/types.ts
```diff
@@ -0,0 +1,1572 @@
...
+// ====================================================
+// START: Typescript template
+// ====================================================
+
+// ====================================================
+// Scalars
+// ====================================================
+
+/** A lower kebab case string */
+export type Slug = any;
```
nathggns: Its my understanding that TypeScript doesn't do any differentiation between types that have different identities but are functionally the same? That's what https://github.com/gcanti/newtype-ts is created to address. Now I'm not suggesting we use that, but I'm wondering what the advantage of all of these `any` type aliases is otherwise?

packages/fixture-generator/src/types.ts
```diff
@@ -0,0 +1,1572 @@
+/* tslint:disable */
```
nathggns: This is a pretty comprehensive list of types used across the project. Is there any possibility of making these cross package in the future?

packages/fixture-generator/index.ts
L0wry: sorted

packages/fixture-generator/src/types.ts
L0wry: the type file is generated from using schema introspection. You're gonna have to talk to the Tpa chaps about the `any` types

packages/fixture-generator/src/mock-article.ts
L0wry: this has been addressed! 
are you thinking about when we have different article types for the interface??

packages/fixture-generator/src/types.ts
L0wry: so this is how TPA has defined types

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,99 @@
+import MockMarkup from "../mock-markup"
+
+describe("get markup", () => {
+
+  it("returns an array", () => {
+    expect(new MockMarkup().create()).toEqual([])
```
Belema: `new Blah().create()` is a bit odd

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,99 @@
...
+  })
+
+  it("should be able to generate paragraphs of markup", () => {
+    const mockMarkup = new MockMarkup().withParagraph().withParagraph()
```
Belema: ~`.withParagraph().withParagraph().withParagraph()`~
`.withParagraphs(3)`
?

packages/fixture-generator/src/__tests__/mock-article-slice.test.ts
```diff
@@ -0,0 +1,10 @@
+import ArticleSlice from "../mock-article-slice";
+
+describe("The Mock Image", () => {
+  it("returns the minimum articleSlice requirements", () => {
+    const articleSlice = ArticleSlice(3);
```
Belema: This suggests that you are creating 3 items? but only asserting on the first one..

packages/fixture-generator/src/__tests__/mock-article.test.ts
```diff
@@ -0,0 +1,20 @@
...
+describe("The mock Article", () => {
+  it("should return the minimum article type requirements", () => {
+    const mockArticle = new MockArticle().create();
+    expect(mockArticle).toHaveProperty("id");
+    expect(mockArticle).toHaveProperty("keywords");
+    expect(mockArticle).toHaveProperty("publicationName");
+  });
+
+  it("should return a sunday times article", () => {
+    const mockArticle = new MockArticle().withSundayTimes().create();
```
Belema: `MockArticle().`~`with`~`SundayTimes().create()`

packages/fixture-generator/src/__tests__/mock-markup.test.ts
L0wry: mm agree this is a bit odd, it's more around following a convention that can be reused across the package so there is some continuity.

Open to suggestions

packages/fixture-generator/src/__tests__/mock-article.test.ts
L0wry: again this was more about convention. Happy to change if you feel it's not required

packages/fixture-generator/src/__tests__/mock-markup.test.ts
Belema: I am referring to the `new` + `.create()`. One or the other is fine, but both?...

packages/fixture-generator/src/__tests__/mock-article.test.ts
Belema: What do you mean by convention?
I do think that `MockArticle().SundayTimes()`, `MockArticle().forSundayTimes()`, `MockArticle().ofSundayTimes()`, `MockArticle().inSundayTimes()`, etc... all sound better than `MockArticle().withSundayTimes()`. No worries if you disagree though..

packages/fixture-generator/tsconfig.json
```diff
@@ -0,0 +1,62 @@
+{
+  "compilerOptions": {
+    /* Basic Options */
```
Belema: Any chance we can remove the comments?

packages/fixture-generator/src/__tests__/mock-markup.test.ts
luciagore: changed this for you!

packages/fixture-generator/src/__tests__/mock-article-slice.test.ts
luciagore: changed this

packages/fixture-generator/src/__tests__/mock-article.test.ts
L0wry: so my thoughts were that any generator we have should follow a similar convention such as. I think this is kind of what @nathggns  was saying about having a defined interface for gens.

packages/fixture-generator/src/__tests__/mock-markup.test.ts
nathggns: I agree with @Belema

packages/fixture-generator/src/types.ts
nathggns: Gotcha.

packages/fixture-generator/src/types.ts
nathggns: Okay!

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,98 @@
...
+      {
+        name: "ad",
+        attributes: {},
+        children: []
+      }
+    ];
+
+    const mockMarkup = new MockMarkup().withXParagraphs(5).withAd().create()
+
+    expect(mockMarkup).toEqual(expected)
```
nathggns: This test may be more readable / maintainable with an [inline snapshot test](https://jestjs.io/docs/en/snapshot-testing#inline-snapshots)?

packages/fixture-generator/src/mock-article.ts
```diff
@@ -0,0 +1,72 @@
...
+      section: SectionName.business,
+      shortHeadline: "shortheadline",
+      shortIdentifier: "37b27qd2s",
+      standfirst: "standfirst",
+      title: "title",
+      summary: new MockMarkup().withParagraph().create()
+    };
+  }
+
+  withRelatedArticles(count = 3) {
```
nathggns: The `with` notation typically suggests that you're creating a new instance of a thing with the changes, not modifying the existing instance.

packages/fixture-generator/tsconfig.json
kaliabadi: this is fixed

packages/fixture-generator/src/mock-article.ts
L0wry: any other suggestions @nathggns ?? 
i'm open to anything as you guys will be using it, but in struggling think of better naming as it is 
`withRelatedArticles`

nathggns: In this case, `set` is a more accurate prefix than `with`. But I'm not completely fussed.

Belema: ..on a different note, I wonder why there is what looks like a random default value (3?)

packages/fixture-generator/src/__tests__/mock-markup.test.ts
L0wry: I did not know this existed for things outside of components, thanks!

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+describe("get markup", () => {
+
+  it("returns an array", () => {
+    expect(new MockMarkup().create()).toEqual([])
+  })
+
+  it("should be able to generate paragraphs of markup", () => {
+    const mockMarkup = new MockMarkup().withParagraph().create()
+    for (const markup of mockMarkup) {
```
Belema: We are looping over one element.. wouldn't it be better to write an expectation on the whole shape of the object?

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,98 @@
...
+      expect(markup.name).toBe('paragraph')
+    }
+  })
+
+  it("should be able to generate ads of markup", () => {
+    const mockMarkup = new MockMarkup().withAd().create()
+    for (const markup of mockMarkup) {
```
Belema: Same as above???

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  })
+
+  it("should be able to generate paragraphs of markup", () => {
+    const mockMarkup = new MockMarkup().withParagraph().create()
+    for (const markup of mockMarkup) {
+      expect(markup.name).toBe('paragraph')
+    }
+  })
+
+  it("should be able to generate ads of markup", () => {
```
Belema: `should be able to generate markup with ads` ?

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    const mockMarkup = new MockMarkup().withAd().create()
+    for (const markup of mockMarkup) {
+      expect(markup.name).toBe('ad')
+    }
+  })
+
+  it("should be able to generate inline of markup", () => {
```
Belema: `should be able to generate markup with inlines`?

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    const mockMarkup = new MockMarkup().withInline().withInline().create()
+    for (const markup of mockMarkup) {
```
Belema: Similar situation as above even though there are two inlines in this test? (Why btw?)

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,98 @@
+import MockMarkup from "../mock-markup"
+
+describe("get markup", () => {
+
+  it("returns an array", () => {
+    expect(new MockMarkup().create()).toEqual([])
+  })
+
+  it("should be able to generate paragraphs of markup", () => {
```
Belema: `should be able to generate markup with paragraphs`?

packages/fixture-generator/src/mock-markup.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+};
+
+class MockMarkup {
+  markup: Markup;
+  constructor() {
+    this.markup = [];
+  }
+
+  withXParagraphs(length: number) {
```
Belema: `withParagraphs`?

please

L0wry: i have done this x

packages/fixture-generator/src/__tests__/mock-markup.test.ts
L0wry: lol 🤦‍♂️

packages/fixture-generator/src/mock-markup.ts
nathggns: I might have done `setParagraphCount`, but that's just me. ;)

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,141 @@
+import MockMarkup from "../mock-markup";
+
+describe("get markup", () => {
+  it("returns an array", () => {
+    expect(new MockMarkup().get()).toEqual([]);
+  });
+
+  it("should be able to generate paragraph of markup", () => {
+    expect(new MockMarkup().setParagraph().get()).toMatchInlineSnapshot(`
```
Belema: That expectation is not as nice as,
```JavaScript
const mockMarkup = new MockMarkup().setParagraph().get();
expect(mockMarkup).toMatchObject([ { name: "paragraph" } ]);
```

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      },
+    ],
+    "name": "paragraph",
+  },
+]
+`);
+  });
+
+  it("should be able to generate markup with ads", () => {
+    expect(new MockMarkup().setAd().get()).toMatchInlineSnapshot(`
```
Belema: That expectation is not as nice as,
```JavaScript
const mockMarkup = new MockMarkup(). setAd().get();
expect(mockMarkup).toMatchObject([ { name: "ad" } ]);
```

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    ],
+    "name": "inline",
+  },
+]
+`);
+  });
+
+  it("should generate large markup shapes", () => {
+    expect(
+      new MockMarkup()
```
Belema: This new `setXXX` syntax is very odd to read here. You are not setting anything, but adding elements to the markup. The former `withXXX` read better...

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,141 @@
...
+        .setParagraphs(5)
+        .setAd()
+        .get()
+    ).toMatchInlineSnapshot(`
```
Belema:
```JavaScript
expect(mockMarkup).toMatchObject([
  { name: "paragraph" },
  { name: "paragraph" },
  { name: "paragraph" },
  { name: "paragraph" },
  { name: "paragraph" },
  { name: "ad" }
]);
```

packages/fixture-generator/src/__tests__/mock-markup.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    "attributes": Object {},
+    "children": Array [],
+    "name": "ad",
+  },
+]
+`);
+  });
+
+  it("should be able to generate inline of markup", () => {
+    expect(new MockMarkup().setInline().get()).toMatchInlineSnapshot(`
```
Belema: Same as above..

packages/fixture-generator/src/__tests__/mock-markup.test.ts
nathggns: Find the discussion and why I encouraged moving from `with` here https://github.com/newsuk/times-components/pull/1484#discussion_r235965328

Belema: I saw that, but I really dislike the result, as it is not fluent anymore now. `add` would have been a much better choice than `set`.
```JavaScript
new MockMarkup().setParagraphs(3).setAd().setParagraphs(2).get();
```
makes be believe that there will only be 2 paragraphs in the result given that `setParagraphs(2)` seems to override `setParagraphs(3)`.
I didn't like `with` much, but it was better. Even better would be,
```JavaScript
new MockMarkup().addParagraphs(3).addAds(1).addParagraphs(2).get();
```
or even,
```JavaScript
new MockMarkup([Paragraphs(3), Ads(1), Paragraphs(2)]).get();
```
or anything else really...

kaliabadi:
🥇

Belema:
This PR has turned off most of our e2e tests. It should not have been merged.

PR #1485, opened 2018-11-14
fix: interactives

nathggns:
Fixes interactives on web by manually adding them to the page using DOM apis, as React does not work well with them. 

<s>Depends on https://github.com/newsuk/times-components/pull/1481</s>

I've rebased this to master, so it no longer depends on the above (to make comparison much easier), but SSR won't actually work until the above is merged too.


packages/interactive-wrapper/__tests__/web/__snapshots__/interactive-wrapper.test.js.snap
```diff
@@ -1,15 +1,3 @@
...
-    rel="import"
-  />,
-  <chapter-header
-    chaptercounter="Chapter%20one"
-    heading="A heading"
-    standfirst="A standfirst"
-  />,
-]
-`;
+exports[`renders correctly 1`] = `"<link href=\\"//components.timesdev.tools/lib2/times-chapter-header-1.0.0/chapter-header.html\\" rel=\\"import\\"><chapter-header chaptercounter=\\"Chapter%20one\\" heading=\\"A heading\\" standfirst=\\"A standfirst\\"></chapter-header>"`;
```
colincclark: why is it all on one line?

packages/article/__tests__/web/__snapshots__/article.web.test.js.snap
```diff
@@ -18,13 +18,7 @@ exports[`1. a full article with all content items 1`] = `
...
       </p>
       <div>
-        <link
-          href="https://interactive.io"
-          rel="import"
-        />
-        <times-datawrapper
-          chart-id="csmgb"
-        />
+        <div />
```
colincclark: is there anything else we can show in the snapshot? what would showing the ref look like?

packages/interactive-wrapper/__tests__/web/__snapshots__/interactive-wrapper.test.js.snap
nathggns: Because its a string, as I'm snapshotting the innerHTML of the node this is mounted within. If you snapshot the react tree, all you will get is "div".

packages/article/__tests__/web/__snapshots__/article.web.test.js.snap
nathggns: Yeah its a tough one isn't in. I wish it just showed `<InteractiveWrapper />`. Really no idea how to make that happen though. 

I don't think showing the ref would particularly help.

tuncaulubilge: You should be able to mock InteractiveWrapper like this:
`jest.mock("@times-components/interactive-wrapper", () => "InteractiveWrapper");`

nathggns: 👍

packages/interactive-wrapper/__tests__/web/__snapshots__/interactive-wrapper.test.js.snap
colincclark: Just snapshot the container, not the innerHTML bit - then it will be formatted nicely with the div

nathggns: Ah I had no idea this would work. Thanks for the suggestion!

PR #1486, opened 2018-11-14
fix: add missing ad config method to ssr topic

colincclark:


packages/ssr/src/lib/graphql-logging-link.js
```diff
@@ -6,6 +6,7 @@ class LogLink extends ApolloLink {
     this.uri = uri;
     this.logger = logger;
   }
+
```
Belema: Do we need this?

packages/ssr/src/lib/run-server.js
```diff
@@ -8,9 +8,9 @@ const { fragmentMatcher } = require("@times-components/schema");
 const { getDataFromTree } = require("react-apollo");
 const { InMemoryCache } = require("apollo-cache-inmemory");
 const ReactDOMServer = require("react-dom/server");
-const safeStringify = require("./safe-stringify");
 const { ServerStyleSheet } = require("styled-components");
 const { ApolloLink } = require("apollo-link");
+const safeStringify = require("./safe-stringify");
```
Belema: Why?

packages/ssr/src/lib/graphql-logging-link.js
colincclark: yep, that is what happens when you format that package

packages/ssr/src/lib/run-server.js
colincclark: formatting

packages/ssr/src/lib/graphql-logging-link.js
Belema: No

colincclark: it looks better

packages/ssr/src/lib/run-server.js
colincclark: Changed back, must have been a deps issue

PR #1487, opened 2018-11-14
feat: Add referral url to article analytics

tuncaulubilge:
Analytics team wants to get the referralUrl whenever user launches an article through deeplinking, especially to see query params such as utm_tracking etc.

This "referralUrl" is passed through the native apps into our article component. 
Part of the Replat-4348 ticket

### DIFF
--- a/packages/article-main-standard/src/article-main-standard.js
+++ b/packages/article-main-standard/src/article-main-standard.js
@@ -88,7 +88,8 @@ class ArticlePage extends Component {
       onTwitterLinkPress,
       onVideoPress,
       onViewed,
-      receiveChildList
+      receiveChildList,
+      referralUrl
     } = this.props;
 
     return (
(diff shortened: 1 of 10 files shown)


packages/article-main-standard/src/article-main-standard.js
```diff
@@ -107,6 +108,7 @@ class ArticlePage extends Component {
         onVideoPress={onVideoPress}
         onViewableItemsChanged={onViewed ? this.onViewableItemsChanged : null}
         receiveChildList={receiveChildList}
+        referralUrl={referralUrl}
```
nathggns: In general I don't like this kind of prop-baton-passing. Not a reason to reject this here, but something to consider.

tuncaulubilge: 10 files changed just to pass down a prop is not a good sign anyway :) but not sure what's the alternative in here

PR #1493, opened 2018-11-16
Feat: Comments template

jimhunty:
Adding support for the comment template as stated here: https://nidigitalsolutions.jira.com/browse/REPLAT-599


packages/article-main-comment/src/article-prop-types/article-prop-types.js
```diff
@@ -0,0 +1,25 @@
+import PropTypes from "prop-types";
+import {
+  articlePropTypesBase,
+  articleDefaultPropsBase
+} from "./article-prop-types.base";
+
+const articlePropTypes = {
+  ...articlePropTypesBase,
+  onAuthorPress: PropTypes.func.isRequired,
```
nathggns: I wonder if in the future these links can go in the context rather than being passed around?

packages/article-main-comment/src/article-prop-types/article-prop-types.web.js
```diff
@@ -0,0 +1,14 @@
+import {
+  articlePagePropTypes,
+  articlePageDefaultProps
+} from "./article-prop-types.base";
+
+const articlePropTypes = {
```
nathggns: Why does this file need to exist?

packages/article-main-comment/src/article-standfirst/article-standfirst.web.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
```
nathggns: Not blocking the PR for it, but if you want to this component looks like a *great* place to try using styled components on native. The native / web files are almost identical.

packages/article-main-comment/src/styles/responsive.web.js
```diff
@@ -0,0 +1,109 @@
...
+    height: ${spacing(3)};
+    margin: 0 ${spacing(2)};
+    width: 1px;
+  }
+`;
+
+export const StandfirstContainer = styled(Text)`
+  @media (min-width: ${breakpoints.medium}px) {
+    font-size: ${fontSizes.infoTitle}px;
+    lline-height: ${lineHeight({
```
nathggns: Eeek this seems wrong.

packages/article-main-comment/src/article-prop-types/article-prop-types.web.js
jimhunty: RN is known to get confused. This could become `.web` and be imported into the "index" but in the past, this has caused a cyclical effect. Maybe ok now, but didn't want to risk it and wanted to follow the patterns in other packages

nathggns: It seems to me that this file only imports and re-exports what is coming from `.base`. Just import that file in the consumer of this file instead, and you can delete this file.

packages/article-main-comment/src/article-prop-types/article-prop-types.js
jimhunty: I think you want to use `Context` sparingly, opting in when you have to and at the lowest possible layer. This would add `Context` at quite a high level which could lead to unnecessary rerenders. These can be managed though. Worth a dicsussion.

packages/article-main-comment/src/article-prop-types/article-prop-types.web.js
nathggns: "The pattern" seems to be that we have `.web` and `.base` files when they're different. But these are the same. Not going to reject this PR but I genuinely think we should look at reducing the amount of code we have that just adds layers of indirection and doesn't actually do anything.

tuncaulubilge:
I think we should focus on centralising article tests and showcases, as it'll get out of hand quite soon. It might not be worth blocking this PR but we should keep template specific tests to minimum (maybe just headers with a huge mocked ArticleContent) and move shared tests to the article package itself (or the new article-index package we discussed)

jimhunty:
Regarding centralising tests, I raised this in the last mega PR. The alternative opinion is that repeating tests is not necessarily a bad thing, especially in a monorepo environment. However, I think it would improve the quality of the testing on the template packages in general and should be tackled as part of tech debt.

tuncaulubilge:
> Regarding centralising tests, I raised this in the last mega PR. The alternative opinion is that repeating tests is not necessarily a bad thing, especially in a monorepo environment. However, I think it would improve the quality of the testing on the template packages in general and should be tackled as part of tech debt.

Maintaining tests and updating snapshots in 3+ packages every time can be really cumbersome imho. I find myself spending 80% of my time fixing snaps, tests and showcases. I think this is worth a second discussion. (article-test-tools package incoming :))

PR #1494, opened 2018-11-16
fix: pull quotes markup

tuncaulubilge:
A couple of features were missing in the pull-quotes. 

Added features:
- Support markups inside pull-quotes
- Add "text" field that's used in the ast for extra caption text
- Added section colouring to the pull-quote quotation marks and caption

Also fixed the keys inside article body to clear up the warnings

screen shot 2018-11-16 at 16 13 11
screenshot_1542385151

### DIFF
--- a/packages/article/__tests__/web/__snapshots__/article-with-style.web.test.js.snap
+++ b/packages/article/__tests__/web/__snapshots__/article-with-style.web.test.js.snap
@@ -387,7 +387,9 @@ exports[`full article with style 1`] = `
           <div
             className="c10 S1"
           >
-            <PullQuote />
+            <PullQuote>
+              The pull quote content
+            </PullQuote>
           </div>
         </div>
         <div
--- a/packages/pull-quote/src/pull-quote.js
+++ b/packages/pull-quote/src/pull-quote.js
@@ -2,8 +2,8 @@ import React from "react";
 import PullQuoteBase from "./pull-quote.base";
 import { propTypes, defaultProps } from "./pull-quote-prop-types";
 
-const PullQuote = ({ caption, ...props }) => (
-  <PullQuoteBase {...props}>{caption}</PullQuoteBase>
+const PullQuote = ({ children, ...props }) => (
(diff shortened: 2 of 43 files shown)


packages/pull-quote/src/pull-quote.js
```diff
@@ -2,8 +2,8 @@ import React from "react";
 import PullQuoteBase from "./pull-quote.base";
 import { propTypes, defaultProps } from "./pull-quote-prop-types";
 
-const PullQuote = ({ caption, ...props }) => (
-  <PullQuoteBase {...props}>{caption}</PullQuoteBase>
+const PullQuote = ({ children, ...props }) => (
```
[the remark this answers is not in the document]
tuncaulubilge: Good point, and in that case we don't even need PullQuote. Updated it to export PullQuoteBase directly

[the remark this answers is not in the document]
tuncaulubilge: See c8f7827

jimhunty:
@tuncaulubilge Is the Pull Quote size correct?

tuncaulubilge:
> @tuncaulubilge Is the Pull Quote "quote" size correct?

I haven't changed the quote mark sizes, so it should be same

packages/article/__tests__/web/__snapshots__/article-with-style.web.test.js.snap
```diff
@@ -387,7 +387,9 @@ exports[`full article with style 1`] = `
           <div
             className="c10 S1"
           >
-            <PullQuote />
+            <PullQuote>
```
colincclark: Shouldn't this be mocked outside of the `PullQuote` package?

tuncaulubilge: What do you mean?

colincclark: As discussed, this is all fine

PR #1495, opened 2018-11-16
fix: SSR - separate ad config for article, topic and author profile

aronshamash:
- separate ad config for article, topic and author profile in `ssr` package


packages/ssr/src/lib/ads/make-author-profile-ad-config.js
```diff
@@ -0,0 +1,38 @@
+const { defaultClient, defaultServer } = require("./base-ad-config");
+
+module.exports = {
+  defaultClient: () => ({
+    ...defaultClient,
+    pageTargeting: {
+      aid: "null",
```
nehasri89: `pageTargeting` object is the same for `author` and `Topic` and marginally diff for articles, you might want to DRY it up

PR #1496, opened 2018-11-19
Feat/test cypress

nehasri89:
This PR adds cypress tests and fixtures for Author Profile and Topic pages.

Tools are separately looking at Mock TPA to figure out how we can get a variety in data.


packages/ssr/__tests__/integration/author-profile.js
```diff
@@ -0,0 +1,34 @@
...
+    cy.url().should(
+      "eq",
+      "http://localhost:3000/article/97c64f20-cb67-11e4-a202-50ac5def393a"
+    );
+  });
+
+  it("Next and Previous Pagination works", () => {
+    cy.get('div[data-testid="pagination-button-next"]')
+      .first()
+      .click();
```
tuncaulubilge: I don't have any experience with Cypress but I think we should do some assertions on the next page to see if the content changes

packages/ssr/__tests__/integration/author-profile.js
```diff
@@ -0,0 +1,34 @@
+describe("AuthorProfile", () => {
+  beforeEach(() => {
+    cy.visit("/profile/fiona-hamilton");
+  });
+
+  it("Author head has required elements", () => {
+    cy.get('div[data-testid="author-head"]');
+    cy.get('h1[data-testid="author-name"]');
```
tuncaulubilge: What are we testing in here? Are we just the existence of these?

packages/ssr/__tests__/integration/author-profile.js
```diff
@@ -0,0 +1,34 @@
...
+
+  it("Click on Author list takes you to an article", () => {
+    cy.get('div[data-testid="97c64f20-cb67-11e4-a202-50ac5def393a.0"]').click();
+    cy.url().should(
+      "eq",
+      "http://localhost:3000/article/97c64f20-cb67-11e4-a202-50ac5def393a"
+    );
+  });
+
+  it("Next and Previous Pagination works", () => {
```
Belema: How is it testing the functionality?

packages/ssr/__tests__/integration/author-profile.js
```diff
@@ -0,0 +1,34 @@
...
+
+  it("Author head has required elements", () => {
+    cy.get('div[data-testid="author-head"]');
+    cy.get('h1[data-testid="author-name"]');
+    cy.get('h2[role="heading"]');
+    cy.get('div[data-testid="author-bio"]');
+  });
+
+  it("Author has article elememts on the page", () => {
+    cy.get('div[data-testid="97c64f20-cb67-11e4-a202-50ac5def393a.0"]');
```
Belema:
```JavaScript
const article = (id) => `'div[data-testid="${id}.0"]'`
```
then ,
```JavaScript
cy.get(article('97c64f20-cb67-11e4-a202-50ac5def393a'));
```

packages/ssr/__tests__/integration/author-profile.js
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+  it("Click on Author list takes you to an article", () => {
```
Belema: `"Click on an article in the author list takes you to the corresponding article"`?

packages/ssr/__tests__/integration/author-profile.js
```diff
@@ -0,0 +1,34 @@
+describe("AuthorProfile", () => {
+  beforeEach(() => {
+    cy.visit("/profile/fiona-hamilton");
+  });
+
+  it("Author head has required elements", () => {
+    cy.get('div[data-testid="author-head"]');
```
Belema: There are no assertions. Reading the docs, I am not sure if a get timeout will make the test fail.
Would it not be better to be explicit about what we expect these elements to be? e.g. that we expect the author name to be Fiona Hamilton?

nehasri89: I know, its not clear in the docs, but I have tested it fails if the element is not present on the page. I can make it more verbose if it is clear. I have referenced the wdio e2e tests in render
https://github.com/newsuk/cps-content-render/blob/master/src/test/e2e/page-objects/author.page.js#L56

packages/ssr/__tests__/integration/author-profile.js
nehasri89: 👍

packages/ssr/__tests__/integration/author-profile.js
nehasri89: yes, we are testing the existence of these fields. I have tried to follow the pattern we have used in render e2e https://github.com/newsuk/cps-content-render/blob/master/src/test/e2e/page-objects/author.page.js#L56

packages/ssr/__tests__/integration/author-profile.js
nehasri89: I have tried to reference the e2e tests in render https://github.com/newsuk/cps-content-render/blob/master/src/test/e2e/author-profile.test.js#L15, can you pls suggest what else we could check?

PR #1499, opened 2018-11-19
feat: Comment template design review

jimhunty:
This includes:

- Changes to the headline line height across all templates
- Fixing the related article label on Android
- Comment header responsive fixes
- Wide breakpoint improvements to improve readability


packages/article-flag/__tests__/ios/__snapshots__/article-flag-with-style.ios.test.js.snap
```diff
@@ -23,10 +23,11 @@ exports[`1. red article flag 1`] = `
       Object {
         "color": "red",
         "fontFamily": "TimesDigitalW04-RegularSC",
-        "fontSize": 12,
+        "fontSize": 14,
         "fontWeight": "400",
         "letterSpacing": 1.4,
         "lineHeight": 12,
+        "marginTop": 5,
```
tuncaulubilge: Do we want this extra margin? On Android, this margin is added so the text is aligned with the diamond. On my branch I added a -5 margin to the containing view, so we don't have a huge space between the flag and the headline. We should double check this one with Jake

jimhunty: As discussed, we will merge the PR and pick up any outstanding items in @tuncaulubilge next PR.

PR #1500, opened 2018-11-19
chore: remove flow

colincclark:
This PR is the next in a series to upgrade our suite of packages. It removes flow. A technical decision was made to move replatforming repos towards TypeScript. In this vein, and to remove confusion about which type enforcement solution we are using, this PR removes flow. The type checking that exists within this project does not give us much anyway, and adds an extra technology to learn which is unwarranted at this time.

- although flow is only really used in three files, this PR is large because flow was added as dependencies across the project
- NOTE: `eslint-plugin-flowtype` has been left in until the babel 7 upgrade otherwise everything 💥


colincclark:
Cannot get this branch to work. Moved to https://github.com/newsuk/times-components/pull/1502

PR #1507, opened 2018-11-22
fix: script to run from bitrise

nehasri89:

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "packages:publish-dry-run": "yarn packages:publish --no-git-tag-version --no-push",
     "clean-snaps": "rm -rf dextrose/snappy/*.png",
     "commit": "git-cz",
-    "test:e2e": "lerna run test:integration --stream"
+    "test:e2e": "lerna run bundle && lerna run test:integration --stream"
   },
   "jest": {
     "preset": "react-native",
--- a/package.json
+++ b/package.json
@@ -52,8 +52,7 @@
     "packages:publish-dry-run": "yarn packages:publish --no-git-tag-version --no-push",
     "clean-snaps": "rm -rf dextrose/snappy/*.png",
     "commit": "git-cz",
-    "bundle": "lerna run bundle",
-    "test:e2e": "lerna run test:integration --stream"
+    "test:e2e": "lerna run bundle && lerna run test:integration --stream"
   },
   "jest": {
     "preset": "react-native",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -51,7 +51,8 @@
     "packages:publish": "lerna publish --conventional-commits --yes --concurrency=1 --exact",
     "packages:publish-dry-run": "yarn packages:publish --no-git-tag-version --no-push",
     "clean-snaps": "rm -rf dextrose/snappy/*.png",
-    "commit": "git-cz"
+    "commit": "git-cz",
+    "test:e2e": "lerna run test:integration --stream"
```
parkjh: @nehasri89 where is this called from?

nehasri89: this will go as a step in bitrise workflow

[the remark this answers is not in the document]
nehasri89: The idea of integration at the moment is for the render as a client, and as SSR would be the new website replacing render, hence the place for integration tests. 
But if we feel we want to have int tests for more packages, we can always have them. Also we need to understand the bits we want to integration test for other packages, are not already covered with  snapshotting.

package.json
```diff
@@ -51,7 +51,8 @@
     "packages:publish": "lerna publish --conventional-commits --yes --concurrency=1 --exact",
     "packages:publish-dry-run": "yarn packages:publish --no-git-tag-version --no-push",
     "clean-snaps": "rm -rf dextrose/snappy/*.png",
-    "commit": "git-cz"
+    "commit": "git-cz",
+    "test:e2e": "lerna run bundle && lerna run test:integration --stream"
```
tuncaulubilge: We do have a bundling step inside "prepublishOnly" in each package that needs bundling. It might be a good idea to split bundle as a separate step so we don't end up bundling everything twice, as it takes time to bundle

packages/ssr/package.json
```diff
@@ -15,7 +15,7 @@
     "start": "webpack --config=src/standalone-renderer/webpack.config.js && node src/standalone-renderer/app.js",
     "start:testservers": "node ./mock-server & wait-on tcp:4000 && GRAPHQL_ENDPOINT=http://localhost:4000/graphql yarn start & wait-on tcp:3000",
     "stop:testservers": "kill $(lsof -t -i:3000,4000)",
-    "test:integration": "yarn bundle:dev && yarn start:testservers; cypress run; yarn stop:testservers"
+    "test:integration": "yarn bundle:dev && yarn start:testservers; cypress run; exitCode=$?; yarn stop:testservers; exit $exitCode"
```
Belema: `"yarn bundle:dev && yarn start:testservers && cypress run; exitCode=$?; yarn stop:testservers; exit $exitCode"`

PR #1510, opened 2018-11-23
feat: Add support for centred captions

aronshamash:
- Create a new `<CenteredCaption />` component in the `caption` package 
- First use will be the `article-magazine-comment` template.

### Android
screenshot 2018-11-23 at 14 48 59

### Web
screenshot 2018-11-23 at 14 51 01

### DIFF
--- a/packages/caption/src/centred-caption.js
+++ b/packages/caption/src/centred-caption.js
@@ -1,0 +1,24 @@
+import React from "react";
+import Caption from "./caption";
+import { defaultProps, propTypes } from "./caption-prop-types";
+
+const CentredCaption = ({ children, credits, style, text }) => (
+  <Caption
+    credits={credits}
+    style={{
+      ...style,
+      text: {
+        ...style.text,
+        textAlign: "center"
+      }
+    }}
+    text={text}
+  >
+    {children}
+  </Caption>
+);
+
+CentredCaption.propTypes = propTypes;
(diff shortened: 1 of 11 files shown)


packages/caption/src/centred-caption.js
```diff
@@ -0,0 +1,23 @@
...
+import Caption from "./caption";
+import { defaultProps, propTypes } from "./caption-prop-types";
+
+const CentredCaption = ({ children, credits, style, text }) => (
+  <Caption
+    credits={credits}
+    style={{
+      ...style,
+      text: {
+        textAlign: "center"
```
tuncaulubilge: Doesn't this override the whole text style? I think it should be:
```
text: {
  ...style.text,
  textAlign: "center"
}
```

packages/caption/src/centred-caption.js
```diff
@@ -0,0 +1,24 @@
...
+import Caption from "./caption";
+import { defaultProps, propTypes } from "./caption-prop-types";
+
+const CentredCaption = ({ children, credits, style, text }) => (
+  <Caption
+    credits={credits}
+    style={{
+      ...style,
+      text: {
+        ...style.text,
```
nathggns: What happens if `style.text` is undefined? I think this blows up on some platforms?

aronshamash: Example snapshot added.

PR #1512, opened 2018-11-23
BREAKING CHANGE: Article template selector

jimhunty:
This PR changes the purpose of the article package from an Article Skeleton to a template selector.

This in-turn makes changes to `SSR` and `Pages` to use the new template selection and updates the provider query to ask for that field.

Please note that author is not implemented in TPA yet and therefore the author object is added to the story for demo purposes.


packages/article/README.md
```diff
@@ -55,12 +22,24 @@ yarn lint
...
+
+Or the tests for all platforms can be run
+
+```
+yarn test:all
+```
+
+Visit the official
+
+<!-- Add the storybook link here. -->
```
tuncaulubilge: Need the link here

packages/article/src/article.js
```diff
@@ -1,172 +1,25 @@
...
-        width: normaliseWidth(screenWidthInPixels())
-      };
-    } else {
-      this.state = {
-        dataSource: {}
-      };
-    }
+const Article = props => {
+  const { template } = props.article;
+  switch (template) {
```
aronshamash: Not a big fan of switch statements. Can we use an object literal eg:
```
const templates = {
'mainstandard': 'ArticleMainStandard',
'maincomment': 'ArticleMainComment
);
```

tuncaulubilge: These will also be used in theming, so it might be a good idea to put these in a shared package

aronshamash: let's do that in another PR.

packages/article/src/article.js
```diff
@@ -1,172 +1,19 @@
...
-Article.propTypes = {
-  ...articlePropTypes,
-  onAuthorPress: PropTypes.func.isRequired,
-  onCommentGuidelinesPress: PropTypes.func.isRequired,
-  onCommentsPress: PropTypes.func.isRequired,
-  onLinkPress: PropTypes.func.isRequired,
-  onRelatedArticlePress: PropTypes.func.isRequired,
-  onTwitterLinkPress: PropTypes.func.isRequired,
-  onVideoPress: PropTypes.func.isRequired
+  const Component = templates[template];
```
aronshamash: 👍

PR #1516, opened 2018-11-26
Feat/theming

nehasri89:
This  covers the first end to end slice for Theming. This will be the base for future PRs

### DIFF
--- a/packages/styleguide/src/theme/theme-factory.js
+++ b/packages/styleguide/src/theme/theme-factory.js
@@ -1,18 +1,38 @@
 import sectionColours, { secondarySectionColours } from "../colours/section";
 
 const sectionColourPicker = (section, template) => {
-  switch (template) {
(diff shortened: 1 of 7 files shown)


packages/styleguide/src/theme/theme-factory.js
```diff
@@ -0,0 +1,28 @@
...
+    },
+    maincomment: {
+      ...sectionColours
+    },
+    mainstandard: {
+      ...sectionColours
+    }
+  };
+
+  return config[template][section] || config[template].default;
```
tuncaulubilge: This should work when template is null as well, which happens on the loading state. We should have a default theme fallback

nehasri89: 👍

aronshamash: updated. Please re-review.

PR #1518, opened 2018-11-26
feat: article link analytics

tuncaulubilge:
REPLAT-3197

Analytics team wanted to track in article link presses and related article presses on the native apps. This PR adds the capability to track events for related articles and article links


packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -79,6 +79,7 @@ const ArticleRow = ({
         element: (
           <ArticleLink
             key={key}
+            linkType={attributes.type}
```
jimhunty: Looks like we should destructure these?

tuncaulubilge: Done :+1:

PR #1519, opened 2018-11-27
feat: magazine comment template

nathggns:
Replacement for #1511

This adds the magazine comment template, mostly by copying the main comment template and adding the lead asset from main standard.

There's a few things that are outside the scope of this due to the lack of ability to theme based on template at the moment, and some other limitations, including:

- Centered captions on the lead assets
- Correcting the fixtures used in the template showcase
- Fonts
- Colours
- Drop Caps


packages/article-magazine-comment/__tests__/mocks.native.js
```diff
@@ -0,0 +1,33 @@
...
+jest.mock("@times-components/brightcove-video", () => "BrightcoveVideo");
+jest.mock("@times-components/button", () => "Button");
+jest.mock("@times-components/image", () => ({
+  __esModule: true,
+  default: "Image",
+  ModalImage: "ModalImage"
+}));
+jest.mock("@times-components/pull-quote", () => "PullQuote");
+jest.mock("@times-components/related-articles", () => "RelatedArticles");
+jest.mock("@times-components/tracking", () => ({
```
tuncaulubilge: I deleted the tracking mocks in my analytics pr, as they are breaking things and not really needed. Can you also delete it from this template?

nathggns: 👍

PR #1520, opened 2018-11-27
fix: remove ifs

parkjh:


parkjh:
@tuncaulubilge will test the android app with these changes

PR #1523, opened 2018-11-28
chore: update main comment with improvements from magazine comment

nathggns:
Main comment had a couple of changes that were made after magazine comment was created, and magazine comment improved a couple of things that can and should be copied back to main comment. That's what this PR does.

### DIFF
--- a/packages/article-main-comment/src/article-prop-types/article-prop-types.base.js
+++ b/packages/article-main-comment/src/article-prop-types/article-prop-types.base.js
@@ -12,7 +12,7 @@ const articlePagePropTypes = {
     })
   }),
   isLoading: PropTypes.bool,
-  receiveChildList: PropTypes.func.isRequired,
+  receiveChildList: PropTypes.func.isRequired
 };
 
 const articlePageDefaultProps = {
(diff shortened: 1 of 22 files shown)


aronshamash:
Looks fine but I also noticed that in `article` storybook we only have `mainstandard` and `maincomment`

nathggns:
> Looks fine but I also noticed that in `article` storybook we only have `mainstandard` and `maincomment`

@aronshamash I'll update this now

packages/article-main-comment/src/article-prop-types/article-prop-types.base.js
```diff
@@ -12,7 +12,7 @@ const articlePagePropTypes = {
     })
   }),
   isLoading: PropTypes.bool,
-  receiveChildList: PropTypes.func.isRequired
```
colincclark: articles should always be wrapped in the track scroll depth HOC

nathggns: Oops… think I was over eager in removing things to remove warnings. Will fix for both templates now!

PR #1525, opened 2018-11-28
Fix: Section list in article storybook

aronshamash:
- Get section list from `storybook`


packages/article-magazine-comment/article-magazine-comment.showcase.js
```diff
@@ -126,9 +126,7 @@ const mockArticle = ({
 
 const selectScales = select => select("Scale", scales, scales.medium);
 const selectSection = select =>
-  invert(colours.section)[
```
nehasri89: does magazine comment has a diff section?

aronshamash: we could make it something else like `The Sunday Times Magazine` as that is more likely to use this template than the main `Comment` section. I'll update

PR #1528, opened 2018-11-28
fix: Support null article data on article package

tuncaulubilge:
Native apps can use the article package with no data to show an error view, so the article package should work when article is undefined. Added a default value and a test

### DIFF
--- a/packages/article/src/article.js
+++ b/packages/article/src/article.js
@@ -10,8 +10,8 @@ export const templates = {
 };
 
 const Article = props => {
-  const { article = {} } = props;
-  const { template = "mainstandard" } = article;
+  const { article } = props;
+  const { template = "mainstandard" } = article || {};
 
   const Component = templates[template];
   return <Component {...props} />;
--- a/packages/article/src/article.js
+++ b/packages/article/src/article.js
@@ -11,7 +11,7 @@ export const templates = {
 
 const Article = props => {
   const {
-    article: { template = "mainstandard" }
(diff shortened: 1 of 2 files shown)


packages/article/src/article.js
```diff
@@ -11,7 +11,7 @@ export const templates = {
 
 const Article = props => {
   const {
-    article: { template = "mainstandard" }
+    article: { template = "mainstandard" } = { template: "mainstandard" }
```
nathggns: This syntax was ugly to begin with, and it doesn't actually add support for `null` articles. (null is considered a value by the default argument mechanism). Also, the default value is duplicated which is unnecessary 

It's probably time to just improve the syntax:

```
const { template = "mainstandard" } = props.article || {};
```

nathggns: [Here's what I mean about the default argument mechanism](https://codepen.io/nathggns/pen/jQeYwr?editors=0010#0). It's using function arguments, but it applies to destructuring too.

tuncaulubilge: That looks much better, but airbnb lint is not happy with is as it forces destructing props. I could do this:
```
  const { article = {} } = props;
  const { template = "mainstandard" } = article;
```

nathggns: 🤦‍♂️  _remind me why we use a linter that enforces destructing? (Which often makes code less clear, not more)_ 

Yeah that's fine.

PR #1531, opened 2018-11-29
feat: magazine standard

nathggns:
This is mostly just cloning magazine comment but removing the author image. 

**We really need to find a way to reduce duplication between templates.**


packages/article-magazine-standard/package.json
```diff
@@ -0,0 +1,87 @@
+{
+  "name": "@times-components/article-magazine-standard",
+  "version": "0.1.0",
```
jimhunty: 0.0.1?

nathggns: 👍

PR #1532, opened 2018-11-29
feat: Toggle React Strict Mode

jimhunty:
Ability to turn off strict mode for your storybook stories. This is to aid debugging for the things we are unable to fix for the time being.

### DIFF
--- a/packages/storybook/src/showcase-to-storybook.js
+++ b/packages/storybook/src/showcase-to-storybook.js
@@ -11,7 +11,13 @@ React.Fragment.displayName = "React.Fragment";
 // eslint-disable-next-line react/prop-types
 const StrictWrapper = ({ children }) => <StrictMode>{children}</StrictMode>;
 
-const addStories = (builder, knobs, actions, [child, ...children]) => {
+const addStories = (
+  builder,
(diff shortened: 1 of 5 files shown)


packages/storybook/src/showcase-to-storybook.js
```diff
@@ -11,17 +11,29 @@ React.Fragment.displayName = "React.Fragment";
 // eslint-disable-next-line react/prop-types
 const StrictWrapper = ({ children }) => <StrictMode>{children}</StrictMode>;
 
-const addStories = (builder, knobs, actions, [child, ...children]) => {
+const addStories = (
+  builder,
+  knobs,
+  actions,
+  [child, ...children],
+  strictMode = true
```
nathggns: It would be probably be good to set the default to something coming from `process.env`, perhaps like:

```
strictMode = process.env.NO_STRICT_MODE ? false : true
```

jimhunty: I think that is nicer. Any other thoughts?

nathggns: Just that.

PR #1534, opened 2018-11-30
fix: article list perf issues

tuncaulubilge:
The whole article list is re-rendering whenever an item state is updated, creating a huge memory/cpu monster on topic & author profile pages. Implemented "shouldComponentUpdate" so we don't do unnecessary re-renders on ArticleList and ArticleListItem.

P.S.: No idea how to test this but suggestions are welcome. Current state & prop update tests are working with the update.

Fixes: https://nidigitalsolutions.jira.com/browse/REPLAT-4574

Paired with: @nathggns


packages/article-list/src/article-list-item.js
```diff
@@ -12,73 +12,83 @@ import { propTypes, defaultProps } from "./article-list-item-prop-types";
...
-  const {
-    article,
-    highResSize,
-    imageRatio,
-    isLoading,
-    onPress,
-    showImage
-  } = props;
+class ArticleListItem extends Component {
+  shouldComponentUpdate(nextProps) {
```
nathggns: I think it's regrettable that we've had to use `shouldComponentUpdate` but I understand why we have. It would be good to look into how we can keep prop identity the same in the future but I don't think its worth the time right now.

tuncaulubilge:
@jimhunty react-native suggests using PureComponents for flat list and its children for smooth scrolling. Since we clone the article object down the line, the props are no longer shallow equal, so `shouldComponentUpdate` is required. We could use PureComponents instead but that'll require a good spring cleaning

PR #1535, opened 2018-12-03
Feat/Dropcap Integration

nehasri89:
- This PR mutates AST of the first para and creates a dropcap child. 
- Dropcap is always only the first letter even if it starts with a double quotes. 
- Having a dropcap is an attribute of a template


packages/markup-forest/__tests__/markup-forest.test.js
```diff
@@ -261,6 +262,99 @@ iterator([
...
     }
   },
+  {
+    name: "paragraph for not a dropcap",
+    test: () => {
+      const output = TestRenderer.create(
+        <div>
+          {renderTrees(
+            paragraphAST,
+            {
```
nathggns: You can probably share this object between tests.

packages/markup-forest/src/markup-forest.js
```diff
@@ -1,21 +1,67 @@
...
+    // mutate AST
+    newChildren = mutateAST(children[0], children);
+  }
 
   const renderer = renderers[name];
 
   if (!renderer) return null;
 
-  const initialResult = renderer(key, attributes, [], indx, tree);
+  newTree.children = newChildren;
```
nathggns: This modifies the tree that is passed in, making `renderTree` a mutative function (which I don't think is the intention). That's because `newTree` is assigned by reference to `tree` rather than cloned on line [36](https://github.com/newsuk/times-components/pull/1535/files#diff-c1e79a259fdf40f946ce2af4a3bcb4d3R36).

packages/article-main-comment/article-main-comment.showcase.js
```diff
@@ -49,7 +49,8 @@ const renderArticle = ({
           image:
             "https://feeds.thetimes.co.uk/web/imageserver/imageserver/image/methode%2Ftimes%2Fprod%2Fweb%2Fbin%2F0694e84e-04ff-11e7-976a-0b4b9a1a67a3.jpg?crop=854,854,214,0&resize=400"
         },
-        ...article
+        ...article,
+        template: "maincomment"
```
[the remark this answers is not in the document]
nehasri89: 👍

packages/markup-forest/src/markup-forest.js
tuncaulubilge: +1 on this one. I think this logic should live inside the article package and modify the ast once. Modifying the ast on render tree is quite risky. We got "filter-interactives" in pages package for example, which modifies the ast to disable interactives for native (based on a feature flag). We could do something similar, but preferably inside the article package

packages/markup-forest/src/markup-forest.js
```diff
@@ -1,21 +1,67 @@
...
+      children: [],
+      name: "text"
+    };
+    child = [dropCapElement, newTextElement, ...children.slice(1)];
+  }
+  return child;
+};
+
+export const renderTree = (tree, renderers, key = "0", indx = 0, template) => {
+  const newTree = Object.assign({}, tree);
```
nathggns: `{ ...tree }` is how we tend to do it elsewhere.

packages/article-main-comment/article-main-comment.showcase.js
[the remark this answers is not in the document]
nathggns: #1531 is merged now so may want to rebase / do that for magazine standard template too

packages/markup-forest/src/markup-forest.js
nathggns: I think my comment here is resolved. @tuncaulubilge may still want his to be.

packages/article-main-comment/article-main-comment.showcase.js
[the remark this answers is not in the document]
nehasri89: done

packages/article-skeleton/src/dropcap-util.js
```diff
@@ -0,0 +1,49 @@
+const templateWithDropCaps = [
```
tuncaulubilge: We should have these enums in a shared package. Article package uses and exports templates as a object, I think we should move it to a shared location and use the same object here as well. That'll make refactoring and adding/removing new templates much easier

tuncaulubilge: Just noticed we didn't use any constants in styleguide theme factory as well, so I think this deserves a separate pr, won't block this one :)

packages/article-skeleton/src/dropcap-util.js
```diff
@@ -0,0 +1,49 @@
...
+  let child = firstContentChild;
+  const { name, children } = child;
+  let newChildren = children;
+  if (
+    template &&
+    templateWithDropCaps.includes(template) &&
+    name === "paragraph" &&
+    children.length > 0
+  ) {
+    newChildren = insertDropcap(children[0], children);
```
tuncaulubilge: We can return `{ ...child, children: newChildren }` here and we can convert these to constants

packages/article-skeleton/src/dropcap-util.js
```diff
@@ -0,0 +1,49 @@
...
+      name: "dropCap"
+    };
+    const newTextElement = {
+      attributes: {
+        value: value.slice(1, value.length)
+      },
+      children: [],
+      name: "text"
+    };
+    child = [dropCapElement, newTextElement, ...children.slice(1)];
```
tuncaulubilge: We can return here so we don't need child to be a variable

packages/article-skeleton/src/dropcap-util.js
```diff
@@ -0,0 +1,49 @@
...
+  "magazinecomment"
+];
+
+const insertDropcap = (firstTextChild, children) => {
+  const {
+    name,
+    attributes: { value }
+  } = firstTextChild;
+  let child = children;
+  if (name === "text") {
```
tuncaulubilge: Let's check if value is not empty here, just to be sure

tuncaulubilge: After the empty paragraphs issue couple of weeks ago, I don't trust our ast any more :)

PR #1539, opened 2018-12-03
refactor: article list native

tuncaulubilge:
Further performance improvements on article list

- WindowSize on the flatlist has been reduced to 5 (default value is 21, which eagerly renders -10 +10 screens for smooth scrolling. We don't need such a high look-ahead value, so it has been reduced for better performance)
- RenderItem and onPress functions were being created for item on each render. Moved them to the component itself.


packages/article-list/src/article-list-item.js
```diff
@@ -21,13 +26,20 @@ class ArticleListItem extends Component {
     );
   }
 
+  onItemPress(event) {
+    const {
+      article: { id, url },
+      onPress
+    } = this.props;
```
nathggns: I really dislike that we have a lint rule that makes us do this.

packages/article-list/src/article-list.js
```diff
@@ -78,6 +79,38 @@ class ArticleList extends Component {
...
+      imageRatio,
+      onArticlePress,
+      pageSize,
+      showImages
+    } = this.props;
+    const { width } = this.state;
+
+    return (
+      <ErrorView>
+        {({ hasError }) =>
```
nathggns: This function still gets created every time. Is this a concern?

packages/article-list/src/article-list-item.js
tuncaulubilge: I do like destructing but being forced to do it everytime is a bit annoying for sure. Latest version of airbnb eslint had this forced in, but I think it's worth a discussion

packages/article-list/src/article-list.js
tuncaulubilge: :+1: I'll fix this

PR #1547, opened 2018-12-06
fix: diable ads on ssr

parkjh:
- adding checks to disable ads if neccessary
- setting server side prebid config (copied from testad-config.json from render)
- added new yarn commands that were useful

### DIFF
--- a/packages/ad/src/ad.js
+++ b/packages/ad/src/ad.js
@@ -72,6 +72,7 @@ class Ad extends Component {
       config,
       contextUrl,
       debug: adConfig.debug || false,
+      disableAds: adConfig.disabled || false,
       networkId: adConfig.networkId,
       pageTargeting: adConfig.pageTargeting,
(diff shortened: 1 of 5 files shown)


packages/ad/src/ad.js
```diff
@@ -72,6 +72,7 @@ class Ad extends Component {
       config,
       contextUrl,
       debug: adConfig.debug || false,
+      disableAds: adConfig.diabled || false,
```
colincclark: typo

aronshamash:
@parkjh Could you add descriptions to your PRs?
Also I'm pretty sure we want to check ads are working in SSR not disable them.

colincclark:
Conceptually, why are we disabling ads?

parkjh:
@colincclark @aronshamash my mistake... no we do not want to disable ads on ssr... here's take 2!

PR #1549, opened 2018-12-06
fix: Handle empty author object in article comment templates 

aronshamash:
- sets a default `author` object in article comment templates -  `main` and `magazine`
- Fixes 500 errors in render when view an article with this template as no author object is being currently returned.

### DIFF
--- a/packages/article-magazine-comment/src/article-magazine-comment.web.js
+++ b/packages/article-magazine-comment/src/article-magazine-comment.web.js
@@ -29,10 +29,12 @@ class ArticlePage extends Component {
       standfirst
     } = article;
 
+    const authorImage = author && author.image ? author.image : null;
+
     return (
       <Fragment>
         <ArticleHeader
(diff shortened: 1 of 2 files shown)


packages/article-magazine-comment/src/article-magazine-comment.web.js
```diff
@@ -29,10 +29,12 @@ class ArticlePage extends Component {
       standfirst
     } = article;
 
+    const authorImage = author && author.image ? author.image : null;
```
colincclark: you could have destructed this with a default on line 21

nathggns: I think that would be less clean tbh. Destructing multiple levels at once becomes very hard to follow very quickly.

aronshamash: I think I prefer the destructed way too. Less logic.

PR #1550, opened 2018-12-06
fix: article list storybook warnings

colincclark:
- fix prop type linting warning
- supress errors for tests that are supposed to error
- add default props for pagination (covers the article list still loading scenario)


packages/article-list/__tests__/web/__snapshots__/article-list-error-with-styles.web.test.js.snap
```diff
@@ -453,7 +453,7 @@ exports[`2. article list 1`] = `
               <div
                 className="S1"
               >
-                Showing 0 - 0 of 0 results
+                Showing -2 - 0 of 0 results
```
tuncaulubilge: ?

colincclark: The test itself is broken. Will fix, thanks

PR #1551, opened 2018-12-06
fix: Multiple bylines in related articles

tuncaulubilge:
Byline markup was wrapped with a View, which causes it to be laid out vertically. Converted it into a <Text> view so it's all laid into the same line now on all platforms.

Before:
screen shot 2018-12-06 at 16 23 47

After:
screenshot_1544116465

### DIFF
--- a/packages/article-summary/__tests__/web/__snapshots__/article-summary-with-style.web.test.js.snap
+++ b/packages/article-summary/__tests__/web/__snapshots__/article-summary-with-style.web.test.js.snap
@@ -47,11 +47,10 @@ exports[`1. article summary component with a single paragraph 1`] = `
 }
 
 .S6 {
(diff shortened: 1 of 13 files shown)


packages/article-summary/__tests__/web/__snapshots__/article-summary-with-style.web.test.js.snap
```diff
@@ -45,6 +45,15 @@ exports[`1. article summary component with a single paragraph 1`] = `
   line-height: 15px;
   margin-bottom: 5px;
 }
+
+.S6 {
```
tuncaulubilge: RNW uses the default font settings if it's not defined. However, it's overwritten by the children Text views' font settings so it's not creating a real issue.

PR #1557, opened 2018-12-07
feat: Integration tests for all article templates

jimhunty:
This provides identical cypress tests for all article templates that we currently have. They are identical to follow the strategy that we only care that the page loads and performs the client side actions that we require (e.g. Lazy Loading). More specific tests will be duplicated what we do further down the tree.

### DIFF
--- a/packages/ssr/__tests__/integration/article.js
+++ b/packages/ssr/__tests__/integration/article.js
@@ -1,37 +1,43 @@
 import { MockArticle } from "@times-components/fixture-generator";
 
 const relatedArticleCount = 3;
+const articleTemplateTest = template =>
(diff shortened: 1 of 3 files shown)


packages/ssr/__tests__/integration/article.js
```diff
@@ -1,37 +1,43 @@
 import { MockArticle } from "@times-components/fixture-generator";
 
 const relatedArticleCount = 3;
+const articleTemplateTest = template =>
+  xdescribe(`Article - template=${template}`, () => {
```
aronshamash: uncomment.

jimhunty:
This PR is waiting on a discussion with @L0wry over his previous Cypress implementation in SSR.

PR #1558, opened 2018-12-10
fix: article template errors

tuncaulubilge:
There are couple of red screens on the app when we enable all the templates, this PR fixes these

- Unknown template types were causing a crash. Now they fallback to articleMainStandard (indepth were causing an issue)
- Articles with empty content were causing a crash, such as the cartoon article
- First paragraph can be empty or could have a markup, which were causing a crash


packages/article/src/article.js
```diff
@@ -15,7 +15,7 @@ const Article = props => {
   const { article } = props;
   const { template = "mainstandard" } = article || {};
 
-  const Component = templates[template];
+  const Component = templates[template] || ArticleMainStandard;
```
aronshamash: `const Component = templates[template || 'mainstandard']` then can remove the default value of template on line 16.

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -127,7 +127,9 @@ class ArticleSkeleton extends Component {
     }
 
     const newContent = [...dataSource.content];
-    newContent[0] = insertDropcapIntoAST(newContent[0], template);
+    if (newContent && newContent.length > 0) {
```
nehasri89: should we do the same check for web too while we are here

tuncaulubilge: Good point, will do :+1:
this is especially a problem on the cartoon article, I'll update the pr

nehasri89:
should have some tests for empty para and the use cases where it crashed

tuncaulubilge:
Added few more tests @nehasri89 @colincclark

PR #1568, opened 2018-12-12
fix: magazine comment empty author

tuncaulubilge:
Magazine comment template is crashing when author image is null. This pr adds a null check for web and native

### DIFF
--- a/packages/article-magazine-comment/src/article-magazine-comment.web.js
+++ b/packages/article-magazine-comment/src/article-magazine-comment.web.js
@@ -30,11 +30,12 @@ class ArticlePage extends Component {
       shortHeadline,
       standfirst
     } = article;
+    const authorImage = author && author.image ? author.image : null;
 
     return (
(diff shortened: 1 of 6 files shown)


packages/article-magazine-comment/src/article-magazine-comment.web.js
```diff
@@ -30,11 +30,12 @@ class ArticlePage extends Component {
       shortHeadline,
       standfirst
     } = article;
+    const authorImage = author && author.image ? author.image : null;
```
aronshamash: adding the default value of `author.image` on line 22  _should_ have fixed this. If not then need to remove the lines 21-23

tuncaulubilge: Good point! Turns out we didn't have that default in native and ended up copying it to web as well. Now I switched back to using default value for author in both :+1:

PR #1570, opened 2018-12-12
Feat/Article Indepth Template

nehasri89:
This PR implements in Depth template for Web, Android and ios. Although there was an issue running ios simulator therefore this is not tested on ios.

This is a copy of `main-standard` template and changes have been made at various files to introduce the background and text colours. The tests are exactly the same apart from adding the template `attribute`

Todo:
- [x] fix native padding


packages/article-flag/src/article-flag-prop-types.js
```diff
@@ -0,0 +1,12 @@
+import PropTypes from "prop-types";
+import { colours } from "@times-components/styleguide";
+
+const articleFlagTypesPropTypes = {
+  color: PropTypes.string
+};
+
+const articleFlagTypesDefaultProps = {
+  color: colours.functional.white
```
aronshamash: `color` -> `colour`

tuncaulubilge:
Tried it in the native app, looks like it could use some padding around the headline:
screenshot_1544710019

packages/article-in-depth/article-in-depth.showcase.js
```diff
@@ -0,0 +1,209 @@
+/* eslint-disable react/prop-types */
```
tuncaulubilge: Colin just ditched the template specific stories in another PR, we can also ditch these. We got a article story with a template selection knob already

packages/article-in-depth/showcase-helper.js
```diff
@@ -0,0 +1,153 @@
+/* eslint-disable import/no-extraneous-dependencies, no-bitwise, operator-assignment, react/prop-types */
```
tuncaulubilge: 💥

packages/article-in-depth/src/article-lead-asset/article-lead-asset-image.js
```diff
@@ -0,0 +1,72 @@
+import React from "react";
```
tuncaulubilge: It might be a good idea to check this one with Nate as it'll conflict with that PR.

packages/article-magazine-in-depth/article-magazine-in-depth.showcase.js
```diff
@@ -0,0 +1,209 @@
+/* eslint-disable react/prop-types */
```
tuncaulubilge: This seems left over, we should have a "article-magazine-in-depth", just "article-in-depth"

packages/article/package.json
```diff
@@ -61,7 +61,9 @@
   },
   "dependencies": {
     "@times-components/ad": "2.2.8",
+    "@times-components/article-in-depth": "0.0.1",
     "@times-components/article-magazine-comment": "0.2.2",
+    "@times-components/article-magazine-in-depth": "0.0.1",
```
tuncaulubilge: This is also left over (actually breaking yarn install)

packages/article-in-depth/src/article-lead-asset/article-lead-asset-image.js
aronshamash: I guess it depends which goes in first

packages/article-flag/src/article-flag.js
```diff
@@ -21,30 +25,37 @@ const ArticleFlag = ({ title, color }) => (
...
 
-ArticleFlag.defaultProps = {
-  color: "black"
-};
+ArticleFlag.defaultProps = articleFlagTypesDefaultProps;
 
-const NewArticleFlag = () => (
-  <ArticleFlag color={colours.functional.articleFlagNew} title="new" />
+const NewArticleFlag = ({ color }) => (
+  <ArticleFlag color={color || colours.functional.tertiary} title="new" />
```
aronshamash: need to default to correct defaults

packages/article-in-depth/src/article-header/article-header.js
```diff
@@ -0,0 +1,42 @@
...
+  flags,
+  headline,
+  label,
+  standfirst,
+  textColour
+}) => (
+  <View
+    style={[
+      styles.container,
+      { backgroundColor: backgroundColour, width: "100%" }
```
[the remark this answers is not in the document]
aronshamash: it can't as shared with web.

packages/article-in-depth/article-in-depth.showcase.js
aronshamash: updated

packages/article-in-depth/showcase-helper.js
aronshamash: removed

packages/article-magazine-in-depth/article-magazine-in-depth.showcase.js
aronshamash: removed

packages/article/package.json
aronshamash: updated

packages/article-in-depth/src/article-lead-asset/article-lead-asset-image.js
nehasri89: @nathggns was planning on working on a branch based of this and doing his changes

packages/article-flag/src/article-flag.js
```diff
@@ -21,30 +25,43 @@ const ArticleFlag = ({ title, color }) => (
...
-  color: "black"
-};
+ArticleFlag.defaultProps = articleFlagDefaultProps;
 
-const NewArticleFlag = () => (
-  <ArticleFlag color={colours.functional.articleFlagNew} title="new" />
-);
-const UpdatedArticleFlag = () => (
-  <ArticleFlag color={colours.functional.articleFlagUpdated} title="updated" />
+const NewArticleFlag = ({ color }) => <ArticleFlag color={color} title="new" />;
```
nathggns: Probably would've gone for just passing all props through if the only thing you're trying to do is make a new component while setting the value a prop. 

```js
const NewArticleFlag = props => <ArticleFlag {...props} title="new" />;
```

packages/article-in-depth/CHANGELOG.md
```diff
@@ -0,0 +1,44 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
```
nathggns: Shouldn't this file be empty?

aronshamash: it shouldn't exist. removed

PR #1571, opened 2018-12-12
chore: added code owners

parkjh:


parkjh:
@nehasri89 @aronshamash @Belema @nathggns @jimhunty 
I've added your newsuk email addresses, which means that for it to work your associated github account needs to have the same newsuk emails...

tuncaulubilge:
:(

colincclark:
👎

aronshamash:
Why will it not work with my current email address associated with github account ?

parkjh:
@aronshamash because that should be the email address that the github assigns account to. And the reason that it's newsuk email is probably to ensure that all the codeowners are still employee here? I'm not sure, but I have not heard any good argument against and it seems sensible to me...

parkjh:
@colincclark @tuncaulubilge feel free to volunteer to be code owner~~~

PR #1572, opened 2018-12-12
chore: add disable comment message on web

Belema:
This PR adds feature parity between web and add by displaying a message on web when comments are disabled.

See [JIRA ticket](https://nidigitalsolutions.jira.com/browse/REPLAT-3000).

### DIFF
--- a/packages/article-comments/src/article-comments.web.js
+++ b/packages/article-comments/src/article-comments.web.js
@@ -1,84 +1,22 @@
-/* eslint-env browser */
-
-import React, { Component } from "react";
+import React from "react";
 import PropTypes from "prop-types";
-import { CommentContainer, CommentGuidelines } from "./styles/responsive";
-import executeSSOtransaction from "./comment-login";
+import Comments from "./comments";
+import DisabledComments from "./disabled-comments";
 
-class ArticleComments extends Component {
-  constructor() {
-    super();
-    this.container = null;
-  }
-
-  componentDidMount() {
-    this.initialiseComments();
-  }
-
-  componentWillUnmount() {
-    this.disposeComments();
-  }
-
-  initialiseComments() {
-    const { articleId, isReadOnly, spotAccountId } = this.props;
-
-    if (!this.container || !articleId || !spotAccountId) {
-      return;
-    }
-
-    const launcherScript = document.createElement("script");
-    launcherScript.setAttribute("async", "async");
-    launcherScript.setAttribute(
-      "src",
-      `https://launcher.spot.im/spot/${spotAccountId}`
-    );
-    launcherScript.setAttribute("data-spotim-module", "spotim-launcher");
-    launcherScript.setAttribute("data-post-id", articleId);
-    launcherScript.setAttribute(
-      "data-post-url",
-      `https://www.thetimes.co.uk/article/${articleId}`
-    );
-    launcherScript.setAttribute("data-seo-enabled", true);
-    launcherScript.setAttribute("data-livefyre-url", articleId);
-    this.container.appendChild(launcherScript);
-
-    if (!isReadOnly) {
-      if (window.SPOTIM && window.SPOTIM.startSSO) {
-        executeSSOtransaction();
-      } else {
-        document.addEventListener("spot-im-api-ready", executeSSOtransaction);
-      }
-    }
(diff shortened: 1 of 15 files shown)


packages/article-comments/src/comments.web.js
```diff
@@ -0,0 +1,89 @@
...
+      this.container.innerHTML = "";
+    }
+  }
+
+  render() {
+    return (
+      <CommentContainer>
+        <CommentEnabledGuidelines>
+          Comments are subject to our community guidelines, which can be viewed{" "}
+          <a href="//www.thetimes.co.uk/article/f4024fbe-d989-11e6-9063-500e6740fc32">
```
colincclark: can we make this environment aware, so it works on UAT for example?

packages/article-comments/src/article-comments.web.js
```diff
@@ -1,84 +1,22 @@
-/* eslint-env browser */
-
-import React, { Component } from "react";
+import React from "react";
```
colincclark: can you add web snapshot tests for this package?

packages/article-comments/src/comments.web.js
Belema: It is just an article on the website. I think that it makes sense to view it as an external link, no?

packages/article-comments/src/article-comments.web.js
Belema: Done.

Belema:
Thanks @colincclark. PR updated!

packages/article-comments/__tests__/web/__snapshots__/shared-with-style.web.test.js.snap
```diff
@@ -0,0 +1,45 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`comments error 1`] = `
+<div>
```
colincclark: the styles are not showing here because you need to use the correct methods from `jest-serializer`. See https://github.com/newsuk/times-components/blob/master/packages/article-byline/__tests__/shared-with-styles.web.js for inspiration

packages/article-comments/src/comments.web.js
colincclark: just a bit weird to be in UAT and click this and end up in prod. This would be annoying when QAing the site and then realising you were now in prod, for example

colincclark: then again, i am not sure we are doing this elsewhere, so let's agree this is optional?

packages/article-comments/__tests__/web/__snapshots__/shared-with-style.web.test.js.snap
Belema: Done.

packages/article-comments/src/comments.web.js
aronshamash: for one the community guidelines shouldn't just be an article rather a file that is present on all env, but at the moment it is published as an article only in prod. I think we just link to the prod one in all env as @Belema has done.

colincclark: makes sense

packages/article-comments/README.md
```diff
@@ -1,6 +1,8 @@
 # Article Comments
 
-This package provides a comment component for Android & iOS applications. The package has its own provider within the package so it can retrieve comment counts independent from the parent article.
+This package provides a comment component for all platforms. The Android & iOS
```
colincclark: can you also add the `yarn test:web` below in the testing section please?

packages/article-comments/__tests__/shared-with-style.web.js
```diff
@@ -0,0 +1,27 @@
...
+import shared from "./shared-base";
+
+export default Component => {
+  addSerializers(
+    expect,
+    compose(
+      stylePrinter,
+      hoistStyleTransform,
+      minimalWebTransform,
+      rnwTransform(AppRegistry)
```
colincclark: 👍 nice work

packages/article-comments/README.md
Belema: Done -> https://github.com/newsuk/times-components/pull/1573

PR #1576, opened 2018-12-13
feat: adding section and position to prebid vendors

parkjh:
- Passes position and section keywords to appnexus bid parameters http://prebid.org/dev-docs/bidders.html#appnexus-bid-params
- index exchange prebid adaptor does not support this 
- rubicon prebid adaptor only supports section keywords

### DIFF
--- a/packages/ad/src/utils/generate-config.js
+++ b/packages/ad/src/utils/generate-config.js
@@ -35,23 +35,23 @@ const sizeMap = {
   "article-sponsored-ad": sizes.pixel,
   default: sizes.intervention,
   header: sizes.header,
-  "native-inline-ad": sizeMap.native,
+  "native-inline-ad": sizes.native,
   pixel: sizes.pixel,
   pixelskin: sizes.pixel,
   pixelteads: sizes.pixel
 };
 
-const getAdSizes = (sizeMap, width) => {
-  for (let i = sizeMap.length - 1; i >= 0; i -= 1) {
-    if (width >= sizeMap[i].width) {
-      return sizeMap[i].sizes;
+const getAdSizes = (adSizeMap, width) => {
+  for (let i = adSizeMap.length - 1; i >= 0; i -= 1) {
+    if (width >= adSizeMap[i].width) {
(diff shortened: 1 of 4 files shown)


nehasri89:
LGTM, can merge as long as we have tested this branch on a prod like env

packages/ad/src/utils/generate-config.js
```diff
@@ -14,6 +14,24 @@ const getMaxSizes = adSizes => {
   );
 };
 
+const getPosition = slotName => {
```
aronshamash: can you use an object literal instead of a switch statement please

parkjh: @aronshamash just checking can I do defaults in object literal? or do I have to catch undefined values?

aronshamash: check https://github.com/newsuk/times-components/blob/master/packages/article/src/article.js

parkjh: changes made

packages/ad/src/utils/generate-config.js
```diff
@@ -14,42 +14,31 @@ const getMaxSizes = adSizes => {
...
-    case "ad-pixelskin":
-    case "pixelteads":
-    case "ad-pixelteads":
-    case "article-sponsored-ad":
-      return 2;
-    default:
-      return 0;
-  }
+const slotPositions = {
+  "ad-header": 1,
```
aronshamash: what is the reason for needing `header` and `ad-header` ? It seems like a hack

parkjh: @aronshamash it's because different slot ids are manually passed in from render and storybook, which refers to the dom element ids

aronshamash: it is because 1 will be deprecated once render uses TC for all article templates ? ie we are changing the ids but we still have legacy ids we have to support ?

parkjh: correct

PR #1577, opened 2018-12-13
Feat/article template integration tests

jimhunty:
This is a re-open of 1557.

This provides identical cypress tests for all article templates that we currently have. They are identical to follow the strategy that we only care that the page loads and performs the client side actions that we require (e.g. Lazy Loading). More specific tests will be duplicated what we do further down the tree.

Please Note: Tests were commented out by @L0wry and merged when they shouldn't have been meaning we have lost coverage. This PR at least gives us some coverage on the Article templates.

### DIFF
--- a/packages/fixture-generator/src/__tests__/mock-article.test.ts
+++ b/packages/fixture-generator/src/__tests__/mock-article.test.ts
@@ -1,4 +1,5 @@
 import MockArticle from "../mock-article";
+import { TemplateType } from "../types";
 
 describe("The mock Article", () => {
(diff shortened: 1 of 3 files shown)


jimhunty:
A decision has been made that this ticket should not go in due to #1484 being wrongfully merged.

@L0wry @kaliabadi Please revert #1484 so we can get our E2E tests back and resubmit #1484 when it is complete.

FYI @aronshamash @nehasri89 @tuncaulubilge @nathggns

PR #1579, opened 2018-12-13
fix: correct storybook adslot config

parkjh:

### DIFF
--- a/packages/ad/fixtures/article-ad-config.json
+++ b/packages/ad/fixtures/article-ad-config.json
@@ -50,6 +50,6 @@
       }
     }
   },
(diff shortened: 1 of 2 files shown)


parkjh:
screen shot 2018-12-13 at 16 44 05

parkjh:
certain inline ads for article pages were not displaying on storybook, this was due to it's static config having wrong slot name...

parkjh:
screen shot 2018-12-13 at 16 48 04

parkjh:
it seems that we need to activate a github repo settings in order for the codeowners approvals to be requirement... I've asked @araines for now, as I do not have access...

PR #1580, opened 2018-12-13
feat: font theming

tuncaulubilge:
Font theming PR for drop caps, headline and pull quotes

- Created theme factory for font families
- Passed font info to drop cap component
- Implemented font specific font sizes and margins to drop cap
- Updated drop cap stories to use theme factory
- Updated pull quotes to use theme factory
- Updated headlines in magazine templates to use theme factory
- Added snapshots to pull-quote, article-paragraph, article-magazine-standard and article-magazine-comment with each theme


packages/article-skeleton/src/article-body/article-body.web.js
```diff
@@ -50,8 +50,11 @@ const renderers = ({ observed, registerNode }) => ({
     return {
       element: (
         <Context.Consumer key={key}>
-          {({ theme: { sectionColour } }) => (
-            <DropCapView colour={sectionColour || colours.section.default}>
+          {({ theme: { dropCapFont, sectionColour } }) => (
+            <DropCapView
+              colour={sectionColour || colours.section.default}
```
aronshamash: we have 2 styles of setting defaults in this code base. I think I'm liking setting a default value when destructing more `{({ theme: { dropCapFont, sectionColour = colours.section.default } })`

tuncaulubilge: +1, changed it all to this in article body

packages/article-paragraph/src/styles/responsive.web.js
```diff
@@ -7,6 +7,20 @@ import {
   spacing
 } from "@times-components/styleguide";
 
+const dropCapFontSizes = {
+  cultureMagazine: 100,
```
aronshamash: worth grouping the magazines in a sub object @tuncaulubilge

tuncaulubilge: These are the font names, with "dropCap" being the name of the default font for all other templates

PR #1585, opened 2018-12-19
fix: related articles layout

nathggns:
Picked this one up from Colin.

Small design changes.

### DIFF
--- a/packages/article-byline/src/article-byline.js
+++ b/packages/article-byline/src/article-byline.js
@@ -4,16 +4,16 @@ import renderByline from "./render-byline";
 import { propTypes, defaultProps } from "./article-byline-prop-types";
 import styles from "./styles";
 
-const renderAuthorComponent = (children, key) => ({
+const renderAuthorComponent = (children, key, attributes, props) => ({
   element: (
-    <Text key={key} style={styles.nonLinkText}>
+    <Text {...props} key={key} style={styles.nonLinkText}>
       {children}
     </Text>
   )
 });
 
-const ArticleByline = ({ ast }) =>
-  renderByline(renderAuthorComponent, ast, styles.nonLinkText);
+const ArticleByline = ({ ast, ...props }) =>
+  renderByline(renderAuthorComponent, ast, styles.nonLinkText, props);
 
 ArticleByline.displayName = "ArticleByline";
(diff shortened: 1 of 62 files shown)


packages/article-byline/src/article-byline.js
```diff
@@ -4,16 +4,16 @@ import renderByline from "./render-byline";
 import { propTypes, defaultProps } from "./article-byline-prop-types";
 import styles from "./styles";
 
-const renderAuthorComponent = (children, key) => ({
+const renderAuthorComponent = (children, key, attributes, { className }) => ({
```
tuncaulubilge: Can we pass the classname instead of the props object? We don't have flow or typescript so it's not clear what the component should expect. Having "className" explicitly instead of props gives some clarity.

nathggns: It could contain other things too https://github.com/newsuk/times-components/pull/1585/files/11899c336f3b579c28f7fd1199790dd46cd10339#diff-61fc466b21715c3848d6546ca0192c6bR11

The idea is that the renderer can access any other arbitrary props passed to the byline component.

nehasri89:
Can we have screenshots of design changes please. Otherwise its LGTM.

nathggns:
Changes to the Related Article by-line (to always match the headline font size at different widths)

screen shot 2019-01-02 at 13 47 22
screen shot 2019-01-02 at 13 47 16

nathggns:
I believe the above actually shows all of the changes

nehasri89:
Thanks! Is there a jira ticket which lists down the changes that were needed?

packages/article-byline/src/article-byline.js
tuncaulubilge: Link's dead (probably due to rebasing changing the commit hash) but I guess it's the one with onAuthorPress. I think it would be better to have a separate prop for each of those props as it can get messy quite easily this way.

tuncaulubilge: I had to trace the code just to understand what the fields are in that props object, which is never good :) Another reason to have typescript

nathggns: Discussed in person.

PR #1591, opened 2019-01-04
spike: listing page tablet

tuncaulubilge:
Managed to get together a proof of concept for author profile and topic pages on tablet.

- Created a "Responsive" package for native apps, containing utilities to support tablet breakpoints and orientation changes
- Responsive package exposes a component named `Responsive` that has dimension change listener and updates its state whenever there is a screenWidth change. This component exposes layout changes through `ResponsiveContext`.
- Author-Profile and Topic pages uses <Responsive> component
- card and article-list-item consumes `ResponsiveContext` to decide between using tablet designs vs phone designs
- Added a "tabletWidth" value inside styleguide, to be used as a max width and check tablet breakpoint
- Applied a  max width on article list items with a center alignment. This structure will be used quite a lot, so we might consider converting it into a HoC
- Added a max width to list headers inside author-profile and topics
- New styles are added into card package to enable tablet specific layout changes. New styles are created by just adding a "tablet" suffix but we can consider using styled components for this or any alternative approach

Results:

tablet author profiles

Paired with @nathggns on designing Responsive package


packages/styleguide/src/styleguide.js
```diff
@@ -31,6 +31,7 @@ const fonts = timesFonts;
 const fontFactory = timesFontFactory();
 const fontSizes = timesFontSizes();
 const lineHeight = timesLineHeightsFactory();
+const tabletWidth = 660;
```
aronshamash: could get in a widths file but we can do that when we extract the other widths from around the codebase

tuncaulubilge: Moved it inside breakpoints now as it's basically a breakpoint

aronshamash: yes makes sense

packages/responsive/src/context.js
```diff
@@ -0,0 +1,3 @@
+import { createContext } from "react";
+
+export default createContext({ isTablet: false });
```
nehasri89: should this go in context pkg?

tuncaulubilge: It's a critical part of the responsive package, so I wanted to keep them together. This way, other projects can also get this package and use it. It's open to debate of course, is there any value in keeping all contexts in the same package?

nehasri89:
can we have tests to check if orientation change leads to component style change?

packages/responsive/src/context.js
nehasri89: sure, I am okay with that too. I just thought we decided to use context package for all context related stuff so one can look in and figure out all the things we keep in context, but I don't really mind this as well.

jimhunty: I dont like the idea of multiple contexts. Could have a performance implication.

tuncaulubilge:
@jimhunty this works on a row-by-row basis for article-list package. (responsiveness is implemented in the card component). So the scroll works the right way.

The poc I did for articles is on a different branch and needs to be converted to use a row-by-row implementation before I create a PR

packages/responsive/src/context.js
tuncaulubilge: I haven't read any disclaimers against using multiple contexts and this makes our lives so much simpler. It might be a good idea to call the other context as a "ThemeContext" to avoid any confusion. Since anything inside the "ThemeContext" is pretty static through the lifecycle of a single article, I wouldn't expect any perf issues

nehasri89:
LGTM

PR #1592, opened 2019-01-04
Feat/disable dropcap flag

nehasri89:
This PR respects the disableDropcap flag for rendering Dropcap.


packages/article-skeleton/__tests__/shared.base.js
```diff
@@ -292,6 +292,22 @@ export const snapshotTests = renderComponent => [
...
+      const template = "maincomment";
+      const dropcapsDisabled = true;
+      const article = articleFixture({
+        dropcapsDisabled,
+        ...fixtureArgs,
+        template
+      });
+      const output = renderComponent(renderArticle(article));
+
+      expect(output).toMatchSnapshot();
```
tuncaulubilge: We already got snapshots for a "no dropcap" scenario, so do we really need a snapshot in here? This seems like it could be a unit test that asserts if there is a drop cap in there or not

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -121,13 +121,13 @@ class ArticleSkeleton extends Component {
       receiveChildList
     } = this.props;
     const { dataSource, width } = this.state;
-    const { template } = dataSource;
+    const { template, dropcapsDisabled } = dataSource;
```
tuncaulubilge: linter will complain about alphabetical order :)

packages/article-skeleton/__tests__/shared.base.js
nehasri89: yeah, make sense. I will change it

PR #1594, opened 2019-01-04
fix: don't show images on author pages with no lead assets turned on

nathggns:
If authors have `noLeadAssets` enabled, we shouldn't be showing lead assets on their author page. This is a regression that needs fixing.


nathggns:
Need to write a test yet. Posted for @tuncaulubilge's benefit.

nathggns:
This is ready for review / merge now.

PR #1596, opened 2019-01-08
Feat: Video labels for remaining templates

aronshamash:
- Adds support for video label on `article-main-comment`, `article-magazine-standard`, `article-magazine-comment` and `article-in-depth` templates

- https://nidigitalsolutions.jira.com/browse/REPLAT-4659

### DIFF
--- a/packages/article-in-depth/src/article-label/article-label.js
+++ b/packages/article-in-depth/src/article-label/article-label.js
@@ -1,25 +1,28 @@
 import React from "react";
-import PropTypes from "prop-types";
 import { View } from "react-native";
 import ArticleLabel from "@times-components/article-label";
+import VideoLabel from "@times-components/video-label";
 import { colours } from "@times-components/styleguide";
+
+import {
+  articleLabelPropTypes,
+  articleLabelDefaultProps
+} from "./article-label-prop-types";
 import styles from "../styles";
 
-const Label = ({ label, color }) =>
-  label ? (
+const HeaderLabel = ({ color, isVideo, label }) => {
+  if (!isVideo && !label) return null;
+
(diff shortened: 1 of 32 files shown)


packages/article-in-depth/src/article-label/article-label.js
```diff
@@ -1,25 +1,28 @@
...
+
+import {
+  articleLabelPropTypes,
+  articleLabelDefaultProps
+} from "./article-label-prop-types";
 import styles from "../styles";
 
-const Label = ({ label, color }) =>
-  label ? (
+const HeaderLabel = ({ color, isVideo, label }) => {
```
nathggns: Did you consider maybe sharing this label component between templates as they'll be the same in each template now?

tuncaulubilge: I have no idea why we have a separate ArticleLabel and VideoLabel tbh. This could just use ArticleLabel and article-label can then decide whether to use video label or a regular label

aronshamash: yes but they not quite the same as indepth takes `color` and  main-standard needs a web version as it needs a responsive wrapper.

aronshamash: we could do that in this PR or refactor in a separate one.

tuncaulubilge: We still have a lot to refactor on article templates (we never had the time to pick those up in the previous quarter), so I don't mind if we leave this refactor for a separate PR

nathggns: Happy to have this refactored later.

PR #1597, opened 2019-01-08
Feat: REPLAT-4828 Update colours in In Depth article

jimhunty:
Support for the TPA values of `backgroundColour` and `textColour`.

`dropcapsDisabled`, `savingEnabled` and `sharingEnabled` are added to the queries, but not sure of the requirements for these. Didnt want them to be missed but happy to remove them if everyone agrees.


packages/article-in-depth/src/article-header/article-header.js
```diff
@@ -19,31 +19,46 @@ const ArticleHeader = ({
...
-      <View
-        style={[
-          styles.container,
-          { backgroundColor: backgroundColour, width: "100%" }
-        ]}
-      >
-        <Label color={textColour} label={label} />
-        <Text
+}) => {
+  const bgColour =
```
nathggns: This code to convert an rgba colour to a string is duplicated four times, probably a good candidate for abstraction.

packages/provider/__tests__/__snapshots__/article-provider.test.js.snap
```diff
@@ -188,12 +199,24 @@ Object {
       },
     ],
   },
+  "savingEnabled": true,
```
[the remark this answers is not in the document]
jimhunty: Yep - mentioned that in the description. I will remove.

packages/provider/__tests__/__snapshots__/article-provider.test.js.snap
```diff
@@ -31,6 +41,7 @@ Object {
       "name": "paragraph",
     },
   ],
+  "dropcapsDisabled": false,
```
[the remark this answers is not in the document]
nehasri89: I can confirm this is the same name 👍

packages/article-in-depth/src/article-header/article-header.js
```diff
@@ -19,31 +19,46 @@ const ArticleHeader = ({
...
+    backgroundColour && backgroundColour.rgba
+      ? `rgba(${backgroundColour.rgba.red}, ${backgroundColour.rgba.green}, ${
+          backgroundColour.rgba.blue
+        }, ${backgroundColour.rgba.alpha})`
+      : null;
+  const txtColour =
+    textColour && textColour.rgba
+      ? `rgba(${textColour.rgba.red}, ${textColour.rgba.green}, ${
```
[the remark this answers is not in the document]
jimhunty: Now converted to Hex and being passed through where needed

packages/article-in-depth/src/article-header/article-header.js
```diff
@@ -20,31 +21,37 @@ const ArticleHeader = ({
...
-          styles.container,
-          { backgroundColor: backgroundColour, width: "100%" }
-        ]}
-      >
-        <Label color={textColour} isVideo={hasVideo} label={label} />
-        <Text
+}) => {
+  const bgColour = gqlRgbaToStyle(backgroundColour);
+  const txtColour = gqlRgbaToStyle(textColour);
+  const textHex = gqlRgbaToHex(textColour);
```
tuncaulubilge: Can you do this conversion inside the component itself, just before passing it into the svg? That way, this use case stays contained inside the component itself, so from an component api perspective, flags will be able accept colour in any format, reducing the chance of any errors in the future

tuncaulubilge: In other words, the whole app can use either rgba or hex anywhere, and the svg package sorts its own bug within the package itself. What do you think?

jimhunty: Yea that is a much better idea

PR #1598, opened 2019-01-08
fix: multiple article bylines should appear on a single line

nathggns:
I introduced a regression with a previous PR that led to multiple article bylines appearing across multiple lines on native instead of the single line they should appear on. This fixes that regression. 

Before

screenshot_1546965291

After

screen shot 2019-01-08 at 17 36 30

### DIFF
--- a/packages/article-summary/__tests__/web/__snapshots__/article-summary-with-style.web.test.js.snap
+++ b/packages/article-summary/__tests__/web/__snapshots__/article-summary-with-style.web.test.js.snap
@@ -45,6 +45,15 @@ exports[`1. article summary component with a single paragraph 1`] = `
   line-height: 15px;
   margin-bottom: 5px;
 }
+
+.S6 {
+  color: inherit;
+  font-family: system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Ubuntu, "Helvetica Neue", sans-serif;
+  font-size: 14px;
+  font-weight: inherit;
+  line-height: inherit;
+  margin-bottom: 0px;
+}
 </style>
 
 <div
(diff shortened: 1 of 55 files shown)


packages/article-summary/__tests__/web/__snapshots__/article-summary-with-style.web.test.js.snap
```diff
@@ -45,6 +45,15 @@ exports[`1. article summary component with a single paragraph 1`] = `
   line-height: 15px;
   margin-bottom: 5px;
 }
+
+.S6 {
+  color: inherit;
+  font-family: system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Ubuntu, "Helvetica Neue", sans-serif;
```
aronshamash: doesn't look right..

packages/article-summary/src/article-summary.js
```diff
@@ -35,7 +35,11 @@ const ArticleSummary = props => {
       ? ArticleBylineOpinion
       : ArticleByline;
 
-    return <Byline {...bylineProps} className={bylineProps.bylineClass} />;
+    return (
+      <Text>
```
aronshamash: Not possible to use a `view` with additional styling ?

tuncaulubilge: Unfortunately, View has a different wrapping rules at the end of the line

packages/article-summary/__tests__/web/__snapshots__/article-summary-with-style.web.test.js.snap
tuncaulubilge: It looks bad but it's just react-native-web going for default font because none was defined in the wrapping text. The children text views hasvefonts specified so this is just ignored.

PR #1602, opened 2019-01-16
fix: fixing indepth label to not take the color

nehasri89:


packages/article-in-depth/src/article-header/article-header.js
```diff
@@ -34,7 +34,7 @@ const ArticleHeader = ({
             { backgroundColor: bgColour, width: "100%" }
           ]}
         >
-          <Label color={textColour} isVideo={hasVideo} label={label} />
+          <Label color={txtColour} isVideo={hasVideo} label={label} />
```
tuncaulubilge: Can we rename the prop to rgbTextColour and rgbBackgroundColour while destructing? The current naming seems quite open to bugs

aronshamash: yes agree

aronshamash:
Nice one @nehasri89. Looks better.

PR #1603, opened 2019-01-17
fix: Tablet layout of loading spinner keyline

aronshamash:
- Keyline above loading spinner (shown when paginating on apps) was being displayed to the left on landscape tablets. This PR fixes this issue.
- Found as part of design review https://nidigitalsolutions.jira.com/browse/REPLAT-4865

### DIFF
--- a/packages/article-list/src/article-list.js
+++ b/packages/article-list/src/article-list.js
@@ -1,9 +1,10 @@
 import React, { Component } from "react";
 import { ActivityIndicator, FlatList, View } from "react-native";
 import Button from "@times-components/button";
-import { colours } from "@times-components/styleguide";
+import { colours, tabletWidth } from "@times-components/styleguide";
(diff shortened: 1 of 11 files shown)


packages/article-list/src/article-list.js
```diff
@@ -173,14 +174,20 @@ class ArticleList extends Component {
...
       return (
-        <View>
-          <ArticleListItemSeparator />
-          <ActivityIndicator
-            color={colours.functional.keyline}
-            size="large"
-            style={styles.loadingContainer}
-          />
-        </View>
+        <ResponsiveContext.Consumer>
```
tuncaulubilge: Do we need this bit? It should be alright with just the max width. Can you test it without a context around it?

aronshamash: Tried without and it works, so removed that code.

tuncaulubilge: 🚢

PR #1607, opened 2019-01-18
chore: more descriptive pr template

parkjh:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,9 +1,32 @@
-<!--
-Thank you for your PR!
+<!--- Provide a general summary of your changes in the Title above -->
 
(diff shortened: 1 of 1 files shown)


tuncaulubilge:
Personally, I'm not a big fan of PR templates, and quite happy with the current PRs

nathggns:
I agree with @tuncaulubilge . Thanks for the effort of doing this @parkjh but unless a conversation among the team has happened in favour of this that I've missed, I think we should give this a miss.

jimhunty:
This PR supports those who want to make contributions to the project outside of The Times. For those that don't know the internal standards and have the conversations we do. However, external contributions have been limited so far and until this gets promoted, this is unlikely to change. So for that reason, I don't think we should be having this for the time being.

tuncaulubilge:
Three votes down, I'm closing the PR for now.

PR #1608, opened 2019-01-21
BREAKING CHANGE: rename slice to slice-layout

aronshamash:
- rename `slice` - > `slice-layout`
- https://nidigitalsolutions.jira.com/browse/REPLAT-5064
- Need to deprecate `slice` package on npm when PR is merged.


tuncaulubilge:
Won't we just create a new slice package, which would need to pick up from the same version number this one left?

PR #1612, opened 2019-01-22
Feat/tablet key facts

nathggns:


packages/article-skeleton/src/article-skeleton.js
```diff
@@ -152,25 +153,29 @@ class ArticleSkeleton extends Component {
...
-          onTopicPress={onTopicPress}
-          onTwitterLinkPress={onTwitterLinkPress}
-          onVideoPress={onVideoPress}
-          onViewableItemsChanged={onViewed ? this.onViewableItemsChanged : null}
-          pageSize={listViewPageSize}
-          renderRow={renderRow(analyticsStream)}
-          scrollRenderAheadDistance={listViewScrollRenderAheadDistance}
-          width={width}
-        />
+        <Responsive>
```
tuncaulubilge: Do we need this Responsive tag? We got it already in article-main-standard. Double responsive tags might be creating extra re-renders

nathggns: This is moving the Responsive tag from article main standard to the article skeleton, to match the work of @jimhunty. I also remove the Responsive tag from article main standard. https://github.com/newsuk/times-components/pull/1612/files#diff-3a3f5c42c9ff6fcf4785d7fc1c026261

nathggns:
This is ready to review but don't merge this as I will need to update any new packages created in master to use the correct react version when it comes time to merge this.

nathggns:
@tuncaulubilge 

> Other than the in article-skeleton, I think this is good to go. Let's check if double context is creating any perf caveats and then we can move it to each template if it's causing issues.

As I said above, @jimhunty already moved the responsive tag in his branch, so I've done the same. It makes the most sense to do it in article skeleton rather than separately in each template.

>  Let's check if double context is creating any perf caveats

There is no double context.

packages/article-skeleton/src/article-skeleton.js
tuncaulubilge: You are right, I confused it with the showcase change. Nice :+1:

tuncaulubilge:
I just merged slice -> slice-layout renaming PR, so that'll create some conflicts

PR #1613, opened 2019-01-22
Feat: REPLAT-4853 article max width

jimhunty:
https://nidigitalsolutions.jira.com/browse/REPLAT-4853

Bugbears:
* Having to mock Responsive is different packages. Any suggestions around this? Only thing I could think of that would work in all situations is to create a new package, but then this could fall out of sync.
* Snapshop updates of all article templates even though there is no work in there.


packages/article-image/__tests__/mocks.native.js
```diff
@@ -0,0 +1,5 @@
+jest.mock("@times-components/image", () => ({
+  ModalImage: "ModalImage"
+}));
+// eslint-disable-next-line global-require
+jest.mock("@times-components/responsive", () => require("./responsive-mock"));
```
nathggns: We introduced a standardised way of mocking responsive in https://github.com/newsuk/times-components/pull/1612 by mocking the underlying the Dimensions module. 

You can simply do this now


```
import { setIsTablet } from "@times-components/test-utils/dimensions";

setIsTablet(true);

// or

setIsTablet(false); 
```

packages/article-image/src/article-image.js
```diff
@@ -10,9 +11,15 @@ const ArticleImageNative = props => {
   } = props;
 
   return (
-    <View key={uri} style={styles[`${display}Container`]}>
-      <ArticleImage {...props} />
-    </View>
+    <ResponsiveContext.Consumer>
+      {({ isTablet }) => (
+        <View style={isTablet ? styles.imageContainerTablet : null}>
```
nathggns: Pretty sure you can do `isTablet && styles.imageContainerTablet`.

*Not a blocking comment, just a more concise pattern*

packages/article-paragraph/__tests__/mocks.native.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line global-require
+jest.mock("@times-components/responsive", () => require("./responsive-mocks"));
```
nathggns: As mentioned above, you should use the new standardised way of testing responsive design. I know that didn't exist when you wrote this, but worth updating it now it does.

packages/article-skeleton/src/styles/article-body/shared.js
```diff
@@ -33,18 +33,42 @@ const sharedStyles = scale => {
...
     },
+    imageContainerTablet: {
+      alignSelf: "center"
+    },
     interactiveContainer: {
       marginBottom: spacing(4),
       paddingLeft: spacing(2),
       paddingRight: spacing(2)
     },
+    interactiveContainerTablet: {
```
nathggns: This is mostly adding styles to the `interactiveContainer`, so I'm wondering why you apply either `interactiveContainer` OR `interactiveContainerTablet`, rather than applying both on tablets so you don't need to duplicate the styles. 

This happens in a few places actually. 

Essentially just wondering why you don't do this

```js
<View style={[ styles.interactiveContainer, isTablet && styles.interactiveContainerTablet ]} />
```

Instead of 

```js
<View style={isTablet ? styles.interactiveContainer : styles.interactiveContainerTablet } />
```

packages/article-image/__tests__/shared-with-style.native.js
```diff
@@ -22,9 +29,31 @@ export default () => {
...
   );
 
-  shared(component => {
+  const makeTest = component => {
     const testInstance = TestRenderer.create(component);
-
     return testInstance.toJSON();
+  };
+
+  shared(makeTest);
```
nathggns: Just wondering why you abstracted this?

packages/article-image/src/article-image.js
```diff
@@ -10,9 +11,15 @@ const ArticleImageNative = props => {
   } = props;
 
   return (
-    <View key={uri} style={styles[`${display}Container`]}>
-      <ArticleImage {...props} />
-    </View>
+    <ResponsiveContext.Consumer>
+      {({ isTablet }) => (
+        <View style={isTablet && styles.imageContainerTablet}>
```
nathggns: Wondering if you need another view here or whether you can just add the `imageContainerTablet` style to the existing view. Also wondering if there is two views whether the `key={uri}` bit needs to be moved up. Not sure what that key is for, I guess it's to ensure the `ArticleImage` is recreated when the URI changes, but don't know why that's necessary.

packages/article-skeleton/src/styles/article-body/shared.js
```diff
@@ -33,18 +33,37 @@ const sharedStyles = scale => {
...
     primaryContainer: {
       flexDirection: "column",
       paddingBottom: spacing(5),
       width: "100%"
+    },
+    primaryContainerTablet: {
+      alignSelf: "center",
+      width: tabletWidth
+    },
+    pullQuoteContainerTablet: {
```
nathggns: I'm not sure we need like more than 4 of 

```
{
  alignSelf: "center",
  width: tabletWidth
},
```

all with different names.

packages/article-image/src/article-image.js
nathggns: > Article images are not always full width so your comment relating to styling cant wor

I don't understand why that makes this impossible. 

```
 <View key={uri} style={[isTablet && styles.imageContainerTablet, styles[`${display}Container`]]}>
  <ArticleImage {...props} />
</View>
```

PR #1614, opened 2019-01-23
feat: tablet layout for main standard article headers

nathggns:
This adds a max width for the article header on tablet, as well as increases the headline font size. 

screen shot 2019-01-23 at 14 55 40


nathggns:
Blocked until https://github.com/newsuk/times-components/pull/1613 has gone in.

PR #1615, opened 2019-01-23
feat: Add ArticleFlags to article-flag package

aronshamash:
- `article-flags` now exports `ArticleFlags`
- https://nidigitalsolutions.jira.com/browse/REPLAT-5105
- used by `edition-slice` and article template packages


tuncaulubilge:
I merged edition slices branch into master, so this needs a good rebase

packages/article-summary/__tests__/shared.base.js
```diff
@@ -20,12 +20,17 @@ jest.mock("@times-components/article-byline", () => ({
   ArticleBylineOpinion: "ArticleBylineOpinion",
   default: "ArticleByline"
 }));
+jest.mock("@times-components/article-flag", () => ({
+  __esModule: true,
```
tuncaulubilge: Afaik, you only need __esModule:true if you are mocking default exports. This one should work without __esModule

packages/edition-slices/__tests__/shared-tiles.base.js
```diff
@@ -0,0 +1,42 @@
...
+import TestRenderer from "react-test-renderer";
+import { iterator } from "@times-components/test-utils";
+import leadOneAndOneDataGenerator from "../fixtures/leadoneandone";
+import { PrimaryTile } from "../src/tiles";
+
+jest.mock("@times-components/image", () => "Image");
+jest.mock("@times-components/gradient", () => "Gradient");
+
+// eslint-disable-next-line global-require
+jest.mock("@times-components/svgs", () => require("./mock-svg"));
```
tuncaulubilge: Let's mock ArticleFlags here, then we don't need to mock svgs

PR #1618, opened 2019-01-24
fix: Change ts build folder to build from dist

tuncaulubilge:
Currently, fixture generator package can not be used inside Jest tests. This is due to the configuration of our jest, done by `jest-configurator` package. Jest configurator picks replaces source paths from "dist/.." to "src/.." in order to run the test code on non-transpiled code. 

The problem is, fixture generator is written in typescript, so it can't be run with the current jest-configurator. To avoid this and bypass jest-configurator logic, I've moved `dist` folder into a `build` folder for fixture generator.

### DIFF
--- a/packages/fixture-generator/.prettierignore
+++ b/packages/fixture-generator/.prettierignore
@@ -1,5 +1,5 @@
 **/node_modules/*
-dist
+build
 __snapshots__
 android/*
 ios/*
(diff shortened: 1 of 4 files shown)


aronshamash:
@tuncaulubilge Might be worth mentioning in readme / contributing about running transpile before running tests

tuncaulubilge:
@aronshamash Good point, I updated the readme

PR #1624, opened 2019-01-25
feat: use image layout to decide which res image to request

nathggns:
This PR uses `onLayout` on the image placeholder to decide which res image to request, meaning that we won't be requesting huge images in slices on tablet.

### DIFF
--- a/packages/image/src/image.js
+++ b/packages/image/src/image.js
@@ -3,7 +3,7 @@ import { Image, View } from "react-native";
 import {
   addMissingProtocol,
   normaliseWidth,
-  screenWidthInPixels
+  convertToPixels
 } from "@times-components/utils";
 import appendSize from "./utils";
 import { defaultProps, propTypes } from "./image-prop-types";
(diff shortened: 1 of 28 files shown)


packages/utils/src/screen.js
```diff
@@ -3,5 +3,7 @@ import { PixelRatio } from "react-native";
 import { acceptedWidths, normaliseWidth, screenWidth } from "./screen.base.js";
 
 export { acceptedWidths, normaliseWidth, screenWidth };
-export const screenWidthInPixels = () =>
-  PixelRatio.getPixelSizeForLayoutSize(screenWidth());
+
+export const convertToPixels = px => PixelRatio.getPixelSizeForLayoutSize(px);
```
tuncaulubilge: We are passing in points, not pixels, so I suggest something like this to avoid confusion:
```suggestion
export const convertToPixels = points => PixelRatio.getPixelSizeForLayoutSize(points);
```

packages/image/src/image.js
```diff
@@ -15,38 +15,57 @@ class TimesImage extends Component {
...
+    this.onImageLayout = this.onImageLayout.bind(this);
+  }
+
+  onImageLayout({
+    nativeEvent: {
+      layout: { width }
+    }
+  }) {
+    const { highResSize } = this.props;
+    const imageRes = convertToPixels(
```
tuncaulubilge: We should first convertToPixels, then normalise it, since normaliseWidth expects pixel values

PR #1625, opened 2019-01-27
fix: ios storybook app

dadederk:
The iOS Storybook app won't compile when running `yarn ios`. One of the reasons is that, if you open the project and try to compile, it is missing a config.h file.

This file gets properly created once running the configure script from the glog third party dependency. Created a script for going to the right directory and executing the configure script and added a call to that script in the `preios` step in the project's `package.json` file.


tuncaulubilge:
Needs a rebase on master, then we can merge this in

dadederk:
@tuncaulubilge Script updated. 👍 Would you mind to see if it now works for you? Thanks!

PR #1629, opened 2019-01-28
Feat: REPLAT-5155 In depth header width

jimhunty:
https://nidigitalsolutions.jira.com/browse/REPLAT-5155

* Sets a max width of 1180px for items that are considered "100%" width.
* Increases in font size for Headline
* Meta information changes display layout and introduces a seperator

### DIFF
--- a/packages/article-in-depth/__tests__/shared-with-style.native.js
+++ b/packages/article-in-depth/__tests__/shared-with-style.native.js
@@ -10,6 +10,7 @@ import {
   print
 } from "@times-components/jest-serializer";
 import { themeFactory } from "@times-components/styleguide";
+import { setIsTablet } from "@times-components/test-utils/dimensions";
 import "./mocks.native";
 import ArticleInDepth from "../src/article-in-depth";
 import articleFixture, { testFixture } from "../fixtures/full-article";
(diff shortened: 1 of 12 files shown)


packages/article-in-depth/__tests__/shared-with-style.native.js
```diff
@@ -177,4 +178,37 @@ export default () => {
 
     expect(testRenderer).toMatchSnapshot();
   });
+
+  it("tablet full article with style in the culture magazine", () => {
```
tuncaulubilge: Do we need all three styles tested in tablet?

nathggns: I don't think so. I didn't do that on magazine.

jimhunty: The styling is different on tablet, so I think we do. We want to track that this doesn't regress.

nathggns: The tablet-specific styles are not dependent on the section, as far as I'm aware?

jimhunty: They are, the font changes and therefore the font matrix (font, font size, line height) for that changes based on the section

PR #1633, opened 2019-01-28
feat: secondary one slice

aronshamash:
- Secondary One Slice | wireframe only | mobile only
- https://nidigitalsolutions.jira.com/browse/REPLAT-5139
- note summary is using `summary125` which should be `summary250` according to https://docs.google.com/spreadsheets/d/1NBC-dOAGcF1sgfrkfFH7uslKG7eEU4o3n3Og5T1WNC8/edit#gid=1321391883. All the summary text lengths will come in a forthcoming PR.

### DIFF
--- a/packages/edition-slices/src/slices/index.js
+++ b/packages/edition-slices/src/slices/index.js
@@ -1,5 +1,3 @@
-import LeadOneFullWidthSlice from "./leadonefullwidth";
-import LeadOneAndOneSlice from "./leadoneandone";
-import SecondaryOneSlice from "./secondaryone";
-
-export { LeadOneFullWidthSlice, LeadOneAndOneSlice, SecondaryOneSlice };
+export { default as LeadOneFullWidthSlice } from "./leadonefullwidth";
+export { default as LeadOneAndOneSlice } from "./leadoneandone";
+export { default as SecondaryOneSlice } from "./secondaryone";
--- a/packages/edition-slices/src/slices/index.js
+++ b/packages/edition-slices/src/slices/index.js
@@ -1,4 +1,5 @@
 import LeadOneFullWidthSlice from "./leadonefullwidth";
 import LeadOneAndOneSlice from "./leadoneandone";
+import SecondaryOneSlice from "./secondaryone";
 
-export { LeadOneFullWidthSlice, LeadOneAndOneSlice };
+export { LeadOneFullWidthSlice, LeadOneAndOneSlice, SecondaryOneSlice };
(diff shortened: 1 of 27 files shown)


packages/edition-slices/src/tiles/secondary/index.js
```diff
@@ -0,0 +1,55 @@
...
+      leadAsset,
+      section,
+      shortHeadline,
+      summary125
+    }
+  }
+}) => (
+  <View>
+    <ArticleSummary
+      content={() => <ArticleSummaryContent ast={summary125} />}
```
nathggns: I wonder why article summary is implemented in such a way that you have to pass a function for each of these that doesn't seem to do anything.

packages/edition-slices/src/slices/index.js
```diff
@@ -1,4 +1,5 @@
 import LeadOneFullWidthSlice from "./leadonefullwidth";
 import LeadOneAndOneSlice from "./leadoneandone";
+import SecondaryOneSlice from "./secondaryone";
 
-export { LeadOneFullWidthSlice, LeadOneAndOneSlice };
+export { LeadOneFullWidthSlice, LeadOneAndOneSlice, SecondaryOneSlice };
```
nathggns: This file would be cleaner if it was just

```js
export { default as LeadOneFullWidthSlice } from "./leadonefullwidth";
export { default as LeadOneAndOneSlice } from "./leadoneandone";
export { default as SecondaryOneSlice } from "./secondaryone";
```

packages/edition-slices/src/tiles/index.js
```diff
@@ -1,4 +1,5 @@
 import PrimaryTile from "./primary";
+import SecondaryTile from "./secondary";
 
 // eslint-disable-next-line import/prefer-default-export
-export { PrimaryTile };
+export { PrimaryTile, SecondaryTile };
```
nathggns: Same comment as above with the `export { default as ... } from ...` pattern.

packages/edition-slices/src/tiles/secondary/index.js
aronshamash: for flexibility. See the use of the function in `related-articles` https://github.com/newsuk/times-components/blob/master/packages/related-articles/src/related-article-item.base.js#L127

packages/edition-slices/src/slices/index.js
aronshamash: agree will change

packages/edition-slices/src/tiles/index.js
aronshamash: agree again.

packages/edition-slices/src/tiles/secondary/index.js
jimhunty: There was a lot of conversation around this when it was first implemented. Frustrating that we didn't document why. But from what I remember, it was for Flexibility as @aronshamash says.

PR #1647, opened 2019-01-29
feat: REPLAT-4861 Integrate tex-layout and use for dropcaps

jmars:
Implements drop caps using the new text-layout utility library, no native dependencies are used so now dropcaps also work on IOS. This also lays the ground work for using the library for inline images and other text flowing components.

screenshot_20190129_130008


packages/article-paragraph/__tests__/android/__snapshots__/article-paragraph.android.test.js.snap
```diff
@@ -17,18 +17,107 @@ exports[`3. paragraph with a drop cap 1`] = `
...
-  >
-    his being Black History 
-  </Text>
-  <Text
-    selectable={true}
-  >
-    Month, last week was Politicians In Search Of An Eye-Catching Race-Related Policy Week.
+    This being Black History Month, last week was Politicians In Search Of An Eye-Catching Race-Related Policy Week.
   </Text>
+  <View>
```
tuncaulubilge: This looks a bit odd

package.json
```diff
@@ -6,7 +6,7 @@
   "main": "index.js",
   "engines": {
     "node": ">=8.10.0 <10",
-    "yarn": "1.12.3"
+    "yarn": ">=1.12.3"
```
tuncaulubilge: Let's keep this as a fixed version, as yarn had some inconsistencies in the previous upgrades a few times. We should do a separate pr if we want to bump to latest yarn

jmars: will revert this it was not meant to be committed, I just have yarn 1.13 locally

packages/article-paragraph/__tests__/android/__snapshots__/article-paragraph.android.test.js.snap
jmars: Looks like it is taking a snapshot before measurement finishes

jimhunty: I think a `sleep` is used elsewhere in tests. Feels unpredictable to me though.

package.json
jimhunty: `npx yarn@1.12.3`

packages/utils/src/text-layout/index.js
```diff
@@ -0,0 +1,178 @@
...
+  };
+};
+
+export const measureInline = elements => {
+  const els = [];
+  const promises = elements.map(
+    (el, i) =>
+      new Promise(resolve => {
+        els.push({
+          ...el,
```
nathggns: Have you considered using `React.cloneElement` instead? Just a more standardised way of cloning a component with updated props.

packages/utils/src/text-layout/index.js
```diff
@@ -0,0 +1,178 @@
...
+
+export const measureText = texts => {
+  const els = [];
+  const promises = texts.map(
+    (text, i) =>
+      new Promise(resolve => {
+        const key = String(i);
+        els.push(
+          <Text
+            {...text.props}
```
nathggns: Have you considered using `React.cloneElement` instead? Just a more standardised way of cloning a component with updated props.

packages/utils/src/text-layout/index.js
```diff
@@ -0,0 +1,178 @@
...
+    {children}
+  </View>
+);
+
+export const measureText = texts => {
+  const els = [];
+  const promises = texts.map(
+    (text, i) =>
+      new Promise(resolve => {
+        const key = String(i);
```
nathggns: As discussed, don't need to cast these to a string, just disable the lint rule for this file.

packages/utils/src/text-layout/formatter.js
```diff
@@ -0,0 +1,36 @@
...
+
+export const defaults = {
+  space: {
+    shrink: 9,
+    stretch: 6,
+    width: 3
+  }
+};
+
+export default words => ({
```
nathggns: Probably worth naming this function and then `export default name` or whatever just to make call stacks and debugging more clear

packages/article-paragraph/__tests__/shared.native.js
```diff
@@ -40,6 +40,11 @@ export default () => {
         );
 
         await delay(0);
+
+        callAllLayouts(testInstance);
+
+        await delay(0);
```
nathggns: Might be worth including these delays as part of the callAllLayouts function if they're necessary to be able to examine the results of it.

packages/article-paragraph/src/drop-cap.js
```diff
@@ -100,43 +94,46 @@ class DropCapParagraph extends Component {
...
-            >
-              {slicePoint > 0 ? text.slice(slicePoint).trim() : null}
-            </Text>
           </View>
         )}
-      </ResponsiveContext.Consumer>
-    );
+      </InlineElement>,
+      <Text selectable style={stylesThemedAndScaled.articleTextElement}>
+        {text.trim()}
```
nathggns: As discussed, this is an implementation detail and should be taken care of in the layout algorithm.

packages/utils/src/text-layout/index.js
```diff
@@ -0,0 +1,178 @@
...
+
+  const { elements: wordElements, results: wordsResults } = measureText(
+    splitEls
+  );
+  const {
+    elements: inlineElements,
+    results: inlineSizesResults
+  } = measureInline(inlines.map(el => el.props.children({})));
+
+  const results = new Promise(async resolve => {
```
nathggns: This is a bit weird… don't really need the new Promise constructor for this.

packages/utils/src/text-layout/index.js
```diff
@@ -0,0 +1,178 @@
...
+        } else {
+          result.push(node);
+        }
+      }
+      tolerance += 1;
+    }
+    const height = paragraph[0].height * lines;
+    return [result, height];
+  }
+  return undefined;
```
nathggns: This is unneeded.

packages/utils/src/screen.base.js
```diff
@@ -21,5 +20,4 @@ export const normaliseWidth = width => {
   return nWidth || acceptedWidths[acceptedWidths.length - 1];
 };
 
-export const screenWidth = isTablet =>
-  isTablet ? tabletWidth - tabletRowPadding : Dimensions.get("window").width;
+export const screenWidth = () => Dimensions.get("window").width;
```
jimhunty: This change will cause a lot of problems elsewhere. Has the full scope of this change been looked at from a styling perspective?

packages/utils/src/text-layout/align.js
```diff
@@ -0,0 +1,159 @@
+import React from "react";
```
jimhunty: `text-layout` should be a package in itself in my opinion.

packages/article-paragraph/src/drop-cap.js
```diff
@@ -1,94 +1,88 @@
...
+      text !== pText ||
+      dropCap !== pDropCap
+    ) {
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState(
+        {
+          content: [],
+          height: null,
+          needsLayout: true,
+          screenWidth: Dimensions.get("window").width
```
jimhunty: Changes to screen width utility function but then not used here?

packages/utils/src/screen.base.js
jmars: Was a bad rebase, fixed now

PR #1654, opened 2019-02-01
refactor: tiles

tuncaulubilge:
- Created shared tile elements such as TileImage and TileSummary to minimise duplication between tiles
- Removed Secondary Tile as it's identical to primary on mobile breakpoint
- Renamed Tiles as TileA, TileB, TileC as discussed. The full definitions can be found in storybook
- Renamed Slice stories to include the tile names


packages/edition-slices/edition-slices.showcase.js
```diff
@@ -19,23 +19,23 @@ export default {
...
+      name: "Lead One Full Width (A)",
       type: "story"
     },
     {
       component: () => {
         const slice = mockLeadOneAndOneSlice();
         return <LeadOneAndOneSlice lead={slice.lead} support={slice.support} />;
       },
-      name: "Lead One And One",
+      name: "Lead One And One (AB)",
```
aronshamash: Showing the tiles used will have to change when we do tablet so either we do (Mobile - AB) or we leave it out?

packages/edition-slices/edition-tiles.showcase.js
```diff
@@ -1,47 +1,34 @@
...
 export default {
   children: [
     {
       component: () => {
         const slice = mockLeadOneFullWidthSlice();
-        return <PrimaryTile tile={slice.lead} withImage withSummaryMargins />;
+        return <TileA tile={slice.lead} />;
       },
-      name: "Primary (with image, with summary margins)",
+      name: "TileA (Bottom image, 35pt headline, no summary)",
```
aronshamash: Maybe we use the fontSize name from the styleguide config instead of `35pt headline` ?

packages/edition-slices/edition-slices.showcase.js
tuncaulubilge: Probably something like this -> (Mobile: AB Tablet:FG)

packages/edition-slices/edition-tiles.showcase.js
tuncaulubilge: The problem is, the slice font sizes and line heights don't really fit with our current styleguide. If you take a look in tileA's styles, I had to manually enter fontSize and lineHeight values as it's not the same font-lineheight mapping we had in articles.

Tablets will add even more sizes, so I don't want to start naming them "small - medium - large"

jimhunty: We shouldnt have any redundant font sizes out of the styleguide. We should work with design to either choose more appropriate ones, or name the new ones.

tuncaulubilge: Let's catch Matt next week and try to add more structure to the designs. We'll definitely need new font sizes for slices but it'd be nice to limit it to 5-6 new headline styles. (maybe a bit more due to extra breakpoints)

packages/edition-slices/edition-slices.showcase.js
nehasri89: yeah, that sounds about right

packages/edition-slices/edition-tiles.showcase.js
```diff
@@ -1,47 +1,34 @@
...
       type: "story"
     },
     {
       component: () => {
         const slice = mockSecondaryFourSlice();
-        return <SecondaryTileImageReversed tile={slice.secondary1} withImage />;
+        return <TileC tile={slice.secondary1} />;
       },
-      name: "SecondaryTileImageReversed",
+      name: "TileC (Top image, 22pt headline, no summary",
```
nehasri89: I think we can say `no teaser` rather than `no summary`. `no summary` to me indicate tile just an image

nehasri89:
nice one! LGTM apart from a very minor comment.

packages/edition-slices/edition-tiles.showcase.js
tuncaulubilge: You are right, I'll update it :+1: . In tpa, that bit is called "summary" but I guess it doesn't mean much to us or the business

packages/edition-slices/edition-tiles.showcase.js
aronshamash: yes we should be reusing and adding new ones to the styleguide.

PR #1656, opened 2019-02-01
spike: Add borders and elevation to slice layout

tuncaulubilge:
Do not merge, this is just for demo purposes.

This pr adds android specific card layout borders to slice layout without changing the tile itself. We would need an android style on each slice-layout, including different breakpoints.

NOTE: Elevation is not supported on versions older than Android 5.0 (we support 4.4 on play store. Those versions won't have shadows, but will still have the border design)

screenshot_1549026567
screenshot_1549027720


tuncaulubilge:
Closing since this is created for spike purposes only. Please do not delete the branch

PR #1659, opened 2019-02-01
fix: article topics size & layout

nathggns:
Because padding was used to correct the location of text within the article topic, it meant the topic ended up being significantly taller than it should be. This corrects for that by using relative positioning. 

screen shot 2019-02-01 at 17 28 32

screen shot 2019-02-11 at 10 27 54

### DIFF
--- a/packages/article-topics/src/styles/shared.js
+++ b/packages/article-topics/src/styles/shared.js
@@ -6,10 +6,10 @@ const styles = {
     borderColor: colours.functional.keyline,
     borderRadius: 2,
     borderWidth: 1,
-    paddingBottom: 10,
+    paddingBottom: spacing(2),
     paddingLeft: spacing(3),
     paddingRight: spacing(3),
-    paddingTop: 10
+    paddingTop: spacing(2)
   },
   spacer: {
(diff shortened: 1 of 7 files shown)


packages/article-topics/src/styles/shared.js
```diff
@@ -6,10 +6,10 @@ const styles = {
     borderColor: colours.functional.keyline,
     borderRadius: 2,
     borderWidth: 1,
-    paddingBottom: 12,
+    paddingBottom: 10,
     paddingLeft: spacing(3),
     paddingRight: spacing(3),
-    paddingTop: 12
+    paddingTop: 10
```
jimhunty: not using `spacing`

packages/article-topics/src/styles/shared.js
```diff
@@ -6,10 +6,10 @@ const styles = {
     borderColor: colours.functional.keyline,
     borderRadius: 2,
     borderWidth: 1,
-    paddingBottom: 12,
+    paddingBottom: 10,
```
jimhunty: not using `spacing`

PR #1660, opened 2019-02-04
Feat: Article Summary Strapline

jimhunty:
Adding a strapline to Article Summary so that it can be used of edition slices.

Consumers of Article Summary can pass in their own component for Strapline or use the one provided by Article Summary.

Please note: The provided component currently takes a prop of `strapline` rather than the use of `children`. This is to copy what is done in `headline`. I would prefer `children` and may refactor that at a later date for both.


jimhunty:
@tuncaulubilge Yea, let's keep everything nice and small. Get this PR in and i'll sort that bit out. Make sure nothing regresses.

PR #1667, opened 2019-02-05
feat: Tile I and tile J

aronshamash:
- used as part of `LeadOneAndFourSlice`

### Tile I
screenshot 2019-02-05 at 17 05 27

### Tile J
screenshot 2019-02-05 at 17 05 43

### DIFF
--- a/packages/edition-slices/src/tiles/tile-i/styles/index.js
+++ b/packages/edition-slices/src/tiles/tile-i/styles/index.js
@@ -1,0 +1,22 @@
+import { colours, fonts, spacing } from "@times-components/styleguide";
+
+const styles = {
+  headline: {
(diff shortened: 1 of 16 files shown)


packages/edition-slices/src/tiles/tile-i/styles/index.js
```diff
@@ -0,0 +1,22 @@
+import { colours, fonts, spacing } from "@times-components/styleguide";
+
+const styles = {
+  headline: {
+    fontFamily: fonts.headline,
+    fontSize: 35,
```
nehasri89: use `spacing`

tuncaulubilge: It's just a font size, no need for spacing

PR #1675, opened 2019-02-07
feat: secondary 2 and 2

tuncaulubilge:
Secondary 2 and 2 slice, reusing the existing tiles C and G.

screenshot_1549551779

Paired with: Ivo & Val

### DIFF
--- a/packages/slice-layout/src/templates/secondarytwoandtwo/styles.js
+++ b/packages/slice-layout/src/templates/secondarytwoandtwo/styles.js
@@ -1,0 +1,23 @@
+const styles = {
+  container: {
+    flex: 1
+  },
+  itemColumnSeparator: {
+    borderLeftWidth: 1,
+    marginBottom: 12,
+    marginTop: 10
+  },
+  itemContainer: {
+    flexDirection: "row"
+  },
+  itemHalfWidth: {
+    width: "50%"
+  },
+  itemRowSeparator: {
+    borderBottomWidth: 0.5,
+    borderStyle: "solid",
+    marginHorizontal: 12
+  }
+};
+
+export default styles;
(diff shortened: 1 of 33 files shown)


packages/slice-layout/src/templates/secondarytwoandtwo/styles.js
```diff
@@ -0,0 +1,23 @@
...
+  itemContainer: {
+    flexDirection: "row"
+  },
+  itemHalfWidth: {
+    width: "50%"
+  },
+  itemRowSeparator: {
+    borderBottomWidth: 0.5,
+    borderStyle: "solid",
+    marginHorizontal: 12
```
aronshamash: We agree we would use multiples of `spacing` so this should be `spacing(2)`

packages/slice-layout/src/templates/secondarytwoandtwo/styles.js
```diff
@@ -0,0 +1,23 @@
...
+    marginBottom: 12,
+    marginTop: 10
+  },
+  itemContainer: {
+    flexDirection: "row"
+  },
+  itemHalfWidth: {
+    width: "50%"
+  },
+  itemRowSeparator: {
```
aronshamash: We should probably have a shared component for this soon

packages/slice-layout/src/templates/secondarytwoandtwo/styles.js
```diff
@@ -0,0 +1,23 @@
+const styles = {
+  container: {
+    flex: 1
+  },
+  itemColumnSeparator: {
+    borderLeftWidth: 1,
+    marginBottom: 12,
+    marginTop: 10
```
aronshamash: `spacing(2)`

PR #1678, opened 2019-02-08
feat: section page

tuncaulubilge:
Edition query and section page work

- Split article query into a fragment for reuse in editions (This is not used in the edition query for now, since the api is not optimised for the full article load)
- Created an edition query based on edition team's work
- Created a fixture generator for the editions
- Added typenames to edition slices fixture generators. We'll need to do the same for every new slice fixture generator we add.
- Added provider test for editions
- Created a Section Page within pages package for native
- Used a FlatList in section page to support item separators.
- Showcases for sections using storybook knobs on sections
- Snapshot tests on section page using edition fixture generator

**Upcoming Refactors:**
While implementing the section page, I've noticed couple of improvements we can do on the slices package. I'll do those in another pr after we merge other slice PRs

- Each slice component should take the same props. (`slice` and `onPress`). Currently, we are passing each tile data as a separate prop, which is not really necessary.
- edition-slices package should export a key-value map of each slice. That way, Section page wouldn't need a switch-case and it can just render whatever is in that map, as long as the names match.

screenshot_1549627510


packages/provider-test-tools/src/edition.js
```diff
@@ -0,0 +1,98 @@
+import { MockList } from "graphql-tools";
+import { mockEditionSlice } from "@times-components/fixture-generator";
+import { edition as editionQuery } from "@times-components/provider-queries";
+import article from "../fixtures/article.json";
+
+const convertRatio = ratio => {
```
aronshamash: Can be an object literal with a default value

packages/provider-test-tools/src/edition.js
```diff
@@ -0,0 +1,98 @@
...
+export default ({ variables = () => {} } = {}) => {
+  const queryVariables = variables();
+  const sliceCount = 2;
+
+  return [
+    {
+      defaults: {
+        types: {
+          Article: () => article,
+          ArticleSlice: () => ({
```
nehasri89: do we need this if we are defining the individual slices asa well?

tuncaulubilge: Graphql mock provider works in mysterious ways...

nehasri89: 👍

nehasri89:
LGTM apart from the query change as we discussed 👍

packages/provider-test-tools/src/edition.js
tuncaulubilge: This page tries to explain how it works: https://www.apollographql.com/docs/graphql-tools/mocking.html

It'd be nice to start using the latest fixture generators instead of this structure, but that'll be a huge chore

packages/provider-test-tools/src/edition.js
tuncaulubilge: Done :+1:_ _

PR #1682, opened 2019-02-11
feat: edition checker

nathggns:
This adds a tool to the command line that allows you to crawl a specific edition or the past six days for issues in loading articles when using react. 

This requires being logged in to work, so you must set a `ACS_COOKIE` and `SACS_COOKIE` environment variable. We should probably set up credentials specifically for this tool in the long run.

This also requires the use of an internal Times API, and you must set the path to this with the `TIMES_API_ROOT` environment variable.

You can also configure the tool to post to Slack using a Slack incoming webhook with the `TIMES_SLACK_HOOK` and `TIMES_SLACK_USERNAME` environment variables. 

<s>I have not written tests for this tool, but it has been refactored and simplified significantly from its previous implementation to allow for this if deemed necessary.</s>

It may also be important to know this was written entirely separately to times-components initially, and has had to be almost entirely rewritten to fit into times components architecture, which has removed some features.

In the long term, it might be useful to consider adding other kinds of checks to the crawler. Currently, the only checks it employs are checking that the status code is 200.  

```bash
$ ./times-components check edition # check todays edition
$ ./times-components check edition -e 2019-01-01 # check a specific edition
$ ./times-components check past-six-days # check articles from the past six days

$ ./times-components check [method] -m 30 # only check the first 30 articles
$ ./times-components check [method] -c 30 # check 30 articles at a time
$ ./times-components check [method] -v # verbose, display more information
```

### DIFF
--- a/lib/.eslintrc.json
+++ b/lib/.eslintrc.json
@@ -1,5 +1,6 @@
 {
   "rules": {
-    "import/no-extraneous-dependencies": "off"
+    "import/no-extraneous-dependencies": "off",
+    "no-console": "off"
   }
 }
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -40,20 +40,20 @@ function addComponent(opts) {
     '\n\nUsage:\n$ times-components create component ComponentName "Component Description"';
   const argumentOrUsage = (argument, message) => {
     if (!argument) {
-      console.error(message + usage); // eslint-disable-line no-console
+      console.error(message + usage);
       process.exit(1);
     }
   };
 
   const rComponent = /^([A-Z][a-z]*)+$/;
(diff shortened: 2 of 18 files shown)


aronshamash:
out of interest what was reason for almost completely rewriting it to fit into `times-components` @nathggns  ?

nathggns:
@aronshamash making the code testable/more readable + also it was build using commander instead of subcommander, and commander gives you a global variable for configuration, so this was all just functions reading off of that. It also had a bunch of features that were somewhat useful during development of it but aren't really all that useful & are difficult to test. 

When this tool was written it was kind of just thrown together as a quick hack. So just had to make it more suitable for long term use.

PR #1683, opened 2019-02-12
fix: add horizontal padding to the centred caption

dadederk:
The centred caption didn't had any start/end paddings so when the text fitted well into one line, it would be to close to the edges. Added some horizontal padding to it.

screenshot 2019-02-11 at 11 22 32
screenshot 2019-02-11 at 11 24 12


dadederk:
@tuncaulubilge Some tests were failing and I have updated the snapshots for the article image with style. Would you mind having a quick look to see if it makes sense? And if it does, do you mind merging? I'm still not authorised to merge in this repo...

PR #1685, opened 2019-02-12
feat: Move section fragment out and simplify

tuncaulubilge:
Updated edition query to include a fragment, for native apps to re-use

Also simplified the article within the tiles to only include fields needed for the homepage. The articles for article pages will use the data from "article-fragment" instead.


packages/provider-queries/src/section-fragment.js
```diff
@@ -0,0 +1,667 @@
+import gql from "graphql-tag";
+
+export default gql`
+  fragment sectionPageProps on Section {
```
ChristianGarcia: Will these fragments be exported to some raw `.graphql` file? Or will the apps need to read these `.js` files to get the fragments?

tuncaulubilge: I haven't done that bit yet, but eventually these will be imported as .graphql files as a raw asset into the android library (xnative). I'll write a node script to create those .graphql files using these js files, but probably on a separate PR

packages/provider-queries/src/section-fragment.js
```diff
@@ -0,0 +1,667 @@
...
+        }
+      }
+      ... on Image {
+        ...imageProps
+      }
+    }
+    ...articleProps
+  }
+
+  fragment imageProps on Image {
```
ChristianGarcia: In anticipation to extracting fragments for xnative:

There's already an `imageProps` fragment in the article query, which will make our `edition` query fail, as we have to request the full article fragment as it exists in `ArticleQuery` and the two fragments will collide.

Since the two fragments are different, we need a new name for this one (`sectionImageProps`? `tileImageProps`?).

An additional, more radical approach (read _possible overkill_) would be having one file per fragment. This way we'll ensure fragment names won't collide when we mix them in a query. Anyway, renaming this one will suffice for now.

ChristianGarcia: 13 mins late! :(

PR #1693, opened 2019-02-13
fix: Fragment name conflicts

tuncaulubilge:
We were using the same fragment names within section fragment and article fragment, which causes an issue when we use both fragments within the same query. Prefixed section fragments with `section...` to avoid conflicts.

Addresses the pr comment by Christian: https://github.com/newsuk/times-components/pull/1685#discussion_r256454533

+ Also added a native-edition-query for test purposes only, to check how it'll work in the native app when we use both fragments

### DIFF
--- a/packages/provider-queries/src/article-fragment.js
+++ b/packages/provider-queries/src/article-fragment.js
@@ -1,7 +1,7 @@
 import gql from "graphql-tag";
 
 export default gql`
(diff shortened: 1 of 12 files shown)


tuncaulubilge:
I'll also add a native-edition query to test those two fragments together, so do not merge for now

PR #1696, opened 2019-02-13
feat: offline fallback on images

tuncaulubilge:
Improved the offline image support on android

- Tries to render the network image first
- If network image throws any errors, renders the offline image


packages/image/src/lazy-loading-image.android.js
```diff
@@ -1,24 +1,44 @@
...
 import { Image } from "react-native";
 import appendToUrl from "./utils";
 
-export default props => {
-  const { source } = props;
-  return (
-    <Fragment>
-      {source && source.uri && !source.uri.includes("data:image/") ? (
+class LazyLoadingImage extends Component {
+  constructor(props) {
```
jimhunty: Shouldnt need props here?

packages/image/src/lazy-loading-image.android.js
```diff
@@ -1,24 +1,44 @@
...
+  }
+
+  render() {
+    const { source } = this.props;
+    const { error } = this.state;
+    return (
+      <Fragment>
+        {error &&
+        source &&
+        source.uri &&
```
jimhunty: Should we fallback in this case to the placeholder if we dont get a `uri`?

tuncaulubilge: We got a placeholder in the images component itself, so if this doesn't display anything, you see the placeholder

PR #1699, opened 2019-02-14
Fix: ios drop caps

jmars:
Fixes a bunch of issues with drop-caps on iOS

image
image
image
image

### DIFF
--- a/packages/article-paragraph/src/drop-cap.js
+++ b/packages/article-paragraph/src/drop-cap.js
@@ -59,13 +59,18 @@ class DropCapParagraph extends Component {
   }
 
   async calculateLayout() {
+    const { font, scale, isTablet } = this.props;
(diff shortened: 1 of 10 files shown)


packages/article-paragraph/src/drop-cap.js
```diff
@@ -59,13 +59,18 @@ class DropCapParagraph extends Component {
...
+    const {
+      articleMainContentRow: { paddingLeft, paddingRight }
+    } = styleFactory(font, scale);
     const { screenWidth: width } = this.state;
     this.setState({
       content: elements
     });
     const sizes = await results;
-    const [laidOut, height] = layoutText(width, sizes);
+    const padding = isTablet ? width : width - (paddingLeft + paddingRight);
```
jimhunty: This is defined in the styleguide with `screenWidth`

PR #1700, opened 2019-02-14
fix: Update edition snap

tuncaulubilge:
Edition snapshot is breaking master, fixed.


tuncaulubilge:
I was waiting for bitrise to pass @jimhunty :+1: merged

PR #1702, opened 2019-02-14
chore: slice refactor

tuncaulubilge:
"Small" refactor on tiles and slices, changed couple of things and tidied showcases and tests

- Mocked tiles inside slice tests, as we already have tile specific tests. This cleaned up lots of snaps
- Update the api for each slice, so they just take a prop called "slice" instead of things like "lead" "support1" etc
- Instead of exporting each slice separately, now they are all exported as a map. This way, section page does not need to know about each slice we support, it just consumes the map
- Minor refactors in tests to make our fixtures consistent


packages/edition-slices/__tests__/shared-slices.base.js
```diff
@@ -22,26 +26,38 @@ import {
   SecondaryTwoNoPicAndTwoSlice
 } from "../src/slices";
 
+jest.mock("../src/tiles", () => ({
+  TileA: "TileA",
```
aronshamash: can we generate programmatically using `fs` ?

packages/edition-slices/edition-tiles.showcase.js
```diff
@@ -31,145 +27,112 @@ const preventDefaultedAction = decorateAction =>
 export default {
```
aronshamash: We should map over an object literal to generate this maybe ?

packages/fixture-generator/src/index.ts
```diff
@@ -16,8 +16,9 @@ import mockEditionSlice, {
   mockLeadersSlice,
   mockSecondaryOneSlice,
   mockSecondaryFourSlice,
+  mockSecondaryTwoAndTwoSlice,
   mockSecondaryTwoNoPicAndTwoSlice,
-  mockList2AndSixNoPicSlice
+  mockListTwoAndSixNoPicSlice
```
aronshamash: could be a good time to alphabetise here ?

packages/pages/src/section/section.js
```diff
@@ -1,79 +1,15 @@
...
-  ListTwoAndSixNoPicSlice,
-  SecondaryOneSlice,
-  SecondaryFourSlice
-} from "@times-components/edition-slices";
+import sliceMap from "@times-components/edition-slices";
 import SectionItemSeparator from "./section-item-separator";
 import styles from "./styles";
 
 const renderSlice = (slice, onPress) => {
-  switch (slice.name) {
```
aronshamash: nice :)

tuncaulubilge: :) this was pretty much the goal of this whole pr

packages/edition-slices/__tests__/shared-slices.base.js
tuncaulubilge: Done :+1:

PR #1703, opened 2019-02-14
feat: Key line combinations for comments, related articles and topics

jimhunty:
Where there were no related articles, above comments would be missing a key line. To deal with optional topics and related articles, and keeping the complexity to a minimum, it made sense to remove the last keyline from related article slices and add this to the top of comments.

Additionally, within `leadoneandtwo`, there was a `withSeperators` prop that was never used. I removed this for simplicity.

screenshot 2019-02-14 at 16 37 09

===

screenshot 2019-02-14 at 16 38 19


aronshamash:
@jimhunty What about when article has no comments  ?

jimhunty:
@aronshamash There is a no comments block of text. "Be the first the comment" type thing. So there is always something.

PR #1704, opened 2019-02-14
feat: gutters

nathggns:
Articles should have gutters on either side if the native device is wider than 1180.

Also fixed lead asset video widths, and a number of refactors & test simplifications.

screen shot 2019-02-14 at 19 52 36


packages/test-utils/src/mock-article-skeleton.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable react/prop-types,import/no-extraneous-dependencies */
+import React from "react";
+
+const MockArticleSkeletonWrapper = "ArticleSkeleton";
+
+const MockArticleSkeleton = ({ Header }) => <MockArticleSkeletonWrapper><Header width={666} /></MockArticleSkeletonWrapper>;
```
tuncaulubilge: :+1:

PR #1710, opened 2019-02-15
feat: Move article flag under teaser and design tweaks

aronshamash:
- as part of https://nidigitalsolutions.jira.com/browse/REPLAT-5478
- moved article flags under teaser text in `article-summary`
- other design review tweaks and making margins consistent 

screenshot 2019-02-15 at 17 02 17

### DIFF
--- a/packages/article-summary/src/article-summary.js
+++ b/packages/article-summary/src/article-summary.js
@@ -68,7 +68,7 @@ const ArticleSummary = props => {
       {headline()}
       {strapline()}
       {content()}
-      {flags()}
+      {flags() ? <View style={styles.flagWrapper}>{flags()}</View> : null}
       {datePublicationProps ? (
         <Text style={styles.metaText} testID="datePublication">
           <DatePublication {...datePublicationProps} />
--- a/packages/article-summary/src/article-summary.js
+++ b/packages/article-summary/src/article-summary.js
@@ -67,8 +67,8 @@ const ArticleSummary = props => {
       {bylineProps && bylineProps.isOpinionByline ? renderByline() : null}
       {headline()}
       {strapline()}
-      {flags()}
       {content()}
+      {flags()}
       {datePublicationProps ? (
         <Text style={styles.metaText} testID="datePublication">
           <DatePublication {...datePublicationProps} />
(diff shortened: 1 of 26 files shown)


packages/article-summary/src/article-summary.js
```diff
@@ -67,8 +67,8 @@ const ArticleSummary = props => {
       {bylineProps && bylineProps.isOpinionByline ? renderByline() : null}
       {headline()}
       {strapline()}
-      {flags()}
       {content()}
+      {flags()}
```
jimhunty: This will change all the list pages and Related Articles as well as all the tiles. Are we sure we want to do this?

aronshamash: @jimhunty Flags are not in the topic or author profile queries so no flags displayed on list pages.
Related Articles could display flags as they are in the query but that is intentional.

tuncaulubilge: We don't use flags on list pages and related articles for now

packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -27,7 +28,9 @@ const TileSummary = ({
         ? () => <ArticleSummaryContent ast={summary} style={summaryStyle} />
         : undefined
     }
-    flags={() => <ArticleFlags {...flagColour} flags={flags} />}
+    flags={() => (
+      <ArticleFlags {...flagColour} flags={flags} style={styles.flag} />
```
tuncaulubilge: Looks like we can just embed this flag style into article-summary as it's part of the "default" style. We should be passing style as props only for exceptions

aronshamash: yes good point.

packages/article-summary/src/article-summary.js
jimhunty: 👍

packages/edition-slices/src/tiles/shared/tile-summary.js
aronshamash: @tuncaulubilge I've re-jigged it now with no `style` prop on `article-flag`.

PR #1711, opened 2019-02-15
feat: Responsive Slice wrapper for tablet support & lead one 

tuncaulubilge:
This PR introduces a ResponsiveSlice wrapper as a shared util to support different breakpoints on edition slices. Also added LeadOneFullWidth tablet implementation.

- Created all the breakpoints inside styleguide, even though we'll just support small and medium for now
- ResponsiveSlice has gutters around it, so anything above 1366 will have empty margins on both sides
- Added a separate test for tablet slices
- Created TileR for the LeadOne tablet slice
- This slice doesn't use a slice-layout, so no changes in there. But for other tablet breakpoints, we'll pass "editionBreakpoint" as a prop so it can return different layouts for each breakpoint. Passed "editionBreakpoint" to the render function on ResponsiveSlice to enable this.
- Also added a scrollview to tile stories, as this tile is bigger than the viewport

screenshot_1550253930


packages/styleguide/src/breakpoints/index.js
```diff
@@ -1,3 +1,26 @@
+const editionBreakpoints = {
+  large: "large",
+  medium: "medium",
+  small: "small",
+  xlarge: "xlarge"
+};
+
+const editionBreakpointWidths = [768, 1024, 1366];
```
aronshamash: why not put this in an object literal then can get rid of the if statement below?

tuncaulubilge: I still need the if statement below to convert breakpoints to intervals, don't I? Can you show an example?

aronshamash: ok I think I found a way but the code is not as clear. Stick with the ifs

PR #1712, opened 2019-02-18
chore: Improve the coverage on the storybook package slightly

jimhunty:
Small test improvements to help improve coverage.


packages/storybook/src/showcase-to-storybook.js
```diff
@@ -7,9 +7,10 @@ React.Fragment.propTypes = {
   children: PropTypes.node.isRequired
 };
 React.Fragment.displayName = "React.Fragment";
+React.Fragment;
```
ValGeorgiev: Why we need this line?

jimhunty: Whoops! We dont!

PR #1714, opened 2019-02-18
feat: Edition slice tile - Tile T

jimhunty:
Tile T is like Tile D, but with a different ratio image.

screenshot 2019-02-18 at 14 22 40

### DIFF
--- a/packages/edition-slices/edition-tiles.showcase.js
+++ b/packages/edition-slices/edition-tiles.showcase.js
@@ -16,7 +16,11 @@ import {
   TileM,
   TileN,
   TileO,
+<<<<<<< HEAD
   TileR
+=======
+  TileT
+>>>>>>> feat: Edition slice tile - Tile t
 } from "./src/tiles";
 
 const tile = mockEditionSlice(1).items[0];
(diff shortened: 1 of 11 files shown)


packages/edition-slices/src/tiles/tile-t/index.js
```diff
@@ -0,0 +1,29 @@
...
+import { TileImage, TileLink, TileSummary } from "../shared";
+import styles from "./styles";
+
+const TileT = ({ onPress, tile }) => (
+  <TileLink onPress={onPress} tile={tile}>
+    <View style={styles.container}>
+      <TileImage
+        aspectRatio={16 / 9}
+        style={styles.imageContainer}
+        uri={tile.article.leadAsset.crop169.url}
```
tuncaulubilge: Can you use the shared "getCrop" function from tiles/shared for this url? (it was just merged couple of mins ago in another PR)

aronshamash:
@jimhunty needs a rebase

packages/edition-slices/src/tiles/tile-t/index.js
jimhunty: Should be done now.

packages/edition-slices/edition-tiles.showcase.js
```diff
@@ -16,7 +16,11 @@ import {
   TileM,
   TileN,
   TileO,
+<<<<<<< HEAD
```
tuncaulubilge: ☝️

jimhunty: 😱

PR #1715, opened 2019-02-18
fix: section page

tuncaulubilge:
Fixed couple of minor issues blocking the native apps consuming react-native section pages:

- Fixed the exports on the section page
- Added typename to section query
- Added a shared "getCrop" function which has a null check and video lead asset support


packages/pages/src/section/index.js
```diff
@@ -7,6 +7,7 @@ import withNativeProvider from "../with-native-provider";
 
 const onPress = () => {};
 const SectionPage = ({ editionId, section, sectionTitle }) => {
+  console.log("SectionData:", section);
```
ValGeorgiev: Could we remove this bad boy, please?

PR #1719, opened 2019-02-18
feat: Edition Slice Tile - Tile H

jimhunty:
Tile H to support `Supporting 1 and Columnist`. If the left-hand column exceeds the length of the image, the image will bottom align with no padding as in the designs.

https://nidigitalsolutions.jira.com/browse/REPLAT-5144

screenshot 2019-02-18 at 16 18 26

### DIFF
--- a/packages/edition-slices/edition-tiles.showcase.js
+++ b/packages/edition-slices/edition-tiles.showcase.js
@@ -10,6 +10,7 @@ import {
   TileE,
   TileF,
   TileG,
+  TileH,
   TileI,
   TileJ,
   TileL,
@@ -66,6 +67,11 @@ const tileStories = [
   {
     name: "Tile G - Roundel image, 22pt headline, no teaser",
     Tile: TileG
+  },
+  {
+    name:
+      "Tile H - 22pt headline, teaser, byline at bottom all on left. Right hand, bottom aligned 2:3 image",
+    Tile: TileH
   },
   {
     name: "Tile I - Vertical, top image, centered aligned summary",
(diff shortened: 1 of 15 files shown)


packages/edition-slices/src/tiles/tile-h/styles/index.js
```diff
@@ -0,0 +1,33 @@
...
+      font: "headline",
+      fontSize: "infoTitle"
+    }),
+    marginBottom: spacing(2)
+  },
+  image: {
+    alignSelf: "flex-end"
+  },
+  imageContainer: {
+    // display: "flex",
```
tuncaulubilge: ☝️

jimhunty: 🤦‍♂️

packages/edition-slices/edition-tiles.showcase.js
```diff
@@ -66,6 +67,11 @@ const tileStories = [
     name: "Tile G - Roundel image, 22pt headline, no teaser",
     Tile: TileG
   },
+  {
+    name:
+      "Tile H - 22pt headline, teaser, byline at bottom all on left. Right hand, bottom aligned 2:3 image",
```
aronshamash: Think we should try and keep desc shorter if possible 
`Tile H - 22pt headline, teaser, byline at bottom left, image 2:3 right` ?

jimhunty: No the implication is the article summary is on the left, rather than the byline being bottom left. Perhaps there is a better way to phrase it?

PR #1720, opened 2019-02-18
Feat/comment lead and cartoon

nehasri89:
This PR creates the Comment Lead and Cartoon Slice with below changes

- Tile P and Tile Q
- A new slice layout
- A new edition slice 
- Changes in `article-byline` to take style for Opinion byline

_Note_ Changes for placement of byline will be discussed with Matt, and will come in a separate PR

screen shot 2019-02-18 at 17 54 20


packages/article-byline/article-byline-opinion.showcase.js
```diff
@@ -27,13 +27,26 @@ export default {
...
       name: "Byline with a single author",
       type: "story"
+    },
+    {
+      component: (_, { decorateAction }) => (
+        <ComponentWrapper>
+          <ArticleBylineOpinion
+            ast={authorsAST.authorInTheBeginning}
+            {...getProps(decorateAction)}
+            opinionStyle={{ fontSize: 12, letterSpacing: 0.6, lineHeight: 12 }}
```
jimhunty: Not a fan of inline style. Perhaps separate it out and move it to the top of the file?

jimhunty: Would a theme approach be more suitable rather than adding any styles to help restrict it? This seems like a `Comment` theme and therefore would be used in multiple places, not just a tile.

nehasri89: So the comment byline already has a theme, which is used in related articles but this slice needed a bit different, so I am not sure if it will go to theming.

packages/edition-slices/src/slices/commentleadandcartoon/index.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { View } from "react-native";
+import { CommentLeadAndCartoon } from "@times-components/slice-layout";
+import PropTypes from "prop-types";
+import { TileP, TileQ } from "../../tiles";
+
+const CommentLeadAndCartoonSlice = ({ onPress, slice: { lead, cartoon } }) => (
+  <View>
```
ValGeorgiev: Do we need View wrapper here, can we return CommentLeadAndCartoon only

nehasri89: good spot, I will remove it

PR #1721, opened 2019-02-18
feat: secondary 4 tablet

tuncaulubilge:
Secondary 4 tablet layout
- Reuses the same tile
- Passes breakpoint to slice-layout, and lays those slices in a col layout

screenshot_1550512634

### DIFF
--- a/packages/slice-layout/src/templates/secondaryfour/index.js
+++ b/packages/slice-layout/src/templates/secondaryfour/index.js
@@ -1,42 +1,60 @@
 import React from "react";
 import { View } from "react-native";
-import styles from "./styles";
-import propTypes from "./proptypes";
+import { editionBreakpoints } from "@times-components/styleguide";
+import styleFactory from "./styles";
+import { propTypes, defaultProps } from "./proptypes";
 import { ItemColSeparator, ItemRowSeparator } from "../shared";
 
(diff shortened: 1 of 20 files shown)


packages/slice-layout/src/templates/secondaryfour/index.js
```diff
@@ -1,42 +1,74 @@
...
 }) => {
-  const renderRowOne = [renderSecondary1(), renderSecondary2()];
-  const renderRowTwo = [renderSecondary3(), renderSecondary4()];
-  return (
-    <View style={styles.container}>
-      <View style={styles.itemContainer}>
-        <View key={renderRowOne[0].props.id} style={styles.item}>
-          {renderRowOne[0]}
+  const styles = styleFactory(breakpoint);
+  if (breakpoint === editionBreakpoints.small) {
```
aronshamash: Do you think the default should be `small` ? ie switch the logic

tuncaulubilge: We are treating large and xlarge as medium for now, I think we should do the same on your PR

tuncaulubilge: As for defaultProps, default value is small. But while rendering we only check small vs the rest.

PR #1722, opened 2019-02-19
feat: lead one and one tablet layout

aronshamash:
- https://nidigitalsolutions.jira.com/browse/REPLAT-5241
- Create new `Tile U - Horizontal, Flex: 1:3, Summary left, Image 3:2 right`
- removed padding under `Tile A` as it should be added to the slice

screenshot 2019-02-19 at 09 53 11


packages/edition-slices/src/tiles/tile-u/styles/index.js
```diff
@@ -0,0 +1,23 @@
+import { fonts, spacing } from "@times-components/styleguide";
+
+const styles = {
+  container: {
+    flexDirection: "row",
+    padding: spacing(2)
+  },
+  headline: {
+    fontFamily: fonts.headline,
+    fontSize: 25,
```
nehasri89: we have this fontsize in styleguide

tuncaulubilge: I know people have different opinions on this and I think it's worth a team-wide discussion. Personally, I think fontSizes should only be reused if the naming makes sense, not just based on value match. These are all homepage specific sizes, so we shouldn't be picking some random font from styleguide just because it's value is 25. 

For bigger tablet breakpoints, we'll need a font factory based on breakpoint for headline fontsizes, so I'd refactor these together with that larger breakpoint implementation

nehasri89: cool, I am happy with that 👍

jimhunty: I have been giving relevant naming when I have implemented it.

aronshamash: I was thinking same as @tuncaulubilge. I could have used one from styleguide but the name was not appropriate.

PR #1725, opened 2019-02-19
feat: Secondary One and Columnist Slice

jimhunty:
Secondary One and Columnist Slice: https://nidigitalsolutions.jira.com/browse/REPLAT-5144

screenshot 2019-02-19 at 15 03 12


aronshamash:
Looks good 👍. Couple of small design things:

- looks like we need and extra `paddingBottom: spacing(1)` for the headline in the top tile
- looks like there is too much margin between the flags and byline

packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -13,8 +13,9 @@ const TileSummary = ({
     article: { flags, hasVideo, headline, label, section, shortHeadline }
   },
   byline,
-  headlineStyle,
   bylineStyle,
+  headlineStyle,
+  isOpinionByline = true,
```
nehasri89: this has changed to `bylineStyle` now in master

packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -26,7 +27,7 @@ const TileSummary = ({
   <ArticleSummary
     bylineProps={
       byline
-        ? { ast: byline, isOpinionByline: true, opinionStyle: bylineStyle }
+        ? { ast: byline, isOpinionByline, opinionStyle: bylineStyle }
```
nehasri89: `opinionStyle` has changed to `bylineStyle` now in master

packages/edition-slices/src/tiles/tile-h/index.js
```diff
@@ -10,6 +10,7 @@ const TileH = ({ onPress, tile }) => (
       <TileSummary
         byline={tile.article.byline}
         headlineStyle={styles.headline}
+        isOpinionByline={false}
```
nehasri89: should not need this

jimhunty:
@aronshamash My understanding is that these tweaks, which are part of `ArticleSummary` will come in a design review. I agree it doesn't look great.

aronshamash:
Personally I think we should get as close as possible to the designs in the initial PR

jimhunty:
@aronshamash This would have been the tile PR not this slice one. A tile that I didn't work on. I would suggest getting this in and going to design review after.

PR #1726, opened 2019-02-19
fix: tile f text wrap bug on web

aronshamash:
- fixed text wrap issue in web only

### DIFF
--- a/packages/edition-slices/__tests__/android/__snapshots__/tiles-with-style.test.js.snap
+++ b/packages/edition-slices/__tests__/android/__snapshots__/tiles-with-style.test.js.snap
@@ -385,7 +385,6 @@ exports[`6. tile f 1`] = `
   <View
     style={
       Object {
-        "flexDirection": "row",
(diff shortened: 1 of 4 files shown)


packages/edition-slices/__tests__/android/__snapshots__/tiles-with-style.test.js.snap
```diff
@@ -385,7 +385,6 @@ exports[`6. tile f 1`] = `
   <View
     style={
       Object {
-        "flexDirection": "row",
         "paddingHorizontal": 10,
```
jimhunty: Shouldn't you add `flexWrap` here rather than remove the direction? RN is `column` be default so might be adding another regression.

aronshamash: we don't want it to use flexDirection `row` as tile is a vertical tile so the default `column` is fine. I checked on native btw.

PR #1729, opened 2019-02-19
fix: tile p style and byline

nehasri89:
This PR places byline to the bottom of Tile H, also centre aligns the strapline.

screen shot 2019-02-19 at 17 34 58

### DIFF
--- a/packages/article-byline/shared-showcase.js
+++ b/packages/article-byline/shared-showcase.js
@@ -109,7 +109,7 @@ export default (Component, name) => ({
         <ComponentWrapper>
           <Component
             ast={authorsAST.authorInTheBeginning}
-            opinionStyle={styles}
+            bylineStyle={styles}
             {...getProps(decorateAction)}
           />
         </ComponentWrapper>
--- a/packages/article-byline/shared-showcase.js
+++ b/packages/article-byline/shared-showcase.js
@@ -2,6 +2,15 @@
 
 import React from "react";
 import { View } from "react-native";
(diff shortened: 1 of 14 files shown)


packages/article-byline/src/article-byline.js
```diff
@@ -4,9 +4,18 @@ import renderByline from "./render-byline";
 import { propTypes, defaultProps } from "./article-byline-prop-types";
 import styles from "./styles";
 
-const renderAuthorComponent = (children, key, attributes, { className }) => ({
+const renderAuthorComponent = (
+  children,
+  key,
+  attributes,
+  { className, opinionStyle }
```
tuncaulubilge: Can we call this `bylineStyle`? Or is it opinion specific?

nehasri89: it is how a opinion byline would look, but actually thinking because we style a normal byline I guess we could call it `bylineStyle` 👍

packages/article-byline/shared-showcase.js
```diff
@@ -94,6 +103,19 @@ export default (Component, name) => ({
       ),
       name: "Byline that is very long",
       type: "story"
+    },
+    {
+      component: (_, { decorateAction }) => (
+        <ComponentWrapper>
+          <Component
+            ast={authorsAST.authorInTheBeginning}
+            opinionStyle={styles}
```
tuncaulubilge: bylineStyle

PR #1734, opened 2019-02-21
feat: REPLAT-5150 implement daily universal register

ValGeorgiev:
Daily Universal Register:
Slice **SSSS**: 
image

Tile **S** :
image


tuncaulubilge:
There shouldn't be a yarn.lock change on this pr, that's probably why it's failing on bitrise

packages/styleguide/src/colours/functional.js
```diff
@@ -6,12 +6,14 @@ const functionalColours = {
   articleFlagUpdated: "#3C81BE",
   backgroundPrimary: "#F9F9F9",
   backgroundSecondary: "#EDEDED",
+  blackHeadline: "#1D1D1B",
```
tuncaulubilge: Looks like we already have this colour as "brandColour", you can reuse that one

packages/styleguide/src/colours/functional.js
```diff
@@ -6,12 +6,14 @@ const functionalColours = {
...
   border: "#F0F0F0",
   brandColour: "#1D1D1B",
   bullet: "#103546",
   contrast: "#FFFFFF",
   darkSupplement: "#272D34",
   greyLabel: "#CCC",
+  greySeparator: "#DCDCDC",
```
tuncaulubilge: Lets double check with design on this colour, ideally it should be the same colour we use for keylines

packages/edition-slices/__tests__/shared-tiles.base.js
```diff
@@ -35,9 +39,12 @@ jest.mock("@times-components/link", () => "Link");
 jest.mock("@times-components/gradient", () => "Gradient");
 
 const tile = mockEditionSlice(1).items[0];
+const dailyRegisterItem = mockDailyRegisterSlice().birthdaysToday;
 
-const testTile = Tile => {
-  const output = TestRenderer.create(<Tile onPress={() => {}} tile={tile} />);
+const testTile = (Tile, registerItem) => {
```
tuncaulubilge: This is more or less what I had in mind, so it'll allow us to use a different mock on certain tiles if needed as well:
```suggestion
const testTile = (Tile, tile = mockTile) => {
```

But this is not a blocking comment, it works either way :+1:

tuncaulubilge:
Matt added Daily register visuals to zeplin, so you can export all the sizes from zeplin on the head logo.

Also you need to exclude assets from eslint and prettier for the CI to pass

packages/edition-slices/src/tiles/tile-s/styles/index.js
```diff
@@ -0,0 +1,39 @@
...
+const styles = {
+  bold: {
+    color: colours.functional.brandColour,
+    fontWeight: "bold"
+  },
+  byline: {
+    color: colours.functional.secondary,
+    fontFamily: fonts.supporting,
+    fontSize: 12,
+    marginTop: 10
```
nehasri89: use spacing

packages/edition-slices/src/tiles/tile-s/styles/index.js
```diff
@@ -0,0 +1,39 @@
...
+    color: colours.functional.secondary,
+    fontFamily: fonts.headline,
+    fontSize: 14,
+    lineHeight: 20
+  },
+  title: {
+    color: colours.functional.brandColour,
+    fontFamily: fonts.headline,
+    fontSize: fontSizes.smallHeadline,
+    marginBottom: 15
```
nehasri89: use spacing

nehasri89:
Looks great apart from one minor comment, also we have got the full set of images for this slice on zeplin now, so you can them add them here👍

PR #1735, opened 2019-02-21
fix: design review changes

aronshamash:
- as per https://nidigitalsolutions.jira.com/browse/REPLAT-5586

### DIFF
--- a/packages/edition-slices/src/tiles/tile-n/index.js
+++ b/packages/edition-slices/src/tiles/tile-n/index.js
@@ -23,9 +23,8 @@ const TileN = ({ onPress, tile }) => (
         flagColour={styles.flagColour}
         headlineStyle={styles.headline}
         labelColour={colours.functional.greyLabel}
+        strapline={tile.article.strapline}
         style={styles.summaryContainer}
-        summary={tile.article.summary125}
-        summaryStyle={styles.summary}
         tile={tile}
       />
     </View>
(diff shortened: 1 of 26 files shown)


packages/fixture-generator/src/mock-slice.ts
```diff
@@ -169,9 +161,13 @@ function mockSecondaryOneAndColumnistSlice(): SecondaryOneAndColumnistSliceWithN
 
 function mockSecondaryOneAndFourSlice(): SecondaryOneAndFourSliceWithName {
   const tiles = getTiles(5);
+  const secondaryTile = {
+    ...tiles[0],
+    article: { ...tiles[0].article, strapline: "Readers share their top tips" }
```
tuncaulubilge: Article doesn't have a strapline field, the tile does

packages/edition-slices/src/tiles/tile-n/index.js
```diff
@@ -17,9 +17,8 @@ const TileN = ({ onPress, tile }) => (
         flagColour={styles.flagColour}
         headlineStyle={styles.headline}
         labelColour={colours.functional.greyLabel}
+        strapline={tile.article.strapline}
```
tuncaulubilge: This should be tile.strapline as strapline only exists within the tile

nehasri89:
LG apart from tunca's comments 👍

aronshamash:
@nehasri89 updated `strapline` now. Please have another look.

PR #1736, opened 2019-02-21
fix: web - image component hides placeholder and loads images smoothly

Ivaylo1987:
[REPLAT-5595](https://nidigitalsolutions.jira.com/browse/REPLAT-5595)
fix: image component hides placeholder after load
fix: paint image smoothly in with opacity from 0 to 1

screenshot 2019-02-21 at 15 55 22


packages/image/__tests__/shared.web.js
```diff
@@ -141,6 +141,40 @@ export default () => {
...
+          <Image aspectRatio={3 / 2} lowResSize={200} uri="https://image.io" />
+        );
+
+        let [, placeholder] = testRenderer.toJSON().children[0].children;
+        expect(placeholder).toBeDefined();
+
+        const lowResImage = testRenderer.root.findByType("img");
+        lowResImage.props.onLoad();
+
+        [, placeholder] = testRenderer.toJSON().children[0].children;
```
tuncaulubilge: Placeholder might not always be the second child in there. Can we do a find by component name `Placeholder`?

Ivaylo1987: Fixed it with findAllByType()

PR #1745, opened 2019-02-22
fix: Missing images in storybook

jimhunty:
Images were not showing in the native storybook. This change fixes that.


aronshamash:
Is weird I'm sure image server worked without `www` before... anyway approved

PR #1750, opened 2019-02-25
fix: design review updates

aronshamash:
- as per https://nidigitalsolutions.jira.com/browse/REPLAT-5617


packages/edition-slices/src/tiles/tile-w/styles/index.js
```diff
@@ -3,19 +3,20 @@ import { fonts, spacing } from "@times-components/styleguide";
 const styles = {
   container: {
     flexDirection: "row",
-    padding: spacing(2)
+    paddingHorizontal: spacing(4),
+    paddingVertical: spacing(2)
   },
   headline: {
     fontFamily: fonts.headline,
```
jimhunty: Please use a font factory here.

PR #1751, opened 2019-02-26
Fix: Add missing test for fixture generator

ValGeorgiev:


aronshamash:
@ValGeorgiev need to update commit msg

PR #1754, opened 2019-02-27
fix: Roundel images working across ios and android

jimhunty:
https://nidigitalsolutions.jira.com/browse/REPLAT-5625

The issue was the tile was using the screen width to calculate the border-radius because the image was a percentage of the width, not a set value. There is no such thing as a proportional border-radius in React Native.

The Tile won't always fill the screen so this is wrong, so we now calculate the size of the container with which the image sits in. Additionally, the calculation was wrong as well assuming `25%` instead of `30%` so I have made this more explicit.

### DIFF
--- a/packages/edition-slices/src/tiles/tile-p/index.js
+++ b/packages/edition-slices/src/tiles/tile-p/index.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { Component } from "react";
 import { Dimensions, View } from "react-native";
 import PropTypes from "prop-types";
 import {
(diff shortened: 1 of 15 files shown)


packages/edition-slices/src/tiles/tile-p/index.js
```diff
@@ -10,27 +10,45 @@ import {
...
-    </View>
-  </TileLink>
-);
+class TileP extends Component {
+  constructor() {
+    super();
+    this.state = {
+      containerWidth: Dimensions.get("window").width
+    };
+  }
```
ValGeorgiev: Suggestion: 
To make it more clean and nice. You can remove the constructor method and use state like a class property:

`state = {
    containerWidth: Dimensions.get("window").width
}`

jimhunty: 👍 👏 💪

PR #1755, opened 2019-02-27
feat: Tablet 768px supplement lead and 4 stack

Ivaylo1987:
This is for: [REPLAT-5246](https://nidigitalsolutions.jira.com/browse/REPLAT-5246)

What is new:
 - Lead and 4 stack for tablet 768px breakpoint
       New tiles created - Tile AC, Tile AD

- Reviewed with UX

Tablet:
Screenshot 2019-03-11 at 16 36 56


packages/slice-layout/src/templates/leadoneandfour/styles.js
```diff
@@ -0,0 +1,16 @@
+const mediumBreakpointStyles = {
+  container: {
+    flex: 1,
+    flexDirection: "row",
+    paddingHorizontal: 10
+  },
+  leadContainer: {
+    padding: 10,
```
ValGeorgiev: please use **spacing** for padding/margin

Ivaylo1987:
> +1 on James's comment. Tiles should be stand alone components with embedded styles. We got TileSummary, TileImage and similar shared components to dry things up, so we are left with a simple TileX component which is basically a set of layout rules and styles.

OK, will create new Tiles and stick to what the practice currently is.
But still want to express that creating such amounts of duplicated code for such small changes seems odd. And think that a discussion on this would be useful.

packages/slice-layout/src/templates/leadoneandfour/styles.js
Ivaylo1987: Done!

packages/edition-slices/src/tiles/tile-ad/styles/index.js
```diff
@@ -0,0 +1,22 @@
...
+    flexDirection: "row",
+    padding: spacing(2)
+  },
+  headline: {
+    fontFamily: fonts.headline,
+    fontSize: 20,
+    marginBottom: spacing(2)
+  },
+  imageContainer: {
+    width: "45%"
```
aronshamash: Looking at the zeplin designs this 45:55 ratio doesn't look to match the designs. See https://app.zeplin.io/project/5c531dcd1dfb92353b64e7dc/screen/5c5dbb7d88bfb904c07f935b

Ivaylo1987: Sure, it doesn't. Initially I was following the design and reviewed the result with Matt Myers. His comment back then was that he likes 45% to 55% ratio better than the mobile 40% to 60%.
Should we ask him to review it again?

aronshamash: No. if you verbally agreed with him then that is fine but he needs to update Zeplin.

packages/edition-slices/src/tiles/shared/tile-link.js
```diff
@@ -5,19 +5,25 @@ import Link from "@times-components/link";
...
 
 TileLink.propTypes = {
   children: PropTypes.node.isRequired,
   onPress: PropTypes.func.isRequired,
+  style: PropTypes.shape({}),
   tile: PropTypes.shape({}).isRequired
 };
 
+TileLink.defaultProps = {
+  style: {}
```
tuncaulubilge: This can be null

packages/edition-slices/src/tiles/tile-ad/index.js
```diff
@@ -0,0 +1,39 @@
...
+  withTileTracking
+} from "../shared";
+import styles from "./styles";
+
+const TileAD = ({ onPress, tile }) => {
+  const { container, headline, imageContainer, summaryContainer } = styles;
+
+  return (
+    <TileLink onPress={onPress} tile={tile}>
+      <View style={container}>
```
tuncaulubilge: We can use the linkStyle in here as well

packages/edition-slices/src/tiles/tile-ad/index.js
```diff
@@ -0,0 +1,39 @@
...
+        <TileImage
+          aspectRatio={3 / 2}
+          style={imageContainer}
+          uri={getCrop(tile.article.leadAsset, "crop32")}
```
tuncaulubilge: provider-queries/section-fragment query has been updated recently, so please make sure these crops are included for this particular slice's query.

packages/edition-slices/src/tiles/shared/tile-link.js
```diff
@@ -5,19 +5,25 @@ import Link from "@times-components/link";
 const TileLink = ({
   children,
   onPress,
+  style,
   tile: {
     article: { url }
   }
 }) => (
-  <Link onPress={e => onPress(e, { url })} url={url}>
+  <Link linkStyle={style} onPress={e => onPress(e, { url })} url={url}>
```
tuncaulubilge: This is actually quite useful, but got two notes:

1. Make sure styles do work on both platforms. We use different components for android and ios, so you'll need to test both.
2. If it works, let's clean up other tiles as well and get rid of any extra views that was added just to support styling.

packages/edition-slices/src/tiles/shared/tile-link.js
Ivaylo1987: Removed. Dont actually need defaults here.

Ivaylo1987: done!

packages/edition-slices/src/tiles/tile-ad/index.js
Ivaylo1987: done!

packages/edition-slices/src/tiles/tile-ad/index.js
Ivaylo1987: LeadOneAndFourSlice fragment already contains crop 3/2
done!

packages/edition-slices/src/tiles/tile-ad/index.js
Ivaylo1987: I tried but it break the `web` version. This can me cleaned up in a future PR where we use `<TileLink>` styles.

packages/edition-slices/src/tiles/shared/tile-link.js
Ivaylo1987: It works ok for ios and android but is tricky with web. We can clean up in a future PR and make it work for all platforms.

packages/edition-slices/src/tiles/tile-ad/index.js
Ivaylo1987: Validated. Crops are in the query.

PR #1756, opened 2019-02-27
feat: Short Headline to be the primary headline for List Pages and Slices

jimhunty:
https://nidigitalsolutions.jira.com/browse/REPLAT-5646

We were using the longer headline. This PR rectifies this.

It also adds a test for these as well as setting up testing for the shared components of `Tiles` and utils of `ArticleList`. These areas could use some more tests following this PR.

### DIFF
--- a/packages/edition-slices/src/tiles/shared/tile-summary.js
+++ b/packages/edition-slices/src/tiles/shared/tile-summary.js
@@ -33,7 +33,7 @@ const TileSummary = ({
     flags={() => <ArticleFlags {...flagColour} flags={flags} />}
     headline={() => (
       <ArticleSummaryHeadline
-        headline={headline || shortHeadline}
+        headline={shortHeadline || headline}
         style={headlineStyle}
       />
     )}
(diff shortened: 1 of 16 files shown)


packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -33,7 +33,7 @@ const TileSummary = ({
     flags={() => <ArticleFlags {...flagColour} flags={flags} />}
     headline={() => (
       <ArticleSummaryHeadline
-        headline={headline || shortHeadline}
+        headline={shortHeadline || headline}
```
nathggns: Seems potentially weird we have the logic for picking which headline to use in multiple places, but it'd probably not be worth it to change it at this point. I'm fine either way 👍

jimhunty: I agree, its abstracted away in other packages in utility functions, but that in itself seems not needed as its such simple logic. Hence the test being added, if the logic needs to change at least the test will show the state of play before.

PR #1757, opened 2019-02-27
fix: modal image close button should be in a safe area

nathggns:
Using RN's `SafeAreaView`, we can ensure the modal exit button is in the correct place across all devices.

screen shot 2019-02-27 at 16 24 17

### DIFF
--- a/packages/image/src/styles/shared.js
+++ b/packages/image/src/styles/shared.js
@@ -3,7 +3,7 @@ import { colours } from "@times-components/styleguide";
 const styles = {
   buttonContainer: {
     marginLeft: 15,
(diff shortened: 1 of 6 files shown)


packages/image/src/styles/shared.js
```diff
@@ -3,7 +3,7 @@ import { colours } from "@times-components/styleguide";
 const styles = {
   buttonContainer: {
     marginLeft: 15,
-    marginTop: 20
+    marginTop: 15
```
jimhunty: styleguide `spacing()`

nathggns: 👍

PR #1758, opened 2019-02-27
feat: List 2 no pic 6 tablet

aronshamash:
- as per https://nidigitalsolutions.jira.com/browse/REPLAT-5245

### Android tablet Nexus landscape
screenshot 2019-02-27 at 16 27 10

### DIFF
--- a/packages/edition-slices/src/slices/listtwoandsixnopic/index.js
+++ b/packages/edition-slices/src/slices/listtwoandsixnopic/index.js
@@ -1,6 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { ListTwoAndSixNoPic } from "@times-components/slice-layout";
+import { ResponsiveSlice } from "../shared";
 import { TileL, TileC } from "../../tiles";
 
 const ListTwoAndSixNoPicSlice = ({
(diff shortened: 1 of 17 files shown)


packages/edition-slices/src/tiles/tile-l/styles/index.js
```diff
@@ -7,8 +7,8 @@ const styles = {
   },
   headlineStyle: {
     fontFamily: fonts.headline,
-    fontSize: 22,
-    lineHeight: 22,
+    fontSize: 20,
```
Ivaylo1987: Isn't this going to change for the mobile breakpoint, too?

aronshamash: yes it was incorrect before. both mobile and tablet should be 20 according to zeplin

Ivaylo1987: LGTM 🐷

packages/edition-slices/src/tiles/tile-l/styles/index.js
```diff
@@ -7,8 +7,8 @@ const styles = {
   },
   headlineStyle: {
     fontFamily: fonts.headline,
```
jimhunty: Should this use the `fontFactory`?

packages/edition-slices/src/slices/listtwoandsixnopic/index.js
```diff
@@ -15,34 +16,41 @@ const ListTwoAndSixNoPicSlice = ({
...
+        <TileL onPress={onPress} tile={support5} tileName="support5" />
+      )}
+      renderSupport6={() => (
+        <TileL onPress={onPress} tile={support6} tileName="support6" />
+      )}
+    />
+  );
+
+  return (
+    <ResponsiveSlice renderMedium={renderSlice} renderSmall={renderSlice} />
```
jimhunty: How does breakpoint get read in this case? In storybook should the render function have `renderMedium(breakpoint)`?

packages/edition-slices/src/tiles/tile-l/styles/index.js
aronshamash: was just updating an existing tile so prefer to refactor all tiles to use fontFactory in a different PR if we agree to do that.

aronshamash:
@jimhunty rebased so slice layout now has the knobs you added

packages/edition-slices/src/slices/listtwoandsixnopic/index.js
aronshamash: In `shared/responsive-slice.js` the breakpoint is sent as a param to `renderSmall` and `renderMedium`

jimhunty: Of course it is!

PR #1760, opened 2019-02-28
feat: Removing section colours for indepth template

jimhunty:
https://nidigitalsolutions.jira.com/browse/REPLAT-5452.

Pull quotes, Dropcap and Key Facts on an In-Depth template require no change of colours. The feedback from design that the contrasting colours to the background colour of the header were too much. This, therefore, covers further components being added. As a result, its a simple change for in-depth templates to not have access to other section colours.


nathggns:
👏

PR #1763, opened 2019-02-28
Feat/Leaders and Secondary one and Four consume publication

nehasri89:
We did not have the ST svg, so we have reused T svg for now which we will update once Matt is back
 
Leaders with Times publication:
screen shot 2019-03-01 at 15 02 01

Leaders with Sunday 
screen shot 2019-03-01 at 15 04 29
Times publication:


packages/section/src/section.js
```diff
@@ -76,6 +80,7 @@ class Section extends Component {
 Section.propTypes = {
   onPress: PropTypes.func,
   onViewed: PropTypes.func,
+  pubName: PropTypes.string.isRequired,
```
jimhunty: Possibly just me, but the shortening here of the field doesn't sit well. In TPA we use `publicationName` so I think we should follow that.

packages/edition-slices/src/slices/leaders/masthead.web.js
```diff
@@ -1,13 +1,23 @@
...
 
-const MastHead = () => (
-  <Image
-    aspectRatio={1435 / 250}
-    style={styles.mastheadStyle}
-    uri="https://www.thetimes.co.uk/d/img/leaders-masthead-d17db00289.png"
-  />
-);
+const MastHead = ({ publicationName }) => {
+  let uri =
```
jimhunty: Times occurs more often than Sunday Times so I would swap these round for a minimal perf benefit.

packages/edition-slices/src/slices/leaders/masthead.js
```diff
@@ -1,14 +1,22 @@
...
 
-const MastHead = () => (
-  <Image
-    resizeMode="contain"
+const MastHead = ({ publicationName }) => {
+  let style = "mastheadStyleST";
+  // eslint-disable-next-line global-require
+  let source = require("../../../assets/st-masthead.png");
+
+  if (publicationName === "TIMES") {
```
jimhunty: Same as the other comment, swapping round because Times is more likely to be used more often due to 6 editions vs 1 per week.

packages/edition-slices/edition-slices.showcase.js
```diff
@@ -47,10 +48,12 @@ const preventDefaultedAction = decorateAction =>
 const renderSlice = (Component, data) => (_, { decorateAction }) => (
   <Responsive>
     <ScrollView>
-      <Component
-        onPress={preventDefaultedAction(decorateAction)("onPress")}
-        slice={data}
-      />
+      <Context.Provider value={{ pubName: "TIMES" }}>
```
jimhunty: I would love to see a knob for showing the ST

packages/edition-slices/__tests__/shared-slices.base.js
```diff
@@ -130,5 +144,37 @@ export default () => {
     }
   }));
 
-  iterator(tests);
+  const testsForTimes = slicesWithPubLogo.map(({ mock, name, Slice }) => ({
```
jimhunty: Seems unnecessary duplication here?

nehasri89: but how else would you test if ST logo is coming up properly?

packages/edition-slices/src/slices/leaders/masthead.js
nehasri89: 👍

packages/edition-slices/src/slices/leaders/masthead.web.js
nehasri89: 👍

packages/section/src/section.js
nehasri89: done!

tuncaulubilge:
@nehasri89 we got the T svg from render, we can do the same for ST

packages/edition-slices/__tests__/shared-slices.base.js
```diff
@@ -116,8 +117,37 @@ const slices = [
...
+    Slice: SecondaryOneAndFourSlice
+  },
+  {
+    mock: mockLeadersSlice(),
+    name: "leaders slice",
+    Slice: LeadersSlice
+  }
+];
+
+const testsWithPublictaion = publicationName =>
```
tuncaulubilge: Can we move these to a separate test? This one is huge as it is

PR #1764, opened 2019-03-01
feat: 360 video icon

aronshamash:
- https://nidigitalsolutions.jira.com/browse/TOC-29
- New 360 video play button designs https://app.zeplin.io/project/5c77fbe0a481f9bcdf61adb5/dashboard


packages/video/src/inline-video-player.web.js
```diff
@@ -3,23 +3,54 @@ import React, { Component } from "react";
 import { propTypes, defaultProps } from "./video-prop-types";
 import SkySportsBanner from "./sky-sports-banner";
 
+const static360PlayIcon =
```
jimhunty: Should be in the Icons package.

packages/video/src/inline-video-player.web.js
```diff
@@ -3,23 +3,54 @@ import React, { Component } from "react";
...
   border-style: solid;
   border-width: 3px;
   border-color: white;
 }
 
+.video-js video:not([data-player="default"]) ~ button span {
+  display: none;
+}
+
+.video-js video:not([data-player="default"]) ~ button, .video-js video:not([data-player="default"]) ~ button:hover {
```
jimhunty: 🤯

nathggns: Little bit hacky, but it's the only way we can think of to know whether the video is 360 or not.

nehasri89:
can you put a screenshot for this pls @aronshamash

PR #1768, opened 2019-03-04
feat: Additional ad sizes for larger widths

jimhunty:
Within the ad package, you can add more sizes to a particular slot. This has been done now to support billboard and leaderboard.

Please note: billboard seems to give inconsistent results and often shows a leaderboard. This might be due to ad inventory? I would ask that someone with better knowledge on ads can advise?


aronshamash:
Where is the JIRA for this requirement @jimhunty ?

PR #1769, opened 2019-03-04
Feat/360 video

aronshamash:
- 360 Video player (web only)
- At moment we are inferring whether video is 360 or not from the playerId. This will change when boolean `is360Video` comes from Methode - JIRA ticket to follow.
- Zeplin designs: https://app.zeplin.io/project/5c77fbe0a481f9bcdf61adb5
screenshot 2019-03-04 at 17 11 51


packages/video/__tests__/web/__snapshots__/video-sky-sports-with-style.web.test.js.snap
```diff
@@ -4,47 +4,16 @@ exports[`1. video marked as sky sports 1`] = `
...
-    <video
-      style={
-        Object {
-          "height": 200,
-          "width": 300,
-        }
-      }
-    />
+  <div>
+    Video unable to play
```
tuncaulubilge: Is this expected?

PR #1770, opened 2019-03-04
feat: 360 video

aronshamash:
- BAU feature to enable Brightcove 360 Video player (web only)
- At moment we are inferring whether video is 360 or not from the playerId. This will change when boolean `is360Video` comes from Methode
- https://nidigitalsolutions.jira.com/browse/TOC-29
- Zeplin designs: https://app.zeplin.io/project/5c77fbe0a481f9bcdf61adb5
screenshot 2019-03-04 at 17 11 51


packages/video/__tests__/web/inline-video-player.web.test.js
```diff
@@ -31,7 +31,7 @@ describe("InlineVideoPlayer", () => {
     renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
 
     expect(document.body.innerHTML.trim()).toBe(
-      '<script src="//players.brightcove.net/[account id]/[player id]_default/index.min.js"></script>'
+      '<script src="//players.brightcove.net/[account id]/default_default/index.min.js"></script>'
```
nehasri89: this change means we will only test for default player? is that a conscious change?

aronshamash: yes it is a conscious change. Needed so `defaultVideoProps` could have `playerId="default"` and hence the 360videoplayer icon would not appear in all the snapshops.

packages/video/src/inline-video-player.web.js
```diff
@@ -104,7 +109,7 @@ class InlineVideoPlayer extends Component {
   };
```
nehasri89: does the skysports still hide the banner when played?

packages/video/__tests__/default-video-props.js
```diff
@@ -3,7 +3,7 @@ const videoProps = {
   height: 200,
   onVideoPress: () => {},
   paidOnly: false,
-  playerId: "[player id]",
+  playerId: "default",
```
nehasri89: do we have just one player id now?

packages/video/src/inline-video-player.web.js
aronshamash: yes and there is a test for it

packages/video/__tests__/default-video-props.js
aronshamash: how it works is that because we don't have an explicit `is360` boolean flag from Methode but we do get the correct playerId corresponding to the brightcove 360 player.
so
Methode set playerId as:
1. `default` when a standard video is embedded
2. `nPinp4a0B` (in Prod) when a 360 video is embedded

nehasri89: okay, so we only have these two player ids? and skysports is also a variation of default player ?

aronshamash: in practice only 2 playerIds yes. A forthcoming Methode release with do away with need to check for playerId. The `skySports` boolean _could_ be applied to a 360 video but in practice it will not happen.

PR #1774, opened 2019-03-05
feat: Secondary one and Columnist - tablet

nehasri89:
This PR creates two new tiles
- Tile AA 
- Tile AB

and updates the slice-layout and edition slice

screen shot 2019-03-06 at 11 19 29


packages/edition-slices/src/tiles/tile-aa/styles/index.js
```diff
@@ -0,0 +1,22 @@
+import { fonts, spacing } from "@times-components/styleguide";
+
+const styles = {
+  container: {
+    paddingLeft: spacing(2),
+    paddingRight: spacing(2)
+  },
+  headline: {
+    fontFamily: fonts.headline,
```
aronshamash: `fontFactory` ?

packages/edition-slices/src/tiles/tile-aa/styles/index.js
```diff
@@ -0,0 +1,22 @@
+import { fonts, spacing } from "@times-components/styleguide";
+
+const styles = {
+  container: {
+    paddingLeft: spacing(2),
```
aronshamash: `paddingHorizontal`

packages/edition-slices/edition-slices.showcase.js
```diff
@@ -100,7 +100,7 @@ const sliceStories = [
   },
   {
     mock: mockSecondaryOneAndColumnistSlice(),
-    name: "Secondary One and Columnist (TH)",
+    name: "Secondary One and Columnist (Mobile: TH, Tablet: AA AB)",
```
aronshamash: hmm now we have tile names with 2 characters we may need a new pattern to describe the tiles used in a slice

packages/edition-slices/src/tiles/tile-aa/styles/index.js
aronshamash: ok. I would like it as is then.

packages/edition-slices/edition-slices.showcase.js
aronshamash: think should be a separate PR. maybe space to separate tile name as you have done for `AA AB` ?

nehasri89: that's all I could think of for now 👍

PR #1783, opened 2019-03-07
feat: Huge and Wide support for slices

jimhunty:
Adds support for Huge and Wide slices: https://nidigitalsolutions.jira.com/browse/REPLAT-5701

Please note, this essentially nullifies the web snapshots. This is because we don't have an effective way to set screen widths for tests. Something that there is no obvious solution for and requires discussion. The focus is not on web, so I suggest ignoring these for now.

### DIFF
--- a/packages/edition-slices/src/slices/leadonefullwidth/index.js
+++ b/packages/edition-slices/src/slices/leadonefullwidth/index.js
@@ -5,8 +5,10 @@ import { ResponsiveSlice } from "../shared";
 
 const LeadOneFullWidthSlice = ({ slice: { lead }, onPress }) => (
   <ResponsiveSlice
+    renderHuge={() => null}
     renderMedium={() => <TileR onPress={onPress} tile={lead} tileName="lead" />}
     renderSmall={() => <TileA onPress={onPress} tile={lead} tileName="lead" />}
+    renderWide={() => null}
   />
 );
 
--- a/packages/edition-slices/src/slices/leadonefullwidth/index.js
+++ b/packages/edition-slices/src/slices/leadonefullwidth/index.js
@@ -5,10 +5,8 @@ import { ResponsiveSlice } from "../shared";
 
 const LeadOneFullWidthSlice = ({ slice: { lead }, onPress }) => (
   <ResponsiveSlice
-    renderHuge={() => null}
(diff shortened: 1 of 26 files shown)


packages/edition-slices/src/slices/leadonefullwidth/index.js
```diff
@@ -5,8 +5,10 @@ import { ResponsiveSlice } from "../shared";
 
 const LeadOneFullWidthSlice = ({ slice: { lead }, onPress }) => (
   <ResponsiveSlice
+    renderHuge={() => null}
     renderMedium={() => <TileR onPress={onPress} tile={lead} tileName="lead" />}
     renderSmall={() => <TileA onPress={onPress} tile={lead} tileName="lead" />}
+    renderWide={() => null}
```
ValGeorgiev: We can add a default values to the ResponsiveSlice, then we won't need to add everywhere functions which return null

tuncaulubilge: +1 on this. We should use fallbacks instead of default null values.
if renderMedium is null, use renderSmall
if renderWide is null, use renderMedium || renderSmall
if renderHuge is null, use renderWide || renderMedium || renderSmall

That way, we don't need to add all those null functions around, and that was the original goal of passing "breakpoint" value to the render functions.

ValGeorgiev: @tuncaulubilge I think we will have components which are rendered only on Desktop. So I believe, ResponsiveSlice.defaultProps will do the job.

packages/edition-slices/src/slices/shared/responsive-slice.js
```diff
@@ -4,21 +4,46 @@ import { ResponsiveContext } from "@times-components/responsive";
...
   <ResponsiveContext.Consumer>
-    {({ editionBreakpoint }) =>
-      editionBreakpoint === editionBreakpoints.small ? (
-        renderSmall(editionBreakpoint)
-      ) : (
-        <Gutter>{renderMedium(editionBreakpoint)}</Gutter>
-      )
-    }
+    {({ editionBreakpoint }) => {
+      switch (editionBreakpoint) {
```
aronshamash: think this could be an object literal to reduce code logic

PR #1784, opened 2019-03-07
feat: section press events

tuncaulubilge:
Article press and puzzle press events on sections.

+ Also fixed puzzle click analytics

### DIFF
--- a/packages/section/src/section.js
+++ b/packages/section/src/section.js
@@ -32,11 +32,11 @@ class Section extends Component {
       onArticlePress,
       onPuzzlePress,
       publicationName,
-      section: { slices, title },
+      section: { name, slices },
       onViewed,
       receiveChildList
     } = this.props;
-    const isPuzzle = title === "Puzzles";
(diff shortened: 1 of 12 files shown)


packages/section/src/section.js
```diff
@@ -29,14 +29,14 @@ class Section extends Component {
...
+      onArticlePress,
+      onPuzzlePress,
       publicationName,
       section: { slices, title },
       onViewed,
       receiveChildList
     } = this.props;
-
-    const showSeparator = title !== "Puzzles";
+    const isPuzzle = title === "Puzzles";
```
aronshamash: this is horrible but hey!

tuncaulubilge: I'll replace this with a typename check

Ivaylo1987: It was horrible even before :)

tuncaulubilge: Replaced with a typename check, instead of a title match now @aronshamash

aronshamash: nice

PR #1787, opened 2019-03-11
feat: Implement Leaders Slice for Tablet

ValGeorgiev:
Implemet the Leaders Slice for Tablet 
Portrait: 
image

Landscape:
image

### DIFF
--- a/packages/edition-slices/src/slices/leaders/index.js
+++ b/packages/edition-slices/src/slices/leaders/index.js
@@ -3,8 +3,9 @@ import { View, Text } from "react-native";
 import Context from "@times-components/context";
 import { Leaders } from "@times-components/slice-layout";
 import PropTypes from "prop-types";
-import { TileM } from "../../tiles";
+import { TileM, TileAG } from "../../tiles";
 import styles from "./styles";
+import { ResponsiveSlice } from "../shared";
 import MastHead from "./masthead";
 
 const renderHead = publicationName => (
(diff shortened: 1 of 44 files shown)


packages/slice-layout/src/templates/leaders/index.js
```diff
@@ -1,14 +1,38 @@
...
+      <Column
+        style={styles.container}
+        tiles={[renderLeader1, renderLeader2, renderLeader3]}
+      />
+    );
+  }
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.columnItems}>{renderLeader1()}</View>
```
tuncaulubilge: I'm not sure about this ordering.

In web, the middle tile seems to be leader1, so this is ordered as leader2 - leader1 - leader3.

@aronshamash might be able to verify this via render

aronshamash: this (the current render) ordering is correct.

packages/edition-slices/src/slices/leaders/index.js
```diff
@@ -16,27 +17,53 @@ const renderHead = publicationName => (
...
+  );
 
+  return (
+    <Context.Consumer>
+      {({ publicationName }) => (
+        <View style={styles.container}>
+          {renderHead(publicationName)}
+          <ResponsiveSlice
+            renderHuge={renderHugeSlice}
+            renderMedium={renderSmallSlice}
```
tuncaulubilge: shouldn't we renderHude in renderMedium?

ValGeorgiev: Tablet portrait is like mobile, thats why renderMedium and renderSmall are with the same values

tuncaulubilge: Same tile, different layout, got it! 👍

PR #1788, opened 2019-03-11
fix: Add standfirst to leaders article in case strapline is empty

tuncaulubilge:
Strapline on LeadersSlice is usually null, so we are picking standfirst from the article instead.


tuncaulubilge:
Let's not merge this until we clarify the approach on TPA. If this can be resolved on TPA, we can delete this PR

aronshamash:
@tuncaulubilge We should close this. TPA should have a strapline field.

tuncaulubilge:
Closing, as this should be resolved on TPA

PR #1792, opened 2019-03-11
fix: Tablet (768) slices

tuncaulubilge:
Fixing couple of things I noticed while checking the tablet slices:

- Secondary four tries to occupy the whole height of the screen. Fixed by using flex:1 instead of height:100% (this required us passing style to TileLink)
- Lead2NoPicAnd2 was treating "small" as the default breakpoint, causing it to be render small when we have wide and huge widths. Set "renderMedium" as the default behaviour, similar to all other slices


aronshamash:
@tuncaulubilge noticed some of this just now too.

packages/edition-slices/src/tiles/tile-c/styles/index.js
```diff
@@ -15,6 +14,9 @@ const styles = {
   imageContainer: {
     marginBottom: spacing(2),
     width: "100%"
+  },
+  link: {
```
nehasri89: I think TileLink's main responsibility is making the Tile clickable, style should not be part of this.

tuncaulubilge: Yeah, I agree on that in general, but <Link> component itself has a view inside it, and flex:1 needs to be on the outermost view, so we have to pass it into the Link

packages/slice-layout/src/templates/leadtwonopicandtwo/index.js
```diff
@@ -14,28 +14,28 @@ const LeadTwoNoPicAndTwoSlice = ({
...
+      <View style={styles.column}>
+        {renderLead1()}
+        <ItemRowSeparator />
+        {renderLead2()}
+      </View>
+      <ItemColSeparator />
+      <View style={styles.column}>
+        {renderSupport1()}
+        <ItemRowSeparator />
+        {renderSupport2()}
```
ValGeorgiev: I believe we can reuse Column component here as well and remove ItemRowSeparator dependancy

packages/edition-slices/src/tiles/tile-c/styles/index.js
Ivaylo1987: I tried this and it would not composed components to stretch to full height if `<TileLink>` does not have a `flex: 1` as the first parent. Cant think of any smarter way to do it.

packages/slice-layout/src/templates/leadtwonopicandtwo/index.js
tuncaulubilge: Done :+1:

jimhunty:
Regarding your second point. We should always default to the lower breakpoint, not the higher one. The change was on purpose.

This is reflected in my other PR.

tuncaulubilge:
@jimhunty medium breakpoint has gutters around it so it can be used on wide and huge as well. That's not the case for renderSmall, so that's why we are falling back to medium if wide and huge is not implemented

tuncaulubilge:
@jimhunty medium breakpoint has gutters around it so it can be used on wide and huge as well. That's not the case for renderSmall; that's why we are falling back to medium if wide and huge is not implemented

PR #1796, opened 2019-03-12
chore: Improve Article Label coverage in templates

jimhunty:
Non-default templates are not covering everything for their article labels.

Ideally with the Design Review PR this should increase the coverage by a fair bit.


packages/article-in-depth/__tests__/shared-internal-components.base.js
```diff
@@ -27,21 +27,23 @@ const snapshotTests = renderComponent => [
     }
   },
   {
-    name: "article label uses default section colour",
```
nehasri89: we should keep this test

PR #1798, opened 2019-03-13
feat: Add new params to section clicks

tuncaulubilge:
Added a few new params to onArticlePress and onPuzzlePress, as per native teams' requests

- onArticlePress( url, id )
- onPuzzlePress( url, title, id )

### DIFF
--- a/packages/edition-slices/src/slices/puzzle/index.js
+++ b/packages/edition-slices/src/slices/puzzle/index.js
@@ -7,11 +7,11 @@ import propTypes from "./proptypes";
 import styles from "./styles";
 import withTracking from "./puzzle-tracking-events";
 
-const Puzzle = ({ onPress, slice: { title, url, image } }) => {
+const Puzzle = ({ onPress, slice: { id, title, url, image } }) => {
   const { main, header, headLine, body } = styles;
 
   return (
-    <Link onPress={e => onPress(e, { url })} url={url}>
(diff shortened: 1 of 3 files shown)


packages/edition-slices/src/slices/puzzle/index.js
```diff
@@ -7,11 +7,11 @@ import propTypes from "./proptypes";
 import styles from "./styles";
 import withTracking from "./puzzle-tracking-events";
 
-const Puzzle = ({ onPress, slice: { title, url, image } }) => {
+const Puzzle = ({ onPress, slice: { id, title, url, image } }) => {
   const { main, header, headLine, body } = styles;
 
   return (
-    <Link onPress={e => onPress(e, { url })} url={url}>
+    <Link onPress={e => onPress({ id, title, url })} url={url}>
```
nehasri89: it doesn't need to pass the event anymore?

tuncaulubilge: No, native apps doesn't really need that event at all. It was getting filtered in pages anyway, now I'm filtering it out in here. I'll remove `e` from there as well

jimhunty:
@tuncaulubilge Shouldn't there be tests to reflect these changes?

tuncaulubilge:
@jimhunty, I added some click snapshots now :+1:

PR #1799, opened 2019-03-13
feat: Secondary one and four on tablet

aronshamash:
- as per https://nidigitalsolutions.jira.com/browse/REPLAT-5252

### Nexus 7 Landscape
Screenshot 2019-03-13 at 12 37 05

### DIFF
--- a/packages/edition-slices/__tests__/shared-tablet-slices.base.js
+++ b/packages/edition-slices/__tests__/shared-tablet-slices.base.js
@@ -37,22 +37,22 @@ import {
 const slices = [
   {
     mock: mockDailyRegisterSlice(),
-    name: "daily universal register tablet slice",
+    name: "daily universal register",
     Slice: DailyRegisterLeadFourSlice
   },
   {
     mock: mockLeadOneAndOneSlice(),
-    name: "lead one and one tablet slice",
+    name: "lead one and one",
     Slice: LeadOneAndOneSlice
   },
   {
     mock: mockLeadOneFullWidthSlice(),
-    name: "lead one full width tablet slice",
+    name: "lead one full width",
     Slice: LeadOneFullWidthSlice
   },
   {
     mock: mockLeadTwoNoPicAndTwoSlice(),
-    name: "lead two no pic and two tablet slice",
+    name: "lead two no pic and two",
     Slice: LeadTwoNoPicAndTwoSlice
   },
   {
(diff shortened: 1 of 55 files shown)


packages/edition-slices/__tests__/shared-tablet-slices.base.js
```diff
@@ -70,12 +77,12 @@ const slices = [
   },
   {
     mock: mockSecondaryOneAndColumnistSlice(),
-    name: "secondary one and columnist",
+    name: "secondary one and columnist tablet slice",
```
nehasri89: the file is only for tablet slices, we don't need to append tablet to each test case I think

packages/slice-layout/__tests__/sd1and4.base.js
```diff
@@ -1,12 +1,13 @@
...
 import { iterator } from "@times-components/test-utils";
+import { editionBreakpoints } from "@times-components/styleguide";
 import createItem from "./utils";
 import { SecondaryOneAndFourSlice } from "../src/slice-layout";
 
 export default renderComponent => {
   const tests = [
     {
-      name: "four child elements",
+      name: "secondary one and four - small",
```
nehasri89: usually we say default - `secondary one and four` and tablet `tablet - secondary one and four`

aronshamash: there are a few "patterns" for this in different slices. As small isn't the default in most cases I think I prefer this pattern. @tuncaulubilge ?

packages/edition-slices/__tests__/shared-tablet-slices.base.js
aronshamash: just following the existing pattern... can remove though ?

packages/slice-layout/__tests__/sd1and4.base.js
tuncaulubilge: +1 on Neha's comment. Looks like we appended "tablet" if it's a tablet

packages/edition-slices/__tests__/shared-tablet-slices.base.js
nehasri89: I would prefer without

packages/slice-layout/__tests__/sd1and4.base.js
aronshamash: should go in a separate pr anyway to make them all consistent

packages/edition-slices/__tests__/shared-tablet-slices.base.js
aronshamash: updated now

packages/slice-layout/__tests__/sd1and4.base.js
aronshamash: updating

aronshamash: I was using pattern in https://github.com/newsuk/times-components/blob/master/packages/slice-layout/__tests__/sd2np2.base.js

aronshamash: think I prefer `small`, `medium` as that is what we have in breakpoints, rather than `tablet` etc @tuncaulubilge @nehasri89 @jimhunty

PR #1800, opened 2019-03-13
fix: portrait images in modals

nathggns:
Portrait images should appear at full height in modals rather than full width.

Screen Shot 2019-03-13 at 14 37 06

Also some associated fixes around reducing the number of onLayout calculations during image load so that there are less loading states and "flashes" during image load. 

Other changes:

- Gradient can now take a specified width and height to display at instead of the size of its container
- Gradient will now use the startColour as a placeholder on native for the gradient as it loads

### DIFF
--- a/packages/gradient/src/gradient-prop-types-shared.js
+++ b/packages/gradient/src/gradient-prop-types-shared.js
@@ -1,0 +1,19 @@
+import PropTypes from "prop-types";
+import { ViewPropTypes } from "react-native";
+
+const { style: ViewPropTypesStyle } = ViewPropTypes;
+
+export const propTypes = {
+  children: PropTypes.oneOfType([
+    PropTypes.element,
+    PropTypes.arrayOf(PropTypes.element)
+  ]),
+  degrees: PropTypes.number,
+  style: ViewPropTypesStyle
+};
+
+export const defaultProps = {
+  children: null,
+  degrees: 265,
+  style: null
+};
--- a/packages/image/__tests__/ios/__snapshots__/modal-image.ios.test.js.snap
+++ b/packages/image/__tests__/ios/__snapshots__/modal-image.ios.test.js.snap
@@ -28,11 +28,32 @@ exports[`1. modal image 1`] = `
                 <View
                   aspectRatio={1.5}
                 >
-                  <View>
-                    <Gradient
-                      degrees={264}
-                    />
-                  </View>
+                  <Gradient
+                    degrees={264}
+                    height={350}
+                    width={700}
+                  >
+                    <View>
+                      <Svg
+                        height={191.87817258883248}
+                        version="1.1"
+                        viewBox="145 50 108 120"
+                        width={191.87817258883248}
+                      >
+                        <G
+                          fill="none"
+                          fillRule="evenodd"
+                          stroke="none"
+                          strokeWidth="1"
+                        >
+                          <Path
+                            d="6a6c07166e9c40e4f62e57af2b507822"
+                            fill="#FFFFFF"
+                          />
+                        </G>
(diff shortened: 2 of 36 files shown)


packages/gradient/src/gradient-prop-types-shared.js
```diff
@@ -0,0 +1,19 @@
...
+    PropTypes.element,
+    PropTypes.arrayOf(PropTypes.element)
+  ]),
+  degrees: PropTypes.number,
+  style: ViewPropTypesStyle
+};
+
+export const defaultProps = {
+  children: null,
+  degrees: 265,
```
aronshamash: why?

packages/image/src/placeholder.js
```diff
@@ -1,53 +1,40 @@
...
+function Placeholder({ height, style, width }) {
+  const hasDimensions = width !== null && height !== null;
+  return (
+    <Gradient
+      degrees={264}
+      height={height}
+      style={[style, !hasDimensions && { flex: 1 }]}
+      width={width}
+    >
+      {width !== null ? (
```
aronshamash: why the need for this. Why not `width ?` ?

packages/gradient/src/gradient-prop-types-shared.js
nathggns: It's just moving the existing defaultProps to a shared file. https://github.com/newsuk/times-components/pull/1800/files/940c7a61827eac78a7c20c270c70d959f5444008#diff-7d1ac95a8330f98989923192cd585c2dL17

aronshamash: sorry I meant why `265` when you are using `264` everywhere else ?

packages/image/src/placeholder.js
nathggns: It's just more explicit about what the "unhappy" path is. Think I made it more explicit to remind me why that check was there at all. Doesn't make a difference either way as the only valid values of null are a number or a null.

packages/image/__tests__/ios/__snapshots__/modal-image.ios.test.js.snap
```diff
@@ -145,11 +208,32 @@ exports[`2. modal image with custom highressize 1`] = `
...
-          degrees={264}
-        />
-      </View>
+      <Gradient
+        degrees={264}
+        height={350}
+        width={700}
+      >
+        <View>
+          <Svg
```
tuncaulubilge: Can you mock icons package in image tests? We should just see `<T/>` here instead of Svg paths etc.

packages/image/src/placeholder.js
aronshamash: rest of codebase doesn't use this pattern though so would change.

packages/gradient/src/gradient-prop-types-shared.js
nathggns: I didn't change it.

packages/image/__tests__/ios/__snapshots__/modal-image.ios.test.js.snap
nathggns: T isn't part of icons, it's just part of the image package and doesn't have its own tests (these snaps are the only place it is tested I believe). I also didn't change any of this, so not sure it should be a block on this PR.

tuncaulubilge: It wasn't in the snapshots before as onLayout was not called. So your PR fixed that, causing the T to appear everywhere :) 

We should move the T to icons and mock it here, but it doesn't have to be a part of this PR. Can you pick that one up after the design review?

PR #1807, opened 2019-03-14
fix: Fix broken snapshot click event

ValGeorgiev:
Fix broken snapshot click event

image

### DIFF
--- a/packages/edition-slices/src/tiles/shared/tile-link.js
+++ b/packages/edition-slices/src/tiles/shared/tile-link.js
@@ -10,7 +10,7 @@ const TileLink = ({
     article: { id, url }
   }
 }) => (
-  <Link linkStyle={style} onPress={e => onPress(e, { id, url })} url={url}>
+  <Link linkStyle={style} onPress={() => onPress({ id, url })} url={url}>
     {children}
   </Link>
 );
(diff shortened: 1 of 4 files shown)


packages/edition-slices/src/tiles/shared/tile-link.js
```diff
@@ -10,7 +10,7 @@ const TileLink = ({
     article: { id, url }
   }
 }) => (
-  <Link linkStyle={style} onPress={() => onPress({ id, url })} url={url}>
+  <Link linkStyle={style} onPress={e => onPress(e, { id, url })} url={url}>
```
tuncaulubilge: Can you do the same for puzzles as well, which has its own onPress event?

ValGeorgiev: There is no need to add `event` object there. Puzzle onPress is using our custom `Link` component which is using `TouchableOpacity`. `TouchableOpacity` onPress doesn't require or use the event object.

tuncaulubilge: Puzzle uses the same Link component as the TileLink. So we can remove the `event` object from both, and change the storybook so it doesn't depend on the `e` being passed

ValGeorgiev: Yes, we can do that, but we need to preventDefault there

PR #1809, opened 2019-03-14
feat: Flag expiry

aronshamash:
- As per https://nidigitalsolutions.jira.com/browse/REPLAT-5484
- `flags` field on `Article` is now deprecated. This PR uses `expirableFlags`
- Also needed to update `types.ts` which hadn't been done in a while


packages/article-in-depth/src/article-header/article-header.js
```diff
@@ -55,7 +55,7 @@ const ArticleHeader = ({
               {headline}
             </Text>
 
-            {flags.length > 0 && (
+            {getActiveFlags(flags).length > 0 && (
```
nehasri89: ain't we doing this again in `article-flag`, think we can skip this call from here

packages/article/showcase-helper.js
```diff
@@ -192,6 +192,9 @@ const renderArticle = ({
         template,
         textColour: inDepthTextColour
       };
+
+      console.log("data", data);
```
nehasri89: this can go

packages/article-flag/package.json
```diff
@@ -58,7 +58,7 @@
     "react-native": "0.55.4",
     "react-native-web": "0.9.0",
     "react-test-renderer": "16.5.2",
-    "webpack": "4.6.0",
+    "webpack": "4.29.6",
```
nehasri89: why does it need a webpack upgrade?

aronshamash: good spot. not intentional

PR #1810, opened 2019-03-15
feat: Remove showInteractives flag and delete logic to remove flag

tuncaulubilge:
Part of: https://nidigitalsolutions.jira.com/browse/REPLAT-5579

We no longer the feature flag to disable interactives, as we have the versioned whitelisting in place.


jimhunty:
Does there need to be a supporting piece of work in the Android App?

tuncaulubilge:
Android and iOS parts of this are coming soon, but doesn't need to block this PR

PR #1816, opened 2019-03-19
chore: change codeowners

ValGeorgiev:
Replace James as a code owner and use usernames instead of emails


CODEOWNERS
```diff
@@ -1,2 +1,2 @@
 # The following people are responsible for this project and its live instances:
-*    neha.srivastava1@news.co.uk tunca.ulubilge@news.co.uk nathaniel.higgins@news.co.uk james.miller@news.co.uk aron.shamash@news.co.uk jaye.marshall@news.co.uk
+*    nehasri89 tuncaulubilge nathggns aronshamash jmars ValGeorgiev
```
araines: Not the correct format, see the RFC:  https://github.com/newsuk/rfcs/blob/master/0004-codeowners.md

ie. should be like `@nehasri89`

ValGeorgiev: @araines done

PR #1818, opened 2019-03-20
Feat: Slice accessibility

ValGeorgiev:
Add accessibility to slices and tiles

Add accessibilityLabel and accessibilityRole to all texts, buttons, images and etc.

Add dots after every label, because we need pause between texts

### DIFF
--- a/packages/article-comments/__tests__/android/__snapshots__/shared.android.test.js.snap
+++ b/packages/article-comments/__tests__/android/__snapshots__/shared.android.test.js.snap
@@ -9,7 +9,6 @@ exports[`comments error 1`] = `
     You must have a network connection to view and post comments. Please check your connection and retry.
   </Text>
   <View
-    accessibilityComponentType="button"
     accessibilityLabel="Retry"
   >
     <Text
@@ -53,7 +52,6 @@ exports[`enabled comments 1`] = `
     </Text>
   </Text>
   <View
-    accessibilityComponentType="button"
     accessibilityLabel="View comments"
   >
     <Text
@@ -79,7 +77,6 @@ exports[`single comment 1`] = `
     </Text>
   </Text>
   <View
-    accessibilityComponentType="button"
     accessibilityLabel="View comments"
   >
     <Text
@@ -105,7 +102,6 @@ exports[`zero comments 1`] = `
     </Text>
   </Text>
   <View
-    accessibilityComponentType="button"
     accessibilityLabel="Post a comment"
   >
     <Text
--- a/packages/article-comments/__tests__/android/__snapshots__/shared.android.test.js.snap
+++ b/packages/article-comments/__tests__/android/__snapshots__/shared.android.test.js.snap
@@ -11,7 +11,6 @@ exports[`comments error 1`] = `
   <View
     accessibilityComponentType="button"
     accessibilityLabel="Retry"
-    accessibilityTraits="button"
   >
     <Text
       title="Retry"
@@ -56,7 +55,6 @@ exports[`enabled comments 1`] = `
   <View
     accessibilityComponentType="button"
     accessibilityLabel="View comments"
-    accessibilityTraits="button"
   >
     <Text
       title="View comments"
@@ -83,7 +81,6 @@ exports[`single comment 1`] = `
   <View
     accessibilityComponentType="button"
     accessibilityLabel="View comments"
-    accessibilityTraits="button"
   >
     <Text
       title="View comments"
@@ -110,7 +107,6 @@ exports[`zero comments 1`] = `
   <View
     accessibilityComponentType="button"
     accessibilityLabel="Post a comment"
-    accessibilityTraits="button"
   >
     <Text
       title="Post a comment"
--- a/packages/article-label/__tests__/android/__snapshots__/article-label.android.test.js.snap
+++ b/packages/article-label/__tests__/android/__snapshots__/article-label.android.test.js.snap
@@ -2,6 +2,7 @@
 
 exports[`Article Label test on android renders ArticleLabel 1`] = `
 <Text
+  accessibilityLabel="swimming."
   accessible={true}
   allowFontScaling={true}
   ellipsizeMode="tail"
@@ -23,6 +24,7 @@ exports[`Article Label test on android renders ArticleLabel 1`] = `
 
 exports[`Article Label test on android renders ArticleLabel with GraphQL Colour type 1`] = `
 <Text
+  accessibilityLabel="swimming."
   accessible={true}
   allowFontScaling={true}
   ellipsizeMode="tail"
--- a/packages/article-summary/src/article-summary-content.js
+++ b/packages/article-summary/src/article-summary-content.js
@@ -7,7 +7,7 @@ import styles from "./styles";
 
 const ArticleSummaryContent = ({ ast, className, style }) =>
   ast.length > 0 ? (
-    <Text className={className} style={[styles.text, style]}>
+    <Text accessible className={className} style={[styles.text, style]}>
       {renderAst(ast)}
     </Text>
   ) : null;
--- a/packages/article-summary/src/article-summary-headline.js
+++ b/packages/article-summary/src/article-summary-headline.js
@@ -7,6 +7,7 @@ const { style: TextStylePropTypes } = Text.propTypes;
 
 const ArticleSummaryHeadline = ({ className, headline, style }) => (
   <Text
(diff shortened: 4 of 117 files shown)


packages/article-summary/src/article-summary-content.js
```diff
@@ -7,7 +7,7 @@ import styles from "./styles";
 
 const ArticleSummaryContent = ({ ast, className, style }) =>
   ast.length > 0 ? (
-    <Text className={className} style={[styles.text, style]}>
+    <Text accessible className={className} style={[styles.text, style]}>
```
tuncaulubilge: Touchables (the wrapping link) is accessible by default, so I'm not sure if this is required, I'll give it a quick check

packages/article-summary/src/article-summary-headline.js
```diff
@@ -7,7 +7,9 @@ const { style: TextStylePropTypes } = Text.propTypes;
 
 const ArticleSummaryHeadline = ({ className, headline, style }) => (
   <Text
+    accessibilityLabel={headline}
```
tuncaulubilge: This should be the children by default

packages/edition-slices/src/slices/dailyregisterleadfour/index.js
```diff
@@ -15,17 +17,25 @@ const DailyRegisterLeadFour = ({
...
         <Logo
+          accessibilityLabel={`${headline} logo`}
           imageUri="https://www.thetimes.co.uk/d/img/DUR-masthead-40fe00731f.png"
           ratio={1435 / 250}
           style={styles.mastheadLogo}
           type="logo"
         />
-        <Text style={styles.title}>Daily Universal Register</Text>
+        <Text
+          accessibilityLabel={headline}
```
tuncaulubilge: We shouldn't need this label

packages/edition-slices/src/slices/leaders/index.js
```diff
@@ -8,11 +8,19 @@ import styles from "./styles";
...
 
+const headline = "Leading Articles";
+
 const renderHead = publicationName => (
   <View style={styles.mastheadContainer}>
     <MastHead publicationName={publicationName} />
     <View style={styles.headingContainer}>
-      <Text style={[styles.heading, styles.text]}> Leading Articles </Text>
+      <Text
+        accessibilityLabel={headline}
```
tuncaulubilge: We shouldn't need this label

packages/edition-slices/src/tiles/shared/tile-link.js
```diff
@@ -8,21 +8,29 @@ const TileLink = ({
   style,
   tile: {
     article: { id, url }
-  }
+  },
+  accessible
 }) => (
-  <Link linkStyle={style} onPress={() => onPress({ id, url })} url={url}>
+  <Link
+    accessible={accessible}
```
tuncaulubilge: Touchables are accessible by default

packages/edition-slices/src/tiles/tile-a/index.js
```diff
@@ -10,7 +10,7 @@ import {
 import styles from "./styles";
 
 const TileA = ({ onPress, tile }) => (
-  <TileLink onPress={onPress} tile={tile}>
+  <TileLink accessible={false} onPress={onPress} tile={tile}>
```
tuncaulubilge: This should be accessible as it's clickable

packages/edition-slices/src/tiles/tile-s/index.js
```diff
@@ -16,7 +16,13 @@ function renderAst(ast) {
 const TileS = ({ tile }) => (
   <View style={styles.container}>
     <View style={styles.titleWrapper}>
-      <Text style={styles.title}>{tile.title}</Text>
+      <Text
+        accessibilityLabel={tile.title}
```
tuncaulubilge: We shouldn't need this label

packages/image/src/image.js
```diff
@@ -79,6 +79,8 @@ class TimesImage extends Component {
             <Placeholder height={height} width={width} />
             {lowResSize ? (
               <Image
+                accessibilityRole="image"
```
tuncaulubilge: Isn't an RN image's accessibility role `image` by default?

packages/link/src/link.js
```diff
@@ -2,21 +2,23 @@ import React from "react";
 import { TouchableOpacity, ViewPropTypes } from "react-native";
 import PropTypes from "prop-types";
 
-const Link = ({ children, linkStyle, onPress }) => (
-  <TouchableOpacity onPress={onPress} style={linkStyle}>
+const Link = ({ children, linkStyle, onPress, accessible }) => (
+  <TouchableOpacity accessible={accessible} onPress={onPress} style={linkStyle}>
```
tuncaulubilge: Touchables are accessible by default

packages/edition-slices/src/tiles/shared/tile-link.js
ValGeorgiev: You are right, but that is the issue here. We wrap tile in TouchableOpacity and its accessible by default, thats why I need to change it to accessible=false to allow the readers read the childrens (text, images and etc) I describe it in the PR description.

packages/image/src/image.js
ValGeorgiev: Image is not accessible by default in RN

packages/link/src/link.js
ValGeorgiev: Check the description challenge 1

packages/edition-slices/src/tiles/tile-a/index.js
ValGeorgiev: Check the description challenge 1

tuncaulubilge:
I've given it a quick look on both platforms yesterday, and it behaves exactly how we like it to behave actually, without these changes. When you click on a tile, it reads out the whole tile as a whole, and you can even navigate to the next tile with a gesture. We'd lose that behaviour if we add accessibility on each text view

ValGeorgiev:
@tuncaulubilge You are right, but this is not how is working on prod. If Adam is fine with reading everything in one sentence. I will decline the PR and move the story in Done as won't fix

tuncaulubilge:
Adam was also a part of the review yesterday and he is quite happy with the behaviour. What we have right now is better than the production app :) 

The story's goal is to check the accessibility and fix any issues we encounter but it looks like it's fine. You can close the PR and move the story to done 👍 Cheers

packages/article-summary/src/article-summary-content.js
ValGeorgiev: This can be used as a standalone text, not wrapped by touchable. Thats why I added accessible=true

packages/article-comments/__tests__/android/__snapshots__/shared.android.test.js.snap
```diff
@@ -11,7 +11,6 @@ exports[`comments error 1`] = `
   <View
     accessibilityComponentType="button"
     accessibilityLabel="Retry"
-    accessibilityTraits="button"
```
tuncaulubilge: What happened to this?

packages/article-label/__tests__/android/__snapshots__/article-label.android.test.js.snap
```diff
@@ -2,6 +2,7 @@
 
 exports[`Article Label test on android renders ArticleLabel 1`] = `
 <Text
+  accessibilityLabel="swimming."
```
tuncaulubilge: Let's double check with adam if we want to change the article behaviour

packages/article-summary/src/article-summary-content.js
tuncaulubilge: We don't need a separate accessible in here

packages/article-comments/__tests__/android/__snapshots__/shared.android.test.js.snap
ValGeorgiev: This is deprecated with componentType, i will remove both and use `accessibilityRole`. Good catch, thanks

packages/article-label/__tests__/android/__snapshots__/article-label.android.test.js.snap
ValGeorgiev: Its changed because we are using it in slices

ValGeorgiev:
Dots fix is not working, we are fine for now with the default behaviour. We will think about improving in the future.

PR #1819, opened 2019-03-21
feat: Puzzle notification bar

aronshamash:
- as per https://nidigitalsolutions.jira.com/browse/REPLAT-5406

### Mobile:
Screenshot 2019-03-21 at 12 18 16

### Tablet
Screenshot 2019-03-21 at 12 17 59

### DIFF
--- a/packages/section/__tests__/ios/__snapshots__/section.test.js.snap
+++ b/packages/section/__tests__/ios/__snapshots__/section.test.js.snap
@@ -1,5 +1,80 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`puzzle bar with more than one puzzle 1`] = `
+<View
+  style={
+    Object {
+      "borderBottomWidth": 1,
+      "borderColor": "#DBDBDB",
+      "flexDirection": "row",
+      "justifyContent": "center",
+      "paddingVertical": 20,
+    }
+  }
+>
+  <Text
+    style={
+      Object {
+        "color": "#1D1D1B",
+        "fontFamily": "TimesModern-Regular",
+        "fontSize": 20,
+      }
+    }
+  >
+    3
+     recently opened puzzle
+    s
+  </Text>
+  <View
+    style={
+      Object {
+        "paddingLeft": 10,
+        "paddingTop": 5,
+      }
+    }
+  >
+    <IconForwardArrow />
+  </View>
+</View>
+`;
+
+exports[`puzzle bar with one puzzle 1`] = `
+<View
+  style={
+    Object {
+      "borderBottomWidth": 1,
+      "borderColor": "#DBDBDB",
+      "flexDirection": "row",
+      "justifyContent": "center",
+      "paddingVertical": 20,
+    }
+  }
+>
+  <Text
+    style={
+      Object {
+        "color": "#1D1D1B",
+        "fontFamily": "TimesModern-Regular",
+        "fontSize": 20,
+      }
+    }
+  >
+    1
+     recently opened puzzle
+  </Text>
+  <View
+    style={
+      Object {
+        "paddingLeft": 10,
+        "paddingTop": 5,
+      }
+    }
+  >
+    <IconForwardArrow />
+  </View>
+</View>
+`;
+
 exports[`section page 1`] = `
 <RCTScrollView>
   <View>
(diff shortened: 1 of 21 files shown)


tuncaulubilge:
No need for a special logic to hide in Android @aronshamash . We'll just hide it if recentlyOpenedPuzzleCount is 0, which will be the case in android

packages/section/src/section.js
```diff
@@ -46,6 +49,12 @@ class Section extends Component {
 
     return (
       <Responsive>
+        {isPuzzle && (
```
tuncaulubilge: this should be a ListHeaderComponent in the flatlist:
```
          ListHeaderComponent={
            isPuzzle ? <PuzzleBar onPress={onPuzzleBarPress} /> : null
          }
```

packages/section/src/styles/index.js
```diff
@@ -9,6 +9,22 @@ const styles = StyleSheet.create({
   listItemSeparator: {
     backgroundColor: colours.functional.keyline,
     height: 1
+  },
+  puzzleNotificationBarArrow: {
```
tuncaulubilge: Can we just call this a `puzzleBar`? Not a big deal but it's quite a long name atm :)

aronshamash: it is long

packages/section/src/puzzle-bar.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { Text, View } from "react-native";
+import PropTypes from "prop-types";
+import { IconForwardArrow } from "@times-components/icons";
+import { colours } from "@times-components/styleguide";
+import styles from "./styles";
+
+const PuzzleBar = ({ count, onPress }) =>
```
tuncaulubilge: We need some snapshots on this, then it's good to ship

packages/section/src/puzzle-bar.js
```diff
@@ -0,0 +1,26 @@
...
+  count ? (
+    <View style={styles.puzzleBarContainer}>
+      <Text onPress={onPress} style={styles.puzzleBarText}>
+        {count} recently opened puzzle
+        {count > 1 && "s"}
```
Ivaylo1987: I like that 👍

packages/section/src/puzzle-bar.js
aronshamash: sorted

packages/section/__tests__/ios/__snapshots__/section.test.js.snap
```diff
@@ -1,5 +1,80 @@
...
+      Object {
+        "color": "#1D1D1B",
+        "fontFamily": "TimesModern-Regular",
+        "fontSize": 20,
+      }
+    }
+  >
+    3
+     recently opened puzzle
+    s
```
tuncaulubilge: Should this be in three lines?

aronshamash: It doesn't do any harm. Think it is due to the formatting of the JSX.

packages/section/__tests__/shared.js
```diff
@@ -45,11 +50,25 @@ export default () => {
...
           onPuzzlePress={() => {}}
           publicationName="TIMES"
+          recentlyOpenedPuzzleCount={1}
           section={edition.sections[0]}
         />
       ).toJSON()
     ).toMatchSnapshot();
   });
+
+  it("puzzle bar with one puzzle", () => {
```
nehasri89: is having 1 or more than 1 count account for different states?
should there be a test with zero puzzle count?

packages/section/src/section.js
```diff
@@ -57,6 +60,14 @@ class Section extends Component {
             ))
           }
           keyExtractor={item => item.elementId}
+          ListHeaderComponent={
+            isPuzzle ? (
```
nehasri89: do we display the bar even with zero count?

packages/section/__tests__/shared.js
aronshamash: it is needed to test for the logic in the `PuzzleBar` : `{count > 1 && "s"}`

packages/section/src/section.js
aronshamash: no but i was missing a test for that. I have added now.

PR #1821, opened 2019-03-22
feat: ios recent puzzles bridge

tuncaulubilge:
- Added an async function to `SectionEvents` native bridge, allowing native apps to update "recently opened puzzle count" via `getOpenedPuzzleCount` method. Native apps needs to implement a method that receives a promise and resolve that promise with an integer value.
- Renamed the context we use in sections to SectionContext. We can add saved star and read/unread flags to the same context in the future, and update those alongside puzzle count
- Also did a bit of clean up on pages/section and simplified the flow.


packages/section/src/section.js
```diff
@@ -14,6 +13,7 @@ import { splitPuzzlesBySlices, buildSliceData } from "./utils";
 class Section extends Component {
   constructor(props) {
     super(props);
+    this.renderItem = this.renderItem.bind(this);
```
ValGeorgiev: We can remove the constructor here and use arrow functions.

tuncaulubilge: This saves us from instantiating same function on each render loop

ValGeorgiev: i mean, you can declare the function like this:
`renderItem = () => { // ..some logic }` 
this won't affect the performance, because babel will do pretty much the same like you. It will move the renderItem function in the constructor. And won't instantiating the function on every loop. It's prettier and we won't need the constructor only for binding some functions.

tuncaulubilge: It sure does look prettier :) I'll give this a try and check the transpiled code

ValGeorgiev: https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAECCAO9oG8BQ0AuBTCHoF5oAKASgID4V1pp0BfVOoA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2017%2Creact%2Cstage-1%2Cstage-2&prettier=false&targets=&version=7.4.2 

something like this

packages/context/src/section-context.js
```diff
@@ -0,0 +1,6 @@
+import { createContext } from "react";
+
+export default createContext({
+  publicationName: "TIMES",
```
aronshamash: we should probably get these constants eg `TIMES` from somewhere central...

packages/section/src/section.js
tuncaulubilge: This is the way it's done on the rest of the repo, and honestly I'd rather keep it consistent unless there is an obvious benefit to it

PR #1826, opened 2019-03-25
fix: Use listing asset override

nehasri89:
The order or precedence as discussed with backed team is 

tile.leadAsset || tile.article.listingAsset || tile.article.leadAsset


aronshamash:
@nehasri89 coverage has reduced. Have we tested all code paths in the fallback logic?

nehasri89:
@aronshamash yes we have tests for both the overrides in shared component test

ValGeorgiev:
I trigger a rebase to fix the netlify deploy and after that will merge it

PR #1827, opened 2019-03-26
feat: Section pages save star bridge

tuncaulubilge:
This is an early draft on the save star bridge using SectionContext
https://nidigitalsolutions.jira.com/browse/REPLAT-5589

Also check the button POC to see how to use this in a save button: https://github.com/newsuk/times-components/pull/1828

- Added two new fields to SectionContext
1. onSaveStarPress - Callback for saveStar clicks, passed down using context to simplify. This uses `NativeModules.SectionEvents.onSaveStarPress(save:Boolean, articleId:String, promise:Promise)`. If this function is `null`, we should assume the feature is not supported on that platform and don't display the star
2. savedArticles: This is an object, containing the saved articles by id. If this object is `null`, the save stars should be inactive

- Added bridge integration on pages package. There is a two-way bridge for updating saved status

1. Async NativeModules bridge using `NativeModules.SectionEvents.getSavedArticles(promise:Promise)`, which gets resolved with a list of article ids
2. Event push mechanism using react-native's `DeviceEventEmitter`. Components listen for a event type called `updateSavedArticles`. The NativeModule mentioned above is used to fetch the list of saved articles whenever this event is received, and updates the context

P.S For android devs: Here is a short doc on how to use DeviceEventEmitter on the native app:
https://facebook.github.io/react-native/docs/native-modules-android#sending-events-to-javascript


packages/pages/src/section/section.js
```diff
@@ -49,14 +54,27 @@ class SectionPage extends Component {
         this.setState({ recentlyOpenedPuzzleCount: count });
       });
     }
+
+    if (getSavedArticles) {
+      getSavedArticles().then(articleIds => {
```
Ivaylo1987: Should we do some error catching here or we will have another error handling strategy?

tuncaulubilge: We don't care about the error scenarios on this one, and just assume it's still syncing until we get a proper resolution. 

We might implement some error cases on the actual onSaveStarPress events, as the apps might fail saving that article

packages/pages/src/section/section.js
```diff
@@ -49,14 +54,27 @@ class SectionPage extends Component {
...
-    const { recentlyOpenedPuzzleCount } = this.state;
+    const { recentlyOpenedPuzzleCount, savedArticles } = this.state;
     return (
       <SectionContext.Provider
-        value={{ publicationName, recentlyOpenedPuzzleCount }}
+        value={{
+          onSaveStarPress,
+          publicationName,
+          recentlyOpenedPuzzleCount,
+          savedArticles
```
ValGeorgiev: here savedArticles is an array with objects, but in the test file is just an object which contains all articles.. 

What is the reason behind true in [{articleId: true}]

tuncaulubilge: It should just be an object, I'll fix the array :) that should be a reduce, not a map.

Used objects instead of sets as the read time on sets are O(n)

packages/pages/src/section/section.js
Ivaylo1987: 👍

packages/pages/src/section/section.js
tuncaulubilge: Should be fixed now :+1: thanks

packages/pages/src/section/section.js
```diff
@@ -49,14 +54,44 @@ class SectionPage extends Component {
...
+    const { recentlyOpenedPuzzleCount, savedArticles } = this.state;
+
     return (
       <SectionContext.Provider
-        value={{ publicationName, recentlyOpenedPuzzleCount }}
+        value={{
+          onSaveStarPress: onSaveStarPress
+            ? (save, articleId) =>
+                onSaveStarPress(save, articleId).then(() => {
+                  if (save) {
```
aronshamash: can use tertiary here

tuncaulubilge: Done

packages/context/src/section-context.js
```diff
@@ -1,6 +1,8 @@
 import { createContext } from "react";
 
 export default createContext({
+  onSaveStarPress: null,
```
Ivaylo1987: What about `onArticleSavePress` isn't it the one we are gonna use?

PR #1828, opened 2019-03-26
Proof of Concept: dummy tile star with context

tuncaulubilge:
This is a proof of concept, based on saved-star-context work. It consumes SectionContext and renders the button with multiple states

- Hidden: Don't show the star if onSaveStarPress is not implemented on that client
- Inactive: Disable and gray out the star if savedArticles map is not present (this happens while the client is syncing the list of saved articles)
- Saved: Click to unsave
- Unsaved: Click to save

### DIFF
--- a/packages/edition-slices/src/tiles/shared/tile-star.js
+++ b/packages/edition-slices/src/tiles/shared/tile-star.js
@@ -1,0 +1,31 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Button from "@times-components/button";
+import { SectionContext } from "@times-components/context";
+
+const renderStar = (onSaveStarPress, savedArticles, articleId) => {
+  const isInactive = !savedArticles;
+  const isSaved = savedArticles && savedArticles[articleId];
+  const getTitle = () => {
+    if (isInactive) {
+      return "Inactive";
+    } if (isSaved) {
+      return "Unsave";
+    }
+    return "Save";
+  };
+
+  return (<Button disabled={isInactive} onPress={() => onSaveStarPress(!isSaved, articleId)} style={{ width: 100 }} title={getTitle()} />);
+}
+
+const TileStar = ({ articleId }) => (
+  <SectionContext.Consumer>
+    {({ onSaveStarPress, savedArticles }) => onSaveStarPress ? renderStar(onSaveStarPress, savedArticles, articleId) : null}
+  </SectionContext.Consumer>
+);
+
+TileStar.propTypes = {
+  articleId: PropTypes.string.isRequired
+}
+
+export default TileStar;
--- a/packages/edition-slices/src/tiles/shared/tile-summary.js
+++ b/packages/edition-slices/src/tiles/shared/tile-summary.js
@@ -1,4 +1,5 @@
 import React from "react";
+import { View } from "react-native";
 import PropTypes from "prop-types";
 import ArticleSummary, {
   ArticleSummaryContent,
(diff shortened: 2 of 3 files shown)


packages/link/src/link.android.js
```diff
@@ -8,7 +8,7 @@ const Link = ({ children, linkStyle, onPress }) => (
     onPress={onPress}
     useForeground={TouchableNativeFeedback.canUseNativeForeground()}
   >
-    <View pointerEvents="box-only" style={linkStyle}>
+    <View style={linkStyle}>
```
tuncaulubilge: Removed "box-only" pointerevents to enable nested clicks. We need to test the implications of this change on other components.

packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -31,40 +34,43 @@ const TileSummary = ({
...
-              strapline={strapline}
-              style={straplineStyle}
-            />
-          )
-        : undefined
-    }
-    style={style}
-  />
-);
+      <TileStar articleId={id} />
```
tuncaulubilge: Added this component inside TileSummary just for POC purposes. This should be added manually to each tile

nehasri89:
Great stuff! 👍

packages/edition-slices/src/tiles/shared/tile-star.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Button from "@times-components/button";
+import { SectionContext } from "@times-components/context";
+
+const renderStar = (onSaveStarPress, savedArticles, articleId) => {
```
ValGeorgiev: For POC this is totally fine. But we should not implement such components in production react app.  1. In renderSomething method/function we should not create other function, this is impacting the performance
2. We should use a state property and not depending on context update (this is the React way). onPress event can invoke `setState` function and then calls onSaveStarPress. Thats how we can use react comparison and improve the performance. With the current implementation, if you change the "state" of some star, react will rerender all stars, because it cannot know which star you changed exactly. If we have 30 articles on page this is 30 more renderStar invocations. 
3. You don't need to pass whole savedArticles object, you can pass only savedArticles[articleId] value in renderStar function

tuncaulubilge: You are telling me "context" is not the "React way" ?

tuncaulubilge: renderStar is just a function to make the code more readable, it isn't a whole different component. We are not passing savedArticles as a prop around, it's just part of the context's data. As long as the button itself is concerned, props only change for the clicked button and remain the same for all the others

ValGeorgiev: No, I am saying that React context is for accessing data from every component. But updating react context without state in the current situation is not the optimal way. Its working, totally agree, but this doesn't mean its the most performant way. When you want to update some component you should create an instance from this component and this instances should work with the state. Thats how react understand which instance of the component to update. This is one the most complex usage of context in React. Some more info from the React documentation about dynamic context: https://reactjs.org/docs/context.html#dynamic-context

tuncaulubilge: The reason we used context is to sync saved article data with the native apps. Once you add a state inside that component, you create a second source of truth, which will surely get out of sync.

tuncaulubilge:
Closing this POC as the actual PR is being merged soon

PR #1830, opened 2019-03-28
feat: section perf improvements

tuncaulubilge:
Round 1 of section perf improvements (more coming soon, in a separate PR):

- Changed windowSize and initialNumToRender on section FlatList. Defaults are 20 items, which is huge for our use case.
- Got rid of inline function declarations in edition-slices and in tile-summary. However, we still got a million more to inline function declarations in slice-layout, but that'll be the next PR

### DIFF
--- a/packages/edition-slices/src/slices/leaders/index.js
+++ b/packages/edition-slices/src/slices/leaders/index.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { Component } from "react";
 import { View, Text } from "react-native";
 import { SectionContext } from "@times-components/context";
 import { Leaders } from "@times-components/slice-layout";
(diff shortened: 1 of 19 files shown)


packages/edition-slices/src/slices/leaders/index.js
```diff
@@ -8,62 +8,85 @@ import styles from "./styles";
...
+          <TileAG onPress={onPress} tile={leader3} tileName="leader3" />
+        )}
+      />
+    );
+  }
+
+  render() {
+    return (
+      <View style={styles.container}>
+        {renderHead()}
```
ValGeorgiev: What about adding renderHead inside of class ?

tuncaulubilge: Did that bit in my part 2 PR, coming soon :)

PR #1831, opened 2019-03-28
feat: save star button package

Ivaylo1987:
This is for [Tile Star Component / Button](https://nidigitalsolutions.jira.com/browse/REPLAT-5846)

Added:
New package for `star-button` - reusable star button which exposes 3 states. Can be used for various purposes.

`tile-star` in edition-slices which encapsulates logic for the saving of article in a tile.

`Initial`:
Screenshot 2019-03-28 at 16 32 47

`selected`:
Screenshot 2019-03-28 at 16 32 56

`disabled`:
Screenshot 2019-03-28 at 16 33 04


packages/edition-slices/src/tiles/shared/tile-star.js
```diff
@@ -0,0 +1,73 @@
...
+        onArticleSavePress={onArticleSavePress}
+        savedArticles={savedArticles}
+      />
+    )}
+  </SectionContext.Consumer>
+);
+
+TileStar.propTypes = {
+  articleId: PropTypes.string.isRequired,
+  onArticleSavePress: PropTypes.func,
```
ValGeorgiev: this should be required, because we always need it

packages/edition-slices/src/tiles/shared/tile-star.js
```diff
@@ -0,0 +1,73 @@
...
+  constructor(props) {
+    super(props);
+
+    const { articleId, savedArticles } = props;
+    this.setStarState(savedArticles, articleId);
+  }
+
+  setStarState(savedArticles, articleId) {
+    const starState = this.getStarState(savedArticles, articleId);
+    this.state = { starState };
```
ValGeorgiev: Please use setState, this.state = {} must be used only in the constructor

packages/edition-slices/src/tiles/shared/tile-star.js
```diff
@@ -0,0 +1,73 @@
...
+import PropTypes from "prop-types";
+import StarButton, { starStates } from "@times-components/star-button";
+import { SectionContext } from "@times-components/context";
+
+class TileStar extends Component {
+  constructor(props) {
+    super(props);
+
+    const { articleId, savedArticles } = props;
+    this.setStarState(savedArticles, articleId);
```
ValGeorgiev: here you can do something like:

`this.state = . { starState: this.getStarState(atr....)}`

packages/star-button/README.md
```diff
@@ -0,0 +1,45 @@
+# StarButton
+
+<!-- Add the description of the package here. -->
```
ValGeorgiev: Could you add more info here, please

packages/star-button/src/star-button.js
```diff
@@ -0,0 +1,68 @@
...
+  disabled: "disabled",
+  initial: "initial",
+  selected: "selected"
+};
+
+class StarButton extends Component {
+  constructor(props) {
+    super(props);
+
+    const stars = {
```
ValGeorgiev: this can be in the style file, not in the constructor

packages/edition-slices/src/tiles/shared/tile-star.js
Ivaylo1987: 👍

packages/star-button/src/star-button.js
Ivaylo1987: These are not actual styles. Can be moved but not sure if the right place is the style file.

packages/edition-slices/src/tiles/shared/tile-star.js
Ivaylo1987: I think initially not because of the implementation on IOS.

packages/star-button/README.md
Ivaylo1987: 👍

packages/edition-slices/src/tiles/shared/tile-star.js
Ivaylo1987: 👍

packages/edition-slices/src/tiles/shared/tile-star.js
```diff
@@ -0,0 +1,73 @@
...
+    const { starState } = this.state;
+
+    return onArticleSavePress ? (
+      <StarButton onPress={this.onStarPress} starState={starState} />
+    ) : null;
+  }
+}
+
+const TileStarProvider = ({ articleId }) => (
+  <SectionContext.Consumer>
```
tuncaulubilge: This can be part of TileStar, I don't think we need a TileStarProvider

packages/star-button/__tests__/android/__snapshots__/star-button.test.js.snap
```diff
@@ -0,0 +1,329 @@
...
+  onResponderTerminate={[Function]}
+  onResponderTerminationRequest={[Function]}
+  onStartShouldSetResponder={[Function]}
+  style={
+    Object {
+      "opacity": 1,
+    }
+  }
+>
+  <ARTSurfaceView
```
tuncaulubilge: We need to mock star svg here, we don't need to see the internals of the svg

packages/star-button/src/star-button.js
```diff
@@ -0,0 +1,68 @@
...
+    this.state = { star: stars[props.starState] };
+  }
+
+  render() {
+    const { height, onPress, starState } = this.props;
+    const { star: { fillColour, opacity, strokeColour } } = this.state;
+    const disabled = starState === starStates.disabled;
+
+    return (
+      <TouchableOpacity disabled={disabled} onPress={onPress}>
```
tuncaulubilge: We should use the Link package for buttons, which uses `TouchableNativeFeedback` for android, instead of `TouchableOpacity`

packages/star-button/src/star-button.js
```diff
@@ -0,0 +1,68 @@
...
+        />
+      </TouchableOpacity>
+    );
+  }
+}
+
+StarButton.propTypes = {
+  height: PropTypes.number,
+  onPress: PropTypes.func.isRequired,
+  starState: PropTypes.oneOf(Object.keys(starStates))
```
tuncaulubilge: Can we use bool props instead of a enum state? I'd suggest `disabled` and `selected`

yarn.lock
```diff
@@ -2444,6 +2444,13 @@ acorn-jsx@4.1.1:
   dependencies:
     acorn "^5.0.3"
 
+acorn-jsx@^3.0.0:
```
tuncaulubilge: What's going on with your yarn.lock?

packages/star-button/src/star-button.js
Ivaylo1987: We can but this will be confusing when you look at the component:
`<StarButton onPress={this.onStarPress} disabled={disabled}  selected={selected}/>`
You can set them both to `true` and which one has precedence?
Personally think that this is a good example to use `PropTypes.oneOf([...])`

tuncaulubilge: `<StarButton onPress={this.onStarPress} disabled={disabled} selected={selected}/>` this is super clear

packages/edition-slices/src/tiles/shared/tile-star.js
Ivaylo1987: In this case we can not go with only one component,  we need two - one to consume the context and one to implement the logic for rendering the star button.
So If we use 2 functional components as in the POC we cannot take advantage of `state` which we need to maintain the state of the star button. Besides this we have to create a new function every time the star button renders e.g. -> `onPress={() => onArticleSavePress(!isSaved, articleId)}` - not good for performance.
One functional and one class should be the better, more 'react-like' approach here.
_Note:_ in react 16.6 we can take advantage of [Class.contextType](https://reactjs.org/docs/context.html#classcontexttype) which should be the perfect solution.

tuncaulubilge: We don't want a state in there.

tuncaulubilge: You can keep two components in there to avoid inline function creation, but we don't need state, we use context as a state

packages/star-button/src/star-button.js
Ivaylo1987: so is it `disabled` or is it `selected`?

tuncaulubilge: What do you think?

packages/edition-slices/src/tiles/shared/tile-star.js
Ivaylo1987: Dont really understand what you mean by `We don't want a state in there`.
We really need it on the `component level` - it is the official recommended (by React docs) way of maintaining state between rerenders.
We cannot update the sate of the star based on the assumption that `onArticleSavePress` will trigger an update to `savedArticles` object which will then cause the whole start to rerender cause it is a dependency.

packages/star-button/__tests__/android/__snapshots__/star-button.test.js.snap
Ivaylo1987: 👍

packages/star-button/src/star-button.js
Ivaylo1987: 👍

yarn.lock
Ivaylo1987: 👍

packages/star-button/src/star-button.js
Ivaylo1987: I think the enum is ok. Maybe we need some more opinions.

tuncaulubilge: Enum is not okay. And it's quite clear that `disabled` takes precedence over `selected`. We disable the button while the app is fetching the data from the database, so it can't be really selected and disabled at the same time

Ivaylo1987: 👍

packages/edition-slices/src/tiles/shared/tile-star.js
Ivaylo1987: Updated. Needs  review now :)

tuncaulubilge:
Looking good :+1: Can you put a gif for the click animation as well? (giphy's desktop app on mac is quite useful)

packages/star-button/README.md
```diff
@@ -0,0 +1,57 @@
...
+<StarButton onPress={this.onStarPress} disabled={true} />;
+<StarButton onPress={this.onStarPress} selected={true}/>;
+```
+
+It changes colours and opacity based on the different state it has.
+`default` star is clickable and has default colour.
+`disabled` would not allow the star to be clicked.
+`selected` would change the star colour.
+
+If `disabled` and `selected` are both true - `disabled` take precedence.
```
aronshamash: minor type `takes` precedence

packages/star-button/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+// works this way
+<StarButton onPress={this.onStarPress} />;
+<StarButton onPress={this.onStarPress} disabled={true} />;
+<StarButton onPress={this.onStarPress} selected={true}/>;
+```
+
+It changes colours and opacity based on the different state it has.
+`default` star is clickable and has default colour.
+`disabled` would not allow the star to be clicked.
```
aronshamash: `would` -> `does`

packages/star-button/README.md
Ivaylo1987: 👍

packages/star-button/README.md
Ivaylo1987: 👍 1

tuncaulubilge:
@Ivaylo1987 Let's mock StarButton in edition-slices

packages/star-button/star-button.showcase.js
```diff
@@ -0,0 +1,23 @@
...
+      name: "Selected",
+      type: "story"
+    },
+    {
+      component: () => <StarButton disabled onPress={() => {}} />,
+      name: "Disabled",
+      type: "story"
+    }
+  ],
+  name: "Primitives/StarButton"
```
tuncaulubilge: should be "Primitives/Star Button"

tuncaulubilge:
Converted my "request changes" back to an approval. We need to fix the click animations on the star button but we can fix it on a separate pr later on. Let's try to get this merged first

packages/star-button/star-button.showcase.js
Ivaylo1987: 👍

PR #1832, opened 2019-03-28
Feat/lead 2 no pic and 2 tablet

apostolnikov:
Implement LeadTwoNoPicAndTwo for wide and huge tablet breakpoints.
New tile (TileAL) added.

TileAL:
image

Wide: 
image

Huge:
image

### DIFF
--- a/packages/slice-layout/src/templates/leadtwonopicandtwo/index.js
+++ b/packages/slice-layout/src/templates/leadtwonopicandtwo/index.js
@@ -5,7 +5,7 @@ import { editionBreakpoints } from "@times-components/styleguide";
 import Column from "../column";
 import { ItemColSeparator } from "../shared";
 import { propTypes, defaultProps } from "./proptypes";
-import styles from "./styles";
+import styleFactory from "./styles";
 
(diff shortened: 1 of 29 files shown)


packages/slice-layout/src/templates/leadtwonopicandtwo/index.js
```diff
@@ -14,6 +14,30 @@ const LeadTwoNoPicAndTwoSlice = ({
...
+  }
+
+  if (breakpoint === editionBreakpoints.wide) {
+    return (
+      <View style={styles.container}>
+        <Column style={{ width: "42%" }} tiles={[renderLead1, renderLead2]} />
+        <ItemColSeparator />
+        <View style={{ width: "16%" }}>{renderSupport1()}</View>
+        <ItemColSeparator />
+        <View style={{ width: "42%" }}>{renderSupport2()}</View>
```
ValGeorgiev: don't  use inline styling, move this into styles.js

apostolnikov: Got it, changes applied.

PR #1835, opened 2019-03-29
feat: gather data for image gallery

aronshamash:
- Gather image data to pass to the native layer in order to render a native image gallery

### Requirements

- should include video poster image when lead asset is a video
- should include video poster images when article has inline video
- should include image url, caption and credit and the index of the pressed image

### Notes

- inline images are a different format to lead assets which makes things less than ideal
- lead assets (video or image) have an id but inline images do not hence we are mutating the AST (`content`)


packages/article-main-comment/src/article-prop-types/article-prop-types.base.js
```diff
@@ -12,13 +12,15 @@ const articlePagePropTypes = {
...
+  onImagePress: PropTypes.func,
   receiveChildList: PropTypes.func.isRequired
 };
 
 const articlePageDefaultProps = {
   article: null,
   error: null,
-  isLoading: false
+  isLoading: false,
+  onImagePress: () => null
```
tuncaulubilge: Why isn't this just `null`?

packages/image/__tests__/modal-shared.native.js
```diff
@@ -114,6 +114,24 @@ export default () => {
...
+      name: "image with onImagePress prop should not have Modal",
+      test: () => {
+        const onImagePress = () => {};
+        const propWithImagePress = { ...props, onImagePress };
+        const testInstance = TestRenderer.create(
+          <ModalImage {...propWithImagePress} />
+        );
+
+        try {
+          testInstance.root.findByType(Modal);
```
tuncaulubilge: You can use `findAllByType` and check if the array is empty

packages/article-main-comment/src/article-prop-types/article-prop-types.base.js
aronshamash: is consistent with https://github.com/newsuk/times-components/blob/master/packages/article-byline/src/article-byline-prop-types.js#L10

packages/image/__tests__/modal-shared.native.js
aronshamash: I _think_ Neha tried that and it failed. @nehasri89  ?

packages/article-main-comment/src/article-prop-types/article-prop-types.base.js
tuncaulubilge: But for this one, we want it to be null if it's not supported by that platform

aronshamash: good point. I'll update

packages/image/__tests__/modal-shared.native.js
nehasri89: @aronshamash don't remember now, worth a try

tuncaulubilge: We got that working in couple of other components, should be okay

aronshamash: yes it works fine

packages/article-image/src/article-image-prop-types.js
```diff
@@ -11,12 +11,14 @@ export const propTypes = {
...
+  onImagePress: PropTypes.func
 };
 
 export const defaultPropTypes = {
   captionOptions: {
     caption: null,
     credits: null
-  }
+  },
+  onImagePress: () => null
```
tuncaulubilge: This should also be null (sorry I just mentioned it in a single place, but we have the same situation on couple of components)

packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -220,9 +223,14 @@ ArticleRow.propTypes = {
...
   }).isRequired,
   interactiveConfig: PropTypes.shape({}).isRequired,
+  onImagePress: PropTypes.func,
   onLinkPress: PropTypes.func.isRequired,
   onTwitterLinkPress: PropTypes.func.isRequired,
   onVideoPress: PropTypes.func.isRequired
 };
 
+ArticleRow.defaultProps = {
+  onImagePress: () => null
```
tuncaulubilge: Should be `null`

packages/article-image/src/article-image.base.js
```diff
@@ -78,7 +79,9 @@ const ArticleImage = ({ imageOptions, captionOptions }) => {
           aspectRatio={aspectRatio}
           caption={<Caption credits={credits} text={caption} />}
           highResSize={highResSize}
+          index={index}
           lowResSize={lowResSize}
+          onImagePress={onImagePress}
```
nathggns: Not a point for here, but I wonder whether in the future if we're just passing through a ton of props we should just use `{ ...props }` after having destructured out the props we are using at this level. So like this

```js
function MyComponent({ propA, propB, ...props }) {
    // Use propA & propB here somehow
    return (
        <ChildComponent {...props} />
    );
}
```

nathggns: Would save us from having to write these kinds of updates.

packages/article/src/utils/index.js
```diff
@@ -0,0 +1,64 @@
...
+  });
+
+  return mediaList;
+};
+
+const addIdsToInlineImages = (content, leadAsset) => {
+  const mutatedContent = content || [];
+  let index = leadAsset ? 1 : 0;
+
+  mutatedContent.forEach(item => {
```
nathggns: Is the intention that this doesn't mutate the objects passed into the function, because it does.

packages/article-lead-asset/src/article-lead-asset.js
```diff
@@ -9,15 +9,19 @@ import {
...
 
-const ArticleLeadAssetModalImage = ({ aspectRatio, caption, uri }) => (
-  <ModalImage {...{ aspectRatio, caption, uri }} />
-);
+const ArticleLeadAssetModalImage = ({
+  aspectRatio,
+  caption,
+  onImagePress,
+  uri
+}) => <ModalImage {...{ aspectRatio, caption, onImagePress, uri }} index={0} />;
```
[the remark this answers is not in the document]
aronshamash: no because you can't launch the image gallery from pressing on a video as pressing plays the video.

packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -52,9 +53,11 @@ const ArticleRow = ({
               }}
               imageOptions={{
                 display,
+                index: imageIndex,
```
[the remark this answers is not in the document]
aronshamash: ok. you mean having:

```
              imageOptions={{
                display,
                imageIndex,
                ratio,
                uri: url
              }}
```
?

packages/article/__tests__/add-ids-to-inline-images.test.js
```diff
@@ -0,0 +1,90 @@
...
+  },
+  crop1251: null,
+  id: "2e7670f9-e035-48ba-aa69-f4bbbae7df30",
+  title: "Geraint Thomas",
+  type: "Image"
+};
+
+describe("addIdsToInlineImages", () => {
+  it("images and video in content, no leadAsset", () => {
+    const output = addIdsToInlineImages(content, undefined);
```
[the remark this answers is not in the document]
aronshamash: true will update

packages/article/__tests__/add-ids-to-inline-images.test.js
```diff
@@ -0,0 +1,90 @@
+import { addIdsToInlineImages } from "../src/utils";
```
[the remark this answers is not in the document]
aronshamash: will cut down on unnecessary fixture data

packages/article/src/utils/index.js
aronshamash: I intention is to mutate the object passed in.

packages/article/src/article.js
```diff
@@ -14,11 +15,20 @@ export const templates = {
 };
 
 const Article = props => {
-  const { article } = props;
-  const { template = "mainstandard" } = article || {};
+  const { article, onImagePress } = props;
+  const { leadAsset, template = "mainstandard" } = article || {};
+  let { content } = article || {};
```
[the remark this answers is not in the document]
aronshamash: `content` needs to be mutated (to add the indexes)

packages/article-image/src/article-image-prop-types.js
aronshamash: resolved

packages/article-lead-asset/src/article-lead-asset.js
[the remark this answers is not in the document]
nathggns: gotcha.

packages/article-skeleton/src/article-body/article-body-row.js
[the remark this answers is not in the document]
nathggns: Yeah

packages/article/src/article.js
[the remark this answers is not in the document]
nathggns: Using const doesn't freeze the object though. It just prevents reassigning, which you don't need to do. Doesn't matter either way.

packages/article/src/utils/index.js
nathggns: ok.

packages/article/__tests__/add-ids-to-inline-images.test.js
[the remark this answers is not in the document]
nathggns: ok

packages/article-skeleton/src/article-body/article-body-row.js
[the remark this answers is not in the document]
aronshamash: reason I didn't do that was that it is already in the context of  `imageOptions`

PR #1836, opened 2019-03-29
fix: Handle null aspect ratio case in images

tuncaulubilge:
Native apps are crashing due to empty aspect ratio on certain images.

- Handled empty aspect ratio on images
- Added aspect ratio to video cover images


nathggns:
Great catch. My bad aha.

PR #1838, opened 2019-03-29
fix: Android image modals

tuncaulubilge:
`SaferAreaView` implementation is not working android, since `SafeAreaView` is no-op on android. Replaced `SaferAreaView` with `SafeAreaView` in android for future-proofing any android support that might come in the future


nathggns:
Good catch!

PR #1839, opened 2019-04-01
Feat/secondary 2 and 2 wide

pdimova:
Added styles for wide and huge breakpoints:

wide:
Screenshot 2019-04-01 at 10 55 49

huge:
Screenshot 2019-04-01 at 10 54 59


packages/edition-slices/src/tiles/tile-am/index.js
```diff
@@ -0,0 +1,35 @@
...
+} from "../shared";
+import styles from "./styles";
+
+const TileAM = ({ onPress, tile }) => (
+  <TileLink onPress={onPress} tile={tile}>
+    <View style={styles.container}>
+      <TileImage
+        aspectRatio={16 / 9}
+        style={styles.imageContainer}
+        uri={getCrop(tile.leadAsset || tile.article.leadAsset, "crop169")}
```
ValGeorgiev: Please check other tiles for this

packages/edition-slices/src/tiles/tile-an/index.js
```diff
@@ -0,0 +1,36 @@
...
+} from "../shared";
+import styles from "./styles";
+
+const TileAN = ({ onPress, tile }) => (
+  <TileLink onPress={onPress} tile={tile}>
+    <View style={styles.container}>
+      <TileImage
+        aspectRatio={1}
+        style={styles.imageContainer}
+        uri={getCrop(tile.leadAsset || tile.article.leadAsset, "crop11")}
```
ValGeorgiev: same here, as above

packages/edition-slices/src/tiles/tile-am/index.js
pdimova: tile.article.listingAsset added to expression

packages/edition-slices/src/tiles/tile-an/index.js
pdimova: here also tile.article.listingAsset is added

nehasri89:
@pdimova have you tested all the slices which were using `P` and `G` tiles just to make sure we haven't regressed?

pdimova:
All tiles with roundels were checked (tileAN, tileG, tileP and tileAH) and it happened that tileAH (used in Comment Lead and Cartoon for tablet) was not updated.

tuncaulubilge:
Needs a snapshot update, then we can merge this in

PR #1840, opened 2019-04-01
Feat/edition slices strapline override

apostolnikov:
Strapline override added.

https://nidigitalsolutions.jira.com/secure/RapidBoard.jspa?rapidView=978&projectKey=REPLAT&modal=detail&selectedIssue=REPLAT-5877


aronshamash:
@apostolnikov Can we have a snapshot test for this logic?

nehasri89:
@apostolnikov needs to add this field in `article` queries in this package `https://github.com/newsuk/times-components/tree/master/packages/provider-queries` and update corresponding test

tuncaulubilge:
You need to update `section-fragment` query in provider-queries. We already have strapline field in `tile` segment of that query, you'll need to add the same thing into the `article` for those tiles

nehasri89:
needs a test here to test the override `https://github.com/newsuk/times-components/blob/master/packages/edition-slices/__tests__/shared-components.base.js`

PR #1842, opened 2019-04-01
feat: remove unnecessary arrow functions and use objects indstead

ValGeorgiev:
Remove unnecessary arrow functions and use objects instead

Start introducing VerticalLayout and Horizontal Layout


packages/edition-slices/src/slices/commentleadandcartoon/index.js
```diff
@@ -5,8 +5,8 @@ import { TileP, TileQ, TileAH, TileAI } from "../../tiles";
 import { ResponsiveSlice } from "../shared";
 
 class CommentLeadAndCartoonSlice extends Component {
-  constructor(props) {
-    super(props);
+  constructor() {
+    super();
```
nathggns: React docs recommend always passing props into super. 

image

https://reactjs.org/docs/state-and-lifecycle.html#adding-local-state-to-a-class

packages/edition-slices/src/slices/commentleadandcartoon/index.js
```diff
@@ -20,12 +20,10 @@ class CommentLeadAndCartoonSlice extends Component {
     return (
       <CommentLeadAndCartoon
         breakpoint={editionBreakpoint}
-        renderCartoon={() => (
+        renderCartoon={
```
nathggns: This is a good move! 👍

packages/edition-slices/src/slices/leaders/index.js
```diff
@@ -59,23 +59,23 @@ class LeadersSlice extends Component {
...
+        }
       />
     );
   }
 
   render() {
     return (
       <View style={styles.container}>
-        {renderHead()}
+        {renderHead}
```
nathggns: I'm not sure what this does.

nathggns: renderHead still seems to be a function

packages/slice-layout/src/templates/leadoneandfour/proptypes.js
```diff
@@ -3,11 +3,11 @@ import { editionBreakpoints } from "@times-components/styleguide";
 
 const propTypes = {
   breakpoint: PropTypes.string,
-  renderLead: PropTypes.func.isRequired,
-  renderSupport1: PropTypes.func.isRequired,
-  renderSupport2: PropTypes.func.isRequired,
-  renderSupport3: PropTypes.func.isRequired,
-  renderSupport4: PropTypes.func.isRequired
+  renderLead: PropTypes.object.isRequired,
```
nathggns: You can use `PropTypes.element` to specifically match a react component, or `PropTypes.node` to match any renderable content. I suggest using the former.

packages/edition-slices/src/slices/commentleadandcartoon/index.js
ValGeorgiev: Props should be passed to the constructor and super only if its used in the constructor, otherwise there is no point of doing that

packages/edition-slices/src/slices/leaders/index.js
ValGeorgiev: Its an object, i agree naming is not the best, will think about it, thanks

packages/edition-slices/src/slices/commentleadandcartoon/index.js
nathggns: It should always be passed because its easy to miss bugs if you start using props in the constructor (or in methods used by the constructor), and there's no downside to passing them. I don't see what you gain by removing props from the super call.

packages/edition-slices/src/slices/leaders/index.js
nathggns: In this file, `renderHead` still seems to be a function. https://github.com/newsuk/times-components/blob/ca987ecf73878e44b84e742992d1ed44fdcb1ae6/packages/edition-slices/src/slices/leaders/index.js#L11

ValGeorgiev: will refactor, good catch

packages/edition-slices/src/slices/commentleadandcartoon/index.js
ValGeorgiev: Agree with you, I like clean code, no unnecessary things and etc. I see benefit when remove constructor at all from our components, because its useless.. In the new versions of react/babel its not needed. But I see we are using it a lot and we are stuck with it :) so I will revert the props. Thanks

nathggns:
I hate to be that guy but if the passed props are no longer render functions, then the `renderX` prefix doesn't make sense anymore and isn't necessary.

packages/edition-slices/src/slices/standard/index.js
```diff
@@ -20,19 +20,19 @@ class Standard extends Component {
       onPress
     } = this.props;
 
-    return items.map(tile => () => <TileK onPress={onPress} tile={tile} />);
+    return items.map(tile => <TileK onPress={onPress} tile={tile} />);
```
nathggns: Do these need keys?

packages/slice-layout/slice-layout.showcase.js
```diff
@@ -93,16 +98,17 @@ export default {
         <ScrollView>
           <LeadOneAndOneSlice
             breakpoint={breakpointSelect(select)}
-            renderLead={() => (
+            renderLead={
```
nathggns: Should be `lead` and `support`, not `renderLead` and `renderSupport`.

packages/slice-layout/slice-layout.showcase.js
```diff
@@ -240,10 +246,10 @@ export default {
         <ScrollView>
           <SecondaryFourSlice
             breakpoint={breakpointSelect(select)}
-            renderSecondary1={() => <Support1 />}
-            renderSecondary2={() => <Support2 />}
-            renderSecondary3={() => <Support3 />}
-            renderSecondary4={() => <Support4 />}
+            renderSecondary1={<Support1 tileName="secondary1" />}
```
nathggns: Remove render prefix.

packages/slice-layout/slice-layout.showcase.js
```diff
@@ -364,8 +370,8 @@ export default {
         <ScrollView>
           <SecondaryOneAndColumnistSlice
             breakpoint={breakpointSelect(select)}
-            renderColumnist={() => <Support1 />}
-            renderSecondary={() => <Support2 />}
+            clumnist={<Support1 tileName="support1" />}
```
nathggns: Typo.

PR #1845, opened 2019-04-01
feat: Replace gradient and svg with solid background and image in native

tuncaulubilge:
Removed gradient and svg from the image placeholders as it's quite costly on native platforms, especially on Android. Web still uses the same gradient+svg for now.

Before:
Screenshot_1554134224

After:
Screenshot_1554133972


aronshamash:
@tuncaulubilge I thought web could also go to solid background or is it not worth doing now ?

nathggns:
Good call!

PR #1848, opened 2019-04-02
fix: Remove gradient from web background

tuncaulubilge:
Updated the placeholder on web to keep it consistent with the native apps. Replaced gradient with a solid coloured background. Tidied up the styles a bit in the process

Screenshot 2019-04-02 at 12 44 56


packages/image/src/placeholder.web.js
```diff
@@ -1,39 +1,32 @@
 import React from "react";
```
nathggns: Discussed in person about putting the web override in `t.js` instead of here to reduce duplication.

PR #1852, opened 2019-04-03
feat: Add magazine section

ValGeorgiev:
Added magazine section with magazine cover image.

Mobile Portrait:
image

Tablet Portrait:
image

Tablet modal view:
image


tuncaulubilge:
@ValGeorgiev can you add magazine cover into the section-fragment query, and add a story to section storybook for a magazine section?

Also it'd be nice to use an actual magazine cover image taken from TPA on screenshots and storybooks

packages/section/src/magazine-cover.js
```diff
@@ -0,0 +1,77 @@
...
+    const { cover, breakpoint, publicationName } = this.props;
+
+    const ratio = getRatio(cover, breakpoint);
+    const uri = getImageUri(cover, breakpoint);
+    const imageHeight = (height - NavBarHeight) * 0.65;
+    return (
+      <View style={[styles.container]}>
+        <View style={styles.header}>
+          <Text style={styles.title}>
+            {publicationName === "TIMES" ? "T" : "ST"} Magazine
```
nehasri89: As discussed we need to check for this data if it feeds in from tpa

tuncaulubilge: Magazine title is a part of the original images fed by editorial, so we don't need a separate view in there for a title

tuncaulubilge:
We also need to add `StandardSlice` to section-fragment query, but you can do that in a separate PR

tuncaulubilge:
We also still need the cover to be added to the section-fragment query as well. You need to ask for all the crops and use the first one (there'll be only one anyway)
```
fragment magazineSection on MagazineSection {
  name: __typename
  cover {
    crops {
      ratio
      url
    }
  }
```

packages/provider-queries/src/section-fragment.js
```diff
@@ -39,6 +39,24 @@ export default gql`
...
+            }
+            listingAsset {
+              ...listingAsset32
+            }
+          }
+          headline
+          leadAsset {
+            ...leadAsset32
+          }
+          strapline
```
tuncaulubilge: Do we need a strapline?

ValGeorgiev: good catch, removed

packages/section/src/magazine-cover.js
```diff
@@ -0,0 +1,95 @@
...
+  render() {
+    const { height } = Dimensions.get("window");
+    const imageHeight = height * 0.65;
+    const { breakpoint } = this.props;
+
+    const styles = styleFactory(breakpoint);
+
+    return (
+      <View style={[styles.container]}>
+        <TouchableOpacity
```
tuncaulubilge: Can you use our Link package for clickables?

ValGeorgiev: sure

tuncaulubilge:
What's going on with the `yarn.lock`?

PR #1853, opened 2019-04-03
feat: new modal image & related changes

nathggns:
This PR implements a new layout for modal images with full screen images and overlaying caption and x button. 

On iOS, the caption has a gradient background that fades the bottom of the image to black. As Android doesn't yet support overlay gradients, we simply use a translucent black background for the caption.

Swipe down to close the modal and tapping to toggle the modal elements have also been implemented. 

android

click to hide
swipe down
ipad landscape image landscape screen
ipad landscape image portrait screen
ipad portrait image landscape screen
ipad portrait image portrait screen
iphone se  portrait image landscape screen
iphone se landscape iamge landscape screen
iphone se landscape image portrait screen
iphone se portrait image portrait screen


tuncaulubilge:
Did you test these on android as well?

nathggns:
@tuncaulubilge Yup. See the first screenshot. android

tuncaulubilge:
Oh I didn't see that one, alright :+1: Can you try to run this branch by Matt before merging it in?

packages/image/src/styles/index.android.js
```diff
@@ -3,8 +3,9 @@ import sharedStyles, { captionStyles, tabletCaptionStyles } from "./shared";
 
 const styles = StyleSheet.create({
   ...sharedStyles,
-  buttonContainer: {
-    alignItems: "flex-end"
+  bottomSafeView: {
+    ...sharedStyles.bottomSafeView,
+    backgroundColor: "rgba(0, 0, 0, 0.7)"
```
tuncaulubilge: This should be in the styleguide

nathggns: Will leave this here pending discussion with design.

packages/gestures/src/gestures.js
```diff
@@ -83,21 +86,36 @@ class Gestures extends Component {
...
 
     this.onViewLayout = this.onViewLayout.bind(this);
   }
 
+  onGestureMove(evt, gestureState) {
+    if (gestureState.numberActiveTouches > 1) {
+      this.handlePinchChange(evt);
+    } else if (
+      gestureState.numberActiveTouches === 1 &&
+      gestureState.dy >= 50
```
aronshamash: `50` should be a named constant

packages/image/src/styles/shared.js
```diff
@@ -1,35 +1,41 @@
 import { colours, spacing } from "@times-components/styleguide";
 
 const styles = {
+  bottomSafeView: {
+    flex: 0,
+    marginTop: "auto",
+    position: "relative",
+    width: "100%",
+    zIndex: 2
```
aronshamash: would be great to have some zIndex management but for another PR...

packages/image/src/styles/shared.js
```diff
@@ -1,35 +1,41 @@
...
-    marginHorizontal: spacing(4),
-    marginVertical: spacing(4)
+    padding: spacing(4)
   },
   closeButton: {
     alignItems: "center",
     backgroundColor: "rgba(0, 0, 0, 0.5)",
-    borderRadius: 30,
-    height: 30,
+    borderRadius: 25,
```
aronshamash: could add `25` as a const?

packages/image/src/styles/shared.js
```diff
@@ -60,27 +80,23 @@ const styles = {
...
     justifyContent: "center"
   },
-  safeViewContainer: {
-    flex: 1
-  },
-  safeViewInnerContainer: {
-    flexGrow: 0,
-    height: "100%"
+  topSafeView: {
+    backgroundColor: "#000000",
```
aronshamash: we have in styleguide

nathggns: This has been raised by @tuncaulubilge and we're in discussion with design about making this gradient consistent with the sky video banner gradient. I'll put those colours in the styleguide once its consistent and will wait on merging this PR until then.

nathggns: I don't believe we do. We used to use the brand colour for modal backgrounds but that's a grey colour not black. Can put it into the styleguide if necessary?

packages/image/src/styles/shared.js
nathggns: 👍

packages/image/src/styles/shared.js
nathggns: Yeah zIndex always sucks.

packages/gestures/src/gestures.js
nathggns: 👍

packages/gestures/src/gestures.js
```diff
@@ -65,12 +70,10 @@ class Gestures extends Component {
...
 
     this.panResponder = PanResponder.create({
-      onMoveShouldSetPanResponder: (e, { numberActiveTouches }) =>
-        numberActiveTouches > 1,
-      onMoveShouldSetPanResponderCapture: (e, { numberActiveTouches }) =>
-        numberActiveTouches > 1,
-      onPanResponderMove: evt => {
-        this.handlePinchChange(evt);
+      onMoveShouldSetPanResponder: () => true,
+      // onMoveShouldSetPanResponderCapture: () => true,
```
ValGeorgiev: Could we delete this comment if its not needed

packages/gestures/src/gestures.js
```diff
@@ -83,21 +86,36 @@ class Gestures extends Component {
...
       },
       // Required for iOS support
       // https://github.com/facebook/react-native/issues/14295#issuecomment-389971835
       onPanResponderTerminationRequest: () => false,
-      onStartShouldSetPanResponder: (e, { numberActiveTouches }) =>
-        numberActiveTouches > 1,
-      onStartShouldSetPanResponderCapture: (e, { numberActiveTouches }) =>
-        numberActiveTouches > 1
+      onStartShouldSetPanResponder: () => true
+      // onStartShouldSetPanResponderCapture: () => true
```
ValGeorgiev: Same here ^

packages/image/src/image.web.js
```diff
@@ -4,7 +4,7 @@ import { addMissingProtocol } from "@times-components/utils";
 import Placeholder from "./placeholder";
 import { defaultProps, propTypes } from "./image-prop-types";
 import appendSize from "./utils";
-import styles from "./styles";
+import styles from "./styles/index.web";
```
ValGeorgiev: hmm this should be not needed ?

packages/image/src/styles/shared.js
aronshamash: we could have a z-index file in styleguide and just use those pre-defined values ?

packages/image/src/styles/shared.js
aronshamash: put in styleguide if it _must_ be black otherwise better to reuse existing colours from the palette.

packages/gestures/src/gestures.js
nathggns: Good catch. Thanks Val.

packages/image/src/image.web.js
nathggns: It seems to be – because there is a version for all platforms plus an index file, it seems to load the index file in some circumstances instead of the web one. This fixed it.

packages/image/src/styles/shared.js
nathggns: Will put it in styleguide.

packages/image/src/styles/shared.js
nathggns: I don't think this is worth it – because they're not really semantic in that sense (like, I can't really think of names that make sense). Also zIndex's are relative to the current stacking context anyway so the effect of them is different depending on various scenarios.

packages/image/src/styles/index.android.js
nathggns: Updated.

PR #1854, opened 2019-04-03
feat: save star storybook

tuncaulubilge:
This PR will simplify save star development by adding stubs to storybook

- Added SectionEvents to android storybook app
- Fixed the storybook on pages.
- Changed the way we update saved article context. Now, we update save status in a synchronous manner, and undo it if native app rejects the promise. This removed the lag on status updates caused by the async nature of NativeModules

### DIFF
--- a/packages/pages/src/section/section.js
+++ b/packages/pages/src/section/section.js
@@ -26,6 +26,7 @@ class SectionPage extends Component {
       savedArticles: null
     };
     this.onAppStateChange = this.onAppStateChange.bind(this);
+    this.toggleArticleSaveStatus = this.toggleArticleSaveStatus.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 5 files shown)


packages/pages/src/section/section.js
```diff
@@ -69,6 +70,12 @@ class SectionPage extends Component {
     }
   }
 
+  toggleArticleSaveStatus(save, articleId) {
+    const { savedArticles } = this.state;
+    savedArticles[articleId] = save || undefined;
```
Ivaylo1987: why do we need `false` to be `undefined`?

tuncaulubilge: This is a list of saved articles, stored in an object for O(1) reads, so I only keep the saved articles in there and remove anything that's not saved

Ivaylo1987: 👍

PR #1855, opened 2019-04-03
fix: ripple effect is not full width any more

Ivaylo1987:
Small PR for fixing full with ripple effect of the star button on android:

save-star

### DIFF
--- a/packages/star-button/src/styles/index.js
+++ b/packages/star-button/src/styles/index.js
@@ -1,12 +1,13 @@
-import { colours } from "@times-components/styleguide";
+import { colours, spacing } from "@times-components/styleguide";
 
 const styles = {
   container: {
-    alignSelf: "flex-start"
+    alignSelf: "flex-start",
+    padding: spacing(1)
   }
 };
 
-const stars = {
+const lightStar = {
   disabled: {
     fillColour: "none",
     opacity: "0.4",
@@ -24,5 +25,13 @@ const stars = {
   }
 };
 
-export { stars };
+const darkStar = {
+  disabled: { ...lightStar.disabled, strokeColour: colours.functional.greyLabel },
+  initial: { ...lightStar.initial, strokeColour: colours.functional.greyLabel },
+  selected: { ...lightStar.selected, fillColour: colours.functional.articleFlagUpdated },
+}
+
+const getTheme = ({ isDark }) => isDark ? darkStar : lightStar;
+
+export { getTheme };
 export default styles;
--- a/packages/star-button/src/styles/index.js
+++ b/packages/star-button/src/styles/index.js
@@ -26,12 +26,21 @@ const lightStar = {
 };
 
 const darkStar = {
-  disabled: { ...lightStar.disabled, strokeColour: colours.functional.greyLabel },
-  initial: { ...lightStar.initial, strokeColour: colours.functional.greyLabel },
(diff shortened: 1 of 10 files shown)


tuncaulubilge:
We need to put a fixed width & height on the button and a 5pt padding around (also, please put a gif in the animation PRs)

Ivaylo1987:
Currently the start is  set to `height: 18` which is by design.

tuncaulubilge:
> Currently the start is set to `height: 18` which is by design.

The width was the problem when I checked, as it stretched full width. Can you set that to `18` as well (or whatever the design suggests)?

packages/star-button/src/styles/index.js
```diff
@@ -18,4 +25,13 @@ const stars = {
   }
 };
 
-export default stars;
+const darkStar = {
+  disabled: { ...lightStar.disabled, strokeColour: colours.functional.greyLabel },
+  initial: { ...lightStar.initial, strokeColour: colours.functional.greyLabel },
+  selected: { ...lightStar.selected, fillColour: colours.functional.articleFlagUpdated },
```
nehasri89: is the fillcolor for disabled same in light and dark theme?

packages/star-button/src/styles/index.js
```diff
@@ -1,6 +1,13 @@
-import { colours } from "@times-components/styleguide";
+import { colours, spacing } from "@times-components/styleguide";
 
-const stars = {
+const styles = {
+  container: {
```
nehasri89: some tiles have star in center, so we don't want this here, my PR is taking care of it.

packages/star-button/src/styles/index.js
Ivaylo1987: Yep, the difference is in the strokeColour.

packages/star-button/src/styles/index.js
Ivaylo1987: This is to make the star ripple effect not spread full width. Similar to `inline` element in web. If you skip this the star container is full width and ripple effect on android takes full width.

Ivaylo1987: I think this is necessary but if you can show me another solution I am happy to see it.

packages/star-button/src/styles/index.js
aronshamash: Think our colour naming is off when we use the same colour for multiple elements. Could we create a new item in the styleguide specifically for the star states ?

packages/star-button/src/styles/index.js
Ivaylo1987: I set it to fixed width and height cause using `alignSelf` overrides parents styles and this will not be ok.

Ivaylo1987:
Star is now 18pt with 5pt padding.

packages/star-button/src/styles/index.js
Ivaylo1987: 👍

nehasri89:
@Ivaylo1987 nice one! if you put a animation gif we will be able to merge this one

Ivaylo1987:
Yep, gif is there now.

packages/styleguide/src/colours/star.js
```diff
@@ -0,0 +1,12 @@
+const starColours = {
+  dark: {
```
aronshamash: 👍 better

PR #1857, opened 2019-04-04
Feat/secondary 2 no pic and 2 wide and huge breakpoints

pdimova:
https://nidigitalsolutions.jira.com/browse/REPLAT-5488

Added styles for wide and huge breakpoints:

wide:
Screenshot 2019-04-04 at 11 25 42

huge:
Screenshot 2019-04-04 at 11 24 22


packages/edition-slices/src/tiles/tile-ap/index.js
```diff
@@ -0,0 +1,38 @@
...
+} from "../shared";
+import styles from "./styles";
+
+const TileAP = ({ onPress, tile }) => (
+  <TileLink onPress={onPress} tile={tile}>
+    <View style={styles.container}>
+      <TileImage
+        aspectRatio={1}
+        style={styles.imageContainer}
+        uri={getCrop(
```
aronshamash: we should have this logic in a shared util file, but should be in a separate PR.

packages/edition-slices/edition-slices.showcase.js
```diff
@@ -118,12 +118,13 @@ const sliceStories = [
   {
     mock: mockSecondaryTwoAndTwoSlice(),
     name:
-      "Secondary Two And Two (Mobile: C,C,G,G, Tablet: V,V,G,G, Wide Tablet/Desktop: AM,AM,AL,AL)",
+      "Secondary Two And Two (Mobile: C,C,G,G, Tablet: V,V,G,G, Wide Tablet/Desktop: AM,AM,AN,AN)",
```
aronshamash: Is this PR changing the `Secondary Two And Two` slice as well ?

packages/slice-layout/src/templates/secondarytwonopicandtwo/styles.js
```diff
@@ -11,4 +11,42 @@ const mediumBreakpointStyles = {
...
+    width: "16.6%"
+  }
+};
+
+const hugeBreakpointStyles = {
+  container: {
+    alignSelf: "center",
+    flexDirection: "row",
+    marginHorizontal: spacing(2),
+    width: "86%"
```
tuncaulubilge: We got gutters in ResponsiveSlice in edition-slices package, which sets up a max width and center alignment for all the slices. You can remove this 86% and share the same style as the wide breakpoint

packages/edition-slices/src/tiles/tile-ap/index.js
pdimova: ok, will create a new PR for this

packages/edition-slices/edition-slices.showcase.js
pdimova: thought it would be better to fix this mistake here (should have been tile AN, not tile AL) instead of opening a new PR for just changing a single letter for the slice description

packages/slice-layout/src/templates/secondarytwonopicandtwo/styles.js
tuncaulubilge: Agreed to do this refactor on a separate PR

PR #1858, opened 2019-04-04
Feat/supplement lead and 4 stack wide and huge tablet

apostolnikov:
Implement leadOneAndFour font-size change for wide and huge tablet breakpoints.

Wide and Huge: 
image

Medium:
image


packages/edition-slices/src/tiles/tile-ad/index.js
```diff
@@ -8,9 +8,10 @@ import {
   TileSummary,
   withTileTracking
 } from "../shared";
-import styles from "./styles";
+import styleFactory from "./styles";
 
-const TileAD = ({ onPress, tile }) => {
+const TileAD = ({ onPress, tile, breakpoint }) => {
```
ValGeorgiev: Please don't pass the breakpoint to Tiles. If you have only font difference use styleFactory, instead of passing breakpoints

tuncaulubilge: fontFactory is just a function that takes the breakpoint and returns a number, nothing special

packages/edition-slices/src/tiles/tile-ad/styles/index.js
```diff
@@ -18,4 +18,19 @@ const styles = {
   }
 };
 
-export default styles;
+const stylesWide = {
+  headline: {
+    fontFamily: fonts.headline,
+    fontSize: 22
```
tuncaulubilge: Let's just use a headlineFontSizeResolver instead of a stylesResolver, as we want to discourage breakpoint specific styling, and limit this to just headline font differences

packages/edition-slices/src/tiles/tile-ac/index.js
```diff
@@ -32,8 +33,13 @@ const TileAC = ({ onPress, tile }) => {
 };
 
 TileAC.propTypes = {
+  breakpoint: PropTypes.string,
   onPress: PropTypes.func.isRequired,
   tile: PropTypes.shape({}).isRequired
 };
 
+TileAC.defaultProps = {
+  breakpoint: "small"
```
tuncaulubilge: Can you use `editionBreakpoints.small` from styleguide?

packages/edition-slices/src/tiles/tile-ac/styles/index.js
```diff
@@ -24,4 +24,22 @@ const styles = {
   }
 };
 
-export default styles;
+const stylesWide = {
+  headline: {
```
tuncaulubilge: This part is identical to the default style. Just use the default style and only change the fontSize depending on the font size please

packages/edition-slices/src/tiles/tile-ad/styles/index.js
```diff
@@ -18,4 +18,19 @@ const styles = {
...
+  }
+};
+
+const headlineFontSizeResolver = {
+  huge: stylesWide,
```
tuncaulubilge: Same here. We don't want two different styles, just need the font size to change

packages/edition-slices/src/tiles/tile-ad/index.js
```diff
@@ -37,8 +38,13 @@ const TileAD = ({ onPress, tile }) => {
 };
 
 TileAD.propTypes = {
+  breakpoint: PropTypes.string,
   onPress: PropTypes.func.isRequired,
   tile: PropTypes.shape({}).isRequired
 };
 
+TileAD.defaultProps = {
+  breakpoint: "small"
```
tuncaulubilge: `editionBreakpoints.small`

ValGeorgiev:
please rebase and merge this one

PR #1864, opened 2019-04-08
Feat/star in slices

nehasri89:
This PR adds the stars in all tiles. The star is in same level as flag, If there are more than one flag then it will wrap on the second line to give space to the star

PFA the screenshots

Screenshot 2019-04-08 at 16 33 23
Screenshot 2019-04-08 at 16 34 08
Screenshot 2019-04-08 at 16 35 03

For Secondary Four, the stars stick to bottom right

Screen Shot 2019-04-10 at 17 10 18


packages/edition-slices/src/tiles/shared/styles/index.js
```diff
@@ -0,0 +1,30 @@
...
+    marginTop: -5,
+    width: "25%"
+  }
+};
+
+const verticalStyles = {
+  container: {
+    flexDirection: "column"
+  },
+  flagStyle: {},
```
ValGeorgiev: do we need that ^

packages/edition-slices/src/tiles/tile-ac/styles/index.js
```diff
@@ -14,6 +15,9 @@ const styles = {
   imageContainer: {
     width: "100%"
   },
+  star: {
+    ...verticalStyles
+  },
```
ValGeorgiev: do we expect additional styling for every tile or we can do `star: verticalStyles`

packages/pages/src/section/section.js
```diff
@@ -27,6 +27,7 @@ class SectionPage extends Component {
     };
     this.onAppStateChange = this.onAppStateChange.bind(this);
     this.toggleArticleSaveStatus = this.toggleArticleSaveStatus.bind(this);
+    this.syncAppData = this.syncAppData.bind(this);
```
ValGeorgiev: why we need this binding?

tuncaulubilge: I slipped this one in, was missing from the previous PR for data sync. `syncAppData` is using props, so we needed to bind it.

packages/edition-slices/src/tiles/tile-ac/styles/index.js
nehasri89: we might have diff styles per tile, but for now of you if you prefer I can change it.

packages/pages/src/section/section.js
ValGeorgiev: ahh didn't saw ComponentDidMount, okay

packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -26,12 +29,30 @@ class TileSummary extends Component {
   renderFlags() {
     const {
       tile: {
-        article: { expirableFlags }
+        article: { expirableFlags, id }
       },
-      flagColour
+      flagColour,
+      isDarkStar,
+      isStar,
```
tuncaulubilge: Can we call this `withStar`?

yarn.lock
```diff
@@ -1861,6 +1861,51 @@
     webpack-dev-server "3.1.7"
     webpack-merge "^4.1.1"
 
+"@times-components/schema@0.4.2":
```
tuncaulubilge: These should go away if you fix the versions, do a `yarn deps:fix` and yarn again

PR #1865, opened 2019-04-08
Feat/modal slide

nathggns:
The modal will now slide on entry/exit. This makes it feel more fluid when you swipe down to exit the image, and makes it less jarring when the modal appears. 

It also reduces flashes caused by the placeholder appearing initially at the wrong size. 

**Note**: This will need to be reviewed by design post-merge. 

ezgif-2-ae8b834568f4


nathggns:
This has been approved by design.

PR #1872, opened 2019-04-11
feat: Article Extras package (lazy loading related articles & co)

tuncaulubilge:
Ticket: https://nidigitalsolutions.jira.com/browse/REPLAT-5920

This grew up to be quite a big PR, mostly due to tests and showcases moving around between packages. 

`Article Topics`, `comments` and `related articles` have been moved out of `article-skeleton` to a new package called `article-extras` and wrapped by a provider to lazy load these components on the page visit on native apps. Also moved these to article-extras in web to keep things consistent.

- ArticleCommentsProvider is replaced by ArticleExtrasProvider
- Provider logic is removed from ArticleComments and smart logic is moved into ArticleExtras with the improved query
- Showcases are using the new provider with a MockedProvider using a article-extras fixture generator from provider-test-tools
- ArticleExtras tests are using fixtures on web, and fixture generator on native


packages/article-extras/src/article-extras.web.js
```diff
@@ -0,0 +1,47 @@
...
+  commentsEnabled,
+  registerNode,
+  relatedArticleSlice,
+  relatedArticlesVisible,
+  spotAccountId,
+  topics
+}) => (
+  <Fragment>
+    <ArticleTopics topics={topics} />
+    <aside id="related-articles" ref={node => registerNode(node)}>
```
ValGeorgiev: Why you are using aside here?

packages/article-extras/src/article-extras-content.js
```diff
@@ -0,0 +1,66 @@
...
+}) => {
+  const {
+    commentCount,
+    commentsEnabled,
+    relatedArticleSlice,
+    topics
+  } = article;
+
+  return (
+    <View>
```
ValGeorgiev: Can this be a Fragment?

packages/article-extras/src/styles/index.android.js
```diff
@@ -0,0 +1,12 @@
+import { StyleSheet } from "react-native";
+import sharedStyles from "./shared";
+
+const styles = StyleSheet.create({
+  ...sharedStyles,
+  ArticleExtrasBody: {
+    ...sharedStyles.ArticleExtrasBody,
+    color: "green"
```
ValGeorgiev: put that bad boy in the styleguide, please

tuncaulubilge: Lol, these are the templates style I forgot to remove, will get rid of this

packages/article-extras/src/article-extras.web.js
tuncaulubilge: This is moved from article-skeleton to here. We got a lazy loading logic on related articles which is only loaded when you scroll to it

nathggns:
Reviewing this now.

packages/article-extras/README.md
```diff
@@ -0,0 +1,45 @@
+# Article Extras
```
nathggns: I wonder if there's a better name than `ArticleExtras`?

nathggns: Maybe `ArticleFooter`?

packages/article-extras/src/article-extras-error.js
```diff
@@ -6,12 +6,13 @@ import Context from "@times-components/context";
...
-    <Text style={styles.errorBody}>
-      You must have a network connection to view and post comments. Please check
-      your connection and retry.
+const ArticleExtrasError = ({ refetch }) => (
+  <View style={styles.extrasErrorContainer}>
+    <Text style={styles.extrasErrorHeadline}>
+      It looks like you&apos;re offline
+    </Text>
+    <Text style={styles.extrasErrorBody}>
+      Some features, such as related articles and comments, may not be avalaible
```
nathggns: Has this copy been approved?

packages/article-extras/README.md
tuncaulubilge: Not a super fan of `Article extras`, but `Article Footer` is misleading as we have a footer on the page itself as well. Open to suggestions on this one

packages/article-extras/src/article-extras-error.js
tuncaulubilge: Yes, this is from the UX team

packages/article-extras/README.md
nathggns: Sure, which is why I suggested `ArticleFooter` instead of just `Footer`. Esp considering the page footer isn't in TC, I think `ArticleFooter` is probably fine?

tuncaulubilge: Not everyone knows that, and it might also change in the future. These are still part of the article body

PR #1876, opened 2019-04-12
feat: add tracking to save article

Ivaylo1987:
Small PR for adding tacking to article save event.


packages/edition-slices/src/tiles/shared/save-article-tracking-events.js
```diff
@@ -0,0 +1,16 @@
+import { withTrackEvents } from "@times-components/tracking";
+
+export default Component =>
```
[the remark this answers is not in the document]
nathggns: I think you'd probably call this `withArticleSaveTracking`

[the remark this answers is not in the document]
Ivaylo1987: 👍

nathggns:
Reviewing now :)

packages/section/__tests__/shared-tracking.js
```diff
@@ -71,6 +74,31 @@ export default () => {
     expect(onArticlePress.mock.calls).toMatchSnapshot("onArticlePress");
   });
 
+  it("Save/Unsave article click tracking", () => {
```
nathggns: Do you need a test for the second time to make sure `isSaved` is false in the tracking?

packages/section/__tests__/shared-tracking.js
```diff
@@ -71,6 +74,31 @@ export default () => {
...
+          section={edition.sections[0]}
+          stream={stream}
+        />
+      </SectionContext.Provider>
+    );
+
+    const [starButton] = testInstance.root.findAllByType(StarButton);
+    starButton.props.onPress();
+
+    const [[call]] = stream.mock.calls;
```
nathggns: This is a difficult syntax to read with not much benefit – I'd prefer just the classic:

```js
const call = stream.mock.calls[0][0];
```

Ivaylo1987: So, you dont want us to be cool ;)

packages/section/__tests__/shared-tracking.js
Ivaylo1987: We can try this 👍

packages/section/__tests__/shared-tracking.js
nathggns: I think there are benefits when you're destructing multiple elements from an array, or during function argument destructing, but in this case its just harder to read.

Ivaylo1987: Sure, I agree will fix it :)

packages/section/__tests__/shared-tracking.js
Ivaylo1987: Done!

packages/section/__tests__/shared-tracking.js
Ivaylo1987: done!

PR #1881, opened 2019-04-15
feat: Update section data on native event

tuncaulubilge:
This PR enables Section page data to be updated without a need to re-create the view. (https://nidigitalsolutions.jira.com/browse/REPLAT-6100)

To use this feature, native apps needs to implement `getSectionData` method in `SectionEvents` and resolve the promise with a section data (as a raw json string). 
```
fun getSectionData(sectionId: String, promise: Promise) {
  ...
}
```

Then, whenever the native app send an event named `updateSectionData`, components will use `getSectionData` bridge to ask for the updated section data and refresh its view. 

- Section prop is moved to state to enable reloads
- Subscribes and unsubscribes added to listen to `updateSectionData` events through DeviceEventEmitter
- Ask for section data when event is fired and update the state if it's successfully resolved
- Tidied up the pages/section a bit to remove inline functions

### DIFF
--- a/packages/pages/src/section/section.js
+++ b/packages/pages/src/section/section.js
@@ -9,15 +9,19 @@ const {
   getOpenedPuzzleCount,
   getSavedArticles,
   getSectionData,
-  onArticlePress,
+  onArticlePress: onArticlePressBridge,
   onPuzzleBarPress = () => {},
-  onPuzzlePress,
-  onArticleSavePress
+  onPuzzlePress: onPuzzlePressBridge,
+  onArticleSavePress: onArticleSavePressBridge
 } = NativeModules.SectionEvents || {
   onArticlePress: () => {},
   onPuzzleBarPress: () => {},
   onPuzzlePress: () => {}
 };
+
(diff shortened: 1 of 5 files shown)


packages/pages/src/section/section.js
```diff
@@ -8,37 +8,51 @@ import trackSection from "./track-section";
...
     this.toggleArticleSaveStatus = this.toggleArticleSaveStatus.bind(this);
     this.syncAppData = this.syncAppData.bind(this);
+    this.updateSectionData = this.updateSectionData.bind(this);
+    this.onArticleSavePress = this.onArticleSavePress.bind(this);
   }
 
   componentDidMount() {
     AppState.addEventListener("change", this.onAppStateChange);
     DeviceEventEmitter.addListener("updateSavedArticles", this.syncAppData);
+    DeviceEventEmitter.addListener("updateSectionData", this.updateSectionData);
```
Ivaylo1987: Just for clarity when is this event raised in mobile app?

tuncaulubilge: They'll use this for pull to refresh, if the edition data updated

PR #1886, opened 2019-04-17
Feat/hover-and-clicked-states

ValGeorgiev:
Hover and clicked state for share and save component icons


nathggns:
Reviewing now.

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -13,64 +13,216 @@ import SharingApiUrls from "./constants";
...
+      twClicked,
+      twColour
+    } = this.state;
+
+    return (
+      <View style={styles.container}>
+        <View style={styles.rowItem}>
+          <Text style={styles.label}>Share</Text>
+          <View
+            onMouseEnter={() =>
```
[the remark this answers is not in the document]
nathggns: If you hover over the Twitter icon, you'll see what I mean.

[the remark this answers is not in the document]
nathggns: Just updated this a bit. https://codesandbox.io/s/vpkyz5vz7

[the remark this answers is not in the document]
ValGeorgiev: Done, thanks

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -13,64 +14,178 @@ import SharingApiUrls from "./constants";
...
+              <IconEmail
+                fillColour="currentColor"
+                height={styles.svgIcon.height}
+                title="Share by email client"
+              />
+            </HoverIcon>
+          </Link>
+          <Link
+            onPress={this.onTwClick}
+            responsiveLinkStyles={twClicked ? styles.clickedLink : styles.link}
```
nathggns: Have you considered extracting this clicked logic to another component so its not duplicated for each icon? Would make this component a lot simpler.

https://codesandbox.io/s/ym4r8kzzoj

Sorry for not mentioning it sooner, sometimes it's harder to see the possible simplifications when there's still a lot going on.

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -13,64 +14,178 @@ import SharingApiUrls from "./constants";
...
+              />
+            </HoverIcon>
+          </Link>
+        </View>
+      </View>
+    );
+  }
+}
+
+const HoverIcon = styled.div`
```
nathggns: Can you define this before its used? Just a little bit easier to follow that way is all.

packages/save-and-share-bar/src/save-and-share-bar.js
nathggns: I think if you do this, the `SaveAndShareBar` can just become a stateless component.

nathggns: It also might be worth just straight up adding this functionality to the Link package.

tuncaulubilge: Adding hover support to link package is a good idea 👍

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -13,64 +14,178 @@ import SharingApiUrls from "./constants";
...
+            url={`${SharingApiUrls.twitter}?text=${articleUrl}`}
+          >
+            <HoverIcon
+              color={styles.svgIcon.fillColour}
+              hoverColor={styles.svgIcon.hoverFillColour}
+            >
+              <IconTwitter
+                fillColour="currentColor"
+                height={styles.svgIcon.height}
+                title="Share on tweeter"
```
nathggns: Typo – tweeter.

packages/save-and-share-bar/src/save-and-share-bar.js
nathggns: This should give you an idea for how simple its possible to make this component https://gist.github.com/nathggns/f33f0fd4a7fad5e0401988f16b1460da

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -13,64 +14,178 @@ import SharingApiUrls from "./constants";
...
+            }
+          >
+            <HoverIcon
+              color={styles.svgIcon.fillColour}
+              hoverColor={styles.svgIcon.hoverFillColour}
+            >
+              <IconCopyLink
+                fillColour="currentColor"
+                height={styles.svgIcon.height}
+                title="Copy link on clipboard"
```
nathggns: Should probably be `Copy link to clipboard`

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -13,64 +14,178 @@ import SharingApiUrls from "./constants";
...
+            url={`${SharingApiUrls.facebook}?text=${articleUrl}`}
+          >
+            <HoverIcon
+              color={styles.svgIcon.fillColour}
+              hoverColor={styles.svgIcon.hoverFillColour}
+            >
+              <IconFacebook
+                fillColour="currentColor"
+                height={styles.svgIcon.fb.height}
+                title="Share on facebook"
```
nathggns: `Share on Facebook`

packages/save-and-share-bar/src/save-and-share-bar.js
nathggns: Should be `Share on Twitter`

ValGeorgiev:
closing because something weird happen with style-components packages, will create a new one

PR #1887, opened 2019-04-17
spike: helmet proof of concept for topic title

nathggns:
This is a proof of concept on how we can use Helmet to manage head tags from times-components, using the title tag as an example. It has server side rendering support and can be updated dynamically on the client side.

This proof of concept is showing you the integration necessary using SSR's standalone server, but I will shortly link an example of this proof of concept being integrated with render. 

In this proof of concept, we have a dynamically updating title tag on a topic page such as http://localhost:3000/topic/religion?page=2 (you'll need to be running SSR on this branch first) 

We're using `react-helmet-async`, as the original implementation of `react-helmet` is not thread safe, meaning you cannot have any async processes as part of your rendering pipeline (such as Apollo, which we use). Read more about this here https://open.nytimes.com/the-future-of-meta-tag-management-for-modern-react-development-ec26a7dc9183

Helmet has far more features than what we need, so this proof of concept only has support for meta, title, and link tags. It'll be relatively easy to add support for more features. 

As you can see, there's a bit of work necessary in SSR to add server side rendering support, but for the most part the usage of helmet is extremely simple as you can see in the topic package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -164,6 +164,7 @@
     "aws-sdk": "2.238.1",
     "babel-preset-es2015": "6.24.1",
     "react-art": "16.5.2",
+    "react-helmet-async": "1.0.2",
     "react-native": "0.55.4",
     "react-native-showcase-loader": "1.1.0"
   }
--- a/package.json
+++ b/package.json
@@ -164,7 +164,6 @@
     "aws-sdk": "2.238.1",
     "babel-preset-es2015": "6.24.1",
     "react-art": "16.5.2",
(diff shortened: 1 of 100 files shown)


nehasri89:
LGTM 👍

tuncaulubilge:
Looking good :+1:

ValGeorgiev:
LGTM

nathggns:
This is not going to be a spike anymore. Closing this PR and will make a new one later with the actual implementation.

PR #1891, opened 2019-04-18
BREAKING CHANGE feat: manage header tags

nathggns:
This implements the ability to manage `title`, `link`, and `meta` tags from within times-components using React Helmet.

This includes moving over the existing tags from render. 

Todo:

- [x] Topic Page
- [x] Author Page
- [x] Article Page

This will require a later PR to render to make use of this.


nathggns:
Going to replace this with a different PR.

PR #1893, opened 2019-04-22
Feat/share and save bar placement anchoring

apostolnikov:
Implement Share and Save Bar placement and anchoring.
Added mock for ArticleSkeleton in all five article templates. There are tests for ArticleSkeleton only and there is no need to update snapshots for every ArticleSkeleton change for the packages that are using it.

image

With anchoring:
image

image


packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -24,17 +27,33 @@ const adStyle = {
...
+  componentWillUnmount() {
+    window.removeEventListener("scroll", this.handleScroll);
+  }
+
+  handleScroll() {
+    const offsetTop = this.sticky.getBoundingClientRect().top;
+
+    return offsetTop === 1 || offsetTop === 0
+      ? this.setState(state => ({ ...state, isSticky: true }))
+      : this.setState(state => ({ ...state, isSticky: false }));
```
ValGeorgiev: ` const isSticky = offsetTop === 1 || offsetTop === 0;
   this.setState(state => ({ ...state, isSticky }));
`

nathggns:
Could we please do the mocking of ArticleSkeleton in a different PR because 

A) I think there are legitimate reasons for *not* doing this and we should discuss that separately 
B) It makes it impossible to view what you've actually done in this PR because its so massive.

nathggns:
> Looks good! Very good job, only few things with setState and default background color, after this we are ready to merge it and unblock other stories.
> 
> P.S. I am all for mocking this ArticleSkeleton, there is no need to test ArticleSkeleton on multiple places. ArticleSkeleton should be tested only in ArticleSkeleton package

@ValGeorgiev I really don't think we should merge this with the mocked article skeleton. I'm pretty sure this will reduce our test coverage around article template headers and I can't actually review this properly when its so big.

tuncaulubilge:
Mocking article skeleton does make sense but let's do that in a separate pr

packages/article-skeleton/src/styles/responsive.web.js
```diff
@@ -28,3 +30,49 @@ export const BodyContainer = styled(View)`
   display: block;
   order: 3;
 `;
+
+/* --- SaveAndShareBar --- */
+export const SaveShareContainer = styled(View)`
+  ${props =>
+    props.isSticky && windowWidth <= breakpoints.huge
+      ? `width: 100% !important;
```
nathggns: CSS embedded in styled-components should use the CSS template function to work properly. In addition I'd probably abstract it to make it more readable. 

```js
import styled, { css } from "styled-components";

const STICKY_STYLES = css`
    width: 100% !important;
    /* ... */
`; 

export const SaveShareContainer = styled(View)`
    ${props => props.isStick && windowWidth <= breakpoints.huge ? STICKY_STYLES : ""}
    /* ... */
`;

File: packages/article-skeleton/src/styles/responsive.web.js [hunk 3]
```diff
@@ -28,3 +30,49 @@ export const BodyContainer = styled(View)`
   display: block;
   order: 3;
 `;
+
+/* --- SaveAndShareBar --- */
+export const SaveShareContainer = styled(View)`
+  ${props =>
+    props.isSticky && windowWidth <= breakpoints.huge
```
nathggns: I don't think this will work the way you want it to as the styles won't automatically reevaluate when the windowWidth changes. It's probably safer to pass windowWidth as a prop if you want to revaluate the styles.

packages/article-skeleton/src/styles/responsive.web.js
```diff
@@ -28,3 +30,49 @@ export const BodyContainer = styled(View)`
...
+    z-index: 999;
+  }
+`;
+
+export const RefContainer = styled.div`
+  border-top: 1px solid;
+  border-bottom: 1px solid;
+  border-color: ${colours.functional.keyline};
+
+  ${props =>
```
nathggns: This can be cleaner too.

```js
const REF_STICKY_STYLES = css`
    ${windowWidth <= breakpoints.huge ? css`border: 0;` : ""};
    ${windowWidth <= breakpoints.medium ? css`padding: 0 ${spacing(2)};` : ""};
`;

export const RefContainer = styled.div`
    /* ... */
    ${props => props.isSticky ? REF_STICKY_STYLES : "" }
     /* ... */
`;

File: packages/article-skeleton/src/styles/responsive.web.js [hunk 5]
```diff
@@ -28,3 +30,49 @@ export const BodyContainer = styled(View)`
   display: block;
   order: 3;
 `;
+
+/* --- SaveAndShareBar --- */
+export const SaveShareContainer = styled(View)`
```
nathggns: Why do you use `styled(View)` for `SaveShareContainer` and `styled.div` for `RefContainer`?

packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -36,6 +41,18 @@ class ArticleSkeleton extends Component {
...
+    window.addEventListener("scroll", this.handleScroll);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener("scroll", this.handleScroll);
+  }
+
+  handleScroll() {
+    const offsetTop = this.sticky.getBoundingClientRect().top;
+    const isSticky = offsetTop === 1 || offsetTop === 0;
```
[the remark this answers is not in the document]
IvayloAndonov: `offsetTop <= 1` wont work correctly, because for large screens there is no sticky bar and 'offsetTop' can be with negative - values.

packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -36,6 +41,18 @@ class ArticleSkeleton extends Component {
...
+
+    return this.setState({ isSticky });
```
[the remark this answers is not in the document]
IvayloAndonov: Done

packages/article-skeleton/src/article-skeleton.web.js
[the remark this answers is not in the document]
nathggns: But aren't the styles only applied if sticky and within the breakpoint?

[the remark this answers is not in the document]
nathggns: If you need the stickyness not to apply at certain breakpoints, put a condition for that. Don't assume other people can understand implicit conditions like that, because the code just looks weird. Besides, I'm pretty sure this code is just going to be buggy based on scroll speed.

packages/article-skeleton/src/styles/responsive.web.js
IvayloAndonov: Done

packages/article-skeleton/src/styles/responsive.web.js
IvayloAndonov: Done

packages/article-skeleton/src/styles/responsive.web.js
IvayloAndonov: Updated

packages/article-skeleton/src/styles/responsive.web.js
IvayloAndonov: Done

packages/article-skeleton/src/article-skeleton.web.js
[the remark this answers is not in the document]
IvayloAndonov: Updated

packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -14,7 +15,9 @@ import insertDropcapIntoAST from "./dropcap-util";
 import {
   BodyContainer,
   HeaderAdContainer,
-  MainContainer
+  MainContainer,
+  SaveShareContainer,
+  RefContainer
```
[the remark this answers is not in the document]
IvayloAndonov: Done

packages/article-skeleton/src/article-skeleton.web.js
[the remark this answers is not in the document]
IvayloAndonov: Condition is updated. Tested on all screen sizes and different scroll speeds.Works fine.

PR #1894, opened 2019-04-23
feat: implement copy to clipboard functionality

IvayloAndonov:
**Share and Save - Copy link to clipboard**
Refer to Jira : https://nidigitalsolutions.jira.com/browse/REPLAT-6137

Implement copy article url to clipboard functionality.
Tested on : Chrome, Safari, Mozilla, Edge, IE11.

_Note: Implemented only copy to clipboard logic in this PR. The  feedback ‘Link copied’ confirmation message is not developed yet as part of [Generic Message Bar component](https://nidigitalsolutions.jira.com/browse/REPLAT-6144) REPLAT-6144 ._


packages/save-and-share-bar/__tests__/shared.base.js
```diff
@@ -51,6 +61,7 @@ export default () => {
         expect(onShareOnEmail).toHaveBeenCalled();
 
         testInstance.root.findAllByType(BarItem)[3].props.onPress();
+        expect(Clipboard.setString).toHaveBeenCalledWith(articleUrlMock);
```
ValGeorgiev: Could we compare Clipboard value with articleUrlMock as well

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -13,66 +13,76 @@ import styles from "./styles";
...
+              height={styles.svgIcon.height}
+              title="Share by email client"
+            />
+          </BarItem>
+          <BarItem
+            target="_blank"
+            url={`${SharingApiUrls.twitter}?text=${articleUrl}`}
+          >
+            <IconTwitter
+              fillColour={styles.svgIcon.fillColour}
```
ValGeorgiev: fillCoulour should be "currentColor"

IvayloAndonov: Done

packages/save-and-share-bar/__tests__/shared.base.js
IvayloAndonov: I think that Clipboard module functionality as part of "react-native" is not necessary to be tested separately. https://facebook.github.io/react-native/docs/clipboard

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -1,6 +1,6 @@
-import React from "react";
+import React, { Component } from "react";
 import PropTypes from "prop-types";
-import { View, Text } from "react-native";
+import { View, Text, Clipboard } from "react-native";
```
nathggns: Do we know what the browser support of React Native Web's implementation of Clipboard is? Considering this is a web only feature, I'm not sure if it's necessary to use RNW's implementation if it isn't super supported.

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -13,66 +13,81 @@ import styles from "./styles";
...
-);
+class SaveAndShareBar extends Component {
+  constructor(props) {
+    super(props);
+    this.copyToClipboard = this.copyToClipboard.bind(this);
+  }
+
+  copyToClipboard() {
+    const { onCopyLink, articleUrl } = this.props;
+    Clipboard.setString(articleUrl);
```
nathggns: Massively resent that we've had to turn this component into a class component for this. Is there a way we can just call the clipboard set function in the consumer?

tuncaulubilge:
We need to make sure this works on IE 10 (and above)

packages/save-and-share-bar/src/save-and-share-bar.js
IvayloAndonov: The reason that this functional component is transformed into class component is not only that we have to extend it with new methods according next stories. For "save to my articles" functionality we have to keep and local state in this class component to indicate if spinner is shown or not ( async operation ). If it will be used different approach, let`s keep it functional component ?

packages/save-and-share-bar/src/save-and-share-bar.js
IvayloAndonov: First of all I implemented it with https://www.npmjs.com/package/react-copy-to-clipboard library which add additional dependency. After small research I found that Clipboard module from react-native is supported and for RNW. In order to save using of superfluous new dependencies I switched to Clipboard as easiest and cleanest way.

IvayloAndonov: Screenshot 2019-04-23 at 16 51 48

IvayloAndonov:
> We need to make sure this works on IE 10 (and above)

Tested on windows machine on IE11 and Edge - working as expected. We tried to test it on document mode IE10 but the entire Storybook app can`t be loaded.
Screenshot 2019-04-23 at 16 51 48

IvayloAndonov:
> We need to make sure this works on IE 10 (and above)

@tuncaulubilge Tested on windows machine on IE11 and Edge - working as expected. We tried to test it on document mode IE10 but the entire Storybook app can`t be loaded. All others browsers are tested with clipboard api.

image

packages/save-and-share-bar/src/save-and-share-bar.js
nathggns: Ah – glad to see RNW's implementation has cross-browser support.

packages/save-and-share-bar/src/save-and-share-bar.js
Ivaylo1987: You will have to create `copyToClipboard` in some component eventually so it better be incapsulated in the bar itself rather than having to pass copy functionality to every instance.

IvayloAndonov:
> Clipboard has an `isAvailable` function, let's use that before displaying the button, so that we only display the button if the feature is supported.
> 
> See: https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/Clipboard/index.js#L14

Good catch!

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -13,66 +13,83 @@ import styles from "./styles";
...
+            target="_blank"
+            url={`${SharingApiUrls.facebook}?text=${articleUrl}`}
+          >
+            <IconFacebook
+              fillColour="currentColor"
+              height={styles.svgIcon.fb.height}
+              title="Share on Facebook"
+            />
+          </BarItem>
+          {Clipboard.isAvailable() && (
```
nathggns: We need a snapshot test for the scenario where it is not available.

IvayloAndonov: Done

packages/save-and-share-bar/__tests__/web/__snapshots__/save-and-share-bar.test.js.snap
```diff
@@ -76,15 +76,18 @@ exports[`1. save and share bar renders correctly 1`] = `
     </BarItem>
     <BarItem
       target="_blank"
-      url="https://www.facebook.com/sharer/sharer.php?text="
+      url="https://www.facebook.com/sharer/sharer.php?u="
```
nathggns: Just checking this change is completely intentional?

IvayloAndonov: Yes

PR #1897, opened 2019-04-24
BREAKING CHANGE feat: manage meta tags in times-components 

nathggns:
This PR allows pages within times-components to manage title, meta, and link tags on web using `react-helmet-async` with server slide and client side dynamic support.

SSR spits out these tags as the string `headMarkup` and it is up to the consumer to place this on the page for server side rendering. 

I've also copied across all of the meta tags for author & topic pages, and *most* of them for article. There's the default lead asset meta tags that haven't yet been copied, as this is blocked on us having the ability to bundle binary assets for web. 

I've also replaced almost all instances of `Context.Provider` with a new `ContextProviderWithDefaults`, which merges the provided value with the defaults specified (so that tests don't need to declare `makeArticleUrl` and the new `makeTopicUrl`).

This is a breaking change as consumers of SSR must now pass `makeTopicUrl` (similarly to `makeArticleUrl`)

This will require a later PR for integrating with render.

As the data for these meta tags is now coming from TPA instead of from content-api, there'll be some differences in the data. It is possible that the section displayed in the title will be different on articles that are within multiple sections. The description meta will also be 200 characters now instead of the ~170 it was previously. This should not be an issue. 

In addition, as the data used for the canonical URL on articles does not yet appear to be available in TPA, this has not yet been moved across and will be done at a later date instead.

### DIFF
--- a/packages/article-skeleton/__tests__/web/head.web.test.js
+++ b/packages/article-skeleton/__tests__/web/head.web.test.js
@@ -1,0 +1,308 @@
+import React from "react";
+import TestRenderer from "react-test-renderer";
+
+import {
+  addSerializers,
+  compose,
+  enzymeRenderedSerializer,
+  minimaliseTransform,
+  minimalWebTransform,
+  print
+} from "@times-components/jest-serializer";
+
+import Head from "../../src/head.web";
+import articleFixture, { testFixture } from "../../fixtures/full-article";
+
+jest.mock("react-helmet-async", () => ({ Helmet: "Helmet" }));
+
+const omitProps = new Set([
+  "className",
+  "data-testid",
+  "responsiveLinkStyles",
+  "style"
+]);
+
+addSerializers(
+  expect,
+  enzymeRenderedSerializer(),
+  compose(
+    print,
+    minimalWebTransform,
+    minimaliseTransform((value, key) => omitProps.has(key))
+  )
+);
+
+const article = articleFixture({ ...testFixture });
+
+describe("Head", () => {
+  it("outputs correct metadata", () => {
+    const testRenderer = TestRenderer.create(<Head article={article} />);
+
+    expect(testRenderer).toMatchSnapshot();
+  });
+
+  it("uses the first non-news section in title", () => {
+    const testRenderer = TestRenderer.create(
+      <Head
+        article={{
+          ...article,
+          tiles: [
+            {
+              slices: [
+                {
+                  sections: [
+                    {
+                      title: "News"
+                    }
+                  ]
+                }
+              ]
+            },
+            {
+              slices: [
+                {
+                  sections: [
+                    {
+                      title: "Comment"
+                    }
+                  ]
+                }
+              ]
+            },
+            {
+              slices: [
+                {
+                  sections: [
+                    {
+                      title: "Foreign"
+                    }
+                  ]
+                }
+              ]
+            }
+          ]
+        }}
+      />
+    );
+
+    const title = testRenderer.root.findByType("title");
+
+    expect(title.children).toMatchSnapshot();
+  });
+
+  it("uses news in title if news is the only available section", () => {
+    const testRenderer = TestRenderer.create(
+      <Head
+        article={{
+          ...article,
+          tiles: [
+            {
+              slices: [
+                {
+                  sections: [
+                    {
+                      title: "News"
+                    }
+                  ]
+                }
+              ]
+            }
+          ]
+        }}
+      />
+    );
+
+    const title = testRenderer.root.findByType("title");
+
+    expect(title.children).toMatchSnapshot();
+  });
+
+  it("uses news in title if news is the only available section", () => {
+    const testRenderer = TestRenderer.create(
+      <Head
+        article={{
+          ...article,
+          tiles: [
+            {
+              slices: [
+                {
+                  sections: [
+                    {
+                      title: "News"
+                    }
+                  ]
+                }
+              ]
+            }
+          ]
+        }}
+      />
+    );
+
+    const title = testRenderer.root.findByType("title");
+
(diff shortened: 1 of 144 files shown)


packages/article-skeleton/__tests__/web/head.web.test.js
```diff
@@ -0,0 +1,324 @@
...
+          ...article,
+          headline: null,
+          shortHeadline: "short headline"
+        }}
+      />
+    );
+
+    const title = testRenderer.root.findByType("title");
+
+    expect(title.children).toMatchSnapshot();
```
ValGeorgiev: why we need snapshot here, you can easily compare title value with shortHeadline

packages/article-skeleton/__tests__/web/head.web.test.js
```diff
@@ -0,0 +1,324 @@
...
+              ]
+            }
+          ]
+        }}
+      />
+    );
+
+    const title = testRenderer.root.findByType("title");
+
+    expect(title.children).toMatchSnapshot();
```
ValGeorgiev: same in every test here, why we need a new snapshot files?

packages/article-skeleton/src/head.web.js
```diff
@@ -0,0 +1,115 @@
...
+    return acc;
+  }, []);
+
+  return renderTreeAsText({ children });
+}
+
+const PUBLICATION_NAMES = {
+  SUNDAYTIMES: "The Sunday Times",
+  TIMES: "The Times"
+};
```
ValGeorgiev: We should have this one in a something like constants package and reuse it everywhere, but this can be done in different pr

packages/article-skeleton/src/head.web.js
```diff
@@ -0,0 +1,115 @@
...
+
+  const children = bylines.reduce((acc, byline) => {
+    acc.push(...byline.byline);
+    return acc;
+  }, []);
+
+  return renderTreeAsText({ children });
+}
+
+const PUBLICATION_NAMES = {
```
aronshamash: Can we use : https://github.com/newsuk/times-components/blob/master/packages/date-publication/src/publication.js ?

packages/ssr/src/lib/make-urls.js
```diff
@@ -0,0 +1,3 @@
+module.exports.makeArticleUrl = ({ id }) =>
+  `http://localhost:3000/article/${id}`;
```
aronshamash: why is this absolute url and topic is remote?

packages/article-skeleton/src/head.web.js
nathggns: I considered doing that but I can't really think of an obvious place to it.

packages/ssr/src/lib/make-urls.js
nathggns: This was just moving existing implementations of both – `makeArticleUrl` was the existing one within SSR, and `makeTopicUrl` was what `article-topics` did already (it just linked to `/topic/{slug}`). I hadn't noticed the inconsistency because of that. I'll update it.

packages/article-skeleton/__tests__/web/head.web.test.js
nathggns: They're quicker to write, and quicker to update if/when the format of various values change.

nathggns: To explain more clearly, the title tag has a lot of different parts in it, meaning if any of them change, I'll have to update a lot of tests – it's quicker to do that with snapshot tests.

packages/article-skeleton/__tests__/web/head.web.test.js
nathggns: Answered above.

nathggns: Here https://github.com/newsuk/times-components/pull/1897#discussion_r278171139

packages/article-skeleton/src/head.web.js
nathggns: Discussing here https://github.com/newsuk/times-components/pull/1897#discussion_r278165919

packages/ssr/src/lib/make-urls.js
nathggns: This is fixed now.

packages/article-skeleton/src/head.web.js
nathggns: If you can suggest where to put it, I'll move it!

packages/ssr/src/component/author-profile.js
```diff
@@ -2,49 +2,57 @@
...
+        {
+          debounceTimeMs,
+          page,
+          pageSize,
+          slug: authorSlug
+        },
+        ({ author, isLoading, error, refetch }) =>
+          React.createElement(
+            ContextProviderWithDefaults,
+            { value: { makeArticleUrl, makeTopicUrl } },
```
Ivaylo1987: Is `makeTopicUrl` really needed here?

nathggns: Yes. The idea is that context will always have both `makeArticleUrl` and `makeTopicUrl` within them, so that any of our components can be included on any page. It's safer for SSR to include both on every page and we wouldn't gain anything by removing it.

packages/article-skeleton/src/head.web.js
aronshamash: maybe `utils` ?

packages/ssr/src/component/author-profile.js
Ivaylo1987: 👍

packages/article-skeleton/src/head.web.js
nathggns: will do this – can you please review the rest of it?

PR #1898, opened 2019-04-24
feat: improve save and share bar icons state onPress

IvayloAndonov:
**Improve Save and Share bar component** - managing hover and press icons state with pure css instead of creation of separate LinkWithPressedStyle class with state

- add ":active" css selector used for onPress icons state
- refactor BarItem component
- remove obsolete files
- update snapshots


packages/save-and-share-bar/src/bar-item.js
```diff
@@ -12,21 +13,26 @@ const HoverIcon =
...
+const BarItem = ({ children, color, hoverColor, onPress, ...props }) => (
+  <Link onPress={onPress} responsiveLinkStyles={styles.link} {...props}>
     <HoverIcon color={color} hoverColor={hoverColor}>
       {children}
     </HoverIcon>
-  </LinkWithPressedStyle>
+  </Link>
 );
 
+BarItem.defaultProps = {
```
nathggns: I don't think it's a great idea to use defaultProps for function components instead of just default arguments. It's arguably slower (defaulting props in userland is going to be slower than the engine doing it for you) and it means that the defaults are much further away from where you use them. Dan Abramov seems to agree. https://twitter.com/dan_abramov/status/1070768024569987078

packages/save-and-share-bar/src/styles/index.js
```diff
@@ -27,6 +27,9 @@ const styles = {
       &:hover {
         background-color: ${colours.functional.whiteGrey};
       }
+      &:active{
```
nathggns: Missing a space

packages/save-and-share-bar/src/bar-item.js
```diff
@@ -12,21 +13,26 @@ const HoverIcon =
...
+  color: styles.svgIcon.fillColour,
+  hoverColor: styles.svgIcon.hoverFillColour,
+  onPress: () => {}
+};
+
+BarItem.propTypes = {
+  children: PropTypes.node.isRequired,
+  color: PropTypes.string,
+  hoverColor: PropTypes.string,
```
aronshamash: Is there a reason for using US spelling of `Color` ?

IvayloAndonov: Nope, just habit from previous US jobs. Renamed.

packages/save-and-share-bar/src/bar-item.js
IvayloAndonov: Agree, I also prefer default arguments.Just saw in the repo that many others components are developed by using defaultProps so I used it the same in terms of consistency.

nathggns: This isn't resolved.

The committed version in master currently using default arguments and there is no reason to change it to using defaultProps in this PR.

Especially when default arguments seem to be the better option.

I'm all for consistency, but I would implement consistency by changing each instance of functional components using defaultProps to use default arguments instead.

IvayloAndonov: Can you see that the "defaultProps" are not used anymore in this PR with latest update?

PR #1901, opened 2019-04-25
Feat/add teaser page

pdimova:
https://nidigitalsolutions.jira.com/browse/REPLAT-2875

Hiding Topics, Related Articles and Comments from the article when user is not logged in.

User state is passed from render as an object, that object is stored in the context and later used in the ArticleSkeleton where different parts are hidden based on the state

Note: Save and Share bar is handled by another story


packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -79,47 +80,53 @@ class ArticleSkeleton extends Component {
...
-            )}
-          </LazyLoad>
-        </AdComposer>
+                        <ArticleExtras
+                          analyticsStream={analyticsStream}
+                          articleId={articleId}
+                          commentsEnabled={commentsEnabled}
+                          registerNode={registerNode}
+                          relatedArticleSlice={
+                            user.isLoggedIn ? relatedArticleSlice : null
```
nehasri89: This might be part of user state theory but worth mentioning here. The user has multiple user states, but the result would be either to show or hide, so RA can be a boolean here depending on the user state. We can have a util to say if certain user state has RA allowed and then show and hide accordingly instead of having 4 ifs here

packages/article-skeleton/.jestlint
```diff
@@ -0,0 +1,7 @@
+{
+  "maxAttributes": 11,
+  "maxAttributeStringLength": 100,
+  "maxFileSize": 40000,
+  "maxLines": 780
+}
```
ValGeorgiev: please delete this file, we don't use jestlint anymore

pdimova: alright, removed

packages/article-skeleton/src/article-skeleton.web.js
aronshamash: yes think we were going to refactor in a future PR

packages/article-skeleton/__tests__/shared.web.js
```diff
@@ -29,5 +38,303 @@ export default () => {
     )
   );
 
-  shared(TestRenderer.create);
+  const article = articleFixture({
```
aronshamash: Can we put in a separate file

pdimova: moved to a separate file

packages/article-skeleton/src/article-skeleton.web.js
pdimova: Changed the ifs with 4 booleans, for logged out user they are all false, but for expired some are true, other false. When we pick up the rest of the user state stories we can have an util that will check a combo of user state properties and will set the 4 bools. Also now ArticleExtras component is responsible for hiding/showing the components

packages/article-in-depth/__tests__/web/__snapshots__/semantic.web.test.js.snap
```diff
@@ -253,26 +253,8 @@ exports[`1. a full article 1`] = `
...
-    Comments are subject to our community guidelines, which can be viewed
-     
-    <a
-      href="https://www.thetimes.co.uk/article/f4024fbe-d989-11e6-9063-500e6740fc32"
-      role="link"
-    >
-      here
-    </a>
-    .
-  </p>
```
ValGeorgiev: ^ can you double check this?

packages/article-main-standard/__tests__/web/__snapshots__/article.web.test.js.snap
```diff
@@ -133,20 +133,6 @@ exports[`1. an article 1`] = `
...
-              Array [
-                Object {
-                  "name": "Topic",
-                  "slug": "topic",
-                },
-              ]
-            }
-          />
-        </nav>
-      </div>
```
ValGeorgiev: ^ this one as well

packages/article-in-depth/__tests__/web/__snapshots__/semantic.web.test.js.snap
pdimova: semantic tests now represent logged in user

packages/article-main-standard/__tests__/web/__snapshots__/article.web.test.js.snap
pdimova: this was also updated with the correct user state

ValGeorgiev:
Good job!

PR #1903, opened 2019-04-29
feat: replat-6101 edition tile summary padding fix

parkjh:
- removes unnecessary padding of empty element when there are no expiable flags prop set


parkjh:
Screenshot 2019-04-29 at 14 45 04

parkjh:
Screenshot 2019-04-29 at 15 26 23

packages/section/__tests__/shared-tracking.js
```diff
@@ -74,6 +74,35 @@ export default () => {
...
+    link.props.onPress();
+
+    const [[call]] = stream.mock.calls;
+
+    expect(call).toMatchSnapshot();
+    expect(onPuzzlePress.mock.calls).toMatchSnapshot("onPuzzlePress");
+  });
+};
+
+// https://nidigitalsolutions.jira.com/browse/REPLAT-6294
```
aronshamash: We don't really need the jira reference

packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -37,7 +37,10 @@ class TileSummary extends Component {
       starStyle
     } = this.props;
 
-    if (!withStar) {
+    if (!withStar || Platform.OS === "ios") {
```
aronshamash: ideally we wouldn't have a `Platform.OS` test but use a `.android.js` and `.web.js` file (`.js` is the picked up by iOS)

packages/edition-slices/src/slices/shared/gutter.js
```diff
@@ -1,18 +1,24 @@
...
-  >
-    {children}
-  </View>
+  <ResponsiveContext.Consumer>
+    {({ isTablet }) => (
+      <View
+        style={{
+          alignSelf: "center",
+          maxWidth: "100%",
+          paddingHorizontal: isTablet ? spacing(2) : spacing(0),
```
aronshamash: `spacing(2) : 0` is fine

packages/edition-slices/src/slices/shared/gutter.js
```diff
@@ -1,18 +1,24 @@
...
-      width: editionMaxWidth
-    }}
-  >
-    {children}
-  </View>
+  <ResponsiveContext.Consumer>
+    {({ isTablet }) => (
+      <View
+        style={{
+          alignSelf: "center",
```
ValGeorgiev: Please don't use "inline" styling, you can move this in a styling file and use breakpoints there

parkjh: @ValGeorgiev that styling was already there, I thought best not to change it

packages/section/__tests__/shared-tracking.js
parkjh: forgot this was public repo...

packages/edition-slices/src/tiles/shared/tile-summary.js
parkjh: @aronshamash happy to chat about this after standup?

parkjh: so the tile star should not be rendered to iOS at all

packages/edition-slices/src/slices/shared/gutter.js
ValGeorgiev: Hm ok, we will have a ticket for Gutter refactoring and we can leave it for now

PR #1904, opened 2019-04-29
fix: Issues with lead asset size

aronshamash:
- Fixes various issue with lead asset / summary text sizes (https://nidigitalsolutions.jira.com/browse/REPLAT-6298)
- ~Updates types.ts as it was out of date~
- Other issues raised in the ticket should be solved once ios app has the latest queries updated (missing `bylines` and `strapline` field in some slices

### DIFF
--- a/packages/fixture-generator/src/types.ts
+++ b/packages/fixture-generator/src/types.ts
@@ -1,3 +1,11 @@
+export interface PaginateArgs {
+  cursor?: Cursor | null;
+
+  first?: number | null;
+
+  desc?: boolean | null;
+}
+
 export interface DateFilter {
   from: DateTime;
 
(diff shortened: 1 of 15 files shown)


packages/fixture-generator/src/types.ts
```diff
@@ -29,6 +21,8 @@ export interface ArticleInput {
 
   authors: Slug[];
 
+  byline: Markup;
```
ValGeorgiev: Why we need byline?

aronshamash: this a generated file see: https://github.com/newsuk/times-components/blob/master/packages/fixture-generator/README.md#fixturegenerator

aronshamash: as discussed removed `types.ts` for another PR

PR #1905, opened 2019-04-29
Feat/share save

nehasri89:
This PR creates SaveStar component for saving articles on web only. Also integrates with Share and save bar.

We have two separate mocks for save api
- Showcase one has delay in resolving the promise to mimic the network delay and show the indicator.
- Tests version have the promise resolved right away.

Gif for Share and Save bar


nathggns:
Not even a code review but just as a UX thing can you stop the Share text from moving while its saving? It looks really odd.

tuncaulubilge:
+1 on Nate's comment, it'd look better if "Save" also stayed at the same spot

packages/save-and-share-bar/save-and-share-bar.showcase.js
```diff
@@ -1,15 +1,16 @@
...
     {
       component: () => (
         <SaveAndShareBar
-          articleUrl="articleUrl"
+          articleId="96508c84-6611-11e9-adc2-05e1b87efaea"
+          articleUrl="article-url-test"
           onCopyLink={() => {}}
-          onSaveToMyArticles={() => {}}
-          onShareOnEmail={() => {}}
+          saveApi={saveApi}
```
ValGeorgiev: onShareOnEmail looks removed as well, is it intentional?

packages/save-star-web/package.json
```diff
@@ -0,0 +1,78 @@
...
+  "name": "@times-components/save-star-web",
+  "version": "0.0.1",
+  "description": "Save Star for Web",
+  "main": "dist/save-star-web",
+  "dev": "src/save-star-web",
+  "scripts": {
+    "depcheck": "depcheck --ignores='babel-*,depcheck,eslint,@thetimes/jest-lint,jest,prettier,webpack*' --ignore-bin-package=false --skip-missing",
+    "fmt": "prettier --write '**/*.*'",
+    "prettier:diff": "prettier --list-different '**/*.*'",
+    "lint": "eslint . && yarn prettier:diff && yarn depcheck && jest-lint",
```
ValGeorgiev: we don't need jest-lint here anymore

packages/save-star-web/src/make-client-util.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-env browser */
```
ValGeorgiev: I think we won't need this file only in this package, maybe move it to the util package

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,184 @@
...
+        });
+      })
+      .catch(err => {
+        this.setState({ loadingState: false, savedStatus: true });
+        console.error("Error in connecting to api", err);
+      });
+  }
+
+  bookmarkEvents() {
+    let newStatus = null;
```
ValGeorgiev: this is not needed, you can use const in the first if block

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,184 @@
...
+    color: ${props => props.colour};
+    &:hover {
+      color: ${props => props.hoverColour || props.colour};
+    }
+  `;
+
+class SaveStarWeb extends Component {
+  constructor(props) {
+    super(props);
+    this.bookmarkEvents = this.bookmarkEvents.bind(this);
```
ValGeorgiev: you don't need this binding when you call the function directly

packages/save-star-web/src/styles/index.js
```diff
@@ -0,0 +1,57 @@
...
+    ...fontFactory({
+      font: "supporting",
+      fontSize: "meta"
+    }),
+    color: colours.functional.secondary,
+    marginRight: spacing(2)
+  },
+  link: {
+    base: `
+    margin-left: ${spacing(1)};
```
ValGeorgiev: some tabs are missing here, prettier is not working in strings :(

packages/save-and-share-bar/save-and-share-bar.showcase.js
nehasri89: that is a mistake, I will put in back

nehasri89:
@ValGeorgiev the reason for making it a separate package is because it was a very stateful package, also in future we might want to drop it on the homepage slices.

packages/provider-queries/src/unsave-bookmark.js
```diff
@@ -0,0 +1,8 @@
+import { addTypenameToDocument } from "apollo-utilities";
+import gql from "graphql-tag";
+
+export default addTypenameToDocument(gql`
+  mutation($id: UUID!) {
```
IvayloAndonov: Here should we have mutation name ( ex. UnsaveBookmark ) as well?

packages/save-and-share-bar/.jestlint
```diff
@@ -1,3 +1,4 @@
 {
+  "maxAttributes": 6,
```
nathggns: We don't need this .jestlint file anymore.

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,184 @@
...
+      loadingState: null,
+      savedArticles: null,
+      savedStatus: false
+    };
+  }
+
+  componentDidMount() {
+    const { articleId, saveApi } = this.props;
+
+    this.setState({ loadingState: true });
```
nathggns: `this.setState({ loadingState: typeof window !== "undefined" })`

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,184 @@
...
+
+    if (typeof window === "undefined") {
+      this.setState({ loadingState: false });
+    }
+    saveApi
```
nathggns: Can you use an async function and extract this to a method please? Just cleaner to read that way.

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,184 @@
...
+          });
+        }
+      })
+      .catch(err => {
+        this.setState({ loadingState: false, savedStatus: false });
+        console.error("Error in connecting to api", err);
+      });
+  }
+
+  saveLink(saveStatus) {
```
nathggns: `renderSaveButton`

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,184 @@
...
+    const { fillColour, strokeColour } = saveStyle;
+
+    const saveText = saveStatus ? "Saved" : "Save  ";
+
+    return (
+      /* eslint-disable jsx-a11y/anchor-is-valid */
+      <Fragment>
+        <Text style={styles.label}>{saveText}</Text>
+        <Link
+          onPress={e => {
```
nathggns: This probably shouldn't be an inline function.

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,184 @@
...
+      });
+  }
+
+  saveLink(saveStatus) {
+    const { colour, hoverColour } = this.props;
+
+    const saveStyle = getStyles({ saveStatus });
+    const { fillColour, strokeColour } = saveStyle;
+
+    const saveText = saveStatus ? "Saved" : "Save  ";
```
nathggns: Is there a better way to maintain constant space between them (perhaps with a flex layout) rather than using a space character?

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,184 @@
...
+
+    return (
+      /* eslint-disable jsx-a11y/anchor-is-valid */
+      <Fragment>
+        <Text style={styles.label}>{saveText}</Text>
```
nathggns: I'd probably but the save text inside the link just for UX.

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,184 @@
...
+      </Fragment>
+    );
+  }
+
+  saveBookmark() {
+    this.setState({ loadingState: true });
+    const { savedArticles } = this.state;
+    const { articleId: id, saveApi } = this.props;
+
+    saveApi
```
nathggns: Async/await.

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,184 @@
...
+      });
+  }
+
+  unsaveBookmark() {
+    const { savedArticles } = this.state;
+    const { articleId: id, saveApi } = this.props;
+
+    this.setState({ loadingState: true });
+
+    saveApi
```
nathggns: Async/await.

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,184 @@
...
+        this.saveBookmark();
+      }
+    } else {
+      this.saveBookmark();
+    }
+  }
+
+  render() {
+    const { savedStatus, loadingState } = this.state;
+    const saveText = savedStatus ? "Saved" : "Save  ";
```
nathggns: Why is this saveText logic duplicated?

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,184 @@
...
+    if (loadingState) {
+      return (
+        <Fragment>
+          <Text style={styles.label}>{saveText}</Text>
+          <ActivityIndicator size="small" style={styles.activityLoader} />
+        </Fragment>
+      );
+    }
+
+    if (savedStatus) {
```
nathggns: `return this.renderSaveButton(!!savedStatus)`?

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,169 @@
...
+  async saveBookmark() {
+    this.setState({ loadingState: true });
+    const { savedArticles } = this.state;
+    const { articleId: id, saveApi } = this.props;
+
+    try {
+      await saveApi.bookmark(id);
+      this.setState({
+        loadingState: false,
+        savedArticles: [...savedArticles, id],
```
nathggns: Going to approve this but I think you can get rid of `savedArticles`, at which point you can probably merge `saveBookmark` and `unsaveBookmark`.

packages/save-star-web/README.md
```diff
@@ -0,0 +1,45 @@
+# SaveStarWeb
+
+<!-- Add the description of the package here. -->
```
ValGeorgiev: Could you add more description here, please

packages/save-star-web/README.md
```diff
@@ -0,0 +1,45 @@
...
+yarn lint
+```
+
+## Testing
+
+Testing can be done on each platform individually
+
+```
+yarn test:android
+yarn test:ios
```
ValGeorgiev: I think we don't need these commands ^

nehasri89:
@ValGeorgiev In order for the showcase and tests to work we had to mock the Api layer, and hence take it as a prop, also you would imagine the SaveApi is actually fetching the data and the point of integration between any component and data layer is usually the composing component which in our case is the `Save and Share bar`

packages/save-star-web/__tests__/shared.web.js
```diff
@@ -0,0 +1,24 @@
+import {
```
nathggns: This is a web only package – why do we have shared tests and then a separate web file? I know the cli creates that template but think it's unnecessary here.

packages/save-star-web/mock-save-api.js
```diff
@@ -0,0 +1,43 @@
+export default {
```
nathggns: This is good.

packages/save-star-web/src/save-api.js
```diff
@@ -0,0 +1,25 @@
+import {
+  getBookmarks,
+  saveBookmarks,
+  unsaveBookmarks
+} from "@times-components/provider-queries";
+import { makeClient } from "@times-components/utils";
+
+const client = makeClient();
+export default {
```
nathggns: Can we give this a name before exporting? It's better for code intelligence and stack traces.

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,146 @@
...
+import React, { Component, Fragment } from "react";
+import { ActivityIndicator, Text } from "react-native";
+import Link from "@times-components/link";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+import { IconSaveBookmark } from "@times-components/icons";
+import styles, { getStyles } from "./styles";
+
+const HoverIcon =
+  styled.div &&
```
nathggns: Why do you need this?

packages/save-star-web/src/save-star-web.js
nathggns: Why is this resolved? I worked with @nehasri89 on doing this but it doesn't seem to have been pushed.

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,146 @@
...
+        console.error("Error in connecting to api", error);
+      });
+  }
+
+  renderSaveButton(saveStatus) {
+    const { colour, hoverColour } = this.props;
+
+    const saveStyle = getStyles({ saveStatus });
+    const { fillColour, strokeColour } = saveStyle;
+    /* eslint-disable jsx-a11y/anchor-is-valid */
```
nathggns: Think this comment is no longer necessary.

packages/save-star-web/src/save-star-web.js
```diff
@@ -0,0 +1,146 @@
...
+      </Fragment>
+    );
+  }
+}
+
+SaveStarWeb.propTypes = {
+  articleId: PropTypes.string.isRequired,
+  colour: PropTypes.string,
+  hoverColour: PropTypes.string,
+  saveApi: PropTypes.shape({}).isRequired
```
nathggns: Can you specify this shape?

packages/save-star-web/README.md
```diff
@@ -0,0 +1,40 @@
+# SaveStarWeb
+
+<!-- Add the description of the package here. -->
```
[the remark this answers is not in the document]
ValGeorgiev: its added, only the comment is not removed

ValGeorgiev:
@nathggns Review comments are fixed. Please check again. I didn't change internal state tests and test file name. I will do that in a separate PR, because I want to ship this as soon as possible. Also to write it here as well. @nehasri89 refactor the api to use async/await but we faced some errors with old babel version and she reverted it to promises. We can have a look again when we bump the babel version

ValGeorgiev:
I will close this PR, because its messy. New PR: [#1945](https://github.com/newsuk/times-components/pull/1945)

PR #1907, opened 2019-04-30
fix: temporary fix for article showcase

nathggns:
Fixing the fixtures that render the article with template choice showcase. This is a bit of hack because I couldn't figure out how to make the mock provider understand the `descriptionMarkup` summary alias in the article web query.


tuncaulubilge:
Let's make sure article page checks for empty descriptionMarkup as well. We have articles with empty content and we should be defensive

PR #1912, opened 2019-05-01
chore: update readme and cli

parkjh:


package.json
```diff
@@ -49,6 +49,7 @@
     "android:publish-lib": "lerna run --scope @times-components/android-app publish-library --stream",
     "prefmt": "prettier --write '**/*.*'",
     "fmt": "lerna run fmt",
+    "install": "GRAPHQL_ENDPOINT='https://api.thetimes.co.uk/graphql' yarn",
```
tuncaulubilge: We want to keep the ability to change graphql_endpoint (for testing local tpa changes), so this should be a default value in case it's not defined, and should be done in schema package

parkjh:
@tuncaulubilge comment addressed

ValGeorgiev:
@parkjh @tuncaulubilge why we exposed the graphql prod endpoint? This sounds like a security bridge and now everyone can access it and run times-components locally :)

PR #1916, opened 2019-05-02
fix: remove ghost article flags view

parkjh:
- removes a padding on an empty articleflag element as there was an additional view component wrapped outside the flag component in the article-sumamary


packages/article-summary/src/article-summary.js
```diff
@@ -68,7 +68,7 @@ const ArticleSummary = props => {
       {headline()}
       {strapline()}
       {content()}
-      {flags() ? <View style={styles.flagWrapper}>{flags()}</View> : null}
```
parkjh: @aronshamash removes view styling here

packages/article-flag/src/style/shared.js
```diff
@@ -12,7 +12,8 @@ const styles = {
   },
   flags: {
     flexDirection: "row",
-    flexWrap: "wrap"
+    flexWrap: "wrap",
+    marginBottom: spacing(2)
```
parkjh: re added the container styling here in the flag component container

parkjh:
@tuncaulubilge hey tunca let me know when the merge freeze is finished

PR #1921, opened 2019-05-03
chore: tidy image component

parkjh:
- no functionality has changed at all
- notes that all the snapshot remains the same
- there are a lot of components within Image, which begs the question whether they should be a package of its own?


packages/image/modal-image.showcase.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable react/no-danger, react/prop-types */
 import React, { Fragment } from "react";
 import { Text, View } from "react-native";
-import ModalImage from "./src/modal-image";
+import ModalImage from "./src/modalImage";
```
jmars: We usually use snake-case

packages/image/src/safeAreaView/index.js
```diff
@@ -0,0 +1,3 @@
+import SafeArea from "./react-native-safe-area";
+
+export default SafeArea;
```
jmars: Is this necessary, just import react-native-safe-area where you use it?

packages/image/src/utils/index.js
```diff
@@ -0,0 +1,3 @@
+import appendURL from "./appendToURL";
+
+export default appendURL;
```
jmars: ?

packages/image/src/image.js
```diff
@@ -8,10 +8,10 @@ import {
   normaliseWidthForAssetRequestCache,
   convertToPixels
 } from "@times-components/utils";
-import LazyLoadingImage from "./lazy-loading-image";
+import LazyLoadingImage from "./lazy-loading-image/lazy-loading-image";
 import appendToUrl from "./utils";
 import { defaultProps, propTypes } from "./image-prop-types";
-import Placeholder from "./placeholder";
+import Placeholder from "./placeholder/placeholder";
```
ValGeorgiev: you can use only ./placeholder because you have index.js file

packages/image/src/image.web.js
```diff
@@ -1,9 +1,9 @@
 import React, { Component } from "react";
 import { View } from "react-native";
 import { addMissingProtocol } from "@times-components/utils";
-import Placeholder from "./placeholder";
+import Placeholder from "./placeholder/placeholder";
```
ValGeorgiev: same ^

PR #1927, opened 2019-05-08
BREAKING CHANGE fix: image performance

nathggns:
This PR improves the performance of our image package by making a number of changes and removing some features.

- Remove the placeholder in modal images and use the low res cached image as a placeholder instead. This is necessary in order to optimise loading the placeholder. 
- Placeholder no longer requires an `onLayout` call to load at the full size of its container. A necessary consequence of this is that placeholder can no longer be used in modal images.
- Remove the `onImageLayout` API as we are no longer calculating the information it provided
- Fix intermittent fade issues on Android. 
- Add image fade on load support on iOS
- Removed unused borderRadius property
- Supported rounded property on web and native
- Fix bug in image placeholder when using rounded property 

Potential future changes:

- Replace `react-native`'s implementation of `Image` and use `react-native-fast-image` instead. This will mean that the low res image placeholder for modal images should appear more quickly. 
- Use the offline image as a placeholder so that `onLayout` is not required to show the image. 
- Pass highResSize to image in more places so that it's not required to use onLayout to calculate this.


packages/image/src/image.js
```diff
@@ -74,43 +63,48 @@ class TimesImage extends Component {
...
 
     return (
       <View
         aspectRatio={aspectRatio}
         onLayout={this.onImageLayout}
         style={[styles.imageContainer, style]}
       >
+        <LazyLoadingImage
+          {...defaultImageProps}
+          fadeDuration={0}
```
tuncaulubilge: Can we keep the fadeIn on the images and just disable it on the placeholder? Without fadeIn, topic page looks quite bad

nathggns: We can't because it causes a bug where you see the placeholder for a moment. I'm going to reimplement this anyway.

nathggns:
This is now a WIP as I've had a few more ideas for how to improve performance. I'm also going to reimplement image fade in.

packages/image/src/image.js
nathggns: This has been reimplemented now.

packages/image/src/image.js
```diff
@@ -33,36 +32,32 @@ class TimesImage extends Component {
...
     }
-
-    if (onImageLayout) {
-      onImageLayout({ height, width });
-    }
   }
 
   handleLoad() {
-    this.setState({ isLoaded: true });
+    Animated.timing(this.fadeAnim, {
```
tuncaulubilge: Relying on the animation to update a state is not a good idea, it can cause bugs and quite a lot of headache. Let's stick with the default fade duration on Android. We shouldn't be doing such a patch just because of a few transparent images they put in there

nathggns: > Let's stick with the default fade duration on Android

We decided not to do this because it would cause issues and create a non-uniform experience. 

Instead, we've decided not to rely on the animation callback to hide the placeholder but instead used a timer, incase the animation fails.

PR #1947, opened 2019-05-14
feat: make ads 100% width on native platforms

jmars:
image


packages/ad/src/ad.js
```diff
@@ -99,7 +99,10 @@ class Ad extends Component {
       ? { height: 0, width: 0 }
       : {
           height: config.maxSizes.height,
-          width: config.maxSizes.width
+          width:
```
tuncaulubilge: We need to test this on tablet as well, with regular ads. Might be quite huge on tablets

PR #1948, opened 2019-05-14
feat: list-2-and-6-no-pic slice for huge and wide

infreezer:
Implement edition slice List Two And Six No Pic for wide and huge tablet breakpoints.

Wide and huge:

image


packages/slice-layout/src/templates/listtwoandsixnopic/styles.js
```diff
@@ -32,7 +27,36 @@ const mediumBreakpointStyles = {
...
+
+const stylesResolver = {
+  small: smallBreakpointStyles,
+  medium: mediumBreakpointStyles,
+  wide: wideBreakpointStyles,
+  huge: wideBreakpointStyles
+};
+
+export default breakpoint => ({
+  ...(stylesResolver[breakpoint] || {})
```
ValGeorgiev: `export default breakpoint => (stylesResolver[breakpoint] || {})`
I think this ^ should do the job as well

ValGeorgiev:
There is a snap conflict and one minor comment from me. Otherwise, good job

packages/slice-layout/src/templates/listtwoandsixnopic/styles.js
infreezer: Thanks.

PR #1950, opened 2019-05-15
fix: repeat 6185 interactive wrapper height in apps

parkjh:
Screenshot 2019-05-15 at 16 03 43


parkjh:
need some help testing this on device as it's difficult to duplicate the bug on emulator...

aronshamash:
@parkjh  you add screenshot for for iOS too ?

packages/interactive-wrapper/src/interactive-wrapper.js
```diff
@@ -52,8 +52,11 @@ class InteractiveWrapper extends Component {
       (e && e.nativeEvent && e.nativeEvent.data) ||
       e.nativeEvent.data === "0"
     ) {
+      const { height: prevHeight } = this.state;
       const height = Number(e.nativeEvent.data);
-      this.setState({ height });
+      if (height !== prevHeight && height > prevHeight) {
```
nathggns: This is probably a deal breaker because interactives can change height during use, and this means that throughout the lifecycle of a interactive they can get bigger, but never smaller which is probably not a good enough user experience. Maybe discuss with @chrishutchinson?

parkjh: @nathggns @chrishutchinson did not know that the interactives could do that, I will look into it again...

parkjh:
> @parkjh you add screenshot for for iOS too ?

ios app hot reloading is not working for me... :(

PR #1957, opened 2019-05-16
chore: Improve yarn install

ValGeorgiev:
Improve yarn install

add new command for full install `yarn full:install` which triggers `yarn install`, transpile and postinstall commands in all packages, fetch fonts and build schema

`yarn install`: install all packages and run transpile and postinstall commands in all packages

add more info in the contribution readme 

Introduce concurrently package to run concurrently commands and improve the speed


package.json
```diff
@@ -54,7 +54,8 @@
     "clean": "watchman watch-del-all && rm -rf $TMPDIR/metro-* && rm -rf $TMPDIR/haste-* && yarn cache clean && jest --clearCache",
     "bundle": "lerna run bundle",
     "build:local": "yarn install && yarn bundle && yarn prepublishOnly",
-    "postinstall": "yarn run fetch-fonts && lerna run postinstall && lerna run transpile",
+    "full:install": "concurrently \"yarn install\" \"yarn fetch-fonts\" \"lerna run build:schema\"",
```
nathggns: I think `full:install` is a poor name for the command, as the `:` tends to be for commands formatted as `command:modifier` not `modifier:command` – so `install:full` or `full-install` are better names

nathggns: Have you proven that `concurrently` works faster? Sometimes commands run less fast when concurrent because the CPU has to do a lot of switching between tasks which is expensive.

ValGeorgiev: i will remove full:install command, based on a discussion with Tunca. We can use --ignore-scripts

ValGeorgiev: i can add more stats

tuncaulubilge: Afaik, text layout scripts depend on fonts being present before running the python scripts on it

PR #1958, opened 2019-05-16
chore: bump styled components

ValGeorgiev:
The new styled-components is faster and lighter 

Related articles: 
https://medium.com/styled-components/styled-components-v4-new-final-finalest-for-real-final-final-psd-fa4d83398a77

https://medium.com/styled-components/announcing-styled-components-v4-better-faster-stronger-3fe1aba1a112

### DIFF
--- a/packages/article-comments/src/styles/responsive.web.js
+++ b/packages/article-comments/src/styles/responsive.web.js
@@ -7,7 +7,7 @@ import {
   spacing
 } from "@times-components/styleguide";
 
-export const CommentContainer = styled("div")`
+export const CommentContainer = styled.div`
   margin-left: auto;
   margin-right: auto;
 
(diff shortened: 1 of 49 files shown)


packages/article-comments/src/styles/responsive.web.js
```diff
@@ -7,7 +7,7 @@ import {
   spacing
 } from "@times-components/styleguide";
 
-export const CommentContainer = styled.div`
+export const CommentContainer = styled("div")`
```
nathggns: Why have you changed from `styled.div` to `styled("div")`? Docs recommend the former.

ValGeorgiev: https://medium.com/styled-components/styled-components-v4-new-final-finalest-for-real-final-final-psd-fa4d83398a77

Because of 3th step of migration here :)

nathggns: You only need to use the `styled()` form instead of the old `Component.extend`. `styled.div` is still valid and the preferred method for created styled components. Please revert these.

PR #1963, opened 2019-05-16
chore: remove webdriver

parkjh:


parkjh:
I honestly have no idea why these were installed here in the first place...

parkjh:
@tuncaulubilge @aronshamash  any objection before I merge this? I tested storybook web, and not sure what else I should be looking out for?

PR #1964, opened 2019-05-16
chore: tidy script files

parkjh:


parkjh:
this seems minor but this is a prelude to another upcoming pr...

ValGeorgiev:
What will be the upcoming PR, what is the goal here?

parkjh:
@ValGeorgiev extract the dependancies of the cli and fetch font into a separate packages, and ideally there shouldn't be any npm install for fetch font but instead use pure node api
end result = smaller (dev)dependancy of the overall monorepo

lib/scripts/publish.sh
```diff
@@ -67,4 +67,4 @@ lerna publish --conventional-commits --yes --concurrency=1 --exact -m "$MESSAGE"
 echo "Pushing to master"
 git push origin master --tags --quiet > /dev/null 2>&1
 
-./lib/publish_storybook.sh
+./lib/scripts/publish_storybook.sh
```
parkjh: @milesillsley @MarkConway94 is the release cli package looking for this script in a hardcoded path directory?

tuncaulubilge: All the CI scripts should be using yarn commands, not hardcoded paths, but please do check bitrise @parkjh

parkjh: @tuncaulubilge I've updated the all the yarn commands but I could not find any reference to this script, but if it's in bitrise I will create the yarn command to replace it

tuncaulubilge: `"storybook:publish": "./lib/publish_storybook.sh",` in the root package.json

parkjh: https://github.com/newsuk/times-components/blob/0da0a1e74d90fdb6b9c8913f4e304aeb0dc41985/package.json

parkjh: @tuncaulubilge confirmed with @milesillsley that the release cli is not used in bitrise for TC, so I will just delete this I think

milesillsley: Bitrise as no direct reference to this script, and pretty sure we don't have release cli on TC

tuncaulubilge: Does bitrise have any references to any scripts? It should be using commands but I want to make sure we don't miss anything

parkjh: @tuncaulubilge https://app.bitrise.io/app/b82f579b5213b496#/workflow no `sh` references as I can see...

parkjh: they just copied and pasted the script file content into the step itself...

tuncaulubilge: 👍

PR #1967, opened 2019-05-16
chore: Downgrade to react native 0.58.6

tuncaulubilge:
React Native 59 now bundles its own JSC for android, but we have discovered couple of ndk crashes and decided to downgrade to 0.58.6, which seems much more stable.

Had to do couple of fixes on storybook-native due to some changes in react-native-cli between these two versions. Converted storybook-native to a package as "react-native start" only works when it's run in that folder.


ValGeorgiev:
Do RN knows about this issue, if yes could you add a link to the RN issue, please?

tuncaulubilge:
@ValGeorgiev https://github.com/facebook/react-native/issues/24261

PR #1969, opened 2019-05-17
chore: Add react-scripts to fix storybook issues

tuncaulubilge:
Added react-scripts as a dev dependency to fix issues on our published storybook


parkjh:
@tuncaulubilge can we move the babel/core back to dev?

PR #1971, opened 2019-05-17
chore: improve section performance

nathggns:
The data going into the `FlatList` on section pages was being recalculated every render with an inefficient (and buggy) algorithm. I've improved the algorithm, and added some basic memoisation to the calculation, in order to improve re-render speed.


packages/section/src/utils.js
```diff
@@ -13,11 +36,14 @@ const splitPuzzlesBySlices = (puzzles, numberOfTilesPerSlice = 3) =>
     return slices;
   }, []);
 
-const buildSliceData = slices =>
-  slices.map((slice, index) => ({
-    ...slice,
-    elementId: `${slice.id}.${index}`
-  }));
+const buildSliceData = memoizeOne(slices =>
```
tuncaulubilge: Can we do this within the same iteration as the separator and memoize it once?

parkjh:
@nathggns can we also have the screenshot of section say comments as sanity check?

packages/section/src/utils.js
```diff
@@ -1,3 +1,26 @@
...
+const shouldIgnoreSeperator = ({ name }) =>
+  name === "LeadersSlice" || name === "DailyUniversalRegister";
+
+const addSeparatorData = memoizeOne(data =>
+  data.reduce((newSlices, slice, idx) => {
+    const nextSlice = data[idx + 1];
+
+    if (nextSlice && shouldIgnoreSeperator(nextSlice)) {
+      newSlices[idx] = withIgnoredSeperator(slice);
+      newSlices[idx + 1] = withIgnoredSeperator(nextSlice);
```
parkjh: you need to ignore the previous sections' leading separator, as the section name only refers to the following separator

packages/section/src/utils.js
nathggns: good point.

nathggns:
@parkjh seems to be working fine

image

packages/section/__tests__/unit/utils.test.js
```diff
@@ -39,20 +39,115 @@ describe("splitPuzzlesBySlices", () => {
...
+      { id: "f", name: "OtherSlice" },
+      { id: "g", name: "OtherSlice" },
+      { id: "h", name: "LeadersSlice" },
+      { id: "i", name: "OtherSlice" },
+      { id: "j", name: "OtherSlice" },
+      { id: "k", name: "OtherSlice" },
+      { id: "l", name: "DailyUniversalRegister" }
+    ];
+    const newData = buildSliceData(originalData);
+    expect(newData).toMatchInlineSnapshot(`
```
tuncaulubilge: This can be a regular snapshot, no need for inline

nathggns: Isn't this just personal preference? I prefer inline snapshots for anything that isn't too big (saves context switching of going into a different file). Unless you think non inlines have a benefit I'm missing?

PR #1972, opened 2019-05-17
(Replat 6411) turn on clipsubviews for flatlists

jmars:
Just adds the clipsubviews parameter to flatlists, it is off by default because it can cause some rendering issues but I don't see any on our components and it's an easy win for memory usage.


parkjh:
https://facebook.github.io/react-native/docs/flatlist#removeclippedsubviews
`Note: May have bugs (missing content) in some circumstances - use at your own risk.`

nathggns:
Think we need conventional commits on either the commits or the PR title right?

tuncaulubilge:
Title should be conventional commit, but you can change it while squashing

nehasri89:
@jmars have we tested on real device ?

jmars:
I have tested this on the beast/slow devices and it does cause any regressions

PR #1974, opened 2019-05-17
chore: Fix dev bundlers via metro config

tuncaulubilge:
This PR fixes android:app ios:app and storybook-native bundlers so they pick up dev changes instead of transpiled ones.

### DIFF
--- a/android-app/metro.config.js
+++ b/android-app/metro.config.js
@@ -1,0 +1,25 @@
+const path = require("path");
+
+module.exports = (async () => {
+  const root = path.resolve(__dirname, "../");
+  return {
+    resolver: {
+      providesModuleNodeModules: ['react-native'],
+      resolverMainFields: ['react-native', 'dev', 'browser', 'main'],
+      hasteImplModulePath: path.join(root, 'node_modules/react-native/jest/hasteImpl.js'),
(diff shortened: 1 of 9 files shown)


android-app/package.json
```diff
@@ -8,7 +8,7 @@
     "prettier:diff": "prettier --list-different '**/*.*'",
     "publish-library": "./publish.sh",
     "lint": "eslint . && yarn prettier:diff",
-    "start": "react-native start"
+    "start": "react-native start --config ../../../../android-app/metro.config.js"
```
nathggns: this path is wild

tuncaulubilge: They fixed this in the cli in RN 59, but for now, it's relative to node_modules/react-native/blabla.../cli

android-app/metro.config.js
```diff
@@ -0,0 +1,25 @@
...
+    },
+    serializer: {
+      getModulesRunBeforeMainModule: () => [
+        path.join(root, 'node_modules/react-native/Libraries/Core/InitializeCore.js')
+      ],
+      getPolyfills: require(path.join(root, 'node_modules/react-native/rn-get-polyfills')),
+    },
+    server: {},
+    projectRoot: root,
+    watchFolders: [root, path.join(root, 'node_modules'), path.join(__dirname, 'node_modules')],
```
parkjh: @tuncaulubilge are these configs from set template or react native package?

tuncaulubilge: https://stackoverflow.com/questions/53286974/which-metro-bundler-settings-are-needed-to-resolve-react-natives-internal-modul

PR #1979, opened 2019-05-20
chore: improve section performance

nathggns:
Improve section performance by:

- Preventing multiple simultaneous saved article syncing (triggering multiple re-renders)
- Improve performance of image placeholder component
- Improve performance of articule summary component

### DIFF
--- a/packages/image/src/logo/t.js
+++ b/packages/image/src/logo/t.js
@@ -1,8 +1,6 @@
 import React from "react";
 import { Image } from "react-native";
-
-// eslint-disable-next-line global-require
-const logoPath = require("../../assets/t.png");
+import logoPath from "../../assets/t.png";
 
 function T() {
   return (
--- a/packages/image/src/logo/t.js
+++ b/packages/image/src/logo/t.js
@@ -1,13 +1,15 @@
 import React from "react";
 import { Image } from "react-native";
 
+// eslint-disable-next-line global-require
+const logoPath = require("../../assets/t.png");
+
 function T() {
   return (
     <Image
       fadeDuration={0}
       resizeMode="contain"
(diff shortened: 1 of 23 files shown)


packages/image/src/logo/t.js
```diff
@@ -1,13 +1,15 @@
 import React from "react";
 import { Image } from "react-native";
 
+// eslint-disable-next-line global-require
+const logoPath = require("../../assets/t.png");
```
tuncaulubilge: Can this be just a regular import? `import logoPath from "../../assets/t.png";`

packages/related-articles/src/related-article-item.base.js
```diff
@@ -141,13 +141,13 @@ class RelatedArticleItem extends Component {
                 )
               }
               datePublicationProps={{ date: publishedTime, showDay: false }}
-              headline={() => (
```
parkjh: can we have an eslint rule to prevent this?

nathggns: Not as part of this PR – its occasionally a valid pattern too

packages/image/src/logo/t.js
```diff
@@ -1,13 +1,13 @@
...
 import { Image } from "react-native";
+import logoPath from "../../assets/t.png";
 
 function T() {
   return (
     <Image
       fadeDuration={0}
       resizeMode="contain"
-      // eslint-disable-next-line global-require
-      source={require("../../assets/t.png")}
```
ValGeorgiev: Did we test this on web?

nathggns: This file isn't used on web.

PR #1981, opened 2019-05-21
Stop animated values as well as timers onunmount as it can leak memory

jmars:
Animated values use timers internally which can cause javascript objects to be kept in the heap when unmounting a react app.

### DIFF
--- a/packages/section/src/section.js
+++ b/packages/section/src/section.js
@@ -118,7 +118,7 @@ class Section extends Component {
               }
               renderItem={this.renderItem}
               style={isTablet ? styles.tabletSpacing : null}
-              windowSize={5}
+              windowSize={2}
             />
           )}
         </ResponsiveContext.Consumer>
--- a/packages/section/src/section.js
+++ b/packages/section/src/section.js
@@ -105,7 +105,7 @@ class Section extends Component {
           {({ isTablet }) => (
             <FlatList
               data={data}
-              initialNumToRender={5}
+              initialNumToRender={isTablet ? 5 : 2}
               ItemSeparatorComponent={this.renderItemSeperator}
               keyExtractor={item => item.elementId}
               ListHeaderComponent={this.getHeaderComponent(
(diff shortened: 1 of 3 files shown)


packages/section/src/section.js
```diff
@@ -118,7 +118,7 @@ class Section extends Component {
               }
               renderItem={this.renderItem}
               style={isTablet ? styles.tabletSpacing : null}
-              windowSize={5}
+              windowSize={2}
```
tuncaulubilge: Is this intentional?

jmars: Yes, 5 means that on most sections we have all slices mounted all the time

tuncaulubilge: Yes that's the case for mobile, but tablet could fit around 3-4 slices. We already got responsiveContext in there so can we use `isTablet?5:2`

nathggns: Isn't `windowSize` just the amount to show off screen? Why would a bigger screen require a bigger windowSize?

nathggns:
Good catch mate

packages/section/src/section.js
tuncaulubilge: We fit more rows in the viewport in wider breakpoints. Mobile lays everything vertically, but tablet has horizontal tiles as well. 

`windowSize` defines how many slices we render (both onscreen and offscreen)

tuncaulubilge: Also, I think what you want to limit is actually `initialNumToRender` instead of windowSize.
windowSize is the "max items" that can be rendered but `initialNumToRender` adjust the minimum number that'll be kept in memory:

https://facebook.github.io/react-native/docs/virtualizedlist#initialnumtorender
```
Note these items will never be unmounted as part of the windowed rendering in order to improve perceived performance of scroll-to-top actions.
```
we can do `isTablet ? 5 : 2` for it as well

tuncaulubilge: Had a chat with Nate and we made a quick investigation. In summary, ignore my previous message, looks like `windowSize` is the number of screens that's pre-rendered, not items (slices). So I suggest we use 3 for windowSize, (1 onscreen, 1 above and 1 below)
and for initialNumToRender we can go for 5 on tablet and 2 on mobile

jmars: Agreed, will make the change

PR #1985, opened 2019-05-22
fix: Replace webview with community lib

tuncaulubilge:
React Native deprecated webviews and suggests using the community alternative. There seems to be some performance issues and weird layout problems with the webview in 0.58.6, so this PR aims to replace it with the community version: https://github.com/react-native-community/react-native-webview


android-app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@times-components/android-app",
-  "version": "0.14.69",
+  "version": "1.0.0",
```
tuncaulubilge: Test purposes only, will remove

android-app/build.gradle
```diff
@@ -41,7 +41,7 @@ task removePreviousReactArchives(type: Delete) {
 task generateReactArchives {
     subprojects { subproject ->
         apply plugin: 'maven'
-        def reactProjects = ["react-native-svg"]
+        def reactProjects = ["react-native-svg", "react-native-webview"]
```
tuncaulubilge: List of native deps that get bundled with our xnative lib

android-app/settings.gradle
```diff
@@ -2,3 +2,6 @@ include ':xnative'
 
 include ':react-native-svg'	
 project(':react-native-svg').projectDir = new File(settingsDir, '../node_modules/react-native-svg/android')
+
+include ':react-native-webview'
```
tuncaulubilge: Adds the native dep to library project

android-app/xnative/build.gradle
```diff
@@ -78,4 +78,7 @@ dependencies {
     api("react-repo:react-native-svg:${packageInfo.dependencies.'react-native-svg'}") {	
         exclude group: "com.facebook.react"	
     }
+    api("react-repo:react-native-webview:${packageInfo.dependencies.'react-native-webview'}") {	
+        exclude group: "com.facebook.react"	
+    }
```
tuncaulubilge: Adds the native dep to library project
`api` means the consumers of this xnative library can also access webview library

android-app/build.gradle
tuncaulubilge: This is specific to our implementation and bundling/publishing flow

nehasri89:
LGTM 👍 , just wanted to check one thing, there was some issues with interactives not loading properly for which we had fix with postMessageBugWorkAround, just wanted to check we have tested interactives for those issues.

tuncaulubilge:
@nehasri89 yeah, this library has its own post message implementation (`window.ReactNativeWebView.postMessage`) and it seems to work without our fixes. I've tested ads and an article with multiple interactives and it's all fine

android-app/xnative/src/main/java/uk/co/thetimes/xnative/XNative.java
```diff
@@ -0,0 +1,18 @@
...
+import com.facebook.react.ReactPackage;
+import com.horcrux.svg.SvgPackage;
+import com.reactnativecommunity.webview.RNCWebViewPackage;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class XNative {
+    public static List<ReactPackage> getPackages() {
+        List<ReactPackage> packages = new ArrayList<>();
```
ChristianGarcia: Does it need to be mutable? (i.e. is any other calling code adding packages to this list?).

Asking not only for correctness but also in case we Kotlinize this in the future.

tuncaulubilge: It can be immutable. I haven't bothered to Kotlinize the library as the java code there is super minimal, but it can be easily done

tuncaulubilge: Done :+1:

ChristianGarcia: Looking good!
Just FYI, Java already supports immutable lists through `Collections.unmodifiableList()`
```
Collections.unmodifiableList(Arrays.asList(
                new ComponentsPackage(),
                new SvgPackage(),
                new RNCWebViewPackage()
        )
)
```
but yeah, it's a bit more verbose than the one you proposed. Happy to leave it as is

PR #1988, opened 2019-05-23
Feat Add tracking to save star press event

GaboDimitrov:
Add tracking to save star press event. Only articleId is going to be tracked. Save star tests are refactored.

### DIFF
--- a/packages/save-and-share-bar/src/save-and-share-bar.js
+++ b/packages/save-and-share-bar/src/save-and-share-bar.js
@@ -17,12 +17,44 @@ class SaveAndShareBar extends Component {
   constructor(props) {
     super(props);
     this.copyToClipboard = this.copyToClipboard.bind(this);
+    this.onSaveButtonPress = this.onSaveButtonPress.bind(this);
   }
 
   copyToClipboard() {
     const { onCopyLink, articleUrl } = this.props;
     Clipboard.setString(articleUrl);
     onCopyLink();
+  }
+
+
+  onSaveButtonPress(evt) {
+    evt.preventDefault();
+
+    const { savedStatus } = this.state;
+    const { saveApi } = this.props;
+
+    if (savedStatus) {
+      this.saveUnsaveBookmark(saveApi.unBookmark, false, true);
+    } else {
+      this.saveUnsaveBookmark(saveApi.bookmark, true, false);
+    }
+  }
+
+  saveUnsaveBookmark(saveMethod, successStatus, errorStatus) {
+    this.setState({ loadingState: true });
+    const { articleId: id } = this.props;
+
+    saveMethod(id)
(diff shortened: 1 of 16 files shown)


packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -25,6 +26,37 @@ class SaveAndShareBar extends Component {
     onCopyLink();
   }
 
+
+  onSaveButtonPress(evt) {
```
nehasri89: why do we need to copy code from `save-star`, we could have done the tracking in `save-star`

packages/save-star-web/__tests__/shared.base.js
```diff
@@ -1,212 +1,51 @@
...
-        const apiMock = {
-          ...mockSaveApi,
-          getBookmarks: () =>
-            Promise.resolve({
-              loading: true
-            })
-        };
-
-        const testInstance = TestRenderer.create(
-          <SaveStarWeb
```
nehasri89: these scenarios need to be tested, I don't see these being added anywhere else

ValGeorgiev: They are obsolete. As we discuss with Nate, in the first PR, this testing of state is not quite right. We have test for save, unsave and loading scenario. I think that is enough. What you think?

packages/save-and-share-bar/src/save-and-share-bar.js
ValGeorgiev: the tracking is done in SaveStar, could you suggest how to change the tracking to not have onSaveButtonPress in save-and-share-bar ? The tracking looks to works in that way

packages/save-star-web/__tests__/shared.base.js
nehasri89: okay, I understand your point. There was a bug in where the incorrect apis where being invoked, for which I have put few tests. Those tests are pretty imp I think

packages/save-star-web/src/save-star-web.js
```diff
@@ -59,14 +52,13 @@ class SaveStarWeb extends Component {
       })
       .catch(error => {
         this.setState({ loadingState: false, savedStatus: false });
-        console.error("Error in connecting to api", error);
+        console.error(error);
```
IvayloAndonov: _Not related with the PR_

**Disclaimer**: The 'console' is useful only for testing purposes. Has our webpack been configured to remove all 'console' logs across the code on production env ?  Better approach is to use some logger here.

ValGeorgiev: @IvayloAndonov this is very useful ^, could you raise it on the next tech dept session, please?

PR #1989, opened 2019-05-23
REPLAT-6157: Added a new publish script for ios assets

udaypandey:
Why is this needed?
-------------------

- Adds a new publish script for ios assets. The git repo is pointing to a temporary repo right now and will be updated to the right path once the new becomes available.
-  Adds the script to package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "ios": "react-native run-ios --no-packager",
     "ios:logs": "yarn ios && react-native log-ios",
     "ios:app": "cd ios-app && yarn start",
+    "ios:publish-lib": "yarn ios:build-lib && ./lib/publish-ios-assets.sh",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,6 +33,7 @@
     "ios": "react-native run-ios --no-packager",
     "ios:logs": "yarn ios && react-native log-ios",
     "ios:app": "cd ios-app && yarn start",
+    "ios:publish-lib": "yarn ios:build-lib && ./lib/publish-ios-assets.sh",
```
tuncaulubilge: we can keep `ios:build-lib` and publish-lib as separate scripts and call them one after each other on bitrise

udaypandey: Decoupled publish-lib and build-lib scripts.

tuncaulubilge:
If this ready to be merged? Do we need some secret keys in bitrise?

PR #1991, opened 2019-05-23
feat: Revert save icon to save star icon

Ivanov558:
Jira: https://nidigitalsolutions.jira.com/browse/REPLAT-6450
Revert save icon to save star icon.

image


packages/save-and-share-bar/src/styles/index.js
```diff
@@ -41,6 +41,9 @@ const styles = {
     fb: {
       height: 18
     },
+    star: {
+      height: 18
```
ValGeorgiev:
```suggestion
      height: 19
```

PR #1993, opened 2019-05-23
Feat/android offline images

nehasri89:
This PR introduces the new fields introduced in `crop` in Media to be available for offline images

PR can be reviewed but needs to wait for TPA to get merge


packages/article-lead-asset/src/article-lead-asset.js
```diff
@@ -12,9 +12,29 @@ import getRatio from "./get-ratio";
...
+      aspectRatio,
+      caption,
+      crop,
+      onImagePress,
+      uri,
+      relativeWidth,
+      relativeHeight,
+      relativeHorizontalOffset,
+      relativeVerticalOffset
+    }}
```
ValGeorgiev: What do you think about this:

`const ArticleLeadAssetModalImage = props => (<ModalImage {...props} index={0} />)`

packages/article-skeleton/__tests__/android/__snapshots__/article.android.test.js.snap
```diff
@@ -116,6 +124,10 @@ exports[`1. a full article with all content items with dropcap template 1`] = `
                 Object {
                   "display": "secondary",
                   "ratio": "1500:1000",
+                  "relativeHeight": undefined,
+                  "relativeHorizontalOffset": undefined,
+                  "relativeVerticalOffset": undefined,
+                  "relativeWidth": undefined,
```
ValGeorgiev: is this intentional all these fields to be undefined?

packages/image/src/lazy-loading-image/lazy-loading-image.android.js
```diff
@@ -1,44 +1,42 @@
...
-          />
-        ) : null}
+  const queryArray = [
+    ["offline", true],
+    ["rel_height", relativeHeight],
+    ["rel_width", relativeWidth],
+    ["rel_horizontal_offset", relativeHorizontalOffset],
+    ["rel_vertical_offset", relativeVerticalOffset]
+  ];
+  const queryParamMap = new Map(queryArray);
```
ValGeorgiev: why new Map ?

packages/article-lead-asset/src/article-lead-asset.js
nathggns: `ArticleLeadAssetModalImage` specifically chooses which props get passed through, because some unnecessary props were gettting passed through that were causing issues (possible name clashes I think) .

packages/article-skeleton/__tests__/android/__snapshots__/article-header-with-style.android.test.js.snap
```diff
@@ -195,6 +195,10 @@ exports[`1. article with header 1`] = `
                 Object {
                   "display": "primary",
                   "ratio": "1500:1000",
+                  "relativeHeight": undefined,
```
nathggns: Is this intentional?

packages/article-lead-asset/src/article-lead-asset.js
ValGeorgiev: Thats why we shouldn't do this anti-pattern 
`
 const LeadAsset = isVideo
    ? ArticleLeadAssetVideo
    : ArticleLeadAssetModalImage;
`

nathggns: I don't necessarily disagree.

packages/article-skeleton/__tests__/android/__snapshots__/article.android.test.js.snap
nehasri89: yup, so if you do not pass these fields in `Article-Skeleton` which is not mandatory as its only for android it would be undefined, but in ideal world it will never be undefined.

packages/article-lead-asset/src/article-lead-asset.js
nehasri89: yup, Agree with both that one for this we need to pass specific props, not all and two it is due to the way it is build

packages/article-skeleton/__tests__/android/__snapshots__/article-header-with-style.android.test.js.snap
nehasri89: same as above

PR #1995, opened 2019-05-24
fix: save/share bar appearance when sticky; fix performance of becomi…

nathggns:
…ng sticky

image

image


packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -46,31 +46,37 @@ class ArticleSkeleton extends Component {
...
   }
 
   componentWillUnmount() {
-    if (window && window.nuk.user.isLoggedIn) {
-      window.removeEventListener("scroll", this.handleScroll);
-    }
+    window.removeEventListener("scroll", this.handleScroll);
   }
 
+  // Not using state for performance
```
parkjh: 👍

packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -46,31 +46,37 @@ class ArticleSkeleton extends Component {
       articleWidth: this.node && this.node.clientWidth
     });
 
-    if (window && window.nuk.user.isLoggedIn) {
-      window.addEventListener("scroll", this.handleScroll);
-    }
+    window.addEventListener("scroll", this.handleScroll);
```
ValGeorgiev: Please don't remove the checks here, there is no need to have these events when user is not logged in.

packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -46,31 +46,37 @@ class ArticleSkeleton extends Component {
...
   }
 
   componentWillUnmount() {
-    if (window && window.nuk.user.isLoggedIn) {
-      window.removeEventListener("scroll", this.handleScroll);
-    }
+    window.removeEventListener("scroll", this.handleScroll);
```
ValGeorgiev: Same here ^

packages/article-skeleton/src/article-skeleton.web.js
ValGeorgiev: Good job, but could we delete this comment, please

nathggns: This comment is important so that down the line people understand why it's been implemented with imperative DOM APIs instead of state.

packages/article-skeleton/src/article-skeleton.web.js
nathggns: okay.

packages/article-skeleton/src/article-skeleton.web.js
nathggns: You don't need the check on unmount as its a noop if the event isn't attached – and i think its actually a potential source of a memory leak too (unlikely, but I can see it happening). Safer to always remove.

packages/article-skeleton/src/article-skeleton.web.js
nathggns: Going to fix this by abstracting the attachment code to a new component so its only ever triggered if its actually one screen – means we're not duplicating the conditions and removes the chance of bugs where the CSS stickyness applies without the javascript applying the class. (I've seen this happen)

packages/article-skeleton/src/attached-save-and-share.web.js
```diff
@@ -0,0 +1,60 @@
...
+    const sticky = this.stickyRef.current;
+
+    if (sticky) {
+      const offsetTop = sticky.getBoundingClientRect().top;
+      const isSticky = offsetTop <= 1;
+
+      if (isSticky !== this.isSticky) {
+        this.isSticky = isSticky;
+
+        if (isSticky) {
```
apostolnikov: It's hard to read with those 3 nested if statements. Can we just go with something like 
isSticky 
  ? sticky.classList.add("isSticky")
  : sticky.classList.remove("isSticky");

packages/article-skeleton/src/attached-save-and-share.web.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  checkAttachment() {
+    const sticky = this.stickyRef.current;
+
+    if (sticky) {
+      const offsetTop = sticky.getBoundingClientRect().top;
+      const isSticky = offsetTop <= 1;
+
+      if (isSticky !== this.isSticky) {
```
apostolnikov: Why do we need to set isSticky as a class field? Can we get rid of this if statement?

nathggns: Because `classList` will be accessed `add` and `remove` will be triggered on every scroll which isn't performant.

packages/article-skeleton/src/attached-save-and-share.web.js
nathggns: I find that much harder to read personally. Ternaries are usually found as part of an expression, not as part of a statement, which means it can be weird to find them like that. They're also harder to read than if statements so we should only use them if they provide value, and it doesn't here.

packages/article-skeleton/src/attached-save-and-share.web.js
apostolnikov: Did you test the performance difference, because it's really a hack right now...

nathggns: Yes I did test the performance – I originally didn't have it and noticed the scroll lag which is why I introduced this.

I don't understand why you believe it to be a hack. It's just a javascript instance property. It's keeping a record of whether the class is added or not. It's no different to it being in `this.state`, except its more performant as changing it doesn't trigger a re-render. This is a pretty common technique for keeping state in React components that you don't use it a render. The image component does the same thing with a `fadeAnimTimeout` field.

packages/article-skeleton/src/attached-save-and-share.web.js
apostolnikov: Both syntaxes have their own value, and to use them both is to get the best from both worlds which is always better. Three nested blocks with indentation are really harder to read and to it's better to shorten it by representing the last if which is 5 lines of logic with only one...

nathggns: Thanks for your suggestion but as explained above, its an unnecessary change that in my belief makes it harder to read. Therefore I won't be making the change at this time.

nathggns: I've explained why it doesn't provide value to change this.

packages/article-skeleton/src/article-skeleton.web.js
tuncaulubilge: Can we use a prop to check if user is logged in, instead of accessing window.nuk in here?

packages/article-skeleton/src/sticky-save-and-share.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  checkSticky() {
+    const sticky = this.stickyRef.current;
+
+    if (sticky) {
+      const offsetTop = sticky.getBoundingClientRect().top;
+      const isSticky = offsetTop <= 1;
+
+      if (isSticky !== this.isSticky) {
```
tuncaulubilge: We could move this check to a simple function to make it more readable. Something like `isStickyAdded = isSticky => isSticky === this.isSticky` maybe?

packages/article-skeleton/src/article-skeleton.web.js
nathggns: This is outdated. I've now abstracted the attachment code to a new component which is only rendered if the user is logged in, and that's checked using context. https://github.com/newsuk/times-components/pull/1995/files?file-filters%5B%5D=.js#diff-a1b8d9ffc37eb7f18a77636d89c19bbcR121

packages/article-skeleton/src/sticky-save-and-share.js
nathggns: Good idea. Decided to break this class down further.

PR #2002, opened 2019-05-30
fix: replat 6625 android savestar

parkjh:


packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -38,7 +38,7 @@ class TileSummary extends Component {
       starStyle
     } = this.props;
 
-    if (!withStar || isSaveSupported) {
+    if (!withStar || !isSaveSupported) {
```
ValGeorgiev: I cannot see how this change will introduce these changes of the snapshots :/

PR #2003, opened 2019-05-30
fix: message bar consumed and styled in web

infreezer:
[Jira link](https://nidigitalsolutions.jira.com/browse/REPLAT-6594)

Message bar consumed by save and share bar and trigger with click on copy link.
Fixed the component to unmount after 3 seconds.
Styling for all breakpoints applied.

Huge:
huge-message-bar-render

Wide:
wide-message-bar-render

Medium:
medium-message-bar-render

Small:
small-message-bar-render

### DIFF
--- a/packages/message-bar/__tests__/ios/__snapshots__/message-bar.test.js.snap
+++ b/packages/message-bar/__tests__/ios/__snapshots__/message-bar.test.js.snap
@@ -3,88 +3,15 @@
 exports[`1. renders correctly 1`] = `
 <View>
   <View>
-    <Text>
-      test message
-    </Text>
     <View>
+      <Text>
+        test message
+      </Text>
       <View>
-        <ARTSurfaceView>
-          <ARTGroup
-            opacity={1}
-            transform={
-              Array [
-                1,
-                0,
-                0,
-                1,
-                0,
-                0,
-              ]
-            }
-          >
-            <ARTShape
-              d={
-                Array [
-                  0,
-                  15.617,
-                  14,
-                  2,
-                  20.3,
-                  19.838,
-                  2,
-                  NaN,
-                  NaN,
-                  2,
-                  NaN,
-                  NaN,
-                  2,
-                  NaN,
-                  NaN,
-                  2,
-                  NaN,
-                  NaN,
-                  2,
-                  12.383,
-                  14,
-                  2,
-                  7.7,
-                  8.162,
-                  2,
-                  8.162,
-                  7.700000000000001,
-                  2,
-                  14,
-                  12.383,
-                  2,
-                  19.838,
-                  7.7,
-                  2,
-                  NaN,
-                  NaN,
-                ]
-              }
-              fill={
-                Array [
-                  0,
-                  1,
-                  1,
-                  1,
-                  1,
-                ]
-              }
+        <View>
+          <ARTSurfaceView>
+            <ARTGroup
               opacity={1}
-              stroke={
-                Array [
-                  1,
-                  1,
-                  1,
-                  1,
-                ]
-              }
-              strokeCap={1}
-              strokeDash={null}
-              strokeJoin={1}
-              strokeWidth={1}
               transform={
                 Array [
                   1,
@@ -95,9 +22,84 @@ exports[`1. renders correctly 1`] = `
                   0,
                 ]
               }
-            />
-          </ARTGroup>
-        </ARTSurfaceView>
+            >
+              <ARTShape
+                d={
+                  Array [
+                    0,
+                    15.617,
+                    14,
+                    2,
+                    20.3,
+                    19.838,
+                    2,
+                    NaN,
+                    NaN,
+                    2,
+                    NaN,
+                    NaN,
+                    2,
+                    NaN,
+                    NaN,
+                    2,
+                    NaN,
+                    NaN,
+                    2,
+                    12.383,
+                    14,
+                    2,
+                    7.7,
+                    8.162,
+                    2,
+                    8.162,
+                    7.700000000000001,
+                    2,
+                    14,
+                    12.383,
+                    2,
+                    19.838,
+                    7.7,
+                    2,
+                    NaN,
+                    NaN,
+                  ]
+                }
+                fill={
+                  Array [
+                    0,
+                    1,
+                    1,
+                    1,
+                    1,
+                  ]
+                }
+                opacity={1}
+                stroke={
+                  Array [
+                    1,
+                    1,
+                    1,
+                    1,
+                  ]
+                }
+                strokeCap={1}
+                strokeDash={null}
+                strokeJoin={1}
+                strokeWidth={1}
+                transform={
+                  Array [
+                    1,
+                    0,
+                    0,
+                    1,
+                    0,
+                    0,
+                  ]
+                }
+              />
+            </ARTGroup>
+          </ARTSurfaceView>
+        </View>
       </View>
     </View>
   </View>
--- a/packages/message-bar/src/message-bar.js
+++ b/packages/message-bar/src/message-bar.js
@@ -18,18 +18,20 @@ class MessageBar extends Component {
   }
 
   componentDidMount() {
-    const { delay } = this.props;
+    const { delay, close } = this.props;
 
     this.animateOpen(() => {});
     this.setState({
       timeout: setTimeout(() => {
-        this.animateClosed();
+        this.animateClosed(() => {
+          close();
+        });
       }, delay)
     });
   }
 
   componentWillReceiveProps(props) {
-    const { message, delay } = props;
+    const { message, delay, close } = props;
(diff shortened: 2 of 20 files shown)


packages/message-bar/src/styles/shared.js
```diff
@@ -2,40 +2,58 @@ import { colours, fontFactory } from "@times-components/styleguide";
...
+  flexGrow: 1,
+  flexShrink: 0,
+  marginLeft: "auto",
+  marginRight: "auto"
+};
+
+const messageBarCloseButton = {
+  alignItems: "center",
+  backgroundColor: "rgba(255, 255, 255, 0.3)",
+  borderRadius: 28 / 2,
```
ValGeorgiev: why not 14 ?

packages/message-bar/src/styles/shared.js
```diff
@@ -2,40 +2,58 @@ import { colours, fontFactory } from "@times-components/styleguide";
...
+  height,
+  flexGrow: 1,
+  flexShrink: 0,
+  marginLeft: "auto",
+  marginRight: "auto"
+};
+
+const messageBarCloseButton = {
+  alignItems: "center",
+  backgroundColor: "rgba(255, 255, 255, 0.3)",
```
ValGeorgiev: this should be in colours file

packages/message-bar/src/styles/shared.js
```diff
@@ -2,40 +2,58 @@ import { colours, fontFactory } from "@times-components/styleguide";
...
+const messageBarBody = {
+  maxWidth: 1182,
+  width: "100%",
+  flexDirection: "row",
+  alignItems: "center",
+  height,
+  flexGrow: 1,
+  flexShrink: 0,
+  marginLeft: "auto",
+  marginRight: "auto"
```
ValGeorgiev: you can use marginHorizontal here

packages/message-bar/src/styles/shared.js
```diff
@@ -2,40 +2,58 @@ import { colours, fontFactory } from "@times-components/styleguide";
 
 const height = 50;
 
-const sharedStyle = scale => ({
-  messageBarBody: {
-    alignItems: "center",
+const messageBarBody = {
+  maxWidth: 1182,
```
ValGeorgiev: do we need this line ?

packages/message-bar/src/styles/shared.js
```diff
@@ -2,40 +2,58 @@ import { colours, fontFactory } from "@times-components/styleguide";
...
+};
+
+const messageBarCloseButton = {
+  alignItems: "center",
+  backgroundColor: "rgba(255, 255, 255, 0.3)",
+  borderRadius: 28 / 2,
+  height: 28,
+  justifyContent: "center",
+  marginLeft: "auto",
+  marginRight: 20,
```
ValGeorgiev: marginLeft: auto and marginRight 20, this doesn't look right

packages/message-bar/src/styles/shared.js
```diff
@@ -2,40 +2,58 @@ import { colours, fontFactory } from "@times-components/styleguide";
...
       width: 0
     },
     shadowRadius: 5,
-    flexGrow: 1,
-    flexShrink: 0,
     zIndex: 10
   },
+  messageBarBody: {
+    ...messageBarBody
+  },
```
ValGeorgiev: this can be only `messageBarBody`

packages/message-bar/src/styles/shared.js
```diff
@@ -2,40 +2,58 @@ import { colours, fontFactory } from "@times-components/styleguide";
...
   messageBarCloseButton: {
-    alignItems: "center",
-    backgroundColor: "rgba(255, 255, 255, 0.3)",
-    borderRadius: 28 / 2,
-    height: 28,
-    justifyContent: "center",
-    marginLeft: "auto",
-    marginRight: 20,
-    width: 28
+    ...messageBarCloseButton
```
ValGeorgiev: Same ^

packages/message-bar/src/message-manager.js
```diff
@@ -60,13 +61,20 @@ class MessageManager extends Component {
...
         <View style={[styles.messageManager, offsetStyle]}>
           {message && (
-            <MessageBar
-              animate={animate}
-              close={this.removeMessage}
-              delay={delay}
-              message={message}
-              scale={scale}
-            />
+            <Responsive>
```
ValGeorgiev: Why do we need <Responsive> here?

packages/message-bar/src/styles/shared.js
infreezer: @jmars Can you advise why not 14?
Meanwhile I will update it to 14.

packages/message-bar/src/styles/shared.js
infreezer: Done.

packages/message-bar/src/styles/shared.js
infreezer: Good point. Done.

packages/message-bar/src/styles/shared.js
infreezer: Thanks. Applied.

packages/message-bar/src/styles/shared.js
infreezer: Done.

packages/message-bar/src/styles/shared.js
infreezer: Can you advise how to make it better

packages/message-bar/src/message-manager.js
infreezer: Because we don't have Responsive Context Provider as parent to consume
and editionBreakpoint is undefined without <Responsive>

packages/message-bar/src/styles/shared.js
infreezer: This is the max width of the inner container. We need it in order the text to be aligned on huge breakpoint.

packages/message-bar/src/message-manager.js
infreezer: I removed the responsive component from the message manager
and moved the manager inside the Resposive component on Article level.
Thanks.

packages/message-bar/src/message-bar.js
```diff
@@ -18,26 +18,30 @@ class MessageBar extends Component {
...
   componentWillReceiveProps(props) {
-    const { message, delay } = props;
+    const { message, delay, close } = props;
     const { message: oldMessage } = this.props;
 
     if (message === oldMessage) {
       const { timeout } = this.state;
       clearTimeout(timeout);
       this.setState({
         timeout: setTimeout(() => {
```
tuncaulubilge: We need to clear this timeout in componentUnmount as well

infreezer: Good point. Thanks

nathggns:
Reviewing now.

packages/message-bar/__tests__/ios/__snapshots__/message-bar.test.js.snap
```diff
@@ -95,9 +22,84 @@ exports[`1. renders correctly 1`] = `
...
+                    NaN,
+                    2,
+                    NaN,
+                    NaN,
+                    2,
+                    NaN,
+                    NaN,
+                    2,
+                    NaN,
+                    NaN,
```
nathggns: What's the NaN's caused by?

packages/message-bar/src/message-bar.js
```diff
@@ -18,31 +18,42 @@ class MessageBar extends Component {
   }
 
   componentDidMount() {
-    const { delay } = this.props;
+    const { delay, close } = this.props;
 
     this.animateOpen(() => {});
     this.setState({
       timeout: setTimeout(() => {
```
nathggns: Why is the timeout stored in the state? This will trigger a seemingly unnecessary re-render.

nathggns: Just store it in an instance variable – `this.timeout = …`.

packages/message-bar/src/message-bar.js
```diff
@@ -18,31 +18,42 @@ class MessageBar extends Component {
...
   componentWillReceiveProps(props) {
-    const { message, delay } = props;
+    const { message, delay, close } = props;
     const { message: oldMessage } = this.props;
 
     if (message === oldMessage) {
       const { timeout } = this.state;
       clearTimeout(timeout);
       this.setState({
         timeout: setTimeout(() => {
```
nathggns: Again, you don't need to store the timeout in state.

packages/message-bar/src/message-bar.js
infreezer: Sure. Thanks

packages/message-bar/src/message-bar.js
infreezer: I'll fix it in a minute.

PR #2004, opened 2019-05-30
feat: use paidContentClassName from Render when loading paywalled content

Ivaylo1987:
This is for [jira](https://nidigitalsolutions.jira.com/browse/REPLAT-4399)

Google would index paywalled content based on a css class name value defined in the json-dl schema `"cssSelector": ".class"`.
This value is set in the render config files and needs to be passed down to SSR/TC so React pages can be loaded with the correct className set on paywalled paragraphs.

_Note_: This PR has a corresponding PR in `Render` -> [cps-content-render](https://github.com/newsuk/cps-content-render/pull/2757)


Ivaylo1987:
> Little bit wary just cause there's no tests (and the standalone server implementation in SSR doesn't even use it, had to manually add it to test it myself), but also its not a huge change so its probably fine. Add tests if you think they're necessary.

Yep, you are right. Will add a test in TC to follow this.

nathggns:
@apostolnikov this one needs to be merged before the render one does, as the render one will depend on the version bump to SSR that merging this will trigger.

PR #2005, opened 2019-05-30
fix: save share bar should stick below fixed siblings

nathggns:
As there is a fixed navigation bar on render, we need the save share bar to fix below any fixed position siblings. To do that, we now fix the save share bar at the distance from the top of the body to the top of the react root, which should actually for any fixed navigation bars.

To do this, the solution has become sufficiently complex that it should really be its own package, so I have created a generic `@times-components/sticky` package, which is relatively simple to use as shown by the showcases. 

I have also updated SSR to include a "mock" navigation bar so you can see how this will interact with it. 

The component is quite simple, you can simply place `<Sticky />` components (which by default are divs) anywhere and it'll stick to the top of your screen, and add `sticky` class when it has become sticky. You can easily change the component it renders using the `Component` prop and it also supports receiving `className` and `style` props. 

If you need the sticky component to make space for a fixed sibling (like a fixed navigation bar) you need to place a `<StickyProvider>` as close to the top of your component tree as you can (the higher it is, the more likely it is to calculate the correct top offset). 

May-30-2019 17-35-43


packages/article-skeleton/src/styles/responsive.web.js
```diff
@@ -90,6 +89,8 @@ export const SaveShareContainer = styled.div`
   }
 
   @media (min-width: ${breakpoints.huge + 1}px) {
+    position: static !important;
```
parkjh: what does this do? and is this required for the sticky package?

packages/sticky/src/sticky.js
```diff
@@ -0,0 +1,117 @@
...
+    window.addEventListener("resize", this.updateSticky);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener("scroll", this.updateSticky);
+    window.removeEventListener("resize", this.updateSticky);
+  }
+
+  setSticky(container, shouldBeSticky) {
+    const { stickyClassName } = this.props;
```
parkjh: no need to bind this?

packages/article-skeleton/src/styles/responsive.web.js
nathggns: Its required to prevent the bar becoming sticky on widescreen devices.

packages/sticky/src/sticky.js
nathggns: No.

packages/sticky/src/context.js
```diff
@@ -0,0 +1,61 @@
...
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import getDisplayName from "react-display-name";
+import hoistNonReactStatic from "hoist-non-react-statics";
+
+import { getTop } from "./util";
+
+const { Provider, Consumer } = React.createContext({ top: 0 });
+
+export class StickyProvider extends Component {
```
ValGeorgiev: Could we export it at the end of the file, with the other function?

packages/ssr/src/component/author-profile.js
```diff
@@ -21,37 +22,41 @@ module.exports = (client, analyticsStream, data, helmetContext) => {
   } = data;
 
   return React.createElement(
-    HelmetProvider,
-    { context: helmetContext },
+    StickyProvider,
```
ValGeorgiev: Do we need StickyProvider on author-profile and topic pages?

nathggns: Yeah – means we can use sticky wherever we want without having to worry.

nathggns: It doesn't really do very much so there's not a whole lot of harm adding it.

packages/sticky/src/context.js
nathggns: 👍

PR #2006, opened 2019-05-31
Feat Add jsonLD to Helmet on article pages

GaboDimitrov:
Json LD is removed from render and added to TC. This is done, because we need to add some data from TPA to json LD in order to improve the SEO.

Render PR: https://github.com/newsuk/cps-content-render/pull/2759

### DIFF
--- a/packages/article-in-depth/__tests__/web/__snapshots__/article.web.test.js.snap
+++ b/packages/article-in-depth/__tests__/web/__snapshots__/article.web.test.js.snap
@@ -57,6 +57,11 @@ exports[`3. an article with no headline falls back to use shortheadline 1`] = `
       content="https://crop169.io"
       name="twitter:image"
     />
+    <script
+      type="application/ld+json"
+    >
+      {"@context":"http://schema.org","@type":"NewsArticle","headline":"Some Short Headline","publisher":{"@type":"Organization","name":"The Times"},"mainEntityOfPage":{"@type":"WebPage","@id":"https://www.thetimes.co.uk/article/this-is-slug-2k629tpvh"},"dateCreated":"2015-03-13T18:54:58.000Z","datePublished":"2015-03-13","isAccessibleForFree":false,"hasPart":{"@type":"WebPageElement","isAccessibleForFree":false,"cssSelector":".undefined"},"image":{"@type":"ImageObject","url":"https://crop169.io","caption":"Some Caption"},"thumbnailUrl":"https://crop169.io"}
+    </script>
   </Helmet>
   <div>
     <Ad
@@ -265,6 +270,11 @@ exports[`4. an article with ads 1`] = `
       content="https://crop169.io"
       name="twitter:image"
     />
+    <script
+      type="application/ld+json"
+    >
+      {"@context":"http://schema.org","@type":"NewsArticle","headline":"Some Headline","publisher":{"@type":"Organization","name":"The Times"},"mainEntityOfPage":{"@type":"WebPage","@id":"https://www.thetimes.co.uk/article/this-is-slug-2k629tpvh"},"dateCreated":"2015-03-13T18:54:58.000Z","datePublished":"2015-03-13","isAccessibleForFree":false,"hasPart":{"@type":"WebPageElement","isAccessibleForFree":false,"cssSelector":".undefined"},"image":{"@type":"ImageObject","url":"https://crop169.io","caption":"Some Caption"},"thumbnailUrl":"https://crop169.io"}
+    </script>
   </Helmet>
   <div>
     <Ad
--- a/packages/article-skeleton/__tests__/web/__snapshots__/head.web.test.js.snap
+++ b/packages/article-skeleton/__tests__/web/__snapshots__/head.web.test.js.snap
@@ -64,6 +64,11 @@ exports[`Head outputs correct metadata 1`] = `
     content="https://crop169.io"
     name="twitter:image"
   />
+  <script
+    type="application/ld+json"
+  >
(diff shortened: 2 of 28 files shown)


packages/article-skeleton/src/head.web.js
```diff
@@ -5,6 +5,9 @@ import PropTypes from "prop-types";
 import Context from "@times-components/context";
 import { renderTreeAsText } from "@times-components/markup-forest";
 
+const FAVICON_URL =
+  "https://www.thetimes.co.uk/d/img/icons/icon_1_5x-a13ee0bf9a.png";
```
chrishutchinson: This looks like a hashed URL to me `...-a13ee0bf9a`. Are we confident it won't change when render next releases? I'd argue this file should be part of `times-components` going forward. This could break outside of TC's control

nathggns: Yeah – either asset urls need to come from render via SSR, or we need to find a way to package static assets in times-components.

GaboDimitrov: With my latest change Favicon will come from render trough SSR. Please have a second look

packages/article-skeleton/__tests__/web/__snapshots__/head.web.test.js.snap
```diff
@@ -64,6 +64,11 @@ exports[`Head outputs correct metadata 1`] = `
     content="https://crop169.io"
     name="twitter:image"
   />
+  <script
+    type="application/ld+json"
+  >
+    {"@context":"http://schema.org","@type":"NewsArticle","headline":"Some Headline","logo":"https://www.thetimes.co.uk/d/img/icons/icon_1_5x-a13ee0bf9a.png","publisher":{"@type":"Organization","name":"The Times"},"mainEntityOfPage":{"@type":"WebPage","@id":"https://www.thetimes.co.uk/article/this-is-slug-2k629tpvh"},"dateCreated":"2015-03-13T18:54:58.000Z","datePublished":"2015-03-13","isAccessibleForFree":false,"hasPart":{"@type":"WebPageElement","isAccessibleForFree":false,"cssSelector":".class-name"},"image":{"@type":"ImageObject","url":"https://crop169.io","caption":"Some Caption"},"thumbnailUrl":"https://crop169.io"}
```
nathggns: Where is the `faviconPath` in this snap coming from? It doesn't seem to be included in the tests.

packages/article-skeleton/__tests__/web/head.web.test.js
```diff
@@ -33,10 +33,13 @@ addSerializers(
 );
 
 const article = articleFixture({ ...testFixture });
+const paidContentClassName = "class-name";
 
 describe("Head", () => {
   it("outputs correct metadata", () => {
-    const testRenderer = TestRenderer.create(<Head article={article} />);
+    const testRenderer = TestRenderer.create(
+      <Head article={article} paidContentClassName={paidContentClassName} />
```
nathggns: Do you need to add a `faviconPath` here?

packages/article-skeleton/__tests__/web/__snapshots__/head.web.test.js.snap
```diff
@@ -64,6 +64,11 @@ exports[`Head outputs correct metadata 1`] = `
     content="https://crop169.io"
     name="twitter:image"
   />
+  <script
+    type="application/ld+json"
+  >
+    {"@context":"http://schema.org","@type":"NewsArticle","headline":"Some Headline","logo":"https://www.thetimes.co.ukfavicon/path","publisher":{"@type":"Organization","name":"The Times"},"mainEntityOfPage":{"@type":"WebPage","@id":"https://www.thetimes.co.uk/article/this-is-slug-2k629tpvh"},"dateCreated":"2015-03-13T18:54:58.000Z","datePublished":"2015-03-13","isAccessibleForFree":false,"hasPart":{"@type":"WebPageElement","isAccessibleForFree":false,"cssSelector":".class-name"},"image":{"@type":"ImageObject","url":"https://crop169.io","caption":"Some Caption"},"thumbnailUrl":"https://crop169.io"}
```
chrishutchinson:
```suggestion
    {"@context":"http://schema.org","@type":"NewsArticle","headline":"Some Headline","logo":"https://www.thetimes.co.uk/favicon/path","publisher":{"@type":"Organization","name":"The Times"},"mainEntityOfPage":{"@type":"WebPage","@id":"https://www.thetimes.co.uk/article/this-is-slug-2k629tpvh"},"dateCreated":"2015-03-13T18:54:58.000Z","datePublished":"2015-03-13","isAccessibleForFree":false,"hasPart":{"@type":"WebPageElement","isAccessibleForFree":false,"cssSelector":".class-name"},"image":{"@type":"ImageObject","url":"https://crop169.io","caption":"Some Caption"},"thumbnailUrl":"https://crop169.io"}
```

It's a small thing, but seems a shame to have this broken. I see that it's because the mock data in the test doesn't have a leading `/`. Perhaps the `DOMAIN` constant should have a trailing slash, or we should just update the mocked value to represent what actually comes from `render` (assuming render does pass down a value with a leading `/`)

packages/article-skeleton/src/head.web.js
```diff
@@ -5,6 +5,8 @@ import PropTypes from "prop-types";
 import Context from "@times-components/context";
 import { renderTreeAsText } from "@times-components/markup-forest";
 
+const DOMAIN = "https://www.thetimes.co.uk";
```
chrishutchinson: What about non-prod environments?

packages/article-in-depth/__tests__/web/__snapshots__/article.web.test.js.snap
```diff
@@ -265,6 +270,11 @@ exports[`4. an article with ads 1`] = `
       content="https://crop169.io"
       name="twitter:image"
     />
+    <script
+      type="application/ld+json"
+    >
+      {"@context":"http://schema.org","@type":"NewsArticle","headline":"Some Headline","publisher":{"@type":"Organization","name":"The Times"},"mainEntityOfPage":{"@type":"WebPage","@id":"https://www.thetimes.co.uk/article/this-is-slug-2k629tpvh"},"dateCreated":"2015-03-13T18:54:58.000Z","datePublished":"2015-03-13","isAccessibleForFree":false,"hasPart":{"@type":"WebPageElement","isAccessibleForFree":false,"cssSelector":".undefined"},"image":{"@type":"ImageObject","url":"https://crop169.io","caption":"Some Caption"},"thumbnailUrl":"https://crop169.io"}
```
chrishutchinson: This snapshot line contains:

```
"cssSelector":".undefined"
```

Is that a bug, or just a mocking error?

GaboDimitrov: It's a mock error. I've decided not to add additional prop to all article tests. 'cssSelector' is tested in head.test.js

PR #2007, opened 2019-05-31
fix: ads on native

parkjh:
- Reverts newsuk/times-components#1623


parkjh:
awaiting on https://github.com/newsuk/times-components/pull/2018

PR #2008, opened 2019-05-31
Fix/360 video react article

pdimova:
The 360 video article was present only on CAPI.
Also the InlineVideoPlayer component used a static field index to count its instances. The server side rendered component hold one value for this index prop and the client side rendering of the component had another value. Thus the served value id was different than the id value used in the client scripts. That value is also used as part of the video tag id and helps controlling the 360 icon button and other overlays.

Solution: Keep the id value the same in ssr and client rendering. Id is now composed of the following variables: brightcoveVideoID , brightcoveAccountID and TPAVideoID.

Result: 
Story-with-360-_-The-Times-_1_

### DIFF
--- a/packages/video/__tests__/web/inline-video-player.web.test.js
+++ b/packages/video/__tests__/web/inline-video-player.web.test.js
@@ -8,7 +8,6 @@ describe("InlineVideoPlayer", () => {
   afterEach(() => {
     delete window.bc;
     delete window.videojs;
-    InlineVideoPlayer.index = 0;
     InlineVideoPlayer.scriptLoadError = false;
     InlineVideoPlayer.activePlayers = [];
     InlineVideoPlayer.brightcoveSDKLoadedStarted = false;
@@ -95,7 +94,9 @@ describe("InlineVideoPlayer", () => {
     addBrightcoveSDKGlobals();
 
     renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
-    renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
+    renderer.create(
+      <InlineVideoPlayer {...defaultVideoProps} id="[tpa video id 2]" />
+    );
 
     const [component1, component2] = InlineVideoPlayer.activePlayers;
     jest.spyOn(component1.player, "pause");
@@ -109,7 +110,9 @@ describe("InlineVideoPlayer", () => {
 
   it("doesn't hold references to players after they have been unmounted", () => {
     renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
-    const p2 = renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
+    const p2 = renderer.create(
+      <InlineVideoPlayer {...defaultVideoProps} id="[tpa video id 2]" />
+    );
 
     expect(InlineVideoPlayer.activePlayers.length).toBe(2);
     const [component1] = InlineVideoPlayer.activePlayers;
@@ -129,7 +132,9 @@ describe("InlineVideoPlayer", () => {
     jest.spyOn(InlineVideoPlayer, "appendScript").mockImplementation();
 
     renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
-    renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
+    renderer.create(
+      <InlineVideoPlayer {...defaultVideoProps} id="[tpa video id 2]" />
+    );
     const [component1, component2] = InlineVideoPlayer.activePlayers;
 
     jest.spyOn(component1, methodName).mockImplementation();
--- a/packages/video/__tests__/web/video.web.test.js
+++ b/packages/video/__tests__/web/video.web.test.js
@@ -72,7 +72,7 @@ const tests = [
     }
   },
   {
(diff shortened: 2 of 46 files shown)


packages/video/src/inline-video-player.web.js
```diff
@@ -169,11 +166,10 @@ class InlineVideoPlayer extends Component {
       videoId,
       accountId,
       playerId,
-      skySports
+      skySports,
+      is360
     } = this.props;
     const { error, hasVideoPlayed } = this.state;
-    const is360Video = playerId !== "default"; // Will get an explicit boolean from CMS and this inference can then go
```
aronshamash: There is another JIRA ticket specifically for this which can also be closed when this PR goes in.

packages/video/__tests__/web/video.web.test.js
```diff
@@ -75,7 +75,7 @@ const tests = [
     name: "no sky banner displayed on play",
```
aronshamash: this test has same description as above. Can we make them more descriptive of the conditions?

packages/video/video.showcase.js
```diff
@@ -97,13 +111,17 @@ export default {
...
     {
-      component: () => (
-        <View>
-          <Video {...defaultVideoProps} />
-          <View style={{ height: 20 }} />
-          <Video {...defaultVideoProps} videoId="5612887446001" />
-        </View>
-      ),
+      component: () => {
+        const mockId = "3dbfe6b8-680b-11e9-b277-88f3d445182d";
```
aronshamash: no need for the `const`

packages/video/src/inline-video-player.web.js
```diff
@@ -169,11 +166,10 @@ class InlineVideoPlayer extends Component {
...
-
     if (error) {
       throw new Error(); // caught by parent ErrorView
```
apostolnikov: Can we specify the error that we are throwing here? If it's anonymous it will be harder to trace.

packages/video/__tests__/web/inline-video-player.web.test.js
```diff
@@ -95,7 +94,9 @@ describe("InlineVideoPlayer", () => {
     addBrightcoveSDKGlobals();
 
     renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
-    renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
+    renderer.create(
+      <InlineVideoPlayer {...defaultVideoProps} id="[tpa video id 2]" />
```
apostolnikov: Why do we add the id on the second player only?

packages/video/__tests__/web/inline-video-player.web.test.js
```diff
@@ -129,7 +132,9 @@ describe("InlineVideoPlayer", () => {
     jest.spyOn(InlineVideoPlayer, "appendScript").mockImplementation();
 
     renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
-    renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
+    renderer.create(
+      <InlineVideoPlayer {...defaultVideoProps} id="[tpa video id 2]" />
```
apostolnikov: Here as well, the same for the id?

packages/video/__tests__/web/inline-video-player.web.test.js
```diff
@@ -109,7 +110,9 @@ describe("InlineVideoPlayer", () => {
 
   it("doesn't hold references to players after they have been unmounted", () => {
     renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
-    const p2 = renderer.create(<InlineVideoPlayer {...defaultVideoProps} />);
+    const p2 = renderer.create(
+      <InlineVideoPlayer {...defaultVideoProps} id="[tpa video id 2]" />
```
apostolnikov: Here as well, the same for the id?

packages/video/src/inline-video-player.web.js
pdimova: great

packages/video/__tests__/web/video.web.test.js
pdimova: changed the name of the test to `"player icon for 360 video not displayed after play"` , though the snaps don't really show a lot here for these tests

packages/video/video.showcase.js
pdimova: moved the const up since it is used in 3 tests, also changed its value

packages/video/src/inline-video-player.web.js
pdimova: It's good to have an error message , so added one

packages/video/__tests__/web/inline-video-player.web.test.js
pdimova: here both instances use the default props where id has value of "[tpa video id]". When having multiple video players per page they should have unique ids. For the purpose of testing we don't need multiple videos uploaded to Brightcove, we can use one, just the id coming from TPA should be different, thus mocking the tpa id

packages/video/__tests__/web/inline-video-player.web.test.js
pdimova: same as above

PR #2010, opened 2019-06-03
fix: save-star-web tests

ValGeorgiev:
Update save star web tests. Add two more test scenarios to test if the correct method is invoked. 

Add some changes to the android-app readme

### DIFF
--- a/packages/save-and-share-bar/__tests__/shared-tracking.js
+++ b/packages/save-and-share-bar/__tests__/shared-tracking.js
@@ -109,7 +109,7 @@ export default () => {
       shareArticleUrlByEmailBarItem.props.onPress();
 
       const [[call]] = stream.mock.calls;
-      console.log(stream.mock.calls);
+
       expect(call).toMatchSnapshot();
       expect(getTokenisedShareUrl.mock.calls).toMatchSnapshot(
         "getTokenisedShareUrl"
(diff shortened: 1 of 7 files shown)


packages/save-and-share-bar/__tests__/shared-tracking.js
```diff
@@ -109,7 +109,7 @@ export default () => {
       shareArticleUrlByEmailBarItem.props.onPress();
 
       const [[call]] = stream.mock.calls;
-
+      console.log(stream.mock.calls);
```
nehasri89: don't need this

packages/save-star-web/__tests__/shared.base.js
```diff
@@ -39,13 +46,55 @@ export default () => {
...
 
       await delay(0);
       expect(testInstance).toMatchSnapshot();
     });
+
+    it("check if bookmark api method is invoked", async () => {
+      const bookmark = jest.fn().mockResolvedValue();
+      const testInstance = TestRenderer.create(
+        <SaveStarWeb
+          articleId={savedId}
```
nehasri89: If we are trying to `save` a savedArticle, it should call the `unbookmark`?

packages/save-star-web/__tests__/shared.base.js
```diff
@@ -39,13 +46,55 @@ export default () => {
...
+          articleId={notSavedId}
+          saveApi={{
+            ...mockSaveApi,
+            unBookmark
+          }}
+          onSaveButtonPress={jest.fn(cb => cb())}
+        />
+      );
+
+      const [saveStarLink] = testInstance.root.findAllByType(Link);
```
nehasri89: are we trying trying to `save` first and then `unsave`? can we not try to just `unsave` what is already saved?

PR #2012, opened 2019-06-03
chore: add fake nav bar to article storybook

nathggns:
Add a fake nav bar to article storybook to demo the sticky save and share bar sticking below it. 

image

### DIFF
--- a/packages/article/package.json
+++ b/packages/article/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "fmt": "eslint . --fix && prettier --write '**/*.*'",
     "prettier:diff": "prettier --list-different '**/*.*'",
-    "depcheck": "depcheck --ignores='babel-*,depcheck,eslint,jest,prettier,react-art,stylelint*,webpack*' --ignore-bin-package=false --skip-missing",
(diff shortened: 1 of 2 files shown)


packages/article/package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "fmt": "eslint . --fix && prettier --write '**/*.*'",
     "prettier:diff": "prettier --list-different '**/*.*'",
-    "depcheck": "depcheck --ignores='babel-*,depcheck,eslint,jest,prettier,react-art,stylelint*,webpack*' --ignore-bin-package=false --skip-missing",
+    "depcheck": "depcheck --ignores='babel-*,depcheck,eslint,jest,prettier,react-art,stylelint*,webpack*,@times-components/sticky' --ignore-bin-package=false --skip-missing",
```
ValGeorgiev: why this is needed?

ValGeorgiev: It looks like an issue with depcheck, it ignores showcase.web.js files. I reproduce it in other packages as well. But if i add the dependancy in the showcase.js file everything is fine. It will be good to raise it with depcheck owners

ValGeorgiev:
@nathggns What is the status of this PR?

nathggns:
@ValGeorgiev going to close it because it still doesn't quite duplicate what's happening in render. Will submit another one later when I do save & share bar work.

PR #2013, opened 2019-06-04
feat: allow dropcaps that contain markup

jmars:
Allows articles to render correctly when a dropcap contains a link tag.

image


packages/article-skeleton/src/article-body/article-link.js
```diff
@@ -8,11 +8,20 @@ import articleLinkTrackingEvents from "./article-link-tracking-events";
 const ArticleLink = props => (
   <Context.Consumer>
     {({ theme: { scale } }) => {
-      const styles = styleFactory(scale);
+      const styles = {
+        ...styleFactory(scale)
```
parkjh: why?

packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -72,14 +73,38 @@ export default ({
...
+      }
       const height = lineHeight * 3 * fontScale;
-      const text = new Text.Text({
+      let text = new Text.Text({
         font: fonts[dropCapFont],
         lineHeight: height,
         markup: [new Body(value)],
         size: height
       });
+      if (children.length) {
```
parkjh: are you checking if the children is array here or if the array length is greater than 0?

packages/article-skeleton/__tests__/common/dropcap-util.test.js
```diff
@@ -40,12 +74,73 @@ const childWithDropCap = [
   }
 ];
 
+const childWithDropCapAndMarkup = [
```
parkjh: 👍

packages/article-skeleton/src/dropcap-util.js
```diff
@@ -6,32 +6,57 @@ const templateWithDropCaps = [
...
+            dropCap: true
+          }
+        },
+        ...children.slice(1)
+      ]
     };
-    const newTextElement = {
+  }
+  const firstChild = splitNode(children[0]);
+  if (firstChild.attributes.dropCap && node.name !== "paragraph") {
```
parkjh: what's the case where `node.name !== paragraph` and can we please have corresponding test case?

packages/article-skeleton/__tests__/android/__snapshots__/article-with-style.android.test.js.snap
```diff
@@ -617,6 +605,33 @@ exports[`3. an article with a nested markup in first paragraph displays no drop
...
+                  "color": "#FF0000",
+                  "fontFamily": "TimesModern-Regular",
+                  "fontSize": 110,
+                  "marginBottom": -30,
+                  "marginRight": 5,
+                  "marginTop": -14,
+                }
+              }
+            >
+              Q
```
nehasri89: the title of this test should change?

packages/article-skeleton/__tests__/android/__snapshots__/article-with-style.android.test.js.snap
```diff
@@ -766,7 +814,7 @@ exports[`4. an article skeleton with responsive items 1`] = `
                     }
                   }
                 >
-                  his being Black History Month, last week
+                  This being Black History Month, last week
```
nehasri89: is this now supposed to not have a drop cap?

packages/article-skeleton/__tests__/android/__snapshots__/article-with-style.android.test.js.snap
parkjh: what title is the letter `Q` coming from?

parkjh:
@jmars is there no storybook cases where we can quickly tests this?

packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -71,14 +71,17 @@ class ArticleSkeleton extends Component {
...
     if (newContent && newContent.length > 0) {
-      newContent[0] = insertDropcapIntoAST(
-        newContent[0],
-        template,
-        dropcapsDisabled
-      );
+      newContent = insertDropcapIntoAST(newContent, template, dropcapsDisabled);
+    }
+
+    if (newContent.length > 0 && newContent[0].name === "dropCap") {
```
parkjh: no test?

packages/article-skeleton/__tests__/web/__snapshots__/article.web.test.js.snap
```diff
@@ -469,8 +472,13 @@ exports[`3. an article with a nested markup in first paragraph displays a drop c
     </div>
     <div>
       <p>
+        <span>
+          <b>
+            Q
+          </b>
+        </span>
         <b>
-          Q Some years ago I made the decision to take as a friend a woman who is very opinionated.
```
parkjh: 👍

packages/article-skeleton/__tests__/web/dropcap-util.test.js
```diff
@@ -0,0 +1,144 @@
...
+          }
+        ],
+        name: "link"
+      }
+    ],
+    name: "paragraph"
+  }
+];
+
+describe("insertDropcapIntoAST", () => {
```
parkjh: 👍

packages/article-skeleton/__tests__/web/__snapshots__/article-with-user-state.web.test.js.snap
```diff
@@ -142,9 +142,7 @@ exports[`Article with user state Render Metered expired article when user is log
     </div>
     <div>
       <p>
-        <span>
-          T
-        </span>
+        <span />
```
parkjh: is this expected?

jmars: This is caused by the two pass rendering in renderTree, dropcaps in this PR use the markup from the children after it's rendered so for the first pass it's not there. I have added a placeholder for now but this exposes a larger issue we have with the whole two-pass thing in rendertree, it doesn't feel right.

parkjh: @nathggns fyi ^^^

packages/article-skeleton/__tests__/web/__snapshots__/article-with-user-state.web.test.js.snap
```diff
@@ -1240,7 +1246,9 @@ exports[`Article with user state Render teaser article when user is null 1`] = `
   <div>
     <div>
       <p>
-        <span />
+        <span>
+           
+        </span>
```
parkjh: @jmars  what does the article look like at this state? I fear that if the 2 pass rendering somehow got stuck at this state... but then that's an unrelated bigger problem than drop caps...

parkjh: This feels wrong to deliberately accept an incorrect snapshot... do you have any suggestion to work around this @nathggns ?

packages/article-skeleton/__tests__/web/__snapshots__/article-with-user-state.web.test.js.snap
parkjh: @jmars wait what is triggering the 2 pass rendering? you are not based off of Nates work, so is there something else I'm missing?

packages/article-skeleton/src/dropcap-util-common.js
```diff
@@ -0,0 +1,96 @@
...
+    const without = splitNode(children[0]);
+    let childs = withCap.children;
+    while (childs[0].attributes.dropCap && childs[0].name !== "text") {
+      childs = childs[0].children;
+    }
+    childs.splice(1);
+    childs = without.children;
+    while (childs[0].attributes.dropCap && childs[0].name !== "text") {
+      childs = childs[0].children;
+    }
```
ValGeorgiev: Could we refactor this in more functional way, please? Something like:

```
const newChildren = reduceChildren(withCap.children);

const reduceChildren = children => {
   return children.reduce(blabla);
}
```

It looks better structured and is less error prone.

packages/article-skeleton/src/dropcap-util.web.js
```diff
@@ -0,0 +1,12 @@
...
+
+export default (children, template, isDropcapDisabled) => {
+  let newContent = insertDropcapIntoAST(children, template, isDropcapDisabled);
+  if (newContent.length > 0 && newContent[0].name === "dropCap") {
+    // remove the wrapping paragraph
+    newContent[0].children = newContent[0].children[0].children;
+    newContent[1].children = [newContent[0], ...newContent[1].children];
+    newContent = newContent.slice(1);
+  }
+  return newContent;
```
ValGeorgiev: What about: 
```
  const newContent = insertDropcapIntoAST(children, template, isDropcapDisabled);
  if (newContent.length > 0 && newContent[0].name === "dropCap") {
    // remove the wrapping paragraph
    newContent[0].children = newContent[0].children[0].children;
    newContent[1].children = [newContent[0], ...newContent[1].children];
    return newContent.slice(1);
  }
  return newContent;
```

Personally I don't like `let`. Also are we 100% sure the first element has children array with first element who has children prop. It will be good to add more conditions

packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -72,14 +72,10 @@ class ArticleSkeleton extends Component {
...
-    const newContent = [...content];
+    let newContent = [...content];
 
     if (newContent && newContent.length > 0) {
-      newContent[0] = insertDropcapIntoAST(
-        newContent[0],
-        template,
-        dropcapsDisabled
-      );
+      newContent = insertDropcapIntoAST(newContent, template, dropcapsDisabled);
```
ValGeorgiev: could we move this into separate function, like `getNewContent`
```
const newContent = getNewContent([...content]);
```

```
getNewContent(content) {
   if (content && content.length > 0) {
      return insertDropcapIntoAST(content, template, dropcapsDisabled);
    }
    return content
}

File: packages/article-skeleton/src/article-body/article-body-row.js [hunk 15]
```diff
@@ -72,14 +73,38 @@ export default ({
...
-    dropCap(key, { value }) {
+    dropCap(key, attributes, children, indx, node) {
+      let value = node;
+      while (typeof value !== "string") {
+        if (value.name === "text") {
+          value = value.attributes.value;
+        } else {
+          value = value.children[0];
+        }
+      }
```
ValGeorgiev: this can be separated in another function and use it in `new Body(getValue(node))` or attached to some constant. We can easily add unit tests for that as well

packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -72,14 +73,38 @@ export default ({
...
+        text.font = fonts[dropCapFont];
+        text.lineHeight = height;
+        text.size = height;
+        text.layout();
+        value = text.words[0].idealSpans[0];
+        if (value.href) {
+          value = value.href(value);
+        } else {
+          value = value.text;
+        }
```
ValGeorgiev: Same here:
`newValue = getNewValue(text.words[0].idealSpans[0]); `
```
const getNewValue = (val) => {
   if (val.href) {
     return val.href(val);
   } 

   return val.text;
}

```

And we can unit tested it as well.

PR #2014, opened 2019-06-05
chore: add watch for easily linking render

ValGeorgiev:
Ticket: https://nidigitalsolutions.jira.com/browse/REPLAT-6354

Related Render PR: https://github.com/newsuk/cps-content-render/pull/2771

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ To debug our native Storybook:
 
 ## Link times-components to the Render project
 
-Follow this steps [here](https://github.com/newsuk/cps-content-render#locally-mount-your-custom-build-of-times-components)
+Follow these steps [here](https://github.com/newsuk/cps-content-render#locally-mount-your-custom-build-of-times-components)
 
 ## Testing and code coverage
 
--- a/README.md
+++ b/README.md
@@ -98,6 +98,10 @@ To debug our native Storybook:
 6. click _Sources_
 7. expand _debuggerWorker.js_ => _webpack://_ => _._ => _packages_
 
+## Link times-components to the Render project
+
+Follow this steps [here](https://github.com/newsuk/cps-content-render#locally-mount-your-custom-build-of-times-components)
+
 ## Testing and code coverage
 
 | Tests                                      | Covered            | CI                                                            |
--- a/lib/cli/templates.js
+++ b/lib/cli/templates.js
@@ -197,6 +197,13 @@ yarn fmt
 yarn lint
 \`\`\`
 
+## Link to Render project
+\`\`\`
+yarn watch
+\`\`\`
+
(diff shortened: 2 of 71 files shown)


packages/ad/package.json
```diff
@@ -14,6 +14,7 @@
     "test:ios": "jest --config='./__tests__/ios/jest.config.js'",
     "test:web": "jest --config='./__tests__/web/jest.config.js'",
     "test:all": "yarn test:android --coverage && yarn test:ios --coverage && yarn test:web --coverage",
+    "watch": "watch 'yarn prepublishOnly && cd ../ssr && yarn prepublishOnly' ./src --ignoreDotFiles --ignoreUnreadable",
```
IvayloAndonov: My suggestion is to run `yarn bundle` in given package. It will reproduce `rnw.js` file which is our purpose to get new changes. Command `prepublishOnly` run first `yarn transpile`  - clean up and rebuild the 'dist' folder which isn`t needed in our case.
Supposed to be applied for each package.json.

```suggestion
    "watch": "watch 'yarn bundle && cd ../ssr && yarn prepublishOnly' ./src --ignoreDotFiles --ignoreUnreadable",
```

lib/cli/templates.js
```diff
@@ -197,6 +197,13 @@ yarn fmt
 yarn lint
 \`\`\`
 
+## Link to Render project
+\`\`\`
+yarn watch
+\`\`\`
+
+To link the package to render project run \`yarn watch\`, make some code change, than run \`yarn install\` in the render project and restart the server
```
nehasri89: I guess we have agreed we don't need the subsequent `yarn install`

packages/tealium/package.json
```diff
@@ -28,6 +29,7 @@
     "@times-components/eslint-config-thetimes": "0.8.12",
     "@times-components/jest-configurator": "2.5.12",
     "@times-components/test-utils": "2.2.43",
+    "@times-components/webpack-configurator": "2.0.21",
```
nehasri89: does this pkg need this dep?

ValGeorgiev: yes, every package has webpack.config.js which is using webpack-configurator package

lib/cli/templates.js
ValGeorgiev: correct, i will change it

README.md
```diff
@@ -98,6 +98,10 @@ To debug our native Storybook:
 6. click _Sources_
 7. expand _debuggerWorker.js_ => _webpack://_ => _._ => _packages_
 
+## Link times-components to the Render project
+
+Follow this steps [here](https://github.com/newsuk/cps-content-render#locally-mount-your-custom-build-of-times-components)
```
aronshamash: `this` -> `these` type

PR #2015, opened 2019-06-05
fix: pass whole userState object by SSR to the TC consumers

IvayloAndonov:
**Currently**, only field 'isLoggedIn' is passed from SSR to packages consumers : 

```
user: {	   
   isLoggedIn: userState.isLoggedIn	
}
```

**Fix**: Pass full user state data from SSR to consumers.

Format: 

```
userState : {
   isLoggedIn: boolean
   isMetered: boolean
   isMeteredExpired: boolean
   isShared: boolean
  registrationTypes: string
}
```


IvayloAndonov:
> What issue do you think this is going to fix, just out of curiosity?

react-renders in Render passes the whole user state object from through SSR to components. Currently, just 'isLoggedIn' field is passed which is wrong and inconsistent, missed to be updated.

PR #2017, opened 2019-06-05
spike: fix user states

nathggns:
This PR is a spike showing how we can use two pass rendering to ensure that the initial render on the client and the server don't differ even though we need some information that is only available on the client to render properly. 

I've also fixed some of the bugs I've noticed in how we've implemented user states. While this went a bit beyond what I strictly needed to do, I want to ensure this approach could handle ever state. There may be more issues with our implementation of user states even after merging this, but I fixed all of the ones I found.

### DIFF
--- a/packages/save-and-share-bar/src/save-and-share-bar.js
+++ b/packages/save-and-share-bar/src/save-and-share-bar.js
@@ -46,23 +46,46 @@ class SaveAndShareBar extends Component {
 
   handleOnShareEmailPress() {
     /* eslint-env browser */
-    const { articleHeadline, articleId, getTokenisedShareUrl } = this.props;
-    this.setState({ isLoading: true });
+    const {
+      articleId,
+      articleUrl,
+      getTokenisedShareUrl,
+      shouldTokeniseShareLinks
+    } = this.props;
 
-    getTokenisedShareUrl(articleId)
-      .then(res => {
-        const { data } = res;
-        if (data) {
+    if (shouldTokeniseShareLinks) {
+      this.setState({ isLoading: true });
+
+      getTokenisedShareUrl(articleId)
+        .then(res => {
+          const { data } = res;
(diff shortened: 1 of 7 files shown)


packages/utils/src/index.web.js
```diff
@@ -12,3 +12,6 @@ export { default as getHeadline } from "./get-headline";
 export { default as gqlRgbaToHex } from "./gql-rgba-to-hex";
 export { default as gqlRgbaToStyle } from "./gql-rgba-to-style";
 export { default as HoverIcon } from "./hover-icon";
+export {
+  default as ClientUserStateConsumer
+} from "./client-user-state-consumer";
```
IvayloAndonov: Where is `./client-user-state-consumer` file placed ?

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -46,23 +46,46 @@ class SaveAndShareBar extends Component {
...
-        this.setState({ isLoading: false });
-        console.error("Error in connecting to api", error);
-      });
+          console.error("Error in connecting to api", error);
+        });
+    } else {
+      let url = articleUrl;
+      const matches = window.location.search.match(/[?&]shareToken=([^&]+)/);
+
+      if (matches) {
```
apostolnikov: There's really no need to create unnecessary variables here and then mutate them.
const matches = window.location.search.match(/[?&]shareToken=([^&]+)/);
      return matches
          ? this.openMailClient(`${articleUrl}?shareToken=${matches[1]}`)
          : this.openMailClient(articleUrl);

packages/utils/src/index.web.js
nathggns: lol oops. Added now.

nathggns:
This is just a spike. Closing it.

PR #2018, opened 2019-06-05
fix: native ads

parkjh:


parkjh:
all article ssr tests are failing as it cannot find component with id `header`...

but after further debugging see test videos https://drive.google.com/open?id=1ol0nVR_qvQrlUT6lEOddm3C9O1j-6Dlc it seems that the pages are rendered but I wonder if there was any work done recently that changed the article-header component?

packages/image/src/modal-image/modal-image.js
```diff
@@ -130,7 +131,11 @@ class ModalImage extends Component {
           </View>
         </Modal>
         <Button onPress={this.showModal}>
-          <Image {...this.props} onLayout={this.onLowResLayout} />
+          <Image
+            {...this.props}
+            lowResSize={lowResSize}
+            onLayout={this.onLowResLayout}
```
ValGeorgiev: this shouldn't be in this PR, could we clear it, please?

packages/ad/__tests__/utils/ad-init-apstag.js
```diff
@@ -0,0 +1,144 @@
...
+
+export default () => {
+  let mock;
+  let initOptions;
+
+  beforeEach(() => {
+    ({ mock, initOptions } = makeAdInitMocks());
+  });
+
+  it("do not sets up Amazon bidding if amazon account ID is not set", () => {
```
aronshamash: `does not set up`

packages/ad/__tests__/utils/ad-init-apstag.js
```diff
@@ -0,0 +1,144 @@
...
+  it("get Amazon Config with a commercial section", () => {
+    const init = adInit(initOptions);
+    expect(
+      init.apstag.getConfig({
+        slots: [],
+        networkId: "3048",
+        adUnit: "d.thetimes.co.uk",
+        section: "news"
+      })
+    ).toEqual([]);
```
aronshamash: Sorry don't understand this test. are you expecting the config to be empty ?

parkjh: @aronshamash yes

aronshamash: @parkjh shouldn't the test desc reflect you expect empty config ?

PR #2019, opened 2019-06-06
Feat: analytics react article registration type

IvayloAndonov:
**Implement analytics for article registration type being viewed on WEB.**

JIRA ticket : https://nidigitalsolutions.jira.com/browse/REPLAT-6569

With this PR we can track registration type and shared status for react article web page.
Data object example that we send to analytics when open an article: 
```
{
        "articleId": "198c4b2f-ecec-4f34-be53-c89f83bc1b44",
        "bylines": "Camilla Long",
        "eventTime": "2018-01-01T00:00:00.000Z",
        "headline": "Caribbean islands devastated by Hurricane Irma, the worst Atlantic storm",
        "label": "HURRICANE IRMA",
        "pageName": "france-defies-may-over-russia-37b27qd2s",
        "publishedTime": "2015-03-13T18:54:58.000Z",
        "referralUrl": "",
        "registrationType": "logged out",
        "section": "news",
        "shared": "no",
        "template": "mainstandard",
}
```


packages/article-skeleton/src/tracking/article-tracking-context.web.js
```diff
@@ -0,0 +1,25 @@
+import get from "lodash.get";
+import { withTrackingContext } from "@times-components/tracking";
+import { getRegistrationType, getSharedStatus } from "../data-helper";
+
```
nehasri89: `web` and `native` have all same values except `registrationType` I think, so we should have one base object and both platforms should inherit

IvayloAndonov: Agree with that, but when I used one base obkect I faced issue because in the test that we mock the user state : `global.nuk.user = userStateMock;` the global.nuk is not valid when test run on `native` platform. That`s why I separated them into web and native. @nehasri89

packages/article-skeleton/__tests__/web/tracking.web.test.js
```diff
@@ -1,17 +1,50 @@
...
 
 beforeEach(() => {
-  const nuk = {
-    user: {
-      isLoggedIn: true
-    }
-  };
-  global.nuk = nuk;
+  mockDate.set(1514764800000, 0);
+  global.nuk = { user: { registrationType: "logged out" } };
```
nathggns: If you're using user state for your tests would you mind waiting on https://github.com/newsuk/times-components/pull/2021 being merged before merging this?

packages/article-skeleton/src/tracking/article-tracking-context.web.js
IvayloAndonov: Current implementation is only for WEB article react pages - [story](https://nidigitalsolutions.jira.com/browse/REPLAT-6569) and it is consumed directly from `window.nuk.user` object. For `native` part the approach should be different and 'user' object should be passed down as prop to the component in terms to be tracked in Article skeleton tracking data. So the tracking context to `native` is separated till now as part of different story.

packages/article-skeleton/__tests__/web/tracking.web.test.js
IvayloAndonov: The tracking field `registrationType` stays the same in user state so it wont be affected by  #2021  changes.

nathggns:
We're going to wait to merge this until https://github.com/newsuk/times-components/pull/2021 is merged.

PR #2020, opened 2019-06-07
feat: implement lead asset to use the passed width

apostolnikov:
Implement Article Perf - Lead Asset not waiting for onLayout.
Story: https://nidigitalsolutions.jira.com/browse/REPLAT-6648

In this PR we are passing the width from the `article-lead-asset` component down to the image as a `hightResSize` prop. This way the `image` component won't use the onLayout to calculate the size it's being rendered at, which will lead to faster request for the image.


nathggns:
Just going to run this now and check its working as expected.

nathggns:
Unfortunately I can't actually verify this is working due to my local setup but it LGTM. You have two approvals so feel free to merge.

PR #2021, opened 2019-06-07
fix: user state hydration

nathggns:
There's a couple of outstanding tasks for this, but I wanted to create the PR now so thqt you can get a sense of what the API I've settled on for accessing user states in a safe way is. It's a nice API because it lends itself well to being easily mockable, and its declarative meaning it's easy to follow and reason about.

## To do:

- [x] Test `UserState`
- [x] Document new user state apis
- [x] Fix storybooks
- [x] Move user state to new package 
- [x] Write tests for remaining matchers
- [x] Fix version number changes
- [x] Fix conflicts

## Example

```js
import UserState from "@times-components/user-state";

<UserState
    state={UserState.subscriber}
    fallback={<div>Will only render for non-subscribers</div>}
    serverRender={/* Control whether you want this to render on the server at all */}
>
    // Anything you only want to render for subscribers
    <div>Will only render for subscribers</div>
</UserState>
```

## Storybook

To save having to write stories for each user state, we now intelligently detect whether a `UserState` component is being used in the story and display the following knobs that allow you to render the story in different states.

image

You can select one of four user states which are mutually exclusive (`Subscriber`, `Registered Access`, `Registered Access (Expired)`, and `Guest`) and then check whether you're viewing the article via a share token (as that is mutually inclusive with all of the states).


parkjh:
@nathggns can you please provide all the steps to test all the user states?

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -130,19 +98,21 @@ class SaveAndShareBar extends Component {
...
-              articleId={articleId}
-              articleHeadline={articleHeadline}
-              saveApi={saveApi}
-              height={styles.svgIcon.star.height}
-              onSaveButtonPress={this.onSaveButtonPress}
-            />
-          </View>
-        )}
+        {savingEnabled ? (
+          <UserState state={UserState.loggedIn} serverRender={false}>
```
nehasri89: does it need to have a UserState? Because the only time its not there on the page is when the share bar is not there on the page for a RA user

packages/storybook/src/storybook.js
```diff
@@ -22,5 +23,6 @@ export {
   sections,
   select,
   showcaseConverter,
-  WhiteBgColorDecorator
+  WhiteBgColorDecorator,
+  withUserStateKnobs
```
nehasri89: they are coming up on all stories, even in ads, do we need it?

nathggns: Yeah I think its a good idea because any story can use user states and its a global configuration, so we should be able to control the user states from anywhere. It might possible to work out whether user states are being used before showing the knobs if you'd prefer? I just didn't try because I didn't think they caused any harm.

packages/article-in-depth/__tests__/web/__snapshots__/article-with-style.web.test.js.snap
```diff
@@ -2454,75 +2161,6 @@ exports[`full article with style in the sunday times magazine 1`] = `
...
-}
-
-@media (max-width:1320px) {
-  .c11 {
-    left: 0;
-    top: 0;
-    z-index: 999;
-  }
-
-  .c11.sticky {
```
nehasri89: looks like the styling for sticky css is removed, is this intentional?

packages/storybook/src/storybook.js
nathggns: I've made it now so user state knobs only shown if user state is being used - check this out https://github.com/newsuk/times-components/pull/2021/commits/bfef00207ab4533b13dcefc5479ae69f9be66331

packages/article-in-depth/__tests__/web/__snapshots__/article-with-style.web.test.js.snap
nathggns: The sticky bar as a whole is being removed because the mock states aren't setup to show it on those templates. Will fix that.

nathggns: Resolved this.

packages/save-and-share-bar/__tests__/web/__snapshots__/save-and-share-bar.test.js.snap
```diff
@@ -148,39 +174,10 @@ exports[`save and share bar component email icon with loading state while networ
...
-        "WebkitBoxOrient": "horizontal",
-        "WebkitFlexDirection": "row",
-        "flexDirection": "row",
-        "msFlexDirection": "row",
-        "paddingBottom": "10px",
-        "paddingTop": "10px",
-      }
-    }
-  >
-    <SaveStarWeb
```
nehasri89: think this should stay

nehasri89:
good stuff!! Just one minor comment, PR looks great!

packages/utils/src/server-client-render.js
```diff
@@ -0,0 +1,45 @@
...
+ * Without this, you could break React hydration as
+ * it could cause a difference between the initial renders
+ * on the front end and on the server.
+ *
+ * This uses two-pass rendering to prevent differences in the
+ * initial render on server and client.
+ *
+ * https://reactjs.org/docs/react-dom.html#hydrate
+ */
+import { Component } from "react";
```
parkjh: @nathggns why are these all utils functions?

nathggns: I chose `utils` because that is where the user state stuff was already. I can move if preferred but not sure where else I'd put them. Happy to take suggestions!

packages/save-and-share-bar/__tests__/web/__snapshots__/save-and-share-bar.test.js.snap
nathggns: good catch. typo in the user states.

packages/utils/src/server-client-render.js
nathggns: Moving it to a separate package

packages/utils/src/user-state-component.js
```diff
@@ -0,0 +1,55 @@
...
+  isShared,
+  shouldShowFullArticle,
+  isSubscriber,
+  isLoggedInOrShared
+} from "./user-state";
+
+function UserState({ state, children, fallback, serverRender }) {
+  return (
+    <ClientUserStateConsumer serverRender={serverRender}>
+      {({ user }) => (state(user) ? children : fallback)}
```
IvayloAndonov: Really like the approach of implementing kind of _facade_ which take care of all user states logic in one place. However the the components nesting level and complexity going higher. Using `state(user)` is a little bit confusing, what do you think to use for example : `isCurrentStateCompatible(user)` ?

packages/user-state/README.md
```diff
@@ -0,0 +1,127 @@
...
+yarn test:all
+```
+
+Visit the official
+
+<!-- Add the storybook link here. -->
+
+## Future
+
+<!-- Add details of future development here. -->
```
parkjh: @aronshamash we should add any additional notes after the meeting with Jon from Akamai

packages/utils/src/user-state-component.js
nathggns: done.

packages/article-extras/src/article-extras.web.js
```diff
@@ -1,55 +1,46 @@
...
-  topics,
-  topicsAllowed
+  topics
 }) => (
-  <Fragment>
-    {topicsAllowed ? <ArticleTopics topics={topics} /> : null}
+  <UserState state={UserState.fullArticle}>
+    <ArticleTopics topics={topics} />
     <aside id="related-articles" ref={node => registerNode(node)}>
-      {relatedArticleAllowed ? (
```
parkjh: 👍

packages/article-magazine-comment/__tests__/web/__snapshots__/article.web.test.js.snap
```diff
@@ -144,15 +144,31 @@ exports[`3. an article with no headline falls back to use shortheadline 1`] = `
         </figcaption>
       </figure>
     </div>
+    <div>
+      <div>
+        <SaveAndShareBar
```
parkjh: is this expected?

packages/user-state/README.md
aronshamash: This PR can be moved without knowing the exact way we will implement the routing in on the times website. We could add something to the affect of the new architecture of the article page user states should allow us to remove the 2 pass rendering as each page will have its own unique template rendered on the server. @nathggns @parkjh What do you think?

packages/article-magazine-comment/__tests__/web/__snapshots__/article.web.test.js.snap
parkjh: there are a lot of save and share bar added to the article web snapshots, and I'm just wondering how that is affected by the user state work and if it's expected on those why it wasn't added beforehand?

packages/user-state/src/user-state.js
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <ClientUserStateConsumer serverRender={serverRender}>
+      {({ user }) =>
+        shouldRenderInCurrentUserState(user) ? children : fallback
+      }
+    </ClientUserStateConsumer>
+  );
+}
+
+UserState.loggedIn = isLoggedIn;
```
parkjh: not a big fan of this but I understand that this is on purpose so that it's easier to remove in the future...

packages/article-magazine-comment/__tests__/web/__snapshots__/article.web.test.js.snap
nathggns: Yeah its expected. Its just because the tests didn't have the correct user states set up, and they do now.

packages/user-state/src/user-state.js
nathggns: Yeah its just so consumers of the package can just do

```js
import UserState from "@times-components/user-state";

<UserState state={UserState.subscriber}>
   {/* ... */}
</UserState>
```

without having to import the matchers directly.

packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -95,90 +94,73 @@ class ArticleSkeleton extends Component {
...
+                  <Ad
+                    contextUrl={url}
+                    section={section}
+                    slotName="header"
+                    style={adStyle}
+                  />
+                </HeaderAdContainer>
+                <MainContainer>
+                  <Header width={articleWidth} />
+                  {savingEnabled || sharingEnabled ? (
```
aronshamash: doesn't this mean that if sharingEnabled is false user also can't use the saving button ?

packages/save-and-share-bar/src/email-share.js
```diff
@@ -0,0 +1,92 @@
...
+
+      this.openMailClient(
+        matches ? `${articleUrl}?shareToken=${matches[1]}` : articleUrl
+      );
+    }
+  }
+
+  openMailClient(url) {
+    const { articleHeadline } = this.props;
+    const mailtoEmailUrl = `mailto:?subject=${articleHeadline} from The Times&body=I thought you would be interested in this story from The Times%0A%0A${articleHeadline}%0A%0A${url}`;
```
aronshamash: We could be sharing a Sunday Times article and they are very picky about this topic. We need use a look up based on the article `publication` to display the correct wording

packages/save-and-share-bar/src/email-share.js
```diff
@@ -0,0 +1,92 @@
...
+      <Fragment>
+        <Text style={styles.label}>Share</Text>
+        <BarItem onPress={this.onShare}>
+          {isLoading ? (
+            <ActivityIndicator size="small" style={styles.activityLoader} />
+          ) : (
+            <IconEmail
+              fillColour="currentColor"
+              height={styles.svgIcon.height}
+              title="Share by email client"
```
aronshamash: minor `Share by email`

packages/user-state/README.md
aronshamash: nice documentation though :)

packages/user-state/__tests__/matchers.test.js
```diff
@@ -0,0 +1,209 @@
+import {
+  isLoggedIn,
+  isMeteredExpired,
+  isShared,
+  isMetered,
+  isNonExpiredUser,
```
aronshamash: I think you _may_ have gone through this in meeting but there is no such thing as an expired user !

packages/user-state/src/storybook-helper.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-env browser */
+
+import { select, boolean } from "@storybook/addon-knobs/react";
+import { SUBSCRIBER, RA, RA_EXPIRED, GUEST } from "./states";
+
+const activeStatesPerUserState = {
+  [SUBSCRIBER]: ["isLoggedIn"],
+  [RA]: ["isLoggedIn", "isMetered"],
+  [RA_EXPIRED]: ["isLoggedIn", "isMetered", "isMeteredExpired"],
+  [GUEST]: []
```
aronshamash: guest is a better way of putting it that `teaser` and `logged-out` 👍

packages/article-skeleton/src/article-skeleton.web.js
nathggns: This is an `or`, not an `and` – the save and share bar will show in either case. 

This is just moving the existing logic and it should be covered by the existing tests.

packages/save-and-share-bar/src/email-share.js
nathggns: This is just moving the existing email icon code. This is a potential issue, but its not relevant to this PR. Good catch, we should file an issue for it.

aronshamash: Agreed JIRA ticket to fix https://nidigitalsolutions.jira.com/browse/REPLAT-6792

packages/article-skeleton/src/article-skeleton.web.js
aronshamash: sorry yes ignore

PR #2022, opened 2019-06-07
Update readme and install script for fonts

mikefieldmay:
Slight update to the Readme and an install script based on the process I had to go through to get the app running. 

I'm not that familiar with `python`'s `pip` so if this doesn't look right I'm happy to revert

### DIFF
--- a/packages/text-flow/lib/process-fonts.sh
+++ b/packages/text-flow/lib/process-fonts.sh
@@ -1,5 +1,5 @@
 #! /usr/bin/env bash
 
-pip install bs4 lxml
+pip install --user bs4 lxml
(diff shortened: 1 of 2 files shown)


packages/text-flow/lib/process-fonts.sh
```diff
@@ -1,5 +1,5 @@
 #! /usr/bin/env bash
 
-pip install bs4 lxml
+pip install --user bs4 lxml
```
parkjh: was this causing any issue? @mikefieldmay

jmars: For one, it wont work on linux without sudo unless --user is present

PR #2027, opened 2019-06-13
fix: ios storybook fonts issue 

nathggns:
It seems Tunca's upgrade to and subsequent reversion of react native resulted in the loss of font references in the iOS storybook xcode project. This PR restores those references and in doing so fixes iOS storybook.


ValGeorgiev:
LGTM, but could we retest webviews on iOS storybook before merge. Thanks

ValGeorgiev:
Tested it locally, looks good

PR #2028, opened 2019-06-13
fix: make unruly ads work correctly in the new webview component

jmars:
This updates the properties on the webview component has the new one has different defaults to the old one.

image

### DIFF
--- a/packages/ad/__tests__/android/__snapshots__/ad-with-style.android.test.js.snap
+++ b/packages/ad/__tests__/android/__snapshots__/ad-with-style.android.test.js.snap
@@ -14,7 +14,7 @@ Array [
         <RNCWebView
           allowFileAccess={false}
           allowsFullscreenVideo={false}
-          androidHardwareAccelerationDisabled={true}
+          androidHardwareAccelerationDisabled={false}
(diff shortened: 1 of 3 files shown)


ValGeorgiev:
Could you add a description to the pr, with some screenshots, please

packages/ad/__tests__/android/__snapshots__/ad-with-style.android.test.js.snap
```diff
@@ -14,7 +14,7 @@ Array [
         <RNCWebView
           allowFileAccess={false}
           allowsFullscreenVideo={false}
-          androidHardwareAccelerationDisabled={true}
+          androidHardwareAccelerationDisabled={false}
```
parkjh: @jmars why was this required?

packages/ad/src/dom-context.js
```diff
@@ -123,7 +123,7 @@ class DOMContext extends PureComponent {
         }}
       >
         <WebView
-          androidHardwareAccelerationDisabled
+          originWhitelist={["http://.*", "https://.*"]}
```
parkjh: lol

PR #2029, opened 2019-06-14
Fix/inline image small size

pdimova:
https://nidigitalsolutions.jira.com/browse/REPLAT-6744

Seems that the native implementation of the ArticleImage was rendered instead of the web one.
Also the clearfix for the dropcap was causing only one paragraph to wrap around an inline image

Before:
Jun-14-2019 13-46-37

After:
Jun-14-2019 13-50-21


packages/article-image/src/article-image.web.js
```diff
@@ -0,0 +1,3 @@
+import ArticleImage from "./article-image.base";
+
+export default ArticleImage;
```
parkjh: what's the purpose of this? @pdimova

PR #2035, opened 2019-06-17
fix: storybook fixes and knobs for native added

infreezer:
1. [Knobs ondevice addon](https://www.npmjs.com/package/@storybook/addon-ondevice-knobs) added for storybook native: 

image

2. Storybook knobs order reverser removed in order to fix broken dropdowns for both web and mobile ([source](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#knobs-select-ordering)):

storybook-select


packages/storybook/src/knobs.js
```diff
@@ -1,9 +1,7 @@
 import * as builtInKnobs from "@storybook/addon-knobs/react";
-import select from "./select-shim";
 
 const knobs = {
-  ...builtInKnobs,
-  select
+  ...builtInKnobs
 };
 
 export default knobs;
```
ValGeorgiev: Could we export builtInKnobs directly?

packages/article-paragraph/article-paragraph.showcase.js
```diff
@@ -12,7 +12,7 @@ import ArticleParagraph from "./src";
 
 const renderParagraphWithScale = (select, ast) => {
   const scale = select("Scale", scales, scales.medium);
-  const section = sections[select("Section", sections, sections.default)];
+  const section = select("Section", sections, "The Sunday Times Magazine");
```
ValGeorgiev: Please change it to The Times Magazine

packages/pull-quote/pull-quote.showcase.js
```diff
@@ -22,7 +22,11 @@ export default {
   children: [
     {
       component: ({ color, select, text }, { decorateAction }) => {
-        const section = sections[select("Section", sections, sections.default)];
+        const section = select(
+          "Section",
+          sections,
+          "The Sunday Times Magazine"
```
ValGeorgiev: The Times Magazine as well

packages/storybook/src/knobs.js
infreezer: Sure.

packages/article-paragraph/article-paragraph.showcase.js
infreezer: Thanks.

packages/pull-quote/pull-quote.showcase.js
infreezer: Good point.

packages/article-label/article-label.showcase.js
```diff
@@ -11,7 +12,7 @@ export default {
         <ArticleLabel
           color={select(
             "Section",
-            invert(colours.section),
+            pick(colours.section, sections),
```
parkjh: @infreezer was this required for the native knobs to work?

infreezer: Yes. Removing the invert function was required for both web and native.
Reversing the object order was a fix in earlier versions of storybook, which is not required anymore, because in 4.0 the order is the correct one.
About the pick function, I'm using it to filter, beautify and make the object more human readable, because there are duplicates.

PR #2038, opened 2019-06-17
feat:web-add analytics to inline links

Ivaylo1987:
This is for [Add analytics to inline web links](https://nidigitalsolutions.jira.com/browse/REPLAT-4505)

`New:`
Inline links should be tracked on web now.

`Changed:`
Simplified tracking on related articles on both mobile and web.

### DIFF
--- a/packages/related-articles/src/related-articles-tracking-context.js
+++ b/packages/related-articles/src/related-articles-tracking-context.js
@@ -17,48 +17,30 @@ export default Component =>
         supports
           .filter(support => support !== undefined)
           .map(({ article }, index) => {
-            const {
-              bylines,
-              headline,
-              id,
-              publishedTime,
-              shortHeadline
-            } = article;
+            const { headline, id, shortHeadline } = article;
             return {
-              bylines,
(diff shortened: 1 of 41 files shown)


packages/related-articles/src/related-articles-tracking-context.js
```diff
@@ -69,12 +51,10 @@ export default Component =>
 
       const opinionOneAndTwoTracking = () => {
         const { article } = opinion;
-        const { bylines, headline, id, publishedTime, shortHeadline } = article;
+        const { headline, id, shortHeadline } = article;
```
nehasri89: `byline` and `publishedTime` are passed on as trackable object as per the PR, so should stay?

Ivaylo1987: As per the requirements in the [ticket](https://nidigitalsolutions.jira.com/browse/REPLAT-4505) no.
This is the simplification of the track object that was requested by PO.

nehasri89: :+1

PR #2039, opened 2019-06-17
fix: share-save-bar at the bottom of article page

apostolnikov:
Implement https://nidigitalsolutions.jira.com/browse/REPLAT-6742.

In this PR Share-and-Save-bar is added at the bottom of the article page between topic links and related articles. This is what's missing from the implementation of this one:  https://nidigitalsolutions.jira.com/browse/REPLAT-6112.

The repeated bar should only be visible on desktop resolution.

Storybook:
image

Render:
image

image


packages/article-extras/src/styles/responsive.web.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable import/prefer-default-export */
```
ValGeorgiev: could you use the default export here and rename the file, please. Responsive.web.js doesn't sound like a good file name here.

apostolnikov: responsive.web.js is the naming convention for using media queries with styled components for our whole project. You can see such file in every single package that needs to behave differently under different resolution sizes. Renaming the file won't stick to the project structure.

packages/article-extras/src/article-extras.web.js
```diff
@@ -3,20 +3,49 @@ import PropTypes from "prop-types";
...
+  articleHeadline,
   relatedArticleSlice,
   relatedArticlesVisible,
   spotAccountId,
   topics
 }) => (
   <UserState state={UserState.fullArticle}>
     <ArticleTopics topics={topics} />
+    {(savingEnabled || sharingEnabled) && (
+      <UserState state={UserState.loggedInOrShared}>
```
parkjh: @apostolnikov just so that I understand, do you still need this user state check after `<UserState state={UserState.fullArticle}>` check above?

parkjh: @nathggns is there no way to combine multiple user state checks?

ValGeorgiev: @parkjh savingEnabled & sharingEnabled are graphql fields and they don't have anything related to the userstate

packages/article-skeleton/src/sticky-save-and-share-bar.js
```diff
@@ -1,22 +1,17 @@
...
 import SaveAndShareBar from "@times-components/save-and-share-bar";
-import { saveApi as saveArticleApi } from "@times-components/save-star-web";
 import Sticky from "@times-components/sticky";
 import {
   SaveShareContainer,
   SaveShareInnerContainer
 } from "./styles/responsive";
 
-function StickySaveAndShareBar({ saveApi, ...props }) {
-  const api = saveApi && saveApi.bookmark ? saveApi : saveArticleApi;
```
parkjh: 👍

packages/article-extras/src/article-extras.web.js
parkjh: @ValGeorgiev but they are accessed via user state no?

ValGeorgiev: Query: https://github.com/newsuk/times-components/blob/4f3041a47abc0b3b5494689dbe9844cb6fd4ea1c/packages/provider-queries/src/article-fragment.js#L36

No, they are taken from TPA and they are article fields.

parkjh: just to confirm, I don't mean `savingEnabled || sharingEnabled` I mean `UserState.loggedInOrShared`

parkjh: @nathggns just to check, the nested user state components will not cause any issue with 2 pass rendering will it?

packages/article-skeleton/src/sticky-save-and-share-bar.js
nehasri89: Please test `save` on storybook article pages because this change was to make it functional there

parkjh:
@apostolnikov also in the future please link these pull requests in the slack channel rather than just moving the jira ticket to `in review`, I say this because this gives visibility for the rest of the team...

PR #2040, opened 2019-06-17
Feature/replat 3983 add alt tag to images

mikefieldmay:
Jira ticket: https://nidigitalsolutions.jira.com/browse/REPLAT-3983

Add alt text to images for article pages

### DIFF
--- a/packages/article-image/src/inline-image.web.js
+++ b/packages/article-image/src/inline-image.web.js
@@ -38,6 +38,7 @@ const InlineImage = ({ imageOptions, captionOptions }) => {
     <figure style={{ margin: 0 }}>
       <InsetImageStyle key="img">
         <Image
+          alt={caption}
           aspectRatio={aspectRatio}
           highResSize={highResSize}
           lowResSize={lowResSize}
--- a/packages/article-image/src/inline-image.web.js
+++ b/packages/article-image/src/inline-image.web.js
@@ -22,6 +22,7 @@ const renderCaption = (display, caption, credits) => {
 };
 
 const InlineImage = ({ imageOptions, captionOptions }) => {
+  console.log("CAPTION", caption)
   const { display, highResSize, lowResSize, ratio, uri } = imageOptions;
   const { caption, credits } = captionOptions;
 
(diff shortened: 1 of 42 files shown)


parkjh:
@mikefieldmay is this WIP?

mikefieldmay:
@parkjh yeah, sorry should have marked it as such

packages/image/src/modal-image/modal-image-prop-types.js
```diff
@@ -10,6 +10,7 @@ export const modalPropTypes = {
 
 export const modalDefaultProps = {
   ...defaultProps,
+  accessibilityLabel: "",
```
parkjh: is default value really necessary?

packages/article-image/src/inline-image.web.js
```diff
@@ -38,6 +38,7 @@ const InlineImage = ({ imageOptions, captionOptions }) => {
     <figure style={{ margin: 0 }}>
       <InsetImageStyle key="img">
         <Image
+          alt={caption}
```
parkjh: don't you mean `caption.text`?

packages/image/src/image-prop-types.js
```diff
@@ -24,3 +25,8 @@ export const defaultProps = {
   rounded: false,
   uri: null
 };
+
+export const webDefaultProps = {
+  ...defaultProps,
+  accessibilityLabel: ""
```
parkjh: I think as a team we decided against setting component default props... as this was going to be phased out by facebook

parkjh: also why have empty string here?

packages/article-image/src/inline-image.web.js
nehasri89: no, we get the markup as `caption`

packages/image/src/modal-image/modal-image-prop-types.js
mikefieldmay: Yeah you're right I don't think it is

packages/image/src/image-prop-types.js
mikefieldmay: I'll remove these default props too

PR #2043, opened 2019-06-17
fix: legacy iOS

parkjh:
Screenshot 2019-06-17 at 18 04 24

### DIFF
--- a/packages/schema/fetch-gql-schema.js
+++ b/packages/schema/fetch-gql-schema.js
@@ -31,7 +31,7 @@ const writeFragmentMatcher = (cwd, schema) => {
   );
 
   const fm = `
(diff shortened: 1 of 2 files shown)


packages/schema/fetch-gql-schema.js
```diff
@@ -31,15 +31,15 @@ const writeFragmentMatcher = (cwd, schema) => {
   );
 
   const fm = `
-  const { IntrospectionFragmentMatcher } = require("apollo-cache-inmemory");
+  const apolloCacheInmemory = require("apollo-cache-inmemory");
```
aronshamash: good spot

aronshamash: :)

PR #2053, opened 2019-06-20
fix: save star positioning

ValGeorgiev:
New save star positioning:
image

image


aronshamash:
@ValGeorgiev No snapshot changes ?

PR #2059, opened 2019-06-20
fix: replat 6693 twitter metadata

parkjh:
jira: https://nidigitalsolutions.jira.com/browse/REPLAT-6693

- use smaller lead asset image for social media meta tag

### DIFF
--- a/packages/article-skeleton/src/head.web.js
+++ b/packages/article-skeleton/src/head.web.js
@@ -72,7 +72,7 @@ function appendToUrl(uriString, key, value) {
   }
 
   return `${uriString}?${key}=${value}`;
-};
+}
 
 const PUBLICATION_NAMES = {
   SUNDAYTIMES: "The Sunday Times",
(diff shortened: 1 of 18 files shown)


packages/article-skeleton/src/head.web.js
```diff
@@ -96,7 +96,7 @@ function Head({ article, paidContentClassName, faviconUrl }) {
       ? renderTreeAsText({ children: descriptionMarkup })
       : null;
   const sectionname = getSectionName(article);
-  const leadassetUrl = get(leadAsset, "crop169.url", null);
+  const leadassetUrl = get(leadAsset, "crop32.url", null);
```
ValGeorgiev: this will be applied for all social media, not only twitter. Is this what we want. Did we tested it on FB  as well. Could we see any snapshots in the description?

parkjh:
related https://github.com/newsuk/cps-content-render/pull/2806

PR #2060, opened 2019-06-20
feat: add polyfill for web components in interactive wrapper

SunshineEleda:
## To Do
- fix snapshot test(I'm not sure how the original tests work so I'm unable to fix it)
- get it working on render


nathggns:
@ValGeorgiev there's no description because this is a WIP and should not be reviewed. It was created as a PR because it was being handed over to me. 

That being said, I take your point about the size of `webcomponent-line.js`. I'll change it so its dynamically loaded from the CDN only when needed, rather than in thwe pacjage on every page on every browser. That has its trade offs too in terms of time to load interactives on browsers that require it, but I think its better to avoid loading that code when not needed.

nathggns:
Going to make a new PR.

PR #2061, opened 2019-06-20
fix: Backwards dropcaps on the web

jmars:
Fixes the dropcaps being backwards on articles with markup dropcaps.
image


packages/link/src/link.web.js
```diff
@@ -10,12 +10,19 @@ const RespLink = responsiveLinkStyles =>
...
+  target,
+  url,
+  underlined
+}) => {
   const Wrapper =
     responsiveLinkStyles !== null ? RespLink(responsiveLinkStyles) : "a";
 
   const style =
-    responsiveLinkStyles !== null
+    underlined && responsiveLinkStyles !== null
```
GaboDimitrov: Do you need the **!== null** part here?

packages/article-skeleton/src/article-body/article-link.web.js
```diff
@@ -19,9 +20,23 @@ const responsiveLinkStyles = {
...
 };
 
-const ArticleLink = ({ children, target, url, onPress }) => (
+const dropCapLinkStyles = {
+  base: `
+    font-size: inherit;
+    text-decoration: none;
+    color: ${colours.functional.action};
+  `,
+  medium: `
```
aronshamash: can we DRY this up with a common string? Does it need to be repeated for medium anyway ?

packages/article-skeleton/src/dropcap-util.web.js
```diff
@@ -8,7 +8,7 @@ export default (children, template, isDropcapDisabled) => {
   );
   if (newContent.length > 0 && newContent[0].name === "dropCap") {
     // remove the wrapping paragraph
-    newContent[0].children = newContent[0].children[0].children;
+    newContent[0].children = newContent[0].children[0].children.slice(0, 1);
```
aronshamash: I was expecting a new test case as we needed to change the implementation. Is it not needed ?

parkjh: @aronshamash are you still working on this?

PR #2064, opened 2019-06-21
fix/ios-slices-feedback

Ivanov558:
These fixes are for point 2c and 3a. The other issues cannot be reproduce.

2c:
image

3a:
image

### DIFF
--- a/packages/edition-slices/src/tiles/shared/tile-summary.js
+++ b/packages/edition-slices/src/tiles/shared/tile-summary.js
@@ -1,5 +1,5 @@
 import React, { Component } from "react";
-import { View } from "react-native";
+import { View, Platform } from "react-native";
 import PropTypes from "prop-types";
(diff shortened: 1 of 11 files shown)


packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -1,5 +1,5 @@
 import React, { Component } from "react";
-import { View } from "react-native";
+import { View, Platform } from "react-native";
```
aronshamash: Can we not use an android specific file `tile-summary.android.js` with these changes in order to not have to test for `Platform`.

parkjh: @aronshamash I don't think it's good idea to start fragmenting components for the tiles, given how many tiles types there are...

PR #2069, opened 2019-06-24
Takeover pages

nehasri89:
The feature is to allow Takeover pages to be served from Content API, the way to do it is it pretends to be a react page, but TC throws an exception for the this template type `takeoverPage` which the render catches and reverts back to the old rendering mechanism using content api and dust template. 

`takeoverPage`  a new template type introduced to identify takeover pages.

The corresponding Render PR https://github.com/newsuk/cps-content-render/pull/2811


packages/ssr/src/server/article.js
```diff
@@ -62,5 +66,7 @@ module.exports = (
     name: "article"
   };
 
-  return runServer(article, options);
+  return runServer(article, options).catch(error => {
+    throw new Error(error);
```
parkjh: @nehasri89 and is this suppose to be caught in render?

packages/article/src/article.js
```diff
@@ -17,11 +17,22 @@ export const templates = {
   mainstandard: ArticleMainStandard
 };
 
+export class TakeoverBailout extends Error {
```
parkjh: @nehasri89 do we not have a more concise error handling package?

packages/ssr/src/server/article.js
nehasri89: yup

packages/article/src/article.js
nehasri89: not at the moment

ValGeorgiev: We should use ErrorBoundaries https://reactjs.org/docs/error-boundaries.html

parkjh: @ValGeorgiev @nehasri89 maybe as a separate piece of work? but I would have assumed that error handling would've been one of the first thing to build as a package...

PR #2070, opened 2019-06-25
fix: native edition query

ValGeorgiev:
This is fixing the native edition query

Error:
```
{“errors”:[{“message”:“There can be only one fragment named \“cropProps\“.”,“locations”:[{“line”:132,“column”:12},{“line”:1655,“column”:12}],“extensions”:{“code”:“GRAPHQL_VALIDATION_FAILED”}}]}
```

Explanation: 
In the native-edition we are using both section and article query. Thats why these fragments needs to be with different names.

### DIFF
--- a/packages/provider-queries/src/article-fragment.js
+++ b/packages/provider-queries/src/article-fragment.js
@@ -52,31 +52,31 @@ export default gql`
     caption
     credits
     crop169: crop(ratio: "16:9") {
-      ...cropProps
+      ...articleCropProps
     }
     crop32: crop(ratio: "3:2") {
-      ...cropProps
+      ...articleCropProps
     }
     crop1251: crop(ratio: "1.25:1") {
-      ...cropProps
+      ...articleCropProps
     }
     crop11: crop(ratio: "1:1") {
-      ...cropProps
+      ...articleCropProps
     }
     crop45: crop(ratio: "4:5") {
-      ...cropProps
+      ...articleCropProps
     }
     crop23: crop(ratio: "2:3") {
-      ...cropProps
+      ...articleCropProps
     }
     crop2251: crop(ratio: "2.25:1") {
-      ...cropProps
+      ...articleCropProps
     }
     id
     title
   }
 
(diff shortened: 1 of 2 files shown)


benjohnbarnes:
Nice one 👍

A small note from me – is this the right way to resolve this? It may well be! I just noticed that `articleCropProps` and `sectionCropProps` are the same, though. 

The other way to resolve it is to have just `crtopProps` as before, and ensure only one definition for the fragment ends up in the query. I'm not sure how we'd achieve that, though. 🤔

ValGeorgiev:
> Nice one 👍
> 
> A small note from me – is this the right way to resolve this? It may well be! I just noticed that `articleCropProps` and `sectionCropProps` are the same, though.
> 
> The other way to resolve it is to have just `crtopProps` as before, and ensure only one definition for the fragment ends up in the query. I'm not sure how we'd achieve that, though. 🤔

Both of them are used in different queries as well, like article and section queries. In future if we want to change only the section fragment it will be easier and won't impact the article query. Thats the reason behind my solution

parkjh:
@ValGeorgiev under what scenario does this happen?

ValGeorgiev:
> @ValGeorgiev under what scenario does this happen?

When we try to fetch graphql edition on ios app. Probably @benjohnbarnes will need to apply the same change in ios repo to make it works, because native-edition query is only for reference in the TC.

parkjh:
> > @ValGeorgiev under what scenario does this happen?
> 
> When we try to fetch graphql edition on ios app. Probably @benjohnbarnes will need to apply the same change in ios repo to make it works, because native-edition query is only for reference in the TC.

what about android app?

ValGeorgiev:
> > > @ValGeorgiev under what scenario does this happen?
> > 
> > 
> > When we try to fetch graphql edition on ios app. Probably @benjohnbarnes will need to apply the same change in ios repo to make it works, because native-edition query is only for reference in the TC.
> 
> what about android app?

I cannot reproduce the issue there, probably they fixed it

PR #2071, opened 2019-06-25
fix: change share-save-bar and topic-links width

apostolnikov:
JIRA: https://nidigitalsolutions.jira.com/browse/REPLAT-6742

Changed width for topic links and share-save-bar.

Storybook: 
image

Render:
image

Tablet width:
image

### DIFF
--- a/packages/article-extras/src/styles/responsive.web.js
+++ b/packages/article-extras/src/styles/responsive.web.js
@@ -6,7 +6,7 @@ export const ShareAndSaveContainer = styled.div`
   border-top-color: ${colours.functional.keyline};
   border-top-style: solid;
   border-top-width: 1px;
-  width: 80.8%;
+  width: 56.2%;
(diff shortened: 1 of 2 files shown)


packages/article-extras/src/styles/responsive.web.js
```diff
@@ -6,7 +6,7 @@ export const ShareAndSaveContainer = styled.div`
   border-top-color: ${colours.functional.keyline};
   border-top-style: solid;
   border-top-width: 1px;
-  width: 80.8%;
+  width: 56.2%;
```
parkjh: @apostolnikov no snapshots to update? strange...

apostolnikov: @parkjh  Yup, we'll need to think of a solution for those media queries to be updated in the snapshots as well. As for now we don't have such.

PR #2079, opened 2019-06-26
chore: upgrade TC in iOS app readme

parkjh:


parkjh:
@aronshamash see https://github.com/newsuk/nu-digital-projectd-times-smartphone-ios/pull/1587/files

PR #2081, opened 2019-06-26
feat: add polyfill for web components in interactive wrapper

nathggns:
I've added a polyfill for web components that is loaded from the CDN only when necessary, as its unfortunately relatively large. This polyfill is necessary for it to work on Firefox and Safari, and in future versions of Chrome.

Therefore, I've also added a placeholder to show while the polyfill is loaded so that users don't see the interactives flashing in from nowhere, as it may be a few seconds before they can load depending on connection speed. 

Screen Shot 2019-06-26 at 13 34 13
Screen Shot 2019-06-26 at 13 34 22


packages/image/src/placeholder/placeholder.js
```diff
@@ -1,10 +1,11 @@
+/* eslint-disable react/require-default-props */
 import React from "react";
 import PropTypes from "prop-types";
 import { View } from "react-native";
 import styles from "../styles/index";
 import T from "../logo/t";
 
-function Placeholder({ borderRadius }) {
+function Placeholder({ borderRadius = 0 }) {
```
parkjh: please use default props if you must

packages/interactive-wrapper/src/interactive-wrapper.web.js
```diff
@@ -2,44 +2,141 @@
...
+    newElement.outerHTML += "";
 
-    delete this.placeholder.current;
+    placeholder.style.cssText += "display: none !important";
   }
 
   render() {
-    return <div ref={this.placeholder} />;
+    return (
+      <>
```
parkjh: what tag is this?

packages/interactive-wrapper/.eslintignore
```diff
@@ -2,3 +2,4 @@
 **/coverage/*
 dist
 rnw.js
+src/webcomponent-lite.js
```
parkjh: is this required during development?

packages/interactive-wrapper/src/interactive-wrapper.web.js
```diff
@@ -2,44 +2,141 @@
...
     );
 
-    parentNode.replaceChild(newElement, placeholder);
-    parentNode.insertBefore(link, newElement);
+    component.appendChild(newElement);
+
+    // Do not remove this. This seems to notify polymer to correctly
+    // render the web component in more circumstances, specifically,
+    // its required to correctly re-render after a react re-render
+    newElement.outerHTML += "";
```
parkjh: is there a corresponding GitHub issue that track this issue?

packages/interactive-wrapper/src/interactive-wrapper.web.js
```diff
@@ -2,44 +2,141 @@
...
     this.placeholder = React.createRef(null);
+    this.component = React.createRef(null);
+    this.whenReady = null;
   }
 
-  componentDidMount() {
+  async componentDidMount() {
+    const { fetchPolyfill } = this.props;
+
+    this.whenReady = fetchPolyfill();
```
parkjh: no await here?

packages/image/src/placeholder/placeholder.js
nathggns: The makers of React discourage the use of defaultProps, especially for function components, and have signified a number of times that defaultProps will be deprecated in the future. Why not use a built in language feature instead? (It's also just better because the default is declared closer to the use site)

packages/interactive-wrapper/src/interactive-wrapper.web.js
nathggns: Fragment. https://reactjs.org/docs/fragments.html

packages/interactive-wrapper/.eslintignore
nathggns: Ooops. I guess that was added my adele and I forgot to remove it. Good catch.

packages/interactive-wrapper/src/interactive-wrapper.web.js
nathggns: It doesn't need to be awaited here, it gets awaited in `insertComponent` instead. That's because `insertComponent` is the function that has the dependency on the script having been loaded. `insertComponent` is also called from `componentDidUpdate` too. If the await was in `componentDidMount` instead, then there'd a potential for `insertComponent` to be called before the script has actually loaded if props changed while it was loading, unless `componentDidUpdate` also awaited for it. I just thought it made sense to do the await inside `insertComponent` instead as that's the common method and I can do the await in one place instead of two.

packages/interactive-wrapper/src/interactive-wrapper.web.js
nathggns: Not that I could find. I found this fix myself. I don't know how or why it works, but it does. But this whole componeny is inherently hacky in and of itself and by design anyway and will hopefully go away one day, so I didn't mind too much. I don't know or understand Polymer/web components, so I didn't want to do too much of a deep-dive into it. It felt like a waste of time.

packages/image/src/placeholder/placeholder.js
parkjh: hmm... this all seems quite inconsistent...

nathggns: https://twitter.com/dan_abramov/status/1133878326358171650

https://twitter.com/dan_abramov/status/1133883865930567680

nathggns: Here's the RFC proposing deprecating `defaultProps` on function components. https://github.com/reactjs/rfcs/pull/107

parkjh: no I understand that, it's just we still have default props in the repo

nathggns: Discussed on Slack.

packages/interactive-wrapper/.eslintignore
nathggns: Fixed.

packages/interactive-wrapper/src/interactive-wrapper.web.js
parkjh: @chrishutchinson cc

packages/interactive-wrapper/src/interactive-wrapper.web.js
```diff
@@ -2,44 +2,141 @@
...
+}
+
+export function polyfillWCIfNecessary() {
+  const htmlImportsSupported = "import" in document.createElement("link");
+  const registerElementSupported = !!document.registerElement;
+
+  if (!htmlImportsSupported || !registerElementSupported) {
+    return Promise.all([
+      ensureScript(
+        "https://cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/0.7.24/webcomponents-lite.min.js"
```
parkjh: @chrishutchinson is this (webcomponents) something that you guys must use in the interactives?

packages/interactive-wrapper/.prettierignore
```diff
@@ -12,3 +12,4 @@ LICENSE
 yarn.lock
 *.log
 rnw.js
+src/webcomponent-lite.js
```
ValGeorgiev: Could we revert this as well?

packages/interactive-wrapper/__tests__/web/interactive-wrapper.test.js
```diff
@@ -5,24 +5,239 @@ import {
...
+    const promise = Promise.resolve(null);
+
+    const component = mount(
+      <InteractiveWrapper {...props} fetchPolyfill={() => promise} />,
+      { attachTo: container }
+    );
+
+    await component.instance().componentDidMount();
+
+    component.instance().componentDidUpdate();
```
ValGeorgiev: Manually invoking react lifecycle methods? This doesn't look right

packages/interactive-wrapper/src/interactive-wrapper.web.js
```diff
@@ -2,44 +2,141 @@
...
 import React, { Component } from "react";
 import PropTypes from "prop-types";
+import { Placeholder } from "@times-components/image";
+
+function ensureElement(selector, createElement) {
+  let element = document.body.querySelector(selector);
+
+  if (element) {
+    return Promise.resolve();
+  }
```
ValGeorgiev: Could we do
```
if ( document.body.querySelector(selector)) {
   return Promise.resolve()
}

return new Promise((resolve, reject) => {
   const element = createElement();
   element.onload = resolve;
   element.onerror = reject;
   
   document.body.appendChild(element);
});

Comment (nathggns) [reply packages/interactive-wrapper/src/interactive-wrapper.web.js]: 👍

File: packages/interactive-wrapper/.prettierignore (hunk 7 above)
Comment (nathggns) [reply packages/interactive-wrapper/.prettierignore]: 👍

File: packages/interactive-wrapper/__tests__/web/interactive-wrapper.test.js (hunk 8 above)
Comment (nathggns) [reply packages/interactive-wrapper/__tests__/web/interactive-wrapper.test.js]: 👍

File: packages/image/src/logo/t.web.js [hunk 10]
```diff
@@ -4,7 +4,13 @@ import { Svg, G, Path } from "@times-components/svgs";
 
 function T() {
   return (
-    <Svg height="auto" version="1.1" viewBox="145 50 108 120" width="25%">
+    <Svg
```
aronshamash: Just checking this still works on list pages etc now we have a maxHeight ?

packages/interactive-wrapper/__tests__/web/interactive-wrapper.test.js
```diff
@@ -5,24 +5,229 @@ import {
...
+    component.setProps({ element: "another-test-element" });
+
+    document.querySelector("link").onload();
+    await polyfillPromise;
+
+    await delay(0);
+
+    expect(container.querySelectorAll("another-test-element")).toHaveLength(1);
+    expect(container.querySelectorAll("test-element")).toHaveLength(0);
+  });
```
aronshamash: nice tests :)

packages/image/src/logo/t.web.js
nathggns: Yeah looks fine!

PR #2082, opened 2019-06-26
fix: remove textflow storyboook

ValGeorgiev:
JIRA: https://nidigitalsolutions.jira.com/browse/REPLAT-6881

Remove textflow storybook, because the package is used in other packages and there is a good readme for the package for more details


parkjh:
@ValGeorgiev I think @nathggns  said that this is showcased in `article-paragraph`? if so can we just confirm that all the use cases are there?

PR #2085, opened 2019-06-27
Fix/share and save bar remove duplicate keyline

apostolnikov:
Implement https://nidigitalsolutions.jira.com/browse/REPLAT-6874.

Storybook: 
image

Render with user (share-save-bar is visible): 
image

Render without user (share-save-bar is not visible)
image


nathggns:
reviewing now

nathggns:
The meta text doesn't seem to be vertically centered within the two borders- I don't know if that's a concern. 
image

packages/article-in-depth/src/article-in-depth.web.js
```diff
@@ -59,6 +64,10 @@ class ArticlePage extends Component {
             textColour={textColour}
           />
         </HeaderContainer>
+        <UserState
+          state={UserState.loggedInOrShared}
```
nathggns: I have concerns about this as if article skeleton changes what user state matcher it uses, we can end up with really weird results in terms of where borders are/aren't. Maybe save/share bar should be moved out of the skeleton and into each template (as its technically part of the header?). Going to leave this for now but I am concerned.

packages/article-in-depth/src/styles/index.web.js
```diff
@@ -1,8 +1,22 @@
...
 const styles = StyleSheet.create({
-  ...sharedStyles
+  ...sharedStyles,
+  leadAsset: {
+    marginBottom: spacing(4)
+  },
+  metaContainer: {
+    alignItems: "center",
+    marginHorizontal: spacing(2),
+    paddingBottom: spacing(2),
```
aronshamash: `paddingVertical`

aronshamash:
other than minor improvement I'm ok to approve

PR #2088, opened 2019-06-28
Chor/replat 6866 dropcap storybook

GaboDimitrov:
This PR adds drop cap to storybook native and updates the web storybook.

1.Web storybook:
- Normal drop cap
image
- Link drop cap
image
- Quote drop cap
image

2. Native storybook
image


packages/article-paragraph/src/article-paragraph.js
```diff
@@ -19,7 +19,12 @@ const BodyParagraph = props => (
         <Context.Consumer>
           {({ theme: { dropCapFont, scale } }) => {
             const stylesScaled = styleFactory(dropCapFont, scale);
-            return props.children(stylesScaled.articleTextElement);
+
+            if (typeof props.children === "function") {
```
aronshamash: I wasn't expecting a change to the component. Is this a bug fix ?

GaboDimitrov: Without this change, the component was throwing an error on native devices.

PR #2089, opened 2019-06-28
Fix: add underline to dropcap with inline links

SunshineEleda:
## Changes

- added the underline to the link when the first letter is a dropcap

Screen Shot 2019-06-28 at 13 59 06


packages/article-skeleton/src/article-body/article-link.web.js
```diff
@@ -9,7 +9,7 @@ import withTrackEvents from "./article-link-tracking-events";
 
 const ArticleLink = ({ children, target, url, onPress, dropCap }) => (
   <Link
-    underlined={!dropCap}
+    underlined={!(dropCap && children[0].length === 1)}
```
aronshamash: Can you add a snapshot test for this please

packages/article-skeleton/__tests__/web/__snapshots__/article-link.web.test.js.snap
```diff
@@ -0,0 +1,125 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Article Link should render with the dropCap link not underlined 1`] = `
+initialize {
```
nathggns: You need to add the correct serializers so this serializes as JSX. Check some of our other web tests to figure out how to do that.

PR #2090, opened 2019-06-28
fix: ad slot position

parkjh:
jira: https://nidigitalsolutions.jira.com/browse/REPLAT-1407

fixes `inline-ad` for articles

### DIFF
--- a/packages/ad/src/utils/generate-config.js
+++ b/packages/ad/src/utils/generate-config.js
@@ -21,7 +21,8 @@ const slotPositions = {
   "ad-pixelteads": 2,
   "article-sponsored-ad": 2,
   default: 0,
-  header: 1,
+  header: 0,
+  "inline-ad": 1,
   pixel: 2,
   pixelskin: 2,
   pixelteads: 2
--- a/packages/ad/src/utils/generate-config.js
+++ b/packages/ad/src/utils/generate-config.js
@@ -21,8 +21,8 @@ const slotPositions = {
   "ad-pixelteads": 2,
   "article-sponsored-ad": 2,
   default: 0,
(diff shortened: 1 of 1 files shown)


packages/ad/src/utils/generate-config.js
```diff
@@ -21,7 +21,8 @@ const slotPositions = {
   "ad-pixelteads": 2,
   "article-sponsored-ad": 2,
   default: 0,
-  header: 1,
+  header: 0,
```
aronshamash: is this correct having `header` and `default` set to `0` ?

aronshamash: Is there a jira ticket for the bug ?

parkjh: @aronshamash I'm doing it as part of https://nidigitalsolutions.jira.com/browse/REPLAT-1407

aronshamash: thanks. Is it working now on staging ?

parkjh: @aronshamash what is working on staging?

aronshamash: @parkjh does the ad show?

parkjh: @aronshamash so it's only showing for the header but the error is that the slot name for the ad is called `inline-ad` even though it should be `header`... conversely the inline ad is empty as it does not have any ad slot id...

parkjh: so it doesn't really work on staging yet but I'm going to add a screenshot of my local render working correctly

PR #2091, opened 2019-06-28
fix: correct ad loading behaviour

parkjh:
jira: https://nidigitalsolutions.jira.com/browse/REPLAT-1407

This separates googletag page and slot targeting, due to the way ad components are set up in react pages

### DIFF
--- a/packages/ad/__tests__/utils/ad-init.js
+++ b/packages/ad/__tests__/utils/ad-init.js
@@ -46,9 +46,15 @@ export default () => {
   it("calls all the vender functions when initially executed", () => {
     const initFirstRun = adInit(initOptions);
 
-    jest.spyOn(initFirstRun.apstag, "process").mockReturnValueOnce(Promise.resolve());
-    jest.spyOn(initFirstRun.prebid, "process").mockReturnValueOnce(Promise.resolve());
-    jest.spyOn(initFirstRun.gpt, "process").mockReturnValueOnce(Promise.resolve());
+    jest
+      .spyOn(initFirstRun.apstag, "process")
+      .mockReturnValueOnce(Promise.resolve());
+    jest
+      .spyOn(initFirstRun.prebid, "process")
+      .mockReturnValueOnce(Promise.resolve());
+    jest
+      .spyOn(initFirstRun.gpt, "process")
+      .mockReturnValueOnce(Promise.resolve());
     jest.spyOn(initFirstRun.gpt, "bid").mockReturnValueOnce(Promise.resolve());
 
     initFirstRun.init();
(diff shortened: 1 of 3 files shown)


ValGeorgiev:
Description? Why this is needed?

parkjh:
> Description? Why this is needed?

this fixes a bug of missing ads in react articles

aronshamash:
@parkjh Can you make commit msg appropriate to the change? ie not just fixing tests

packages/ad/src/utils/ad-init.js
```diff
@@ -32,21 +32,27 @@ export default ({ el, data, platform, eventCallback, window }) => {
...
+        );
       }
+      window.initCalled = true;
 
-      const gptReady = [
-        this.apstag.process(),
-        this.prebid.process(),
-        this.gpt.process()
-      ];
+      if (!localInitCalled) {
```
aronshamash: `localInitCalled` hasn't been set with an initial value.

packages/ad/src/utils/ad-init.js
```diff
@@ -32,21 +32,27 @@ export default ({ el, data, platform, eventCallback, window }) => {
...
-        return Promise.resolve("skipped");
+      if (data.disableAds) {
+        return Promise.resolve("ads disabled");
       }
-      localInitCalled = true;
-      window.initCalled = true;
 
-      if ((!data.bidInitialiser && isWeb) || !isWeb) {
-        this.loadScripts();
+      const gptReady = [Promise.resolve()];
```
aronshamash: Can you add a few notes on the reason why it wasn't working before and why this now fixes the missing ads. I'm finding hard to work out why this will work now. Thanks

packages/ad/src/utils/ad-init.js
parkjh: @aronshamash look at line 28

packages/ad/src/utils/ad-init.js
parkjh: will do it once the pr is fixed

packages/ad/src/utils/ad-init.js
aronshamash: doh!

packages/ad/__tests__/utils/ad-init.js
```diff
@@ -43,18 +43,79 @@ export default () => {
...
+    expect(init1.gpt.bid).toHaveBeenCalledTimes(1);
+  });
+
+  it("calls all the vender functions only once when initialised multiple times", () => {
+    const init1 = adInit(initOptions);
+
+    jest.spyOn(init1.apstag, "process").mockReturnValueOnce(Promise.resolve());
+    jest.spyOn(init1.prebid, "process").mockReturnValueOnce(Promise.resolve());
+    jest.spyOn(init1.gpt, "process").mockReturnValueOnce(Promise.resolve());
+    jest.spyOn(init1.gpt, "bid").mockReturnValue(Promise.resolve());
```
ValGeorgiev: could we use `mockReturnValueOnce` function here as well

packages/ad/__tests__/utils/ad-init.js
```diff
@@ -43,18 +43,79 @@ export default () => {
...
+    init1.init();
+    init1.init();
+
+    expect(init1.prebid.process).toHaveBeenCalledTimes(1);
+    expect(init1.apstag.process).toHaveBeenCalledTimes(1);
+    expect(init1.gpt.process).toHaveBeenCalledTimes(1);
+    expect(init1.gpt.bid).toHaveBeenCalledTimes(1);
+  });
+
+  it("calls all the vender functions only once when initialised multiple times", () => {
```
ValGeorgiev: This description it doesn't look right, could we change it

PR #2093, opened 2019-07-02
feat:load watermark svg from render replat 6776

todorRadkov:
Moving watermark for web from times component package to render.
https://nidigitalsolutions.jira.com/browse/REPLAT-6776


todorRadkov:
I have moved the SVG to render with this PR https://github.com/newsuk/cps-content-render/pull/2820
Also, i have passed the path to the SVG from render as a parameter and it is working in author-profile and topic.js, but there is a problem with the inline-ad in the article page and they are not displayed propperly.

PR #2094, opened 2019-07-03
feat: implement wide and huge tablet Second1Column

apostolnikov:
Jira: https://nidigitalsolutions.jira.com/browse/REPLAT-5492

Implement wide and huge tablet breakpoint for Secondary 1 and Columnist edition slice.

Wide/Huge:
image

Portrait tablet for ref:
image

### DIFF
--- a/packages/edition-slices/src/tiles/tile-aa/index.js
+++ b/packages/edition-slices/src/tiles/tile-aa/index.js
@@ -1,15 +1,17 @@
 import React from "react";
 import PropTypes from "prop-types";
 import Image from "@times-components/image";
+import { editionBreakpoints } from "@times-components/styleguide";
 import {
   getTileImage,
   TileLink,
   TileSummary,
   withTileTracking
 } from "../shared";
-import styles from "./styles";
+import styleFactory from "./styles";
 
-const TileT = ({ onPress, tile }) => {
+const TileAA = ({ onPress, tile, breakpoint }) => {
+  const styles = styleFactory(breakpoint);
   const crop = getTileImage(tile, "crop169");
 
   return (
@@ -32,9 +34,14 @@ const TileT = ({ onPress, tile }) => {
   );
 };
 
-TileT.propTypes = {
+TileAA.propTypes = {
+  breakpoint: PropTypes.string,
   onPress: PropTypes.func.isRequired,
   tile: PropTypes.shape({}).isRequired
 };
 
-export default withTileTracking(TileT);
+TileAA.defaultProps = {
+  breakpoint: editionBreakpoints.small
+};
+
+export default withTileTracking(TileAA);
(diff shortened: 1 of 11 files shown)


packages/edition-slices/src/tiles/tile-aa/index.js
```diff
@@ -32,9 +34,14 @@ const TileT = ({ onPress, tile }) => {
...
-TileT.propTypes = {
+TileAA.propTypes = {
+  breakpoint: PropTypes.string,
   onPress: PropTypes.func.isRequired,
   tile: PropTypes.shape({}).isRequired
 };
 
-export default withTileTracking(TileT);
+TileAA.defaultProps = {
+  breakpoint: editionBreakpoints.small
```
parkjh: please use the function default parameter instead of `defaultProps` @apostolnikov

parkjh: all good except this so I will approve after you make this change

apostolnikov: @parkjh I agree with the change that you requested  but, I'm not sure if I need to make this change for only two tiles **here** in this PR. 
If I do so, we'll still need to create a general story for applying the change in the whole project as the default props on functional components will eventually get deprecated. https://twitter.com/dan_abramov/status/1133878326358171650

Moreover applying it only for two tiles now, will make them inconsistent with the other functional components in the project using the default props.

parkjh: @apostolnikov  see https://github.com/newsuk/times-components/pull/2081#discussion_r297683058
but I think it's worth discussing this with the team going forward

parkjh: but the overall conclusion previously was that we should not be using anything that is about to be deprecated even though it creates inconsistencies

parkjh: @nathggns cc

aronshamash: We need a strategy on tech debt. Either:
1. When ever you touch a piece a pice of code you update to the new way (pros - we moving in the right direction and eventually will update most of the code, cons - we can a state where codebase is inconsistent, especially bad for new starters)
2. We don't do any tech debt work in feature/bug PRs and have separate tech debt tickets which we developers _must_ force into the backlog and argue the case with PO and business. (pros - consistency, cons - might never get done)

@nehasri89 @chrishutchinson @nathggns

nathggns: I'm supportive of #1 to be honest.

nathggns: I agree. Please use default arguments.

nehasri89: I vote for first one too.

aronshamash: @apostolnikov ^. please update with default parameter then based on the above

PR #2095, opened 2019-07-03
fix: issue with safari browser being launched when scroll to ad

aronshamash:
- fixes https://nidigitalsolutions.jira.com/browse/REPLAT-6914 which blocks beta release.
- it reverts a change made by @jmars here: https://github.com/newsuk/times-components/commit/e5272b5ee55f900cbdafb36fd63ba29f261f5a3a#diff-6bc333c08ceabd74f38d0d732289f9c2 but without the `androidHardwareAccelerationDisabled` change (https://nidigitalsolutions.jira.com/browse/REPLAT-6795)

### DIFF
--- a/packages/ad/src/dom-context.js
+++ b/packages/ad/src/dom-context.js
@@ -123,7 +123,6 @@ class DOMContext extends PureComponent {
         }}
       >
         <WebView
-          originWhitelist={["http://.*", "https://.*"]}
           onMessage={this.handleMessageEvent}
           onNavigationStateChange={this.handleNavigationStateChange}
           ref={ref => {


packages/ad/src/dom-context.js
```diff
@@ -123,7 +123,6 @@ class DOMContext extends PureComponent {
         }}
       >
         <WebView
-          originWhitelist={["http://.*", "https://.*"]}
```
parkjh: @aronshamash this was added by @jmars to fix unruly ads...

aronshamash: yes - as noted in description.

parkjh: @aronshamash so the unruly ads work without this?

aronshamash: Unruly videos are not in production yet so believe we could go to beta without that fix, but we _definitely_ need to fix this bug with safari.

parkjh: @aronshamash so it doesn't work?

aronshamash: @parkjh what doesn't work? Unruly ads are not configured in DFP AFAIK so I can't test if they work with or with this change. This change _does_ fix the blocking issue with Safari being launched though.

parkjh: @aronshamash https://github.com/newsuk/times-components/pull/2028 @jmars managed to add a screenshot of it working...
so I'm not blocking this, just clarifying that potentially in the future the unruly ad will not work specifically due to this change

aronshamash: Raised new Jira to fix unruly ads: https://nidigitalsolutions.jira.com/browse/REPLAT-6947

PR #2096, opened 2019-07-04
Feat/tablet puzzles

Ivanov558:
feat/tablet-puzzles

Add new font sizes for tablet puzzles on landscape and landscape large landscape.

image

image

### DIFF
--- a/packages/edition-slices/src/tiles/tile-ak/styles/index.js
+++ b/packages/edition-slices/src/tiles/tile-ak/styles/index.js
@@ -10,7 +10,7 @@ const headlineFontSizeResolver = {
   [editionBreakpoints.huge]: 35
 };
 
-const styles = breakpoint => ({
+const styles = (breakpoint = editionBreakpoints.medium) => ({
   header: {
     paddingHorizontal: spacing(2),
     paddingTop: spacing(2),
--- a/packages/edition-slices/src/tiles/tile-ak/styles/index.js
+++ b/packages/edition-slices/src/tiles/tile-ak/styles/index.js
@@ -4,27 +4,10 @@ import {
   editionBreakpoints
 } from "@times-components/styleguide";
 
-const getHeadline = breakpoint => {
-  switch (breakpoint) {
-    case editionBreakpoints.wide: {
-      return {
-        fontSize: 30,
-        lineHeight: 30
-      };
-    }
-    case editionBreakpoints.huge: {
-      return {
-        fontSize: 35,
-        lineHeight: 35
-      };
-    }
-    default: {
-      return {
-        fontSize: 22,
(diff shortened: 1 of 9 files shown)


packages/edition-slices/src/tiles/tile-ak/styles/index.js
```diff
@@ -1,15 +1,39 @@
-import { colours, spacing } from "@times-components/styleguide";
+import {
+  colours,
+  spacing,
+  editionBreakpoints
+} from "@times-components/styleguide";
 
-const styles = {
+const getHeadline = breakpoint => {
+  switch (breakpoint) {
```
aronshamash: please use an object literal rather than switch statement to reduce code logic, also no need to store separate value for `fontSize` and `lineHeight` as they are the same.

Ivanov558: Hi. I don't see any benefit to use object literal. I think that using object literal or switch case statement is matter of style here and won't give us any benefit. The code is very simple and readable where we have different cases and return different objects.

aronshamash: it reduces code logic and it is more consistent with the style of storing config values we have already in the codebase

aronshamash: https://github.com/newsuk/times-components/pull/2094/files#diff-6025402c0c791874d7e963a4ab60b173R7

Ivanov558: and what happened if we want to add more styles for different screens?

aronshamash: nicer now with less code.

packages/edition-slices/src/tiles/tile-ak/styles/index.js
```diff
@@ -1,14 +1,24 @@
-import { colours, spacing } from "@times-components/styleguide";
+import {
+  colours,
+  spacing,
+  editionBreakpoints
+} from "@times-components/styleguide";
 
-const styles = {
+const headlineFontSizeResolver = {
```
apostolnikov: Not all cases are handled, this will cause potential error in the future if we use TileAK on the small breakpoint.

Ivanov558: we don't use it on small

apostolnikov: Yes, for now. Leaving it like this opens a potential error vulnerability which is not good idea at all. Especially when you don't pass any default parameter for the breakpoint.

PR #2099, opened 2019-07-04
feat: REPLAT-5483 implement wide huge breakpoints on  lead 1 and 1 slice

pdimova:
Jira: https://nidigitalsolutions.jira.com/browse/REPLAT-5483

- Wide breakpoint: 125 char teaser text on lead tile, headline font size on support tile remains 22 
- Huge breakpoint: 300 char teaser text on lead tile, 125 teaser text on support tile, increased headline font size on lead tile, content width - max 1180

Storybook - wide breakpoint:
Screenshot_1562245855

Storybook - huge breakpoint:
Screenshot_1562341429
Screenshot_1562341405

App - wide breakpoint:
Screenshot_1562342141

App - huge breakpoint:
Screenshot_1562341210
Screenshot_1562341060

### DIFF
--- a/packages/edition-slices/src/tiles/tile-u/index.js
+++ b/packages/edition-slices/src/tiles/tile-u/index.js
@@ -1,3 +1,4 @@
+/* eslint-disable react/require-default-props */
 import React from "react";
 import PropTypes from "prop-types";
 import Image from "@times-components/image";
@@ -11,7 +12,7 @@ import {
 } from "../shared";
 import styleFactory from "./styles";
 
-const TileU = ({ onPress, tile, breakpoint }) => {
+const TileU = ({ onPress, tile, breakpoint = editionBreakpoints.small }) => {
   const styles = styleFactory(breakpoint);
   const crop = getTileImage(tile, "crop32");
 
@@ -55,10 +56,6 @@ TileU.propTypes = {
   breakpoint: PropTypes.string,
   onPress: PropTypes.func.isRequired,
   tile: PropTypes.shape({}).isRequired
-};
-
-TileU.defaultProps = {
-  breakpoint: editionBreakpoints.small
 };
 
 export default withTileTracking(TileU);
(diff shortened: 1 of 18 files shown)


packages/edition-slices/src/tiles/tile-u/index.js
```diff
@@ -39,8 +52,13 @@ const TileU = ({ onPress, tile }) => {
 };
 
 TileU.propTypes = {
+  breakpoint: PropTypes.string,
   onPress: PropTypes.func.isRequired,
   tile: PropTypes.shape({}).isRequired
 };
 
+TileU.defaultProps = {
```
apostolnikov: Use default function params instead of defaultProps.
https://twitter.com/dan_abramov/status/1133878326358171650

pdimova: nice suggestion but lint does not agree with it ->  propType "breakpoint" is not required, but has no corresponding defaultProp declaration
We have to agree to stick to defaultProp or disable the lint rule

nathggns: We've decided to use default arguments. Disable the linter if its complaining.

pdimova: updated

packages/slice-layout/src/templates/leadoneandone/styles.js
```diff
@@ -18,4 +22,14 @@ const styles = {
   }
 };
 
-export default styles;
+const hugeBreakpointStyles = {
+  ...styles,
+  container: {
+    ...styles.container,
+    width: "87.6%",
```
aronshamash: this is an interesting width. Slice specific or can it be same as `dailyuniversalregister`, `secondarytwoandtwo` and `secondarytwonopicandtwo` that uses `86%` ?

pdimova: 1180/1366 = 0.8638...
but since we have `paddingHorizontal: isTablet ? spacing(2) : 0,` in the Gutter component then to have a content of 1180 you need to take into account the 20px padding in total so 1180/(1366-20) = 0.8766... that was my idea behind the width calc

aronshamash: as agreed not to abstract to `styleguide` for now but update to use `86%` to be consistent with other slices

PR #2101, opened 2019-07-04
fix: lead asset position should be responsive on the main standard article template

nathggns:
https://nidigitalsolutions.jira.com/browse/REPLAT-604

We created a regression where lead assets no longer changed position on tablet/mobile on the main standard article template. This happened because that feature depended on flex order, and at some point the parent container was updated to use block layout instead of flex layout on web.

In that time, the save and share bar was introduced, adding extra complexities around double keylines when on mobile vs on desktop (as on mobile, the meta information and the save and share bar are now next to each other). In addition, we recently ran into a similar problem with in depth article templates, where the meta information sits permanently next to the save and share bar. 

Because of this, I decided to introduce a helper for automatically applying the correct keylines. It does this by shifting the keyline item up one pixel so that the borders cover each other when they're next to each other. This is the only reliable way I could figure out how to do it, as some of these items are hidden with `display:none` at certain breakpoints (which `:last-child` and `:first-child` don't take into account)

**article main standard desktop, guest**

article main standard desktop, guest

**article main standard, mobile, guest**

article main standard, mobile, guest

**article main standard desktop, subscriber**

article main standard desktop, subscriber

**article main standard mobile, subscriber**

article main standard mobile, subscriber


packages/article-skeleton/src/keylines.js
```diff
@@ -0,0 +1,32 @@
+import styled from "styled-components";
+import { breakpoints, colours, spacing } from "@times-components/styleguide";
+
+export const KeylineItem = styled.div`
+  border: solid ${colours.functional.keyline};
+  border-width: 1px 0;
+
+  & + & {
```
aronshamash: Can you explain this please?

packages/article-skeleton/src/sticky-save-and-share-bar.js
```diff
@@ -1,21 +1,28 @@
...
+import Sticky, { STICKY_CLASS_NAME } from "@times-components/sticky";
+import { breakpoints, colours } from "@times-components/styleguide";
+import { ArticleKeylineItem } from "./keylines";
+
+export const SaveShareContainer = styled.div`
+  background-color: ${colours.functional.white};
+  height: 60px;
+
+  &.${STICKY_CLASS_NAME} {
+    box-shadow: 0 2px 8px 0 rgba(0, 0, 0, 0.3);
```
aronshamash: this fixes the styling of the sticky bar now ?

packages/article-main-standard/src/article-meta/util.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable import/prefer-default-export */
+
+export const hasBylineData = bylines =>
```
aronshamash: I have named this `has-byline-data.j` in [ this PR](https://github.com/newsuk/times-components/pull/2106) - I guess we should be consistent, as suggested in other PR we could do in :
1. `article-skelton` - might become the place for common article code ?
2. `article-byline` - think this  might make more sesne

packages/article-skeleton/src/keylines.js
nathggns: Applies styles to a direct sibling if that sibling is the same styled component. https://codesandbox.io/s/example-kcqhr

packages/article-skeleton/src/sticky-save-and-share-bar.js
nathggns: No. This is just moving styles.

packages/article-skeleton/src/keylines.js
nathggns: This essentially says if a `KeylineItem` is followed immediately by another KeylineItem, then that `KeylineItem` should be shifted one pixel up so that their borders are overlapping (so that you only see one of them if they're both visible). This "hack" allows you to ensure there is always a single pixel line between `KeylineItem`s, even if one of them has `display:none` applied.

packages/article-main-standard/src/article-meta/util.js
nathggns: I only made a new file for this in this PR because I got rid of the article meta base because the components are now sufficiently different on the different platforms. I think we should merge this, and do the change to moving it to article-byline (my preference) in your PR?

aronshamash: ok I'll update my PR then

PR #2105, opened 2019-07-05
feat: REPLAT 5482 implement wide huge breakpoints on lead 1 full width slice

pdimova:
Jira: https://nidigitalsolutions.jira.com/browse/REPLAT-5482

- Wide breakpoint - font size increased
- Huge breakpoint - same as wide, except the max width of 1180

Storybook - wide breakpoint:
Screenshot_1562326131

Storybook - huge breakpoint:
Screenshot_1562335460
Screenshot_1562335676
*note:  the bordered box is of 1366px and it is to depict the gutter space for huge breakpoint

### DIFF
--- a/packages/edition-slices/src/tiles/tile-r/styles/index.js
+++ b/packages/edition-slices/src/tiles/tile-r/styles/index.js
@@ -10,12 +10,6 @@ const headlineFontSizeResolver = {
   [editionBreakpoints.wide]: 45
 };
 
-const defaultContainerStyle = {
-  paddingBottom: spacing(3),
-  paddingHorizontal: spacing(4),
-  paddingTop: spacing(2)
-};
-
 const hugeScreenContainerStyle = {
   width: "87.6%",
   paddingHorizontal: spacing(0),
@@ -24,8 +18,10 @@ const hugeScreenContainerStyle = {
 
 export default breakpoint => ({
   container: {
-    ...defaultContainerStyle,
-    ...(breakpoint === editionBreakpoints.huge ? hugeScreenContainerStyle : {})
+    paddingBottom: spacing(3),
+    paddingHorizontal: spacing(4),
+    paddingTop: spacing(2),
+    ...(breakpoint === editionBreakpoints.huge && hugeScreenContainerStyle)
   },
   headline: {
     fontFamily: fonts.headline,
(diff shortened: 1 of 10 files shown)


packages/edition-slices/src/tiles/tile-r/styles/index.js
```diff
@@ -1,16 +1,35 @@
...
+} from "@times-components/styleguide";
 
-const styles = {
+const headlineFontSizeResolver = {
+  [editionBreakpoints.medium]: 40,
+  [editionBreakpoints.huge]: 45,
+  [editionBreakpoints.wide]: 45
+};
+
+const defaultContainerStyle = {
```
apostolnikov: Why would you extract this in a new constant? If you don't need it anywhere else, we should probably move it right inside the container.

packages/edition-slices/src/tiles/tile-r/styles/index.js
```diff
@@ -1,16 +1,35 @@
...
+  alignSelf: "center"
+};
+
+export default breakpoint => ({
   container: {
-    paddingBottom: spacing(3),
-    paddingHorizontal: spacing(4),
-    paddingTop: spacing(2)
+    ...defaultContainerStyle,
+    ...(breakpoint === editionBreakpoints.huge ? hugeScreenContainerStyle : {})
```
apostolnikov: ...(breakpoint === editionBreakpoints.huge && hugeScreenContainerStyle)

packages/edition-slices/src/tiles/tile-r/styles/index.js
pdimova: edited

packages/edition-slices/src/tiles/tile-r/styles/index.js
```diff
@@ -1,16 +1,31 @@
...
 
-const styles = {
+const headlineFontSizeResolver = {
+  [editionBreakpoints.medium]: 40,
+  [editionBreakpoints.huge]: 45,
+  [editionBreakpoints.wide]: 45
+};
+
+const hugeScreenContainerStyle = {
+  width: "87.6%",
```
aronshamash: this `87.6%` and `86%` are cropping up on a few slices. Can we double check with Matt Myers on which it should be and store in `styleguide` package as `hugeContainerWidth` and use for all slices that require it ?

pdimova: agreed [here](https://github.com/newsuk/times-components/pull/2099) to keep the width `86%`

PR #2106, opened 2019-07-05
fix: tighten up byline data validation

aronshamash:
- Fixes issue with no bylines: https://nidigitalsolutions.jira.com/browse/REPLAT-6899
- Improve validation for byline data.
- adds `hasBylineData` as an export of `article-bylines` and now used on all article templates
- Fix some old fixture data that was out of date :( (need some TLC on fixtures....)


packages/article-in-depth/src/article-meta/has-byline-data.js
```diff
@@ -0,0 +1,2 @@
+export default bylines =>
+  bylines.length > 0 && bylines[0].byline && bylines[0].byline.length > 0;
```
apostolnikov: Do we have any reason not to add this inside the utils folder?

aronshamash: Well I was thinking utils is for more generic utility functions not something as specific as this. I'm more of the opinion to keep helper functions local to the package now, but I'm open to others opinion. @nathggns @nehasri89 @ValGeorgiev thoughts ?

nathggns:
This will likely conflict with https://github.com/newsuk/times-components/pull/2101 – it might be better to wait for that to merge as yours will be easier to rebase than mine?

packages/article-in-depth/src/article-meta/has-byline-data.js
nathggns: I'd put it in `article-skeleton` or `article-byline`.

packages/article-in-depth/src/article-meta/article-meta.js
```diff
@@ -6,6 +6,7 @@ import Context from "@times-components/context";
 import DatePublication from "@times-components/date-publication";
 import { colours } from "@times-components/styleguide";
 
+import hasBylineData from "./has-byline-data";
```
nathggns: Can we please move the `hasBylineData` function from the new utils file in `@times-compoments/article-main-standard` into `article-byline`, upgrade that function, and share it across the different templates that way? As discussed on Slack .

packages/article-in-depth/src/article-meta/has-byline-data.js
nathggns: Can we please move the hasBylineData function from the new utils file in @times-compoments/article-main-standard into article-byline, upgrade that function, and share it across the different templates that way? As discussed on Slack .

aronshamash: putting in `article-byline` and sharing

PR #2107, opened 2019-07-05
Feat/Fix Scroll Analytic on Storybook

nehasri89:


packages/tracking/storybook-components/box.js
```diff
@@ -13,9 +13,11 @@ export const boxStyles = StyleSheet.create({
 });
 
 const Box = ({ children, color, id }) => (
-  <View id={id} style={[boxStyles.box, { backgroundColor: color }]}>
-    {children}
-  </View>
+  <div id={id}>
```
parkjh: why?

packages/tracking/storybook-components/box.js
```diff
@@ -13,9 +13,11 @@ export const boxStyles = StyleSheet.create({
...
+    <View id={id} style={[boxStyles.box, { backgroundColor: color }]}>
```
aronshamash: duplicate id

packages/tracking/storybook-components/box.js
nehasri89: Because the way scrollTracking works it expects the html to have an `id` attribute, but the React Native `View` does not have it, so the way it is done in components for ex. `Article List` is to have a `div` element which can have an `id`

parkjh:
£5 please

Ivaylo1987:
This PR is fine for Storybook, but please keep in mind that the real problem is in Render where scroll analytics does not work. This is Because it is tied to `id='related-articles'` container which on a react article page disappears client side even though server side rendering returns it correctly.

nehasri89:
> This PR is fine for Storybook, but please keep in mind that the real problem is in Render where scroll analytics does not work. This is Because it is tied to `id='related-articles'` container which on a react article page disappears client side even though server side rendering returns it correctly.

Yes, this is only fixing the storybook, there will be a separate PR for RN article page, although I do see the related articles on client side.

PR #2108, opened 2019-07-05
chore: upgrade react native 0.60

parkjh:
see https://react-native-community.github.io/upgrade-helper/?from=0.59.10&to=0.60.0


android/app/src/main/res/values/styles.xml
```diff
@@ -3,6 +3,7 @@
     <!-- Base application theme. -->
     <style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
         <!-- Customize your theme here. -->
+        <item name="android:textColor">#000000</item>
```
ChristianGarcia: I'm curious now. Are the Android components in times-components using this theme? We're definitely not using them on the native side.

android/build.gradle
```diff
@@ -25,11 +25,11 @@ buildscript {
 allprojects {
     repositories {
         mavenLocal()
-        google()
-        jcenter()
         maven {
             // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
-            url "$rootDir/../node_modules/react-native/android"
+            url("$rootDir/../node_modules/react-native/android")
```
ChristianGarcia: These parentheses are redundant.

PR #2109, opened 2019-07-05
fix: restore fix for android unruly ads

jmars:


aronshamash:
@jmars interesting only android requires it. So Video ads work without whitelisting on iOS ?

jmars:
> @jmars interesting only android requires it. So Video ads work without whitelisting on iOS ?

Yep, that seems to be the case, it is strange.

aronshamash:
> > @jmars interesting only android requires it. So Video ads work without whitelisting on iOS ?
> 
> Yep, that seems to be the case, it is strange.

ok

PR #2110, opened 2019-07-08
Update content in meta tag for journalist and topic list pages

mikefieldmay:
This PR relates to tickets:
https://nidigitalsolutions.jira.com/browse/TOC-128
https://nidigitalsolutions.jira.com/browse/TOC-129

Updating the content in meta tag for journalist and topic list pages.

Any thoughts would be useful.


packages/author-profile/src/author-profile.js
```diff
@@ -111,7 +111,7 @@ const AuthorProfile = ({
 
         return (
           <Responsive>
-            <Head name={name} />
+            <Head metaContent={biography} name={name} />
```
nathggns: Should this prop just be called `biography`?

packages/author-profile/src/head.web.js
```diff
@@ -1,20 +1,24 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { Helmet } from "react-helmet-async";
+import { convertMarkupTreeToString } from "@times-components/utils";
 
-function Head({ name }) {
+import { propTypes as authorProfileHeadPropTypes } from "./author-profile-head-prop-types";
+
+const Head = ({ metaContent, name }) => {
```
nathggns: I'm not sure that there's any need to change the syntax from `function Head()`, which is likely more performant and slightly more readable. I've noticed many of the React team use that syntax too.

packages/topic/src/topic.js
```diff
@@ -88,7 +88,7 @@ const Topic = ({
 
         return (
           <Responsive>
-            <Head {...{ name, slug }} />
+            <Head metaContent={description} {...{ name, slug }} />
```
nathggns: Should this prop just be called description?

packages/topic/src/head.web.js
```diff
@@ -2,25 +2,32 @@ import React from "react";
 import PropTypes from "prop-types";
 import { Helmet } from "react-helmet-async";
 import Context from "@times-components/context";
+import { convertMarkupTreeToString } from "@times-components/utils";
 
-function Head({ name, slug }) {
+import { propTypes as topicHeadPropTypes } from "./topic-head-prop-types";
+
+const Head = ({ metaContent, name, slug }) => {
```
nathggns: I'm not sure that there's any need to change the syntax from function Head(), which is likely more performant and slightly more readable. I've noticed many of the React team use that syntax too.

packages/utils/src/convert-markup-tree-to-string.js
```diff
@@ -0,0 +1,14 @@
+import get from "lodash.get";
+
+const mapChildValues = markupTree =>
+  markupTree.map(child => child.attributes.value).join("");
+
+export default markupTree =>
```
nathggns: We already have `renderTreeAsText` in `@times-components/markup-forest`. Does this do something different?

packages/utils/src/convert-markup-tree-to-string.js
```diff
@@ -0,0 +1,14 @@
...
+  markupTree
+    .map(
+      markupObject =>
+        get(markupObject, "attributes.value") ||
```
jmars: This will fail on markup that is not just text nodes?

packages/utils/src/convert-markup-tree-to-string.js
mikefieldmay: I'll make sure they do the same thing and if they do I'll remove the version I created

packages/topic/src/topic.js
mikefieldmay: Probably. I toyed around with the variable names for ages, but I think description is better

packages/topic/src/head.web.js
mikefieldmay: I would disagree in the readability front. I think if you know what the fat arrow syntax is you should be fine. Plus, when you're using typescript I think function declarations become more obvious again, although I know we're not there yet.
I couldn't work out when I was looking through the codebase why some functions were es5 and others were es6 and initially assumed it was a legacy thing.

This article gave some explanation on the performance front: https://frontarm.com/james-k-nelson/when-to-use-arrow-functions/ so I can see the benefit for older browsers.

I've reverted it back.

Is there a platform here where people discuss coding standards across the group? Where I used to work had a Dev Guild where all the engineers got together and talked about standards, general improvements etc

nathggns: I know what the arrow syntax is, and I use it in many places where its appropriate to do so. I just do think the `const Component = ({ propA, probB}) => { ... }` syntax is better than the `function Component({ propA, propB}) { ... }` syntax on account of being less readable because its literally more characters, and many of them are special characters so it just looks like special character soup. 

There are places where arrow functions are appropriate – one line utilities and lambas are both good examples, event handlers where the this value needs to be lexically bound is another. But when you don't care about the function context, I don't really see the benefit of using them. They're not a replacement for the ES5 function syntax, they're an addition to them and both syntaxes have their uses. It's not a case of "reverting es6 to es5", as the function syntax still exists in es6 and for good reason.

The reason why there's inconsistency is we've never made a decision on which to use and the linter doesn't care. I've always written components with function statements, others have always used const statements with arrow function expressions. I was just saying in this PR that in leue of a specific rule or decision, unless you have a reason to change a bit of code from one to the other, don't. 

If you want to make a decision about this, get a conversation going on `#times-components` on Slack and we can decide which we prefer writing and reading going forward. Personally, for me, it'll always be the one with less characters.

mikefieldmay: Apologies, definitely not trying to suggest you don't know what the arrow syntax is. 
I still disagree with your first comment about readability, but that's fine, we should use what we prefer.

I haven't come across using ES5 and ES6 together before either. I was under the impression that es6 functions were "the way" to write functions, but I can see that I was wrong and both have their uses. At the end of the day, a function is a function regardless of how it's declared.

Where I used to work we would tend to update any component that we touched and out of habit I changed the es5 function to es6, but obviously the decision was made for a reason, and as such I will leave things be in future.

PR #2111, opened 2019-07-09
feat: (REPLAT-6797) add viewport tracking for unruly ads

jmars:
This ads viewport tracking for unruly so that they can know when they are in view before starting or stopping a video ad.

I am not sure of a good way to test this? Any suggestions?

Also currently waiting for an approved list of device data to pass to them without violating user data confidentiality.

### DIFF
--- a/packages/ad/src/dom-context.js
+++ b/packages/ad/src/dom-context.js
@@ -2,6 +2,7 @@ import React, { PureComponent } from "react";
 import { View, Linking, Platform } from "react-native";
 import { WebView } from "react-native-webview";
 import { Viewport } from "@skele/components";
+import DeviceInfo from "react-native-device-info";
 import webviewEventCallbackSetup from "./utils/webview-event-callback-setup";
 import logger from "./utils/logger";
 import { propTypes, defaultProps } from "./dom-context-prop-types";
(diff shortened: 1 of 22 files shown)


nathggns:
Might be worth using something like https://blog.netcetera.com/viewport-aware-components-for-react-native-apps-a28ea605e89e instead of a timer.

packages/ad/src/dom-context.js
```diff
@@ -55,6 +63,37 @@ class DOMContext extends PureComponent {
     }
   };
 
+  startWatchingViewport() {
+    if (this.interval) {
+      return;
+    }
+    this.interval = setInterval(() => {
```
parkjh: is this the only possible way?

packages/ad/src/dom-context.js
```diff
@@ -135,7 +175,17 @@ class DOMContext extends PureComponent {
             baseUrl,
             html
           }}
-          style={{ backgroundColor: "transparent" }}
+          style={{ position: "absolute" }}
+        />
+        <ViewportAwareView
+          onViewportEnter={() => this.inViewport()}
+          onViewportLeave={() => this.outViewport()}
```
ValGeorgiev: Could we not use arrow functions in render, please use arrow functions in the declaration of inViewport and outViewport or bind them in the constructor

packages/ad/src/dom-context.js
```diff
@@ -135,7 +175,17 @@ class DOMContext extends PureComponent {
...
+          style={{
+            width: 10,
+            height: height / 10,
+            top: height / 2 - height / 20,
```
ValGeorgiev: could we use constants for these magic numbers

ValGeorgiev:
@jmars Nice one 🚀 Could you rebase this one and check my comments. Lets merge this, because  its here for a while :)

parkjh:
@jmars closing due to inactivity

ValGeorgiev:
> @jmars closing due to inactivity

@parkjh @jmars rebased 18h ago ???

parkjh:
@ValGeorgiev oops didn't see that... but to be fair this pr has been around for ages...

jmars:
Yes it got put aside for a while due to other things needed for release, fixing up Vals comments now

parkjh:
@jmars are there no additional functional tests? only the snapshots are updated...

PR #2113, opened 2019-07-10
fix: REPLAT-6946 video label disappears when centred layout

pdimova:
Jira story: https://nidigitalsolutions.jira.com/browse/REPLAT-6946
A regression was created where the video label was not shown in some articles or slices. This happened when we tried to get the video label text not to overlap its container (which could have been observed for label text longer than a row) .`flex:1` was added to the video label styles that gives enough space for short labels, makes the longer labels wrap properly but when the layout is centred
the label text width is lost.

Video label disappears on:
- article templates like article-in-depth, article-main-comment 
- slices where we have tile with centred layout 

As per RN documentation: 

> When flex is -1, the component is normally sized according width and height. However, if there's not enough space, the component will shrink to its minWidth and minHeight. 

Before:
Screenshot_1562678895
Screenshot_1562679489
Screenshot_1562679965

After:
Screenshot_1562678938
Screenshot_1562679459
Screenshot_1562680016

Storybook example with short and long text:
Screenshot_1562678585
Screenshot_1562680650

IOS:
Simulator Screen Shot - iPhone Xʀ - 2019-07-11 at 18 40 45
Simulator Screen Shot - iPhone Xʀ - 2019-07-11 at 18 41 18
Simulator Screen Shot - iPhone Xʀ - 2019-07-12 at 11 11 58

Web still works as expected:
Screenshot 2019-07-09 at 17 27 05
Screenshot 2019-07-09 at 17 19 41
Screenshot 2019-07-09 at 17 25 53

### DIFF
--- a/packages/video-label/src/style/shared.js
+++ b/packages/video-label/src/style/shared.js
@@ -16,7 +16,7 @@ const styles = {
       font: "supporting",
       fontSize: "cardMetaMobile"
     }),
-    flex: 1,
+    flex: -1,
     fontWeight: "400",
     letterSpacing: 1.2,
     lineHeight: 11,
--- a/packages/fixture-generator/src/mock-slice.ts
+++ b/packages/fixture-generator/src/mock-slice.ts
@@ -111,9 +111,19 @@ function getDailyRegister(count: number): Array<DailyUniversalRegisterItem> {
 
 function mockLeadOneAndFourSlice(): LeadOneAndFourSliceWithName {
   const tiles = getTiles(5);
+
+  const leadTile = {
+    ...tiles[0],
+    article: {
+      ...tiles[0].article,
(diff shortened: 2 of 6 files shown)


packages/video-label/src/style/shared.js
```diff
@@ -16,7 +16,7 @@ const styles = {
       font: "supporting",
       fontSize: "cardMetaMobile"
     }),
-    flex: 1,
+    flex: -1,
```
aronshamash: this affects web too. Can you post a screenshot of web too ? BTW there is another ticket re: alignment issue of video label in Chrome on Windows..

pdimova: added 2 more screenshots of web.
what is the Jira id for the other task ?

PR #2114, opened 2019-07-10
fix: links should not recreate a styled component on every render

nathggns:
Currently, links re-create a styled-component on every single render, forcing both styled-components and react to do more work than necessary. I've updated the component to use `styled-components`'  prop interpalation feature instead.

In the long term, I think we should deprecate and/or remove `responsiveLinkStyles` as it doesn't seem like a useful abstraction and isn't great for performance, and is also not supported on native platforms, but I don't have a great view of how and why it is used. 

I also think we should add "cursor:pointer" as a default for all links, but didn't want to do that in this PR as it impacts more places. 

Todo: 

- [x] Remove the `!important`

### DIFF
--- a/packages/link/src/link.web.js
+++ b/packages/link/src/link.web.js
@@ -1,63 +1,56 @@
+/* eslint-disable react/require-default-props */
 import React from "react";
 import styled from "styled-components";
 import { breakpoints } from "@times-components/styleguide";
 import PropTypes from "prop-types";
 
-const RespLink = responsiveLinkStyles =>
-  styled.a`
-    ${responsiveLinkStyles.base} @media (min-width: ${breakpoints.medium}px) {
-      ${responsiveLinkStyles.medium};
-    }
-  `;
+const respStylesSelector = selector => ({ responsiveLinkStyles }) =>
+  (responsiveLinkStyles && responsiveLinkStyles[selector]) || "";
+
+const RespLink = styled.a`
+  text-decoration: ${props =>
+    props.underlined && props.responsiveLinkStyles
+      ? "underline"
+      : "none"} !important;
+
+  ${respStylesSelector("base")};
+
+  @media (min-width: ${breakpoints.medium}px) {
+    ${respStylesSelector("medium")};
+  }
+`;
 
 const Link = ({
   children,
-  onPress,
-  responsiveLinkStyles,
-  target,
   url,
-  underlined
+  onPress = () => {},
+  target = null,
+  underlined = true,
+  responsiveLinkStyles = null
 }) => {
-  const Wrapper =
-    responsiveLinkStyles !== null ? RespLink(responsiveLinkStyles) : "a";
+  const props = Object.assign(
+    { underlined },
+    target ? { target } : {},
+    responsiveLinkStyles ? { responsiveLinkStyles } : {}
+  );
 
-  const style =
-    underlined && responsiveLinkStyles
-      ? { textDecoration: "underline" }
-      : { textDecoration: "none" };
-
-  const props = {
-    href: url,
-    onClick: onPress,
-    style
-  };
-
-  return target ? (
-    <Wrapper {...props} target={target}>
+  return (
+    <RespLink href={url} onClick={onPress} {...props}>
       {children}
-    </Wrapper>
-  ) : (
-    <Wrapper {...props}>{children}</Wrapper>
+    </RespLink>
   );
 };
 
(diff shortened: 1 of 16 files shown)


nathggns:
Created an issue for deprecating `linkResponsiveStyles`. https://github.com/newsuk/times-components/issues/2115

packages/link/src/link.web.js
```diff
@@ -1,63 +1,56 @@
...
-    }
-  `;
+const respStylesSelector = selector => ({ responsiveLinkStyles }) =>
+  (responsiveLinkStyles && responsiveLinkStyles[selector]) || "";
+
+const RespLink = styled.a`
+  text-decoration: ${props =>
+    props.underlined && props.responsiveLinkStyles
+      ? "underline"
+      : "none"} !important;
```
chrishutchinson: `!important`? Is there a way for us to avoid doing this?

packages/link/src/link.web.js
```diff
@@ -1,63 +1,56 @@
...
-    underlined && responsiveLinkStyles
-      ? { textDecoration: "underline" }
-      : { textDecoration: "none" };
-
-  const props = {
-    href: url,
-    onClick: onPress,
-    style
-  };
+  const props = Object.assign(
```
chrishutchinson: Can we use object spread to neaten this `assign` up?

packages/link/src/link.web.js
```diff
@@ -1,63 +1,56 @@
...
-    ${responsiveLinkStyles.base} @media (min-width: ${breakpoints.medium}px) {
-      ${responsiveLinkStyles.medium};
-    }
-  `;
+const respStylesSelector = selector => ({ responsiveLinkStyles }) =>
+  (responsiveLinkStyles && responsiveLinkStyles[selector]) || "";
+
+const RespLink = styled.a`
+  text-decoration: ${props =>
+    props.underlined && props.responsiveLinkStyles
```
chrishutchinson: So if `underlined={true}` but `responsiveLinkStyles={something-falsey}` I wouldn't get an underlined link? That seems counter intuitive

packages/link/src/link.web.js
nathggns: It's not strictly necessary. I only included it because it seemed that if a consumer uses the `underline property, it should override everything. Will remove.

packages/link/src/link.web.js
nathggns: This wouldn't actually be neater with object spread, which is why I used Object.assign.

```
const props = Object.assign(
    { underlined },
    target ? { target } : {},
    responsiveLinkStyles ? { responsiveLinkStyles } : {}
);
```

vs

```
const props = {
    underlined,
    ...(target ? { target } : {}),
    ...(responsiveLinkStyles ? { responsiveLinkStyles } : {})
};
```

The former seems cleaner to me. Do you disagree? 
```

File: packages/link/src/link.web.js (hunk 3 above)
Comment (nathggns) [reply packages/link/src/link.web.js]: I'm 100% sure this is just moving existing logic. 

Previously, there was the following lines of code.

```js
  const style =
    underlined && responsiveLinkStyles
      ? { textDecoration: "underline" }
      : { textDecoration: "none" };
```

I've just moved it from a style prop into the styled component.

I don't claim to understand why it is this way, but I wasn't intending to fix everything wrong with this package either. Just the issue of the styled component being recreated. It's my intention to remove `responsiveLinkStyles` anyway in the long run.

packages/link/src/link.web.js
nathggns: Found a cleaner way to do it, without Object.assign:

```js
const props = {
    underlined,
    ...(!!target && { target }),
    ...(!!responsiveLinkStyles && { responsiveLinkStyles })
  }
```

chrishutchinson: I may be oversimplifying, so sorry if so, but would this be sufficient, leaving null values for `target`, and `responsiveLinkStyles`.

```js
const props = {
  underlined,
  target,
  responsiveLinkStyles
};
```

resulting in:
```js
const props = {
  underlined,
  target: null,
  responsiveLinkStyles: null
};

nathggns: Will try that.

nathggns: Okay, that works. I'd assumed react-dom wouldn't like null being passed for target, and that styled components would blow up trying to compute styles from a null prop. But it works.

PR #2116, opened 2019-07-11
fix: headline overlapping save star

apostolnikov:
Jira: https://nidigitalsolutions.jira.com/browse/REPLAT-6943

**Before:** 
before
before2

======================================================================

**After:**
after
after2

======================================================================

**More screenshots on different screens with different resolutions:**
diff
diff2
diff3
diff4
diff5


nathggns:
👀

packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -64,13 +64,22 @@ class TileSummary extends Component {
...
         article: { headline, shortHeadline }
       },
-      headlineStyle
+      headlineStyle,
+      withStar,
+      summary
     } = this.props;
 
+    const shouldAddBottomPadding =
+      Platform.OS === "android" && !withStar && !summary;
```
nathggns: why is this dependent on platform? Did the bug not happen on iOS?

packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -97,6 +106,9 @@ class TileSummary extends Component {
       labelColour
     } = this.props;
 
+    const shouldAddBottomPadding =
```
nathggns: This "android only" logic now seems duplicated.

packages/edition-slices/src/tiles/shared/tile-summary.js
apostolnikov: We don't have the star on IOS thus adding the padding there messes up the layout.

packages/edition-slices/src/tiles/shared/tile-summary.js
apostolnikov: There is a difference. On the headline we have the **!summary**  and on the article content we need to test for the existing of flags instead... **expirableFlags.length === 0**

apostolnikov:
> Is this an after image? It looks a little too close to the text still.
> 
> image

@nathggns This is an after image indeed. Adding bigger padding looked weird on the other slices... with a lot of empty white space.

packages/edition-slices/src/tiles/shared/tile-summary.js
parkjh: @apostolnikov there's a react native module that does this, which is in the utils function in the same package I believe...

packages/edition-slices/src/tiles/shared/tile-summary.js
parkjh: same as before

parkjh:
@apostolnikov nice screenshots!

packages/edition-slices/src/tiles/shared/styles/index.js
```diff
@@ -17,8 +17,8 @@ const tileStar = {
   bottom: spacing(1)
 };
 
-const starPadding = {
-  paddingBottom: spacing(5)
+const starPaddingBottom = {
+  paddingBottom: spacing(3)
```
parkjh: this doesn't change any snapshots?

apostolnikov: I think it's because it was never applied. This is what we currently have. Platform.os... is not a valid syntax and it always fallbacks to the default style.

image

parkjh: git blame? I hope it wasn't me...

PR #2118, opened 2019-07-12
Feat/Trigger scroll event on scroll

nehasri89:
The issue was that the scroll depth event was triggered even on page views without scrolling the page. This is fixed in the PR.

### DIFF
--- a/packages/tracking/__tests__/tracking.web.test.js
+++ b/packages/tracking/__tests__/tracking.web.test.js
@@ -3,6 +3,7 @@ import { View, Text } from "react-native";
 import PropTypes from "prop-types";
 import renderer from "react-test-renderer";
 import { mount } from "enzyme";
+import { delay } from "@times-components/test-utils";
 import { withTrackScrollDepth } from "../src/tracking";
 import withTestContext from "./test-tracking-context";
 import sharedTrackingTests from "./shared-tracking-tests";
@@ -95,12 +96,24 @@ class ListComponent extends Component {
 module.exports = () => {
   describe("WithTrackScrollDepth", () => {
     let realGetElementById;
+    let eventMap;
 
     beforeEach(() => {
       global.window.IntersectionObserver = FakeIntersectionObserver;
 
       realGetElementById = global.window.document.getElementById;
       global.window.document.getElementById = stubGetElementById;
+      eventMap = {};
+
+      global.window.addEventListener = jest.fn((eventName, callback) => {
+        eventMap[eventName] = callback;
+      });
+
+      global.window.removeEventListener = jest.fn(eventName => {
+        delete eventMap[eventName];
+      });
+
+      global.window.pageYOffset = 0;
     });
 
     afterEach(() => {
@@ -109,9 +122,12 @@ module.exports = () => {
       global.window.document.getElementById = realGetElementById;
 
       FakeIntersectionObserver.clearObserving();
+
+      global.window.addEventListener = global.window.addEventListener;
+      global.window.removeEventListener = global.window.removeEventListener;
     });
 
-    it("tracks scroll depth", () => {
+    it("does not tracks scroll depth without scroll", () => {
       const reporter = jest.fn();
       const ListWithChildTracking = withTestContext(
         withTrackScrollDepth(ListComponent),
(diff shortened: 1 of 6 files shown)


packages/tracking/src/track-scroll-depth.web.js
```diff
@@ -16,21 +16,25 @@ export default (
...
+      this.isOnScroll = false;
       this.viewed = new Set();
       if (typeof window !== "undefined" && window.IntersectionObserver) {
         this.observer = new window.IntersectionObserver(
           this.onObserved.bind(this),
           {
             root: null,
             rootMargin: "0px",
-            threshold: 1.0
+            threshold: 0.5
```
nehasri89: Debugging another issue related to Scroll depth on web, I have realised this is the optimum value which works in all cases. Sometimes even though the object is in viewport while you are scrolling the Intersection Observer callback is invoked with a value lesser than `1` but greater than `0.5` so we have a chance to miss it if we keep the threshold too high.

packages/tracking/__tests__/web/__snapshots__/tracking.web.test.js.snap
```diff
@@ -75,50 +77,6 @@ exports[`Tracking tests on web WithTrackScrollDepth renders when tracking contex
 </div>
 `;
 
-exports[`Tracking tests on web WithTrackScrollDepth tracks scroll depth 1`] = `
```
ValGeorgiev: This doesn't look promising, could we check why these snapshots are deleted

packages/tracking/__tests__/tracking.web.test.js
```diff
@@ -181,6 +197,12 @@ module.exports = () => {
         <ListWithChildTracking analyticsStream={reporter} items={items} />
       );
 
+      global.window.pageYOffset = 9;
+
+      await delay(1);
```
nathggns: Does `delay(0)` work?

packages/tracking/src/track-scroll-depth.web.js
```diff
@@ -16,21 +16,25 @@ export default (
...
           }
         );
       }
       this.childList = [];
     }
 
     componentDidMount() {
+      window.addEventListener("scroll", () => {
```
nathggns: You need to remove this event listener on unmount.

packages/tracking/__tests__/web/__snapshots__/tracking.web.test.js.snap
nathggns: As Val said, this doesn't look right!

packages/tracking/__tests__/tracking.web.test.js
nathggns: Actually, I'm not entirely sure these delays are necessary at all.

packages/tracking/__tests__/tracking.web.test.js
```diff
@@ -95,12 +96,24 @@ class ListComponent extends Component {
...
+
+      global.window.addEventListener = jest.fn((eventName, callback) => {
+        eventMap[eventName] = callback;
+      });
+
+      global.window.removeEventListener = jest.fn(eventName => {
+        delete eventMap[eventName];
+      });
+
+      global.window.pageYOffset = 0;
```
nathggns: What's this necessary for?

packages/tracking/__tests__/web/__snapshots__/tracking.web.test.js.snap
nehasri89: so no, the tests was initially testing if the scroll event was fired on page load, but now because we have changed the implementation to trigger scroll only when the user scrolls, I have changed the title to check it should trigger the scroll event, like you can see on the right `Tracking tests on web WithTrackScrollDepth does not tracks scroll depth without scroll`

packages/tracking/__tests__/tracking.web.test.js
nehasri89: removed

packages/tracking/src/track-scroll-depth.web.js
```diff
@@ -39,6 +42,7 @@ export default (
     }
 
     componentWillUnmount() {
+      window.removeEventListener("scroll", this.updateOnScroll);
```
nathggns: Is there a test for this?

nehasri89: nope, I will put one

PR #2119, opened 2019-07-12
fix: Fix ad placeholder size - REPLAT-4733

todorRadkov:
The ad placeholder was not getting its correct size, because it was depending on the screen size, but the ssr don't have screen size, so it returned only zeroes, and the placeholder was 0px with 1px border in the middle of the keylines. The fix was to move the loading of the ads to the client side rendering.

Jira: https://nidigitalsolutions.jira.com/browse/REPLAT-4733

Before:
image

After:
image


packages/ad/src/ad.js
```diff
@@ -29,10 +29,17 @@ class Ad extends Component {
     this.state = {
       config: getSlotConfig(slotName, screenWidth()),
       hasError: false,
-      isAdReady: false
+      isAdReady: false,
+      isClientSide: false
```
nathggns: Please move `ServerClientRender` out of `@times-components/user-state` and use that.

todorRadkov: Ok, this is done.

PR #2120, opened 2019-07-12
feat: save and share bar styles + animated transition 

nathggns:
JIRA ticket: https://nidigitalsolutions.jira.com/browse/REPLAT-6878

The save & share bar has had its styles updated, including:

- Making icon sizes more unified
- Making the hover bubbles circles rather than ovals
- Making the icons centered vertically and horizontally within their bubbles
- Making the icons and labels vertically centered within the save and share bar
- Reducing the height of the save and share bar when sticky
- Making the drop shadow less harsh 
- The save/saved label is now right aligned, so that the gap between it and the icon is consistent. 

You can see the effects of these changes in the below screenshots.

### Non sticky mode:

Screen Shot 2019-07-16 at 11 24 21

### Sticky mode:

Screen Shot 2019-07-16 at 11 12 32

### Hover states:

Screen Shot 2019-07-16 at 11 23 57

## Transition

However, as the save and share bar now changes height between its sticky and non-sticky states, a decision was made to make it smoothly transition between those two heights as you scroll. In order to do that, a "scroll progress" tracking feature was introduced as part of the sticky package, allowing you to create styles that are computed as a proportion of the users scroll progress through the sticky elements original position on the page. 

You can see the transition in the GIF below. It may take a while to load and the quality may be too low, so you can also [download a video copy from slack here](https://newscorp.enterprise.slack.com/files/WEUELNSD7/FL4BJ10UA/save_and_share_bar_transition.mp4)

ezgif com-optimize

## Todo

- [ ] Cross browser testing


packages/article-skeleton/src/sticky-save-and-share-bar.js
```diff
@@ -1,28 +1,74 @@
 import React from "react";
-import styled from "styled-components";
+import styled, { css } from "styled-components";
+import { rgba } from "polished";
```
aronshamash: what does this give us?

nathggns: I have to be able to fade out the keyline colour from the styleguide programatically. This allows us to do `rgba(hex, opacity)`. It's used here. https://github.com/newsuk/times-components/blob/feat/save-share-bar/packages/article-skeleton/src/sticky-save-and-share-bar.js#L33

PR #2121, opened 2019-07-15
fix: (REPLAT-6982) update slices width for huge breakpoint

pdimova:
This PR is to implement a simple slice content wrapper used to restrict the content width to 1180px for huge breakpoints

This is by design:
Screenshot 2019-07-15 at 16 57 25

**Before:**
Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2019-07-15 at 17 07 31

**After:**
Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2019-07-15 at 17 08 32

### DIFF
--- a/packages/edition-slices/src/slices/shared/content-wrapper.js
+++ b/packages/edition-slices/src/slices/shared/content-wrapper.js
@@ -1,18 +1,15 @@
+/**
+ * ContentWrapper restricts the content area of each slice. It is used together with the Gutter component,
+ * which sets the gutter space around the content. This component is currently used only for huge breakpoints.
+ */
+
 import React from "react";
 import PropTypes from "prop-types";
 import { View } from "react-native";
-import { sliceContentMaxWidth } from "@times-components/styleguide";
+import styles from "./styles";
 
 const ContentWrapper = ({ children }) => (
-  <View
-    style={{
-      flex: 1,
-      alignSelf: "center",
-      width: sliceContentMaxWidth
-    }}
-  >
-    {children}
-  </View>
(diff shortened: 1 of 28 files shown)


packages/edition-slices/src/slices/shared/content-wrapper.js
```diff
@@ -0,0 +1,22 @@
...
+import { View } from "react-native";
+import { sliceContentMaxWidth } from "@times-components/styleguide";
+
+const ContentWrapper = ({ children }) => (
+  <View
+    style={{
+      flex: 1,
+      alignSelf: "center",
+      width: sliceContentMaxWidth
+    }}
```
ValGeorgiev: Please move this to style.js file

packages/edition-slices/src/slices/shared/content-wrapper.js
```diff
@@ -0,0 +1,22 @@
...
+  >
+    {children}
+  </View>
+);
+
+ContentWrapper.propTypes = {
+  children: PropTypes.node.isRequired
+};
+
+export default ContentWrapper;
```
ValGeorgiev: Could we have a description at the top of the file? (What is ContentWrapper, why is needed, where is used and etc)

pdimova: I added a couple of words at the beginning, you can check

packages/edition-slices/src/slices/shared/content-wrapper.js
pdimova: Done

PR #2122, opened 2019-07-16
fix: paragraph with dropcap margin bottom fixed

infreezer:
[Jira](https://nidigitalsolutions.jira.com/browse/REPLAT-6707).

What was done:
Spacing below paragraph with drop cap updated.

What's affected: 
iOS and Android app

Before:
para-spacing-before

After:
para-spacing-after

### DIFF
--- a/packages/article-skeleton/src/article-skeleton.js
+++ b/packages/article-skeleton/src/article-skeleton.js
@@ -98,7 +98,12 @@ const renderText = (block, inlined = false) => {
     ));
   }
   return block.getComponent(style => (
-    <View style={{ height: block.measuredHeight }}>
+    <View
+      style={[
+        { height: block.measuredHeight },
+        styles.inlineParagraphContainer
+      ]}
+    >
       {block.block.children.map(line =>
         line.idealSpans.map(span => {
           if (span.href) {
--- a/packages/article-skeleton/src/article-skeleton.js
+++ b/packages/article-skeleton/src/article-skeleton.js
@@ -98,7 +98,7 @@ const renderText = (block, inlined = false) => {
     ));
   }
   return block.getComponent(style => (
-    <View style={[{ height: block.measuredHeight }]}>
+    <View style={{ height: block.measuredHeight }}>
       {block.block.children.map(line =>
         line.idealSpans.map(span => {
           if (span.href) {
--- a/packages/article-skeleton/src/article-skeleton.js
+++ b/packages/article-skeleton/src/article-skeleton.js
@@ -257,6 +257,7 @@ class ArticleSkeleton extends Component {
     const { fontScale } = Dimensions.get("window");
 
     const textFlow = new Layout.TextFlow({
+      padding: 20,
       flow: rows.map(rowData =>
         ArticleRowFlow({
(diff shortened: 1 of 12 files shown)


packages/article-skeleton/src/styles/shared.js
```diff
@@ -14,6 +14,9 @@ const globalStyle = {
     backgroundColor: "#ffffff",
     maxWidth: "100%",
     width: maxWidth
+  },
+  inlineParagraphContainer: {
+    marginBottom: spacing(-2)
```
nathggns: This seems like a hack. Why was it done this way?

infreezer: The height of the inline paragraph is calculated in the text-flow component - Text.js file.
If the Text.js is updated the pull quote and caption on tablet are be also affected.
So the solution with spacing(-2) is safer, isolated and less time consuming for testing.

nathggns: You're not actually fixing the miscalculation though that leads to it being calculated as 10px too high. I think if possible we should fix that bug, but happy to be overruled on that. @jmars might be able to make some guesses about whether its possible to fix the bug.

nathggns: Certainly wouldn't immediately approve this without at least that conversation happening. Has it?

infreezer:
Do not merge tag added.
Because the change is affecting the pull quote.
Where the spacing is also currently 30px, instead of 20.

current-pull-quote


After the change the pull quote spacing doesn't look good:
after-pull-quote-broken

nathggns:
@infreezer might be worth closing the PR as it seems a whole new approach to fixing this issue might be necessary.

ValGeorgiev:
I am going to investigate if **text-flow** can help here, if not and it looks like it is working with the current implementation, I think we should merge it

jmars:
Hi, I have added the ability to do this in text-flow, please have a look at the diff

ValGeorgiev:
Thanks @jmars, looks good. Could you check the build and do we need @infreezer fix in this case?

ValGeorgiev:
@jmars I know you are very busy. Just wanna check what is the status here?

packages/article-skeleton/src/styles/shared.js
infreezer: Hacky margin removed.

infreezer:
The spacing between paragraphs looks good after the commit from @jmars.
article-para-drop-cap-padding-android

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -93,7 +93,7 @@ const renderText = (block, inlined = false) => {
     ));
   }
   return block.getComponent(style => (
-    <View style={{ height: block.measuredHeight }}>
+    <View style={[{ height: block.measuredHeight }]}>
```
parkjh: why?

infreezer: Missed. I'll clean it.

infreezer:
Do not merge added. Still in progress.
image

PR #2123, opened 2019-07-17
fix: (REPLAT-6787) article video label alignment

pdimova:
Story: https://nidigitalsolutions.jira.com/secure/RapidBoard.jspa?rapidView=1116&modal=detail&selectedIssue=REPLAT-6787

Video icon and label text should be vertically aligned but currently they are not. This is observed mainly in Chrome on Windows, but actually it is an issue for all browsers despite the OS. 

Currently on all browsers once we put the items' borders:
Screenshot 2019-07-16 at 14 09 07 copy

The incorrect alignment is better noticed on Windows browsers because the used font `GillSansMTStd-Medium` behaves differently on Windows vs Mac/Linux and all the mobile simulators

*[Some fonts on apple devices has more padding than Windows font](https://stackoverflow.com/questions/25209060/some-fonts-on-apple-devices-has-more-padding-than-windows-font)

GillSansMTStd-Medium on macOS Mojave/Chrome 75
Screenshot 2019-07-16 at 17 40 31 copy

GillSansMTStd-Medium on Windows 10/Chrome 75
Capture-Example copy

The current PR suggests vertically aligning the video icon in the centre and adjusting its padding size so it looks acceptable on both Mac/Win. This is the easiest/faster approach but not 100% accurate.
Other solutions include fixing the vertical alignment font metrics or detecting and applying custom CSS for each OS but all of them will need further investigation.

Attaching screenshots with the current code change:
Before:
**Windows/Chrome**
Capture2-Chrome

After:
**Windows/Chrome 75**
Capture-Chrome-Solution

**Mac/Chrome 75**
Screenshot 2019-07-17 at 13 47 21

### DIFF
--- a/packages/video-label/src/style/shared.js
+++ b/packages/video-label/src/style/shared.js
@@ -10,7 +10,7 @@ const styles = {
   },
   iconContainer: {
     paddingBottom: spacing(1),
-    alignSelf: 'center'
+    alignSelf: "center"
   },
   title: {
     ...fontFactory({
--- a/packages/video-label/src/style/shared.js
+++ b/packages/video-label/src/style/shared.js
@@ -9,7 +9,8 @@ const styles = {
     marginTop: -1
   },
   iconContainer: {
-    paddingBottom: spacing(1)
+    paddingBottom: spacing(1),
+    alignSelf: 'center'
   },
   title: {
     ...fontFactory({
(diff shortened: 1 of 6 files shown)


nathggns:
The Gill Sans font is causing us serious problems. We should make a tech debt ticket to address this.

packages/video-label/src/style/shared.js
```diff
@@ -9,7 +9,8 @@ const styles = {
     marginTop: -1
   },
   iconContainer: {
-    paddingBottom: spacing(1)
+    paddingBottom: spacing(1),
+    alignSelf: 'center'
```
nathggns: Need to use double quotes and then the build should pass (also may need to update more snaps)

PR #2124, opened 2019-07-18
fix: article scroll tracking

nathggns:
https://nidigitalsolutions.jira.com/browse/REPLAT-7031

In this PR, I've fixed article scroll tracking. I've done that by updating the two pass rendering mechanism to remove the "wrapper" Client and Server components. These were originally included to prevent React rendering both the Client and the Server "paths" whilst only one was needed. However, it turns out this breaks React's reconciling mechanism, and it leads to React unnecessarily destroying and recreating dom nodes on the second "pass" of rendering, even when this is unnecessary.

I thought for performance reasons it was necessary to continue to not render both "paths" whilst only using one, and so I have removed the "wrapper" components from `ClientUserStateConsumer`, and updated `ServerClientRender` to take render functions which it only calls whilst necessary. This is unfortunately a less elegant API, but it feels necessary.

### DIFF
--- a/packages/utils/src/server-client-render.js
+++ b/packages/utils/src/server-client-render.js
@@ -28,13 +28,13 @@ class ServerClientRender extends Component {
     const { isClient } = this.state;
     const { client, server } = this.props;
 
-    return isClient ? client : server;
+    return isClient ? client && client() : server && server();
   }
 }
 
 ServerClientRender.propTypes = {
-  server: PropTypes.node,
-  client: PropTypes.node
+  server: PropTypes.func,
+  client: PropTypes.func
 };
 
 ServerClientRender.defaultProps = {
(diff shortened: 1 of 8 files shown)


packages/utils/src/server-client-render.js
```diff
@@ -28,13 +28,13 @@ class ServerClientRender extends Component {
     const { isClient } = this.state;
     const { client, server } = this.props;
 
-    return isClient ? client : server;
+    return isClient ? client && client() : server && server();
```
ValGeorgiev: Why you are using functions instead of nodes?

nathggns: I explained in the PR description! 

> I've done that by updating the two pass rendering mechanism to remove the "wrapper" Client and Server components. These were originally included to prevent React rendering both the Client and the Server "paths" whilst only one was needed. However, it turns out this breaks React's reconciling mechanism, and it leads to React unnecessarily destroying and recreating dom nodes on the second "pass" of rendering, even when this is unnecessary.

> I thought for performance reasons it was necessary to continue to not render both "paths" whilst only using one, and so I have removed the "wrapper" components from ClientUserStateConsumer, and **updated ServerClientRender to take render functions which it only calls whilst necessary. This is unfortunately a less elegant API, but it feels necessary.**

nehasri89:
LGTM apart from adding the tests as discussed with @nathggns

PR #2126, opened 2019-07-19
fix: uppercase font on style magazine templates

nathggns:
https://nidigitalsolutions.jira.com/browse/REPLAT-7122

image

### DIFF
--- a/packages/styleguide/src/theme/theme-factory.js
+++ b/packages/styleguide/src/theme/theme-factory.js
@@ -50,6 +50,12 @@ const magazineFontPicker = (
   return config[template][section] || defaultFont;
 };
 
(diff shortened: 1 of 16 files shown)


packages/styleguide/src/theme/theme-factory.js
```diff
@@ -50,6 +50,12 @@ const magazineFontPicker = (
   return config[template][section] || defaultFont;
 };
 
+const headlineCasePicker = (section = "default", template = "mainstandard") =>
+  section.toLowerCase() === "style" &&
```
aronshamash: would probably put this in config like the above

nathggns: It's easier to read this way as discussed on Slack.

PR #2127, opened 2019-07-19
Fix: Sponsored ads

aronshamash:
- In order to make injecting Sponsored Articles easier for Nativo, commercial have asked for a wrapping div with id `sponsored-article`.


aronshamash:
@nathggns exactly. same here but they insisted at the time of putting in the script tag it wasn't needed....

PR #2130, opened 2019-07-23
Feat/lead one full

Ivanov558:
Add new padding  and update snapshots.

image


packages/edition-slices/src/tiles/tile-r/styles/index.js
```diff
@@ -13,8 +13,8 @@ const headlineFontSizeResolver = {
 export default breakpoint => ({
   container: {
     paddingBottom: spacing(3),
-    paddingHorizontal: spacing(4),
-    paddingTop: spacing(2)
+    paddingHorizontal: spacing(8),
+    paddingTop: spacing(3)
```
pdimova: we could have `paddingVertical:spacing(3)` instead. Also `paddingHorizontal: spacing(8)` would give 40pt, but the parent element has also 10pt vertical padding, so as a whole you have 50pt instead of 40. In addition, the StarButton needs more padding at the bottom and the headline has a bigger gap when the star is disabled.

Screenshot 2019-07-23 at 14 34 35

PR #2132, opened 2019-07-23
feat/puzzles-768: Add new styles for puzzles and snapshots

Ivanov558:
Add new styles for puzzles and snapshots

image


packages/edition-slices/src/slices/puzzle/styles.js
```diff
@@ -14,8 +14,8 @@ const mediumBreakpointStyles = {
   container: {
     flex: 1,
     flexDirection: "row",
-    paddingHorizontal: spacing(3),
-    paddingTop: spacing(2)
+    paddingHorizontal: spacing(8),
```
pdimova: Can you make the vertical padding 30pt (`spacing(6)` )? Because now you have again 50pt on the left/right. Also maybe some bottom padding? 
Simulator Screen Shot - iPad Air 2 - 2019-07-24 at 16 59 37

PR #2133, opened 2019-07-24
Fix/spot label

ValGeorgiev:
Fix comments label rendering on web

Tested on Fast 3G:
giphy (1)

### DIFF
--- a/packages/article-comments/__tests__/web/__snapshots__/shared.web.test.js.snap
+++ b/packages/article-comments/__tests__/web/__snapshots__/shared.web.test.js.snap
@@ -1,5 +1,23 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Render comments label, when comments are loaded 1`] = `
+<div>
+  <p>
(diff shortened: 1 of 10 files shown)


aronshamash:
@ValGeorgiev can we have a test for this ?

nehasri89:
nice work! LGTM apart from the test as @aronshamash mentioned

packages/article-comments/__tests__/web/__snapshots__/shared.web.test.js.snap
```diff
@@ -34,42 +35,23 @@ exports[`enabled comments 1`] = `
...
+
+exports[`enabled comments 1`] = `
+<div>
   <div />
 </div>
 `;
 
 exports[`single comment 1`] = `
 <div>
-  <p>
```
aronshamash: @ValGeorgiev is this expected ?

ValGeorgiev: @aronshamash yes, we don't render comment label, until they are presented

PR #2136, opened 2019-07-24
fix: REPLAT-7145, REPLAT-7130

jmars:
Fixes a bunch of issues causing high cpu usage in some articles, depending on template.


packages/text-flow/src/Layout/TextFlow.js
```diff
@@ -17,11 +17,17 @@ export default class TextFlow extends Container {
   constructor(props = {}) {
     super();
     Object.assign(this, props);
-    this.layout();
+  }
+
+  get block() {
```
parkjh: 👍

PR #2137, opened 2019-07-25
fix: the copy link banner should appear above the save and share bar 

nathggns:
https://nidigitalsolutions.jira.com/browse/REPLAT-6967

In this PR, I've made it so that the copy link banners appears above the save and share bar. I've done that by making the save & share bar be inserted at the same level in the DOM as the message bar.

Because of how our "open in new tab" links in storybooks were implemented before, they were causing the message bar (and now the save & share bar) to "stick" about ten pixels off from where they should be, causing a gap at the top only in storybook. I've reimplemented those links as a fixed position button to avoid this issue.

## New tab button

image

## Copy link banner

copy link banner


packages/storybook/src/new-tab.js
```diff
@@ -0,0 +1,36 @@
...
+
+function NewTab() {
+  return (
+    typeof document === "object" &&
+    window !== window.top && (
+      <a
+        href={`/iframe.html${window.top.location.search}`}
+        rel="noopener noreferrer"
+        target="_blank"
+        dangerouslySetInnerHTML={{ __html: svg }}
```
ValGeorgiev: Why we are rendering the svg with dangerouslySetInnerHTML and not using it as a regular html <a> <svg /> </a> ?

nathggns: 👍  done

nathggns:
@ValGeorgiev screenshot & gif added.

PR #2140, opened 2019-07-26
Feat: (replat 7163) fix issues from design review

jmars:

### DIFF
--- a/packages/article-skeleton/src/article-body/article-body-row.js
+++ b/packages/article-skeleton/src/article-body/article-body-row.js
@@ -89,7 +89,7 @@ export default ({
         new Body(value)
       ]);
       const dropCapGap =
-        typeof value === 'string' && ['M', 'W'].includes(value) ? 18 : 10;
+        typeof value === "string" && ["M", "W"].includes(value) ? 18 : 10;
       const capWidth = (cap[0].measuredWidth + dropCapGap) * fontScale;
 
       return {
--- a/packages/article-skeleton/src/article-body/article-body-row.js
+++ b/packages/article-skeleton/src/article-body/article-body-row.js
@@ -88,7 +88,9 @@ export default ({
       const cap = getDropCap(children, fonts[dropCapFont], height, [
         new Body(value)
       ]);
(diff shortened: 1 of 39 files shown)


packages/article-skeleton/src/article-skeleton.js
```diff
@@ -30,7 +34,7 @@ const viewabilityConfig = {
 };
 
 const convertStyles = ({ font, size }) => ({
-  fontFamily: font,
+  fontFamily: "TimesDigitalW04",
```
parkjh: why is it hard coded?

parkjh: if you remove it or have it as default parameter value for font I am happy to approve it

packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -88,7 +88,9 @@ export default ({
       const cap = getDropCap(children, fonts[dropCapFont], height, [
         new Body(value)
       ]);
-      const capWidth = (cap[0].measuredWidth + 20) * fontScale;
+      const dropCapGap =
+        typeof value === "string" && ["M", "W"].includes(value) ? 18 : 10;
```
ValGeorgiev: What are these values here? 10 & 18

packages/caption/src/centred-caption.js
```diff
@@ -10,6 +10,7 @@ const CentredCaption = ({ children, credits, style, text }) => (
       ...style,
       text: {
         ...style.text,
+        marginTop: 5,
```
ValGeorgiev: could we use spacing?

packages/caption/src/styles/index.android.js
```diff
@@ -10,11 +10,13 @@ const styles = StyleSheet.create({
...
     ...sharedStyles.credits,
-    lineHeight: 20
+    lineHeight: 20,
+    marginTop: 0
   },
   text: {
     ...sharedStyles.text,
-    lineHeight: 20
+    lineHeight: 20,
+    marginTop: -5
```
ValGeorgiev: spacing here as well

PR #2141, opened 2019-07-26
fix: theme factory handling of null sections

aronshamash:
- default values don't handle `null` values well so need the extra check for section.
- Fix Render 500 errors in CI with latest TC.

### DIFF
--- a/packages/styleguide/src/theme/theme-factory.js
+++ b/packages/styleguide/src/theme/theme-factory.js
@@ -51,6 +51,7 @@ const magazineFontPicker = (
 };
 
 const headlineCasePicker = (section = "default", template = "mainstandard") =>
+  section &&
   section.toLowerCase() === "style" &&
   ["indepth", "magazinestandard", "magazinecomment"].includes(template)
     ? "uppercase"


packages/styleguide/src/theme/theme-factory.js
```diff
@@ -51,6 +51,7 @@ const magazineFontPicker = (
 };
 
 const headlineCasePicker = (section = "default", template = "mainstandard") =>
+  section &&
```
nehasri89: but we are specifying a default value in the params

aronshamash: yes it is good for `undefined` but not `null` see https://medium.com/@bmb21/default-parameters-and-nulls-in-javascript-9149c8d26b0a

parkjh: no test...?

ValGeorgiev: Unfortunately.. no one function in theme-factory.js has unit tests... Lets merge this one and i will put in my to do list to increase the test coverage in this file. Thanks

PR #2144, opened 2019-07-29
spike: responsive API

nathggns:
This is very messy as I changed my mind about how to do it a number of times, but I think something like this is a good alternative for our media query issue mentioned on Slack today. 

The only important changes are inside the Responsive package. 

The API is defined here https://github.com/newsuk/times-components/blob/spike/responsive-api/packages/responsive/src/hooks.js (ignore the name of this file)

There is an example here https://github.com/newsuk/times-components/blob/spike/responsive-api/packages/responsive/responsive.showcase.js

The API is very `styled-components`-y.

```js
const MyComponent = Responsive.View`
    background-color: blue;

    ${Responsive.minWidth.huge`
        background-color: red;
    `};
`;
```

### DIFF
--- a/packages/article-comments/package.json
+++ b/packages/article-comments/package.json
@@ -42,7 +42,7 @@
     "@times-components/link": "3.4.31",
     "@times-components/styleguide": "3.28.36",
     "prop-types": "15.7.2",
-    "styled-components": "4.2.0"
+    "styled-components": "4.3.2"
   },
   "devDependencies": {
     "@times-components/eslint-config-thetimes": "0.8.12",
--- a/packages/article-extras/package.json
+++ b/packages/article-extras/package.json
@@ -69,7 +69,7 @@
     "@times-components/styleguide": "3.28.36",
     "@times-components/user-state": "0.0.15",
     "prop-types": "15.7.2",
-    "styled-components": "4.2.0"
+    "styled-components": "4.3.2"
   },
   "peerDependencies": {
     "react": ">=16.5",
--- a/packages/article-image/package.json
+++ b/packages/article-image/package.json
@@ -75,7 +75,7 @@
     "@times-components/responsive": "0.4.54",
     "@times-components/styleguide": "3.28.36",
     "prop-types": "15.7.2",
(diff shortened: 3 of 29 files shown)


packages/responsive/src/hooks.js
```diff
@@ -0,0 +1,59 @@
...
+const minWidth = min => {
+  const parsed = parse(min);
+  return mq(width => width >= parsed);
+};
+const maxWidth = max => {
+  const parsed = parse(max);
+  return mq(width => width <= parsed);
+};
+
+function createResponsiveViewTag(as) {
```
parkjh: @nathggns just so that I understand the context ( ;p ) what use case do you think it will benefit? or is this part of a tech debt ticket?

parkjh:
https://nidigitalsolutions.jira.com/browse/REPLAT-7158

packages/responsive/src/hooks.js
nathggns: We currently used web styled components with React Native web components. This only works because RNW components currently support styling via className. This is deprecated and is going away. So in some places we need to change to using `styled.div` instead of `styled(View)`. We can't do that in some places, because we're using style arrays or RN styles (`margin-horizontal`) for example (`key-facts` is an example of a package that does this). 

For these packages, we need to change from using web styled components to native styled components. However for obvious reason, native styled components do not support media queries, so we need a replacement API for that. That's what this is a proof of concept for. 

It's the prerequisite for a tech debt ticket that you linked below.

parkjh:
https://github.com/styled-components/styled-components/issues/2061
https://github.com/styled-components/styled-components/issues/1874

packages/responsive/src/hooks.js
parkjh: @nathggns does the example showcase currently work on storybook on this branch?

nathggns: Should do yes.

parkjh:
@nathggns closing in favour of https://github.com/newsuk/times-components/pull/2147

PR #2146, opened 2019-07-31
fix: return nothing if adblock is detected

todorRadkov:
JIRA:
https://nidigitalsolutions.jira.com/browse/REPLAT-7192

I have added detection for adblock in render:
https://github.com/newsuk/cps-content-render/pull/2860
Also there was a change in the flag and the way it is added to the project in this PR:
https://github.com/newsuk/cps-content-render/pull/2862

So if the user has been detected using adBlock, the Ad component will return nothing (null).
Also I have created a snapshot test for this case.

### DIFF
--- a/packages/ad/__tests__/web/ad-with-style.web.test.js
+++ b/packages/ad/__tests__/web/ad-with-style.web.test.js
@@ -66,8 +66,6 @@ const tests = [
   {
     name: "multiple ad slots",
     test: () => {
-      window.isAdBlockDisabled = true;
-
       const wrapper = mount(
         <AdComposer adConfig={adConfig}>
           <Fragment>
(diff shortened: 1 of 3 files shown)


nehasri89:
you will need to update ssr client side

ValGeorgiev:
@nehasri89  I don't think this is needed here

packages/ad/__tests__/web/ad-with-style.web.test.js
```diff
@@ -66,6 +66,8 @@ const tests = [
   {
     name: "multiple ad slots",
     test: () => {
+      window.isAdBlockDisabled = true;
```
ValGeorgiev: is this needed?

todorRadkov: Yes this is needed for the test, because the component can't take this value from the included file ads.js in render, so we set it manually.

todorRadkov: otherwise the test fails

todorRadkov: The variable has changed so this is not needed anymore.

PR #2147, opened 2019-07-31
feat: native media queries

nathggns:
This PR creates a package for supporting media queries with React Native's View & Text components, and shows you how to use it in `key-facts`. 

### [View Package Readme](https://github.com/newsuk/times-components/blob/feat/native-media-queries/packages/responsive-styled-components-native/README.md)

I decided not to use the responsive context in the end as it was suggested that the performance implications of letting each styled component manafge its own event listener are better than that of vhaving a global one that re-renders a top level context. 

There's an example of the API below here. 

```js
import responsiveStyled, {
  mediaQuery
} from "@times-components/responsive-styled-components-native";

const Box = responsiveStyled.View`
  width: 80px;
  height: 80px;
  background-color: red;
  
  ${mediaQuery.maxWidth.medium`
    background-color: blue; 
  `} 
   
  ${mediaQuery.minWidth.huge`
    width: 100%; 
  `}
`;
```

## Todo

Because we don't want styles that are behind a media query to show up in the `jest-styled-components` output (as we've got our own serialisation methods), `markupMediaQuery` currently swaps out its implementation when in a test environment. This means that the tests for the package itself may not work, so we need to think of a smarter way of handling this.

<s>I think the serialisation doesn't quite work perfectly yet. It seems to work for one test, but not the other. Need to spend more time looking into that.</s>

Performance needs testing properly. 

Needs testing on native platforms (ironic, consider its designed for use with styled components native)

I wish the serialisation didn't require reaching into react native web internals. Consider alternatives to this. 

Confirm that serialisation code doesn't ship to production

<s>Seperate the serialisation process from jest-serialiser somehow. Probably by working with jest-styled-components instead.</s>

<s>Styles that are not hidden behind a media query should show up in the snapshot, but currently don't seem to. I think jest-styled-snapshots` and `styled-components/native` don't work together. </s>


packages/responsive-styled-components-native/README.md
```diff
@@ -0,0 +1,52 @@
+# ResponsiveStyledComponentsNative
```
ValGeorgiev: I think this could live in the styleguide, what do you think?

packages/responsive-styled-components-native/src/responsive-styled-components-native.js
```diff
@@ -0,0 +1,105 @@
...
+  }, []);
+
+  return screenWidth;
+}
+
+// Creates a template tag which you can use inside our Responsive styled-components
+// wrapper to selectively apply styles based on screen width.
+function mediaQuery(shouldApplyStyles) {
+  function mediaQueryTag(...templateStringArguments) {
+    function styledComponentPropMapper(componentProps) {
```
ValGeorgiev: could we not create a function in another function in another function please

packages/responsive-styled-components-native/src/responsive-styled-components-native.js
```diff
@@ -0,0 +1,105 @@
+import React, { forwardRef, useEffect, useState } from "react";
```
ValGeorgiev: So with this approach we are going to use media queries only for web? Or this is supported for web as well?

bidah:
A comment. I think that it could be great to include the API with the css prop. You use styled components interchangeably so missing this API with props in mind would be incomplete.

packages/responsive-styled-components-native/README.md
nathggns: I'm not so sure – I've tried to make this as self contained a package as possible because it's almost like a straight forward wrapper around styled-components with the same API.

packages/responsive-styled-components-native/src/responsive-styled-components-native.js
nathggns: Why? The `mediaQuery` function is a factory for ES6 template tag functions. The template tag function (similarly to how styled-components itself works) is a factory itself for a function that maps props into the final CSS. You need the multiple layers as we're closing over `shouldApplyStyles` and `...templateStringArguments` for use in the prop mapping function. The first layer is a factory and is only called once – the second is called wherever a styled component with a media query is defined, and the third is called by styled components whenever it wants to generate the CSS.

packages/responsive-styled-components-native/src/responsive-styled-components-native.js
nathggns: This is currently only used on web but within RNW's version of RN components (View, Text, etc). It should work in native with no changes I believe.

nathggns:
@bidah that is a fair point and I will update it to export the `css` helper that styled-components exports.

parkjh:
@nathggns were there no packages that we could have used?

nathggns:
@parkjh I searched. There's packages that replace React Native's Stylesheet API to add media query support, but you can't make styled-components use a specific stylesheet API as it imports it directly from RN. There doesn't seem to be any that do what we're doing here, and I did a lot of searching. I'm happy for you to suggest alternatives :)

nathggns:
There's also a replacement for styled-components that have media query support on native but it didn't seem supported anymore. Also we're probably way too far down the line to start replacing styled-components.

parkjh:
> There's also a replacement for styled-components that have media query support on native but it didn't seem supported anymore. 

can you link it?

> Also we're probably way too far down the line to start replacing styled-components.

agreed

parkjh:
@nathggns fyi this is happening...
https://github.com/newsuk/times-components/pull/2170
https://nidigitalsolutions.jira.com/browse/PPDSC-142

nathggns:
> can you link it?

@parkjh  https://jacobp100.github.io/cssta//native/

parkjh:
> > can you link it?
> 
> @parkjh https://jacobp100.github.io/cssta//native/

hmm the last commit was last august... not sure if that's `not supported`... but ok
basically my concern is building something that may will be replaced later...

harrymansworth:
> I decided not to use the responsive context in the end as it was **_suggested_ **_that the performance implications_ of letting each styled component manafge its own event listener are better than that of vhaving a global one that re-renders a top level context.

Did we undertake any benchmarking to confirm this suggestion to be correct before commencing? IMO it's important we have metrics to support any decision towards an engineering solution; especially one to fix a (perceived) performance issue (which, anecdotally, tend to resolve to be negligible).

nathggns:
> > I decided not to use the responsive context in the end as it was **_suggested_**  _that the performance implications_ of letting each styled component manafge its own event listener are better than that of vhaving a global one that re-renders a top level context.
> 
> Did we undertake any benchmarking to confirm this suggestion to be correct before commencing? IMO it's important we have metrics to support any decision towards an engineering solution; especially one to fix a (perceived) performance issue (which, anecdotally, tend to resolve to be negligible).

No, for three reasons:

1. The difference is likely to be neglible and dependent entirely on how frequently this package is used
2. It's incredibly easy to change this at a later stage with little to no impact on the consumer. All of our components will be wrapped in a Responsive context *anyway*. 
3. The performance of the implementation will largely be dependent on how/where/how frequently its used. Because the same API is implementable with both "methods" for accessing the screen width, I decided to use what felt like the cleanest implementation for now. Once we've implemented this, if we decide that the event listeners are causing a problem we can very easily revert to a shared context. Whether that's a shared context specific to this package or using the Responsive package is a question for a later time.

My feeling as I've been developing this API is that it should probably entirely replace the responsive package anyway. Anywhere that is doing styling dependent on screen size should use this. Once that migration has been made, we can see what limited uses of the responsive package remain and see whether a future exists for the responsive package or whether all code can be migrated to using this package.

packages/responsive-styled-components-native/src/responsive-styled-components-native.js
nathggns: We're doing the same thing for the actual responsive views too if you hadn't noticed.

nathggns:
Closing this now as it is stale and will not be continued on for a while.

PR #2160, opened 2019-08-06
chore: remove lib cli

parkjh:


parkjh:
to be replaced by react native cli `react-native init <package-name> --version 0.59.10 --template typescript`

ValGeorgiev:
> to be replaced by react native cli `react-native init <package-name> --version 0.59.10 --template typescript`

Did you try that? What are the pros and cons of this ?

parkjh:
> > to be replaced by react native cli `react-native init <package-name> --version 0.59.10 --template typescript`
> 
> Did you try that? What are the pros and cons of this ?

pro: we do not need to maintain our own cli and react native cli should be more stable
con: may need additional steps to adjust for custom non react native package

PR #2161, opened 2019-08-06
fix: revert design changes for web

jmars:
Revert design changes that are breaking ayespy snapshots on web until there is confirmed signoff


parkjh:
@jmars is this the Firefox issue?

PR #2162, opened 2019-08-06
chore: remove flowtype eslint preset

parkjh:


ValGeorgiev:
Could you add more description why you are removing that, please? We need to agree on such changes, if you describe why you are removing it and what benefits you see, it will be enough. Thanks

parkjh:
> Could you add more description why you are removing that, please? We need to agree on such changes, if you describe why you are removing it and what benefits you see, it will be enough. Thanks

we do not use flow...

ValGeorgiev:
> > Could you add more description why you are removing that, please? We need to agree on such changes, if you describe why you are removing it and what benefits you see, it will be enough. Thanks
> 
> we do not use flow...

I am talking overall for all chore PRs

parkjh:
> > > Could you add more description why you are removing that, please? We need to agree on such changes, if you describe why you are removing it and what benefits you see, it will be enough. Thanks
> > 
> > 
> > we do not use flow...
> 
> I am talking overall for all chore PRs

will do

parkjh:
@jmars text-flow packages has a lot of `error  Parsing error: Unexpected token =` issues...

PR #2169, opened 2019-08-07
fix: REPLAT-7374 disable RNDeviceInfo on ios

jmars:


packages/ad/src/dom-context.android.js
```diff
@@ -0,0 +1,229 @@
...
+          height,
+          width
+        }}
+      >
+        {loaded && (
+          <WebView
+            onMessage={this.handleMessageEvent}
+            onNavigationStateChange={this.handleNavigationStateChange}
+            originWhitelist={
+              Platform.OS === "android"
```
parkjh: why?

packages/ad/src/dom-context.android.js
```diff
@@ -0,0 +1,229 @@
...
+import { Viewport } from "@skele/components";
+import DeviceInfo from "react-native-device-info";
+import webviewEventCallbackSetup from "./utils/webview-event-callback-setup";
+import logger from "./utils/logger";
+import { propTypes, defaultProps } from "./dom-context-prop-types";
+import { calculateViewportVisible } from "./styles/index";
+
+const ViewportAwareView = Viewport.Aware(View);
+
+class DOMContext extends PureComponent {
```
parkjh: no tests?

packages/ad/src/dom-context.android.js
```diff
@@ -0,0 +1,229 @@
...
+      return;
+    }
+    const { type, detail } = JSON.parse(json);
+    const { loaded } = this.state;
+    const { isVisible } = this;
+    switch (type) {
+      case "renderFailed":
+        onRenderError();
+        break;
+      case "unrulyLoaded": {
```
parkjh: what's firing this message over the bridge?

packages/ad/src/dom-context.android.js
```diff
@@ -0,0 +1,229 @@
...
+  };
+
+  loadAd = () => {
+    this.setState({
+      loaded: true
+    });
+  };
+
+  inViewport = () => {
+    this.isVisible = true;
```
parkjh: can you separate this out so that you don't have to check if `this.webView` ref is set or not?

PR #2171, opened 2019-08-08
fix: (REPLAT-7336) save star alignment

pdimova:
https://nidigitalsolutions.jira.com/browse/REPLAT-7336

Prior to this time, the tile star had an absolute positioning that caused issues with the star itself overlapping the content and insufficient space around it.

The new slices' designs set a rule for the star positioning - 15pt above the bottom line of the slice.

This PR suggests replacing the absolute positioning with flex-end alignment. PositionedTileStar component is introduced that helps to push the star to the bottom of its parent's container.
The default place for the tile star would be the tile summary container. And when a tile needs a dynamic teaser text, the tile star has to be moved out of its default place and lifted up to the TileLink component.

Screenshots:
Screenshot_1565609958
Screenshot_1565610023
Screenshot_1565610008
Screenshot_1565609995
Screenshot_1565609980
Screenshot_1565610048
Screenshot_1565610056
Screenshot_1565610068
Screenshot_1565610082
Screenshot_1565610134

### DIFF
--- a/packages/edition-slices/src/slices/leaders/index.js
+++ b/packages/edition-slices/src/slices/leaders/index.js
@@ -112,15 +112,12 @@ class LeadersSlice extends Component {
 
   render() {
     return (
-      <View style={styles.container}>
-        {renderHead()}
-        <ResponsiveSlice
-          renderHuge={this.renderWide}
-          renderMedium={this.renderSmall}
-          renderSmall={this.renderSmall}
-          renderWide={this.renderWide}
-        />
-      </View>
+      <ResponsiveSlice
+        renderHuge={this.renderWide}
+        renderMedium={this.renderSmall}
+        renderSmall={this.renderSmall}
+        renderWide={this.renderWide}
+      />
     );
   }
 }
--- a/packages/edition-slices/src/slices/leaders/index.js
+++ b/packages/edition-slices/src/slices/leaders/index.js
@@ -25,7 +25,7 @@ class LeadersSlice extends Component {
   constructor(props) {
     super(props);
     this.renderSmall = this.renderSmall.bind(this);
-    this.renderHuge = this.renderHuge.bind(this);
+    this.renderWide = this.renderWide.bind(this);
   }
 
   renderSmall(breakpoint) {
@@ -69,7 +69,7 @@ class LeadersSlice extends Component {
     );
   }
 
-  renderHuge(breakpoint) {
+  renderWide(breakpoint) {
     const {
       onPress,
       slice: { leader1, leader2, leader3 }
@@ -112,12 +112,15 @@ class LeadersSlice extends Component {
 
   render() {
     return (
-      <ResponsiveSlice
-        renderHuge={this.renderHuge}
-        renderMedium={this.renderSmall}
-        renderSmall={this.renderSmall}
-        renderWide={this.renderHuge}
-      />
+      <View style={styles.container}>
+        {renderHead()}
+        <ResponsiveSlice
+          renderHuge={this.renderWide}
+          renderMedium={this.renderSmall}
+          renderSmall={this.renderSmall}
+          renderWide={this.renderWide}
+        />
+      </View>
     );
   }
 }
--- a/packages/edition-slices/src/tiles/shared/positioned-tile-star.js
+++ b/packages/edition-slices/src/tiles/shared/positioned-tile-star.js
@@ -7,37 +7,35 @@ import TileStar from "./tile-star";
 import { isSaveSupported } from "./utils";
 
 import {
-  starDefaultPosition,
-  starCenterPosition,
-  starAfterContentPosition
+  starDefaultStyles,
+  starCenterStyles,
+  starUnderneathTextStyles
 } from "./styles";
 
 const PositionedTileStar = ({
-  articleId,
-  isAfterContentStar = false,
-  isCenteredStar = false,
-  isDarkStar = false,
-  starStyle = {}
+  centeredStar = false,
+  underneathTextStar = false,
+  ...props
 }) => {
   if (!isSaveSupported) return null;
 
-  const starPositionStyles = {
-    ...starDefaultPosition,
-    ...(isCenteredStar && starCenterPosition),
-    ...(isAfterContentStar && starAfterContentPosition)
-  };
-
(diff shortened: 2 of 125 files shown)


packages/edition-slices/src/tiles/tile-al/index.js
```diff
@@ -30,17 +27,21 @@ const TileAL = ({ onPress, tile }) => {
...
-          <TileSummary
-            headlineStyle={styles.headline}
-            summary={getTileSummary(tile, 800)}
-            summaryStyle={styles.summaryContainer}
-            tile={tile}
-            whiteSpaceHeight={whiteSpaceHeight}
-          />
-        )}
-      />
+      <View style={styles.summaryContainer}>
```
apostolnikov: Is it necessary to have the **<WithoutWhiteSpace>** and **<PositionedTileStar>** elements wrapped with a **<View>** . Could we avoid the addition of the <View> by passing the summaryContainer styles to the <WithoutWhiteSpace> element?

pdimova: For the star to be aligned at the bottom it necessary to have the wrapper indeed with `flex:1.` For tiles with dynamic teaser text, the star could not stay within the TileSummary component since this container can't have full width. Therefore there are 2 options - the star to be inside WithoutWhiteSpace (which is not ok IMO) or to be a sibling of that component.

apostolnikov: They can be sibling elements without them being wrapped in a **View**. And their wrapper will be indeed with **flex: 1** which in our situation is **TileLink**.

pdimova: This scenario would work when the tile does not have an image

apostolnikov: This is not exactly true. It will work perfectly fine with both scenarios.

pdimova: In cases with image, TIleLink has `flexDirection: "row"`, without the wrapper all 3 elements - TileSummery, Star and Image will be on one line

apostolnikov: Not all Tiles with images have the **row** direction. For those that do, we can leave the **View** and proceed with removing it in all the other places where there is no need for it.

packages/edition-slices/src/tiles/tile-al/styles/index.js
```diff
@@ -1,10 +1,12 @@
 import { fontFactory, spacing } from "@times-components/styleguide";
+import { fullHeightSummaryContainer } from "../../shared/styles";
 
 const styles = {
   container: {
     flex: 1,
     flexDirection: "column",
-    paddingHorizontal: spacing(2)
+    paddingHorizontal: spacing(2),
+    paddingVertical: spacing(2)
```
ValGeorgiev: padding: spacing(2)

packages/edition-slices/src/slices/leaders/index.js
```diff
@@ -44,7 +44,7 @@ class LeadersSlice extends Component {
     );
   }
 
-  renderHuge(breakpoint) {
+  renderWide(breakpoint) {
```
nathggns: Why are you changing the breakpoint?

nathggns:
Reviewing now

packages/edition-slices/src/tiles/tile-ae/styles/index.js
```diff
@@ -12,8 +13,8 @@ const styles = {
     lineHeight: 30,
     marginBottom: spacing(2)
   },
-  star: {
-    bottom: spacing(1)
+  summaryContainer: {
```
nathggns: what's wrong with just doing `summaryContainer: fullHeightSummaryContainer`?

packages/edition-slices/src/slices/leaders/index.js
pdimova: This was the initial : `renderWide={this.renderHuge}` 
For the sake of consistency, it is a name change more than anything else. Moving from smaller to bigger breakpoints `renderWide` method should be introduced first. Here both wide and huge breakpoints have the same layout and use the same tiles

packages/edition-slices/src/tiles/tile-al/styles/index.js
pdimova: done

packages/edition-slices/src/tiles/shared/positioned-tile-star.js
```diff
@@ -0,0 +1,45 @@
...
+
+  const starPositionStyles = {
+    ...starDefaultPosition,
+    ...(isCenteredStar && starCenterPosition),
+    ...(isAfterContentStar && starAfterContentPosition)
+  };
+
+  return (
+    <View style={starPositionStyles}>
+      <TileStar articleId={articleId} isDark={isDarkStar} style={starStyle} />
```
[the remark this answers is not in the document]
pdimova: thanks, looks much better

packages/edition-slices/src/tiles/tile-ae/styles/index.js
pdimova: nothing .. on second thought decided to get rid of **fullHeightSummaryContainer**

packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -148,29 +128,35 @@ TileSummary.propTypes = {
...
-  isDarkStar: false,
   labelColour: null,
   strapline: null,
   straplineStyle: null,
   style: null,
   summary: null,
   summaryStyle: null,
-  withStar: false
+  withStar: true,
+  isAfterContentStar: false,
```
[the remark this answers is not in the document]
pdimova: all renamed

packages/edition-slices/src/tiles/shared/styles/index.js
```diff
@@ -34,10 +20,24 @@ const verticalStyles = {
   }
 };
 
+const starDefaultStyles = {
+  marginTop: "auto",
+  alignItems: "flex-end",
```
parkjh: 👍

PR #2174, opened 2019-08-09
feat/comment-lead-cartoon-slie Add padding and font size and update s…

Ivanov558:
feat/comment-lead-cartoon-slice

Add padding and font size and update snapshots

cartoon

With Label:

comment-cartoon-label


packages/edition-slices/src/tiles/tile-ah/styles/index.js
```diff
@@ -37,7 +37,7 @@ const styles = (breakpoint = editionBreakpoints.medium) => ({
   },
   imageContainer: {
     overflow: "hidden",
-    width: "40%"
+    width: "10%"
```
infreezer: Why? According to the design this should be 97 pixels

infreezer:
image

The byline is misplaced.
The slices are not aligned. (the spacing is not the same as the slice above - 40px)
The padding of tileAH is not according to the design.
The itemColSeparator margin vertical should be 15 instead of 10.

PR #2175, opened 2019-08-09
chore: remove eslint disable react default props

parkjh:


ValGeorgiev:
Please remove the lint rule from the config as well

PR #2176, opened 2019-08-09
chore: add a11y tests using cypress axe for all ssr pages

parkjh:


packages/ssr/__tests__/helpers/article-helper.js
```diff
@@ -98,6 +98,34 @@ const articleTemplateTest = template =>
...
+
+    it("should pass basic a11y test", () => {
+      cy.task("startMockServerWith", {
+        Article: sundayTimesArticleWithThreeRelatedArticles,
+        User: userWithBookmarks
+      })
+        .visit("/article/8763d1a0-ca57-11e8-bde6-fae32479843d")
+        .wait(1000)
+        .injectAxe()
+        .wait(200)
```
jmars: Is there no callback to listen for instead of waiting a magic amount of time?

PR #2178, opened 2019-08-12
feat: REPLAT-7068 Leaders slice redesigned for medium breakpoint

infreezer:
[Jira story.](https://nidigitalsolutions.jira.com/browse/REPLAT-7068)

Before:
image

After:
image

### DIFF
--- a/packages/edition-slices/src/tiles/tile-m/styles/index.js
+++ b/packages/edition-slices/src/tiles/tile-m/styles/index.js
@@ -45,7 +45,6 @@ const mediumBreakpointStyles = {
   star: verticalStyles,
   strapline: {
     color: colours.functional.primary,
-    fontFamily: "TimesModern-Regular",
     fontSize: 16,
     lineHeight: 19,
     textAlign: "center",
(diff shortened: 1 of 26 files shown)


packages/edition-slices/src/tiles/tile-m/styles/index.js
```diff
@@ -24,4 +29,31 @@ const styles = {
...
+    fontSize: 20,
+    lineHeight: 20,
+    marginBottom: spacing(1),
+    marginTop: spacing(6),
+    textAlign: "center"
+  },
+  star: verticalStyles,
+  strapline: {
+    color: colours.functional.primary,
+    fontFamily: "TimesModern-Regular",
```
ValGeorgiev: Could we use existing fonts and not hardcode this here

infreezer: Sure. I'll update it to use a variable instead of hardcoded string.

infreezer: After I checked I find out that the default font of the strapline is TimesModern-Regular.
That's why I'm deleting rule in the tile.

PR #2181, opened 2019-08-13
feat: (REPLAT-7091) redesign dailyregister slice for medium breakpoint

pdimova:
https://nidigitalsolutions.jira.com/browse/REPLAT-7091

Screenshots:
Android
Screenshot_1565691971
Screenshot_1565696256

iOS
Simulator Screen Shot - iPad Air 2 - 2019-08-13 at 14 37 39


packages/edition-slices/src/slices/dailyregisterleadfour/index.js
```diff
@@ -52,6 +56,56 @@ class DailyRegisterLeadFour extends Component {
...
+    const natureLogo = (
+      <Logo
+        imageUri="https://www.thetimes.co.uk/d/img/DUR-nature-80d36dd1cd.png"
+        ratio={1 / 1}
+        style={styles.imageWrapper}
+        type="nature notes"
+      />
+    );
+
+    const birthdayLogo = (
```
ValGeorgiev: Please move logos outside of render functions, they shouldn't be rebuild every time we render.

ValGeorgiev: Ratio could be 1

packages/edition-slices/src/tiles/tile-at/markup.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
```
ValGeorgiev: Could we reuse this file from tile-s? Probably to move it in the shared folder

packages/edition-slices/src/tiles/tile-at/markup.web.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
```
ValGeorgiev: Same here

packages/edition-slices/src/slices/dailyregisterleadfour/index.js
pdimova: The way the logo looks like depends on the breakpoint that is passed within the render method.

packages/edition-slices/src/tiles/tile-at/markup.js
pdimova: dropping tileAT

PR #2184, opened 2019-08-15
Feat/analytics fixes

nehasri89:
This PR captures the analytics fixes required as part of https://nidigitalsolutions.jira.com/browse/REPLAT-6992


packages/article-skeleton/src/tracking/article-tracking-context.web.js
```diff
@@ -6,6 +6,7 @@ export default Component =>
   withTrackingContext(Component, {
     getAttrs: ({ data, pageSection, referralUrl = "" }) => ({
       articleId: get(data, "id", ""),
+      article_topic_tags: get(data, "topics", "").map(topic => topic.name),
```
nathggns: Shouldn't the get defailt here be an array? strings don't have a map method

packages/save-star-web/src/save-star-web.js
```diff
@@ -1,129 +1,45 @@
...
-
-const IconContainer = styled(HoverIcon)`
-  width: 100%;
-  height: 100%;
-  display: flex;
-  align-items: center;
-  justify-content: center;
-`;
+import styles from "./styles";
+import SaveStarWithTracking from "./save-star-with-tracking";
```
nathggns: Is this file missing?

packages/save-star-web/src/tracking/with-track-events.js
```diff
@@ -6,10 +6,14 @@ export default Component =>
       {
         actionName: "Pressed",
         eventName: "onSaveButtonPress",
-        getAttrs: ({ articleId, articleHeadline }) => ({
-          articleId,
-          articleHeadline
-        }),
+        getAttrs: props => {
+          const { articleId, articleHeadline, savedStatus } = props;
```
nathggns: Didn't need to change this from a destructured argument right?

packages/article-skeleton/src/tracking/article-tracking-context.web.js
nehasri89: topics is an array structure

packages/save-star-web/src/save-star-web.js
nehasri89: added now

packages/article-skeleton/src/tracking/article-tracking-context.web.js
nathggns: Yes, I know. But the third argument to get ia a default for if data or topics is not defined. In this case, it will error as you will be trying to call map on a string…

packages/save-star-web/__tests__/web/__snapshots__/save-star-web-tracking.test.js.snap
```diff
@@ -4,8 +4,9 @@ exports[`save star tracking events when press save star 1`] = `
 Object {
   "action": "Pressed",
   "attrs": Object {
-    "articleHeadline": "Article-headline",
+    "articleHeadline": undefined,
```
nathggns: What's happened here?

packages/save-star-web/src/tracking/with-track-events.js
```diff
@@ -6,9 +6,10 @@ export default Component =>
       {
         actionName: "Pressed",
         eventName: "onSaveButtonPress",
-        getAttrs: ({ articleId, articleHeadline }) => ({
+        getAttrs: ({ articleId, articleHeadline, savedStatus }) => ({
           articleId,
-          articleHeadline
+          articleHeadline,
+          savedStatus: !savedStatus
```
nathggns: Why do we need to invert this?

packages/article-skeleton/src/tracking/article-tracking-context.web.js
nehasri89: updated

packages/save-star-web/src/tracking/with-track-events.js
nehasri89: because the point at which tracking receives the props, the status needs to be updated in the component itself.

packages/save-star-web/__tests__/web/__snapshots__/save-star-web-tracking.test.js.snap
nehasri89: So, I am not sure how this was not undefined before, because its not a prop for `SaveStarWeb` , actually I am going to remove it from here

PR #2187, opened 2019-08-16
feat: update secondary2NoPicAnd2 for 768px

apostolnikov:
Jira: https://nidigitalsolutions.jira.com/browse/REPLAT-7498

Update SecondaryTwoNoPicAndTwo so it can have minimum 5 lines of teaser text for both upper Tiles (TileAE).

**Before:** 
beforeS

**=============================================================**

**After:** 
afterS

### DIFF
--- a/packages/edition-slices/src/tiles/shared/tile-summary.js
+++ b/packages/edition-slices/src/tiles/shared/tile-summary.js
@@ -23,7 +23,7 @@ class TileSummary extends Component {
       summary,
       summaryStyle,
       whiteSpaceHeight,
-      linesOfTeserToRender
+      linesOfTeaserToRender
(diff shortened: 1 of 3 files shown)


packages/edition-slices/src/tiles/shared/tile-summary.js
```diff
@@ -19,13 +19,19 @@ class TileSummary extends Component {
   }
 
   renderContent() {
-    const { summary, summaryStyle, whiteSpaceHeight } = this.props;
+    const {
+      summary,
+      summaryStyle,
+      whiteSpaceHeight,
+      linesOfTeserToRender
```
ValGeorgiev: Teaser* typo

PR #2191, opened 2019-08-20
feat: Updating the gutter of the edition slices and the item separator

infreezer:
Wide Before:
wide-gutter-before

Wide After:
wide-gutter-after

The item separator of the huge breakpoint is also updated
Before:
wide-separator-before

After:
huge-separator-after

### DIFF
--- a/packages/edition-slices/src/slices/shared/gutter.js
+++ b/packages/edition-slices/src/slices/shared/gutter.js
@@ -2,22 +2,36 @@ import React from "react";
 import PropTypes from "prop-types";
 import { View } from "react-native";
 import { ResponsiveContext } from "@times-components/responsive";
-import { editionMaxWidth, spacing } from "@times-components/styleguide";
+import {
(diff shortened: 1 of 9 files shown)


packages/edition-slices/src/slices/shared/gutter.js
```diff
@@ -2,22 +2,36 @@ import React from "react";
...
-      >
-        {children}
-      </View>
-    )}
+    {({ isTablet, editionBreakpoint }) => {
+      const tabletPaddingHorizontalResolver = {
+        [editionBreakpoints.medium]: spacing(2),
+        [editionBreakpoints.wide]: spacing(6),
+        [editionBreakpoints.huge]: spacing(2)
+      };
```
ValGeorgiev: this doesn't need to be in the gutter component

infreezer: Please review again.

PR #2194, opened 2019-08-20
Feat/extract graphql files

Plamen5kov:
EDIT: ticket: https://nidigitalsolutions.jira.com/browse/REPLAT-7371

The purpose of this PR is to extract all `gql` queries to  `.graphql` files. The extracted files can be then packed inside the `xnative` library and reused as is from the android application.

In order for that to happen:
 * had to add specific loaders to enable the `.graphql` files to be loaded into the `js` files on webpack and jest
 * android-app package had to be updated to find all `.graphql` files from the `provider-queries` folder and pack them when building before publishing
 * updated readme where necessary

Tested:
 * android will be able to load the `graphql` files  and use them to make queries (tested and will be done in a later PR)
 * no changed behaviour in renderer with `react` feature flag enabled


parkjh:
@Plamen5kov please add the corresponding ticket to the description?

parkjh:
> * android-app package had to be updated to find all `.graphql` files from the `provider-queries` folder and pack them when building before publishing
> * updated readme where necessary
> 
> Tested:
> 
> * android will be able to load the `graphql` files  and use them to make queries (tested and will be done in a later PR)
> * no changed behaviour in renderer with `react` feature flag enabled

what about iOS? did you test the latest changes to make sure it won't break?

Plamen5kov:
> > * android-app package had to be updated to find all `.graphql` files from the `provider-queries` folder and pack them when building before publishing
> > * updated readme where necessary
> > 
> > Tested:
> > 
> > * android will be able to load the `graphql` files  and use them to make queries (tested and will be done in a later PR)
> > * no changed behaviour in renderer with `react` feature flag enabled
> 
> what about iOS? did you test the latest changes to make sure it won't break?

The made change hasn't affected neither iOS nor Android yet. Currently the `graphql` files are extracted but they're not used by the mobile applications yet.
My comment about testing it on Android is true, but it doesn't matter until the Android app starts using the `graphql` files.
After this PR is merged, I'll work on integrating the extracted files in Android and iOS in separate PRs. The current workflow stays the same. Developers can still copy/paste the `graphql` queries in the native apps until I finish the integration, but in my rough estimation ... integrations can be done in a couple of days in Android after this PR is merged and for iOS, i've already talked to some iOS guys that will lend me a hand, and we've discussed how it will be done.

PR #2197, opened 2019-08-20
Fix/offline images

jmars:
Offline images were not loading, now they do.


packages/image/src/image.js
```diff
@@ -105,7 +117,7 @@ class TimesImage extends Component {
     const radius = width ? width / 2 : 9999;
     const borderRadius = rounded ? radius : 0;
     const placeholder = disablePlaceholder ? null : (
-      <Placeholder borderRadius={borderRadius} />
+      <Placeholder borderRadius={borderRadius} onLayout={this.getLowRes} />
```
nathggns: How does adding an onLayout here affect performance? We worked hard to remove as many as possible from the image package (and specifically the placeholder) because we had serious issues with performance and time-to-display-placeholder (leaving lots of unsightly gaps).

jmars: It's not ideal, I'm open to a better alternative if you have any ideas? I also tried adding a default for lowResSize however that caused placeholders to never show.

nathggns:
Spoke to Jaye about this. Going to go for a different approach.

packages/image/src/utils/index.js
```diff
@@ -1,26 +1,18 @@
...
-    ) {
-      const separator = finalUri.includes("?") ? "&" : "?";
-      finalUri = `${finalUri}${separator}${item.name}=${item.value}`;
-    }
-  });
-
-  return finalUri;
+  const uri = url.parse(uriString, true);
+  uri.search = undefined;
+  uri.query = {
```
nathggns:
```js
delete uri.search;
Object.assign(uri.query, paramMap);
```

might be cleaner, but I don't really mind either way

packages/image/src/lazy-loading-image/lazy-loading-image.js
```diff
@@ -12,27 +12,27 @@ const LazyLoadingImage = props => {
...
+    offline: true,
+    rel_height: relativeHeight,
+    rel_width: relativeWidth,
+    rel_horizontal_offset: relativeHorizontalOffset,
+    rel_vertical_offset: relativeVerticalOffset
+  };
+
+  Object.keys(queryObject).forEach(k => {
+    if (!queryObject[k]) {
+      delete queryObject[k];
```
nathggns: Quick note: this will remove any params which are zero. I'm sure you know this, just want to make sure it is expected behaviour.

packages/image/src/utils/index.js
```diff
@@ -1,26 +1,18 @@
+import url from "url";
```
nathggns: Is this a new dependency, I don't see it added anywhere.

jmars: It's not new, it's in the times-components package.json, weirdly adding it to the image package.json causes lint to say its unused. Webpack is polyfilling it maybe?

packages/image/src/utils/index.js
jmars: I like your way better too here, changed

packages/image/src/lazy-loading-image/lazy-loading-image.js
jmars: Actually that was an oversight, changed it to a undefined check

PR #2198, opened 2019-08-20
feat: full width interactives

nathggns:
[JIRA](https://nidigitalsolutions.jira.com/browse/REPLAT-7225)

image

image

I believe the alignment issues on native are an issue with the interactive (as its inside a webview)


ValGeorgiev:
On mobile the button is center align, but the text and graphic are not? Is this intentionally?

PR #2208, opened 2019-08-21
feat: secondary1andColumnist for 1024px

apostolnikov:
Jira: https://nidigitalsolutions.jira.com/browse/REPLAT-7518

image


apostolnikov:
> image
> 
> According to the [design](https://app.zeplin.io/project/5d2849a2b4a4605645afb4be/screen/5d5189622dba3f9bc1a8ef52)
> the ratio of each tile should be 1/3.

@infreezer  In our situation we are dealing with 2 tiles. Therefore they can't use ratio of 1/3. Otherwise, good catch, I'll bump up the width of the right column.

PR #2211, opened 2019-08-21
fix: update styles for fullwidth image

nehasri89:
Jira https://nidigitalsolutions.jira.com/browse/REPLAT-7426

Screenshot 2019-08-21 at 17 34 52


nathggns:
I don't think this is correct. The text caption isn't supposed to be able to go 100% wide as far as I know. I think you need to figure out why alignSelf: "center" didn't work instead.

nathggns:
As discussed on Slack, this is not the correct design. Closing to prevent it being merged accidentally.

Reference: https://app.zeplin.io/project/5c2df4097501c0242a8ddd92/screen/5c2df943173f9020588bff71

PR #2216, opened 2019-08-22
Feature/Fix Indepth FullWidth Image

nehasri89:
**Desktop**
Screenshot 2019-08-22 at 17 29 30

**Mobile**
Screenshot 2019-08-22 at 17 29 48

**Tablet**
Screenshot 2019-08-22 at 17 29 40

### DIFF
--- a/packages/article-image/src/article-image.base.js
+++ b/packages/article-image/src/article-image.base.js
@@ -6,6 +6,7 @@ import { ModalImage } from "@times-components/image";
 import InsetCaption from "./inset-caption";
 import InsetCenteredCaption from "./inset-centered-caption";
 import InlineImage from "./inline-image";
+import FullWidthCaption from "./fullwidth-caption";
 import { propTypes, defaultPropTypes } from "./article-image-prop-types";
 import styles from "./styles";
 
@@ -24,7 +25,11 @@ function getCaptionComponent(display, { imageCaptionAlignment = {} }) {
       : InsetCaption;
   }
 
-  if (display === "fullwidth" || imageCaptionAlignment[display] === "center") {
+  if (display === "fullwidth") {
+    return FullWidthCaption;
+  }
+
+  if (imageCaptionAlignment[display] === "center") {
     return CentredCaption;
   }
 
@@ -34,7 +39,17 @@ function getCaptionComponent(display, { imageCaptionAlignment = {} }) {
 const renderCaption = (caption, credits, display, theme) => {
   const CaptionComponent = getCaptionComponent(display, theme);
 
-  return caption || credits ? (
+  if (!caption && !credits) {
+    return null;
+  }
+
+  return display === "fullwidth" ? (
+    <CaptionComponent
+      credits={credits}
+      style={captionStyle[display]}
+      text={caption}
+    />
+  ) : (
     <View style={styles[`${display}Caption`]}>
       <CaptionComponent
         credits={credits}
@@ -42,7 +57,7 @@ const renderCaption = (caption, credits, display, theme) => {
         text={caption}
       />
     </View>
-  ) : null;
+  );
 };
 
 const ArticleImage = ({ imageOptions, captionOptions, onImagePress }) => {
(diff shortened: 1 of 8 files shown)


packages/article-image/src/styles/responsive.web.js
```diff
@@ -57,3 +57,20 @@ export const InsetCaptionStyle = styled(View)`
     padding-left: 0px;
   }
 `;
+
+export const FullWidthCaptionStyle = styled(View)`
```
nathggns: Can you make this a div please? It doesn't need to be a View and it'll introduce another incompatibility that we'll have to fix later. Using web styled components with RN View's is deprecated.

packages/article-image/src/fullwidth-caption.web.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import Caption from "@times-components/caption";
+import { propTypes, defaultProps } from "./fullwidth-caption-prop-types";
+import { FullWidthCaptionStyle } from "./styles/responsive";
+
+const FullWidthCaptionWeb = ({ text, credits }) => (
```
nathggns: You don't consume any of these props in this component so should just pass them through, which will make maintanence easier. (If Caption requires any further props in the future you won't need to update this file) 

```js
function FullWidthCaptionWeb(props) {
    return (
        <FullWidthCaptionStyle>
            <Caption {...props} />
        </FullWidthCaptionStyle>
    );
}
```

packages/article-image/src/fullwidth-caption.web.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import Caption from "@times-components/caption";
+import { propTypes, defaultProps } from "./fullwidth-caption-prop-types";
+import { FullWidthCaptionStyle } from "./styles/responsive";
```
nathggns: Should this be called `FullWidthCaptionContainer` or something like that instead?

packages/article-image/src/article-image.base.js
```diff
@@ -34,15 +39,25 @@ function getCaptionComponent(display, { imageCaptionAlignment = {} }) {
 const renderCaption = (caption, credits, display, theme) => {
   const CaptionComponent = getCaptionComponent(display, theme);
 
-  return caption || credits ? (
+  if (!caption && !credits) {
+    return null;
+  }
+
+  return display === "fullwidth" ? (
+    <CaptionComponent
```
nathggns: Can you extract this to a variable so its not duplicated? Will make maintaence easier.

packages/article-image/src/fullwidth-caption.web.js
```diff
@@ -0,0 +1,15 @@
...
+
+const FullWidthCaptionWeb = ({ text, credits }) => (
+  <FullWidthCaptionStyle>
```
nathggns: Why is it necessary for web / native to have completely different styles for this component?

packages/article-image/src/styles/responsive.web.js
nathggns: Might need to specify display flex manually.

packages/article-image/src/fullwidth-caption.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
```
nathggns: It's unfortunate that this file is duplicated between web / native and is essentially the same. There must be a better way to change the specified styles (if that is even necessary) rather than duplicating the entire file. Maybe using `Platform.select`.

packages/article-image/src/fullwidth-caption.web.js
nehasri89: Because we need to support different viewports for web which we do through responsive styles whereas its not the case for native

nathggns: I know. That's also the case in other places. We share the base styles and bolt on the responsive behaviour rather than having entirely separate styles.

PR #2221, opened 2019-08-23
chore: upgrade react 16.9.0

parkjh:
- see https://reactjs.org/blog/2019/08/08/react-v16.9.0.html#new-features
- next step https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#migrating-from-legacy-lifecycles


packages/storybook/src/styles.js
```diff
@@ -15,9 +15,11 @@ export function createScopedGlobalStyle(strings, ...attrs) {
 
   const InnerStyleComponent = createGlobalStyle(copiedStrings, ...attrs);
   class GlobalStyle extends Component {
-    componentWillMount() {
+    /* eslint-disable */
+    UNSAFE_componentWillMount() {
```
ValGeorgiev: Did you try componentDidMount

packages/message-bar/src/message-bar.js
```diff
@@ -27,7 +27,8 @@ class MessageBar extends Component {
     }, delay);
   }
 
-  componentWillReceiveProps(props) {
+  /* eslint-disable */
+  UNSAFE_componentWillReceiveProps(props) {
```
ValGeorgiev: these needs to be refactored

packages/image/src/safeAreaView/react-native-safe-area.js
```diff
@@ -175,9 +175,11 @@ class SafeView extends Component {
     this._isMounted = false;
   }
 
-  componentWillReceiveProps() {
+  /* eslint-disable */
+  UNSAFE_componentWillReceiveProps() {
```
ValGeorgiev: this needs to be refactored

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -162,9 +162,11 @@ class ArticleSkeleton extends Component {
     }
   }
 
-  componentWillMount() {
+  /* eslint-disable */
+  UNSAFE_componentWillMount() {
```
ValGeorgiev: componentDidMount

parkjh:
@ValGeorgiev refactoring is not the purpose of this pr

parkjh:
> This upgrade is safe, no breaking changes from React, so we can move this pr to a regular one. But please lets refactor the deprecated lifecycles

@ValGeorgiev 
see https://reactjs.org/blog/2019/08/08/react-v16.9.0.html#renaming-unsafe-lifecycle-methods
tldr `React 16.9 does not contain breaking changes, and the old names continue to work in this release. But you will now see a warning when using any of the old names:`

ValGeorgiev:
> @ValGeorgiev refactoring is not the purpose of this pr

I am not agree with that. If we upgrade a library we should remove all deprecated code there and replace it with the new versions :) If we don't do that now, when?

ValGeorgiev:
Please don't close it, its fine. I will add label "help needed" and we all can contribute to it, when we have time :)

parkjh:
> Please don't close it, its fine. I will add label "help needed" and we all can contribute to it, when we have time :)

no it's ok I will reopen when it's ready

PR #2222, opened 2019-08-23
fix: broken flags and article summary

aronshamash:
- Fixes issue with flags and article summary on web, which we supposed to be for edition slices only.
- I have retained the original code but made it for native only as edition slices are only for native for now.

### Profile page (`article-summary` fix)
Screenshot 2019-08-23 at 17 41 43

### Article page (`article-flag` fix)
Screenshot 2019-08-23 at 17 41 34

### DIFF
--- a/packages/article-flag/src/style/index.js
+++ b/packages/article-flag/src/style/index.js
@@ -4,6 +4,7 @@ import sharedStyles from "./shared";
 const styles = {
   ...sharedStyles,
   flags: {
+    ...sharedStyles.flags,
     marginBottom: 0,
     marginTop: spacing(2)
   },
--- a/packages/article-flag/src/style/index.js
+++ b/packages/article-flag/src/style/index.js
@@ -3,6 +3,10 @@ import sharedStyles from "./shared";
 
 const styles = {
   ...sharedStyles,
+  flags: {
+    marginBottom: 0,
+    marginTop: spacing(2)
+  },
   title: {
     ...sharedStyles.title,
     fontSize: fontSizes.meta,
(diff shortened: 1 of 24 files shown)


ValGeorgiev:
Please, don't merge until we test the save star

packages/article-flag/src/style/index.js
```diff
@@ -1,8 +1,12 @@
-import { fontSizes } from "@times-components/styleguide";
+import { fontSizes, spacing } from "@times-components/styleguide";
 import sharedStyles from "./shared";
 
 const styles = {
   ...sharedStyles,
+  flags: {
+    marginBottom: 0,
```
pdimova: we need to have the shared styles destructed here `...sharedStyles.flags,`

packages/article-summary/src/styles/index.js
```diff
@@ -6,6 +6,9 @@ const styles = StyleSheet.create({
   headlineWrapper: {
     ...sharedStyles.headlineWrapper,
     lineHeight: 25
+  },
+  text: {
+    marginBottom: 0
```
pdimova: ` ...sharedStyles.text,` here as well.
Also, can you add the same `text` property to the android styles (index.android.js)?

parkjh:
> Please, don't merge until we test the save star

@ValGeorgiev @pdimova  can you guys please make the corresponding changes and see if it fixes the save star?

ValGeorgiev:
> > Please, don't merge until we test the save star
> 
> @ValGeorgiev @pdimova can you guys please make the corresponding changes and see if it fixes the save star?

This is already done, requested changes are tested by @pdimova. I want to check with my team, who is responsible for this change. We are happy to do it, if no one will. :)

PR #2190, opened 2019-08-19
chore: remove babel 6 polyfill and use core-js 2

parkjh:
- trying to remove babel polypill as they are deprecated

PR #2224, opened 2019-08-26
Revert "chore: remove babel 6 polyfill and use core-js 2"

infreezer:
Changes in [this PR](https://github.com/newsuk/times-components/pull/2190) fully reverted.
In order to fix the error below on ios-app and have a stable master.
image


ValGeorgiev:
I could not reproduce this issue, android and ios apps are working as expected. I will close it

PR #2225, opened 2019-08-26
feat: list2and6nopic for 1024px

apostolnikov:
[Story](https://nidigitalsolutions.jira.com/browse/REPLAT-7517)

image

### DIFF
--- a/packages/edition-slices/src/tiles/tile-as/styles/index.js
+++ b/packages/edition-slices/src/tiles/tile-as/styles/index.js
@@ -1,6 +1,15 @@
-import { fonts, spacing } from "@times-components/styleguide";
+import {
+  fonts,
+  spacing,
+  editionBreakpoints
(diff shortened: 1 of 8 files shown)


packages/edition-slices/src/tiles/tile-as/styles/index.js
```diff
@@ -1,15 +1,24 @@
...
+  editionBreakpoints
+} from "@times-components/styleguide";
 
-const styles = {
+const headlineFontSize = {
+  [editionBreakpoints.medium]: 18,
+  [editionBreakpoints.wide]: 20
+};
+
+export default breakpont => ({
```
ValGeorgiev: typo **breakpoint**

infreezer:
> LGTM, why there are no images for list2 in the image?

Because the GraphQL query isn't synced between TC and the apps.

PR #2227, opened 2019-08-27
feat/PPEC-23: Add optimizely script to topic head

tobenna:
Adding optimizely MVT script to topic head for POC.

### DIFF
--- a/packages/topic/src/head.web.js
+++ b/packages/topic/src/head.web.js
@@ -16,6 +16,10 @@ function Head({ description, name, slug }) {
     <Context.Consumer>
       {({ makeTopicUrl }) => (
         <Helmet>
+          {/**
+            @todo: Remove this script tag and the corresponding proxy in Akamai
+            after the PoC is done.
+          */}
           <link
             as="script"
             href="/assets/optimizely/16079460199.js"
--- a/packages/topic/src/head.web.js
+++ b/packages/topic/src/head.web.js
@@ -16,6 +16,8 @@ function Head({ description, name, slug }) {
     <Context.Consumer>
       {({ makeTopicUrl }) => (
         <Helmet>
+          <link rel="preload" href="/assets/optimizely/16079460199.js" as="script" />
+          <script src="/assets/optimizely/16079460199.js" />
           <title>{name} | The Times &amp; The Sunday Times</title>
           <meta content={content} name="description" />
           <link href={makeTopicUrl({ slug })} rel="canonical" />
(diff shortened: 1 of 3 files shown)


packages/topic/src/head.web.js
```diff
@@ -16,6 +16,12 @@ function Head({ description, name, slug }) {
     <Context.Consumer>
       {({ makeTopicUrl }) => (
         <Helmet>
+          <link
+            as="script"
+            href="/assets/optimizely/16079460199.js"
```
tobenna: this will exist as soon as Tim deploys his Akamai changes.

packages/topic/src/head.web.js
```diff
@@ -16,6 +16,16 @@ function Head({ description, name, slug }) {
     <Context.Consumer>
       {({ makeTopicUrl }) => (
         <Helmet>
+          {/**
+            @todo: Remove this script tag and the corresponding proxy in Akamai
+            after the PoC is done.
+          */}
+          <link
+            as="script"
+            href="/assets/optimizely/16079460199.js"
```
chrishutchinson: We have a Storybook for components, I have some concerns that this won't work there. I'm not sure we care, but it'd be good to confirm with one of the codeowners

tobenna: We've tested this locally using cps-render and it seems to be working fine.

chrishutchinson: Was it tested when running the storybook?

tobenna: we also tested the local version of storybook and saw the script show up in the head of the `topic` page

packages/topic/src/head.web.js
```diff
@@ -16,6 +16,16 @@ function Head({ description, name, slug }) {
...
+            rel="preload"
+          />
+          <script src="/assets/optimizely/16079460199.js" />
```
ValGeorgiev: could we add defer?

ValGeorgiev: Basically we add scripts in the end of the body, not in the head

tbradbury: It's recommend to get best results and to stop flicker (as content changes) to add the script as hight as possible in the head. As part of the POC we want to see how this effects performance and if necessary switch and test with async. I don't believe it would work correctly with defer.

nathggns: With helmet, the script will be added in the head, not in the body.

tobenna: We want the script in the head

PR #2229, opened 2019-08-27
feat: (REPLAT-7502) redesign lead 2no pic and 2 slice for wide breakpoint

pdimova:
ticket: https://nidigitalsolutions.jira.com/browse/REPLAT-7502

- small updates for medium breakpoint
- tileZ removed since not used for huge breakpoint or in any other slice

Screenshot_1566906012


packages/edition-slices/src/tiles/tile-x/styles/index.js
```diff
@@ -19,28 +19,14 @@ const mediumBreakpointStyles = {
...
-    lineHeight: 40
-  },
-  strapline: {
-    ...fontFactory({
-      font: "bodyRegular",
-      fontSize: "pageComponentHeadline"
-    }),
-    paddingBottom: spacing(1)
-  }
+  ...mediumBreakpointStyles
```
ValGeorgiev: in this case we don't need breakpoint and we can always return wideBreakpointStyles (which can be renamed as styles)

packages/edition-slices/src/tiles/tile-y/styles/index.js
```diff
@@ -19,17 +19,7 @@ const mediumBreakpointStyles = {
...
-  },
-  headline: {
-    fontFamily: fonts.headline,
-    fontSize: 30,
-    lineHeight: 30
-  },
-  summaryContainer: {
-    marginHorizontal: spacing(2)
-  }
+  ...mediumBreakpointStyles
```
ValGeorgiev: same here

packages/edition-slices/src/tiles/tile-x/styles/index.js
pdimova: indeed, but prefer to delete the wideBreakpointStyles logic once we know the styles for huge will be the same as the ones for medium. Don't want to have the case where this code will be removed today and brought back tomorrow. What do you think?

ValGeorgiev: This is not good, if we don’t deleted bow, probably we will forgot it. I will approve it only because we have demo and it will go to different than master branch. We can fix it when we are doing a pr to master

PR #2244, opened 2019-08-30
fix: fixing alignment on indepth primary image

nehasri89:

### DIFF
--- a/packages/styleguide/src/theme/theme-factory.js
+++ b/packages/styleguide/src/theme/theme-factory.js
@@ -57,9 +57,7 @@ const headlineCasePicker = (section = "default", template = "mainstandard") =>
     ? "uppercase"
     : null;
 
-const imageCaptionAlignment = {
-  indepth: { primary: "center" }
-};
+const imageCaptionAlignment = {};
 
 export default (section, template) => ({
   dropCapFont: magazineFontPicker("dropCap", section, template),


nathggns:
Reviewing now…

packages/styleguide/src/theme/theme-factory.js
```diff
@@ -57,9 +57,7 @@ const headlineCasePicker = (section = "default", template = "mainstandard") =>
     ? "uppercase"
     : null;
 
-const imageCaptionAlignment = {
-  indepth: { primary: "center" }
```
nathggns: If you're removing the only scenario where the themed caption alignment is used, it may be worth removing it altogether.

nathggns:
Isn't this PR a "fix", not a "chore", btw?

nathggns:
It's odd – this was definitely an intentional change to make primary images centre aligned, wasn't a bug. I'm trying to find the ticket that led to this change being made. https://github.com/newsuk/times-components/pull/1740

packages/styleguide/src/theme/theme-factory.js
nathggns: As in, this should be removed from the theme system altogether (and out of the components that read from it)

nehasri89:
> It's odd – this was definitely an intentional change to make primary images captions centre aligned on in depth, wasn't a bug. I'm trying to find the ticket that led to this change being made. #1740

Right, but according to the theming sheet https://docs.google.com/spreadsheets/d/1CDedFyfaI0YreLUgyuKHu05_zZGWI91A6M8lcKRWJUk/edit#gid=0 inly fullwidth images should have center align caption, you think this sheet is not updated?

packages/styleguide/src/theme/theme-factory.js
aronshamash: confirmation it should be left aligned so `imageCaptionAlignment` should be removed the theme factory.

nehasri89: I need to check how the native behaves for this too, the theming sheet is web specific

nehasri89: @aronshamash the zeplin https://app.zeplin.io/project/5c2df4097501c0242a8ddd92/screen/5c2df97562b80caee662764b for native does not have a `primary` image example, does that mean native `primary image caption` is left aligned too?

PR #2246, opened 2019-08-30
fix: fix a crash where a dropcap-enabled article has a newline as the…

jmars:
… first element

Some articles have weird markup like a newline as the first element and dropcaps enabled, we should just fail back to rendering without dropcaps (This is what web and native do)  in these cases instead of trying to make every possible markup case work.

### DIFF
--- a/packages/article-skeleton/__tests__/common/dropcap-util.test.js
+++ b/packages/article-skeleton/__tests__/common/dropcap-util.test.js
@@ -33,45 +33,6 @@ const invalidChild = {
   name: "paragraph"
 };
 
-const invalidChildOut = [
-  {
-    attributes: {},
-    children: [
-      {
-        attributes: {},
-        children: [
-          { attributes: { dropCap: true }, children: [], name: "break" },
-          {
-            attributes: {
-              value:
-                "As I follow Chris Reynolds Gordon down the first f…fty Shades of Grey but, he says, “I’m living it.”"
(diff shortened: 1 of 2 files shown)


packages/article-skeleton/__tests__/common/dropcap-util.test.js
```diff
@@ -104,7 +122,7 @@ const childWithDropCapAndMarkup = [
     ]
   },
   {
-    attributes: [],
```
parkjh: how did it work with attributes being an array before? or was this just not used in the test at all?

packages/article-skeleton/src/dropcap-util-common.js
```diff
@@ -73,32 +73,36 @@ const findFirstTextNode = children => {
...
+        {
+          name: "dropCap",
+          attributes: {},
+          children: [withCap]
+        },
+        withoutCap,
+        ...children.slice(1)
+      ];
+    }
+  } catch (error) {
```
parkjh: what exact error case are you catching here? and should it not fail genuinely in some case?

packages/article-skeleton/__tests__/common/dropcap-util.test.js
jmars: It compares input vs expected output, both had the wrong type so it still passed

PR #2248, opened 2019-09-02
fix: deps

nathggns:


packages/save-star-web/__tests__/web/__snapshots__/save-star-web.test.js.snap
```diff
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+g// Jest Snapshot v1, https://goo.gl/fbAQLP
```
ValGeorgiev: this looks like a typo

nathggns: Wow. Bizarre how that happened. Good catch.

PR #2250, opened 2019-09-02
fix: show correct saved status when user has more than 20 saved articles

nathggns:
[JIRA](https://nidigitalsolutions.jira.com/browse/REPLAT-8223)

Currently, in order to determine the saved status of an article for a user, we request a list of the user's bookmarked articles and check whether the article is contained within it. However, it is not possible (or desirable) to get the full list of a user's bookmarked articles consistently, as the list is paged and the query must provide a page size (which defaults to 20).

This means that under the current approach, there will always be a maximum number of saved articles a user can have before they start to experience bugs. There are also undesirable side effects on increasing the page size, as it is essentially wasted network traffic for the client to compute this, and it requires both the server and the client to load into memory increasingly large lists. This is unnecessary.

Instead, I'm using the `isBookmarked` field on articles. This continues to be lazy loaded as with the previous approach. It is unclear why this was not used in the first place, but my understanding is that this field was not available when we first implemented saved status for articles.  

I'm interested to know if there were any other reasons for why `isBookmarked` was not used.

### DIFF
--- a/packages/provider-queries/src/article-bookmarked.graphql
+++ b/packages/provider-queries/src/article-bookmarked.graphql
@@ -1,0 +1,7 @@
+query ArticleBookmarkedQuery($id: ID!) {
+  article(id: $id) {
+    __typename
+    id
+    isBookmarked
+  }
+}
--- a/packages/provider-queries/src/article-bookmarked.js
+++ b/packages/provider-queries/src/article-bookmarked.js
@@ -1,0 +1,4 @@
+import { addTypenameToDocument } from "apollo-utilities";
+import articleBookmarkedFragment from "./article-bookmarked.graphql";
+
+export default addTypenameToDocument(articleBookmarkedFragment);
--- a/packages/provider-queries/src/provider-queries.js
+++ b/packages/provider-queries/src/provider-queries.js
@@ -1,32 +1,24 @@
-import articleQuery from "./article";
-import articleExtrasQuery from "./article-extras";
-import getTokenisedArticleUrlQuery from "./get-tokenised-article-url";
-import * as authorArticlesNoImagesQuery from "./author-articles-no-images";
-import * as authorArticlesWithImagesQuery from "./author-articles-with-images";
-import authorQuery from "./author";
-import getBookmarksQuery from "./get-bookmarks";
-import saveBookmarksQuery from "./save-bookmark";
-import unsaveBookmarksQuery from "./unsave-bookmark";
-import editionQuery from "./edition";
-import nativeEditionQuery from "./native-edition";
-import topicQuery from "./topic";
-import * as topicArticlesQuery from "./topic-articles";
+export article from "./article";
+export articleBookmarked from "./article-bookmarked";
+export articleExtras from "./article-extras";
+export author from "./author";
+export edition from "./edition";
+export getBookmarks from "./get-bookmarks";
+export getTokenisedArticleUrl from "./get-tokenised-article-url";
+export nativeEdition from "./native-edition";
+export saveBookmarks from "./save-bookmark";
+export topic from "./topic";
+export unsaveBookmarks from "./unsave-bookmark";
 
-export const article = articleQuery;
-export const articleExtras = articleExtrasQuery;
-export const getTokenisedArticleUrl = getTokenisedArticleUrlQuery;
-export const authorArticlesNoImages = authorArticlesNoImagesQuery.default;
-export const authorArticlesNoImagesPTV =
-  authorArticlesNoImagesQuery.propsToVariables;
-export const authorArticlesWithImages = authorArticlesWithImagesQuery.default;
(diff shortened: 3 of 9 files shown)


parkjh:
> Currently, in order to determine the saved status of an article for a user, we request a list of the user's bookmarked articles and check whether the article is contained within it. However, it is not possible (or desirable) to get the full list of a user's bookmarked articles consistently, as the list is paged and the query must provide a page size (which defaults to 20).

@nathggns can you not just increase the offset paramemter in your query to fetch the next batch of bookmarked article until you recieve all the bookmarked articles?

chrishutchinson:
> > Currently, in order to determine the saved status of an article for a user, we request a list of the user's bookmarked articles and check whether the article is contained within it. However, it is not possible (or desirable) to get the full list of a user's bookmarked articles consistently, as the list is paged and the query must provide a page size (which defaults to 20).
> 
> @nathggns can you not just increase the offset paramemter in your query to fetch the next batch of bookmarked article until you recieve all the bookmarked articles?

If I've saved many hundreds of articles, that becomes a _very_ slow query chain, especially if the one I'm looking at is the last one in the list. The `isBookmarked` field is surely the most sensible (and correct) option?

nathggns:
> > Currently, in order to determine the saved status of an article for a user, we request a list of the user's bookmarked articles and check whether the article is contained within it. However, it is not possible (or desirable) to get the full list of a user's bookmarked articles consistently, as the list is paged and the query must provide a page size (which defaults to 20).
> 
> @nathggns can you not just increase the offset paramemter in your query to fetch the next batch of bookmarked article until you recieve all the bookmarked articles?

Yes that's possible. I explained why I chose not to do that.

> There are also undesirable side effects on increasing the page size, as it is essentially wasted network traffic for the client to compute this, and it requires both the server and the client to load into memory increasingly large lists

Also, we have a field that is clearly the most efficient way to do this now. Why not use it?

parkjh:
> > > Currently, in order to determine the saved status of an article for a user, we request a list of the user's bookmarked articles and check whether the article is contained within it. However, it is not possible (or desirable) to get the full list of a user's bookmarked articles consistently, as the list is paged and the query must provide a page size (which defaults to 20).
> > 
> > 
> > @nathggns can you not just increase the offset paramemter in your query to fetch the next batch of bookmarked article until you recieve all the bookmarked articles?
> 
> If I've saved many hundreds of articles, that becomes a _very_ slow query chain, especially if the one I'm looking at is the last one in the list. The `isBookmarked` field is surely the most sensible (and correct) option?

is there a jira ticket for this? because I feel like if this is a production issue then 

> > > Currently, in order to determine the saved status of an article for a user, we request a list of the user's bookmarked articles and check whether the article is contained within it. However, it is not possible (or desirable) to get the full list of a user's bookmarked articles consistently, as the list is paged and the query must provide a page size (which defaults to 20).
> > 
> > 
> > @nathggns can you not just increase the offset paramemter in your query to fetch the next batch of bookmarked article until you recieve all the bookmarked articles?
> 
> Yes that's possible. I explained why I chose not to do that.
> 
> > There are also undesirable side effects on increasing the page size, as it is essentially wasted network traffic for the client to compute this, and it requires both the server and the client to load into memory increasingly large lists
> 
> Also, we have a field that is clearly the most efficient way to do this now. Why not use it?

it's not more efficient though...

parkjh:
@nathggns what's the ticket? what exactly are you trying to adddress here? if the issue is that the user have more than 20 bookmarked article and we are not querying past the default 20 then we should address it by incrementing the offset param. paginated queries exist for a reason...

nathggns:
Apologies – the ticket should have been included in the PR description. My mistake. Fixed that.

nathggns:
@parkjh it  is true that it is wasted network traffic to load the whole list of bookmarked articles just to find out if the current article is bookmarked. Sure, it's possible to implement it so that neither the server or client has the *entire* list in memory simultaneously (also very easy to naively implement it so it that they do load the whole thing), but it's still a waste of network and CPU time to continously download the list on the client when the server can just tell us…

chrishutchinson:
> @nathggns what's the ticket? what exactly are you trying to adddress here? if the issue is that the user have more than 20 bookmarked article and we are not querying past the default 20 then we should address it by incrementing the offset param. paginated queries exist for a reason...

In a GraphQL API, a field likes `article.isBookmarked` exist exactly to answer the question "is this article bookmarked or not?". Expecting a client to page through a potentially infinitely long list, just to find a single record, is a complete anti-pattern.

nathggns:
Discussed offline. I believe @parkjh now agrees with this approach.

parkjh:
apologies - i was missing context from the ticket, one thing to note though, were there no functional tests for save-api component?

nathggns:
@parkjh there are tests of the save API component. They didn't need to be updated because they used the mocks in the provider-test-tools package, which has been updated in this PR.

PR #2251, opened 2019-09-02
feat: add missing tracking attribute

nehasri89:
PR adds tracking attr https://nidigitalsolutions.jira.com/browse/REPLAT-7399


yarn.lock
```diff
@@ -3257,6 +3257,76 @@
     fs-extra "6.0.0"
     glob "7.1.2"
 
+"@times-components/context@1.0.39":
```
todorRadkov: Do we need to update yarn.lock too?

nathggns:
Did you consider putting navigationMode in the context?

PR #2253, opened 2019-09-03
feat: REPLAT-7557 add sponsored-article div for all user states of...

todorRadkov:
article page

Adding nativo sponsored ads div in all article pages.
The div with id "sponsored-article" is placed on corresponding place to the one in render.

JIRA:
https://nidigitalsolutions.jira.com/browse/REPLAT-7557

### DIFF
--- a/packages/article-extras/src/article-extras.web.js
+++ b/packages/article-extras/src/article-extras.web.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from "react";
+import React from "react";
 import PropTypes from "prop-types";
 import UserState from "@times-components/user-state";
 import ArticleComments from "@times-components/article-comments";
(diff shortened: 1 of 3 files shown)


packages/article-extras/src/article-extras.web.js
```diff
@@ -22,43 +22,49 @@ const ArticleExtras = ({
...
     {/* Nativo inserts Sponsored Articles in this div */}
     <div id="sponsored-article" />
-    <ArticleComments
-      articleId={articleId}
-      isEnabled={commentsEnabled}
-      spotAccountId={spotAccountId}
-    />
-  </UserState>
+
+    <UserState state={UserState.fullArticle}>
```
parkjh: what is this change for?

nathggns:
Reviewing now…

packages/article-extras/src/article-extras.web.js
```diff
@@ -22,43 +22,49 @@ const ArticleExtras = ({
   spotAccountId,
   topics
 }) => (
-  <UserState state={UserState.fullArticle}>
```
nathggns: The `UserState` component should not be considered "free" to use. Each additional use has an increasing amount of performance concerns and its use should be limited. Additionally, `UserState` has a `fallback` prop that can be used to avoid needing to use an extra `UserState` component in this case. 

Instead of the following, which you are essentially doing:

```js
function MyComponent() {
    return (
        <Fragment>
            <UserState state={UserState.fullArticle}>
                {/*...*/}
            </UserState>
            <div id="sponsored-article" />
            <UserState state={UserState.fullArticle}>
                {/*...*/}
            </UserState>
        </Fragment>
    );
}
```

you can do the following – the key should help React know not to destroy the DOM node when transitioning between user states (on the second render of the two pass rendering). *You should verify that this is the case, or we may have a race condition between when the second render is, and when Nativo inserts the ad*

```js
function MyComponent() {
    const sponsoredArticle = <div id="sponsored-article" key="sponsored-article" />;

    return (
        <UserState
            state={UserState.fullArticle}
            fallback={sponsoredArticle}
        >
            {/*...*/}
            {sponsoredArticle}
            {/*...*/}
        </UserState>
    );
}
```

todorRadkov: This change is applied.

packages/article-extras/src/article-extras.web.js
todorRadkov: The idea was the div to be visible for all user states, but @nathggns suggested better implementation, which is now applied.

PR #2255, opened 2019-09-03
fix: REPLAT-7354 unruly expect an object always

jmars:
A small change, unruly have stated they expect an object with the same properties as when in-viewport for the out-of-viewport event.

### DIFF
--- a/packages/ad/src/dom-context.android.js
+++ b/packages/ad/src/dom-context.android.js
@@ -93,7 +93,10 @@ class DOMContext extends PureComponent {
     this.isVisible = false;
     this.webView.injectJavaScript(`
         if (typeof unrulyViewportStatus === "function") {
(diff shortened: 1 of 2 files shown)


parkjh:
> A small change, unruly have stated they expect an object with the same properties as when in-viewport for the out-of-viewport event.

can you please link it? and does `unrulyViewportStatus` even exist before mounting?

jmars:
> > A small change, unruly have stated they expect an object with the same properties as when in-viewport for the out-of-viewport event.
> 
> can you please link it? and does `unrulyViewportStatus` even exist before mounting?

Even worse, that function is a part of their own script that runs nested in a few iframes after initialisation. The requirement was conveyed by Loki in the tnl-unruly channel after he investigated issues on their side.

PR #2258, opened 2019-09-03
chore: upgrade react

parkjh:
- upgrade react to v16.9.0


yarn.lock
```diff
@@ -17367,7 +17367,17 @@ react-docgen@^3.0.0:
...
-react-dom@16.8.6, react-dom@^16.7.0:
+react-dom@16.9.0:
+  version "16.9.0"
+  resolved "https://registry.yarnpkg.com/react-dom/-/react-dom-16.9.0.tgz#5e65527a5e26f22ae3701131bcccaee9fb0d3962"
+  integrity sha512-YFT2rxO9hM70ewk9jq0y6sQk8cL02xm4+IzYBz75CQGlClQQ1Bxq0nhHF6OtSbit+AIahujJgb/CPRibFkMNJQ==
+  dependencies:
+    loose-envify "^1.1.0"
+    object-assign "^4.1.1"
+    prop-types "^15.6.2"
+    scheduler "^0.15.0"
```
ValGeorgiev: this is good, scheduler is coming <3

packages/image/src/safeAreaView/react-native-safe-area.js
```diff
@@ -154,6 +154,10 @@ class SafeView extends Component {
     _customStatusBarHeight = height;
   };
 
+  static getDerivedStateFromProps() {
+    this._onLayout();
```
nathggns: This isn't going to work right? `_onLayout` is an instance method but `getDerivedStateFromProps` is a class method.

Besides, this is third party code anyway and it has been updated to remove componentWillReceiveProps. Copy it from here https://github.com/react-native-community/react-native-safe-area-view/blob/legacy/index.js

nathggns: This is used in modal images if you want to check it works, btw.

parkjh:
> It's unclear why all the snaps now have extra views in too?

yea me neither... checked with jaye and it's to do with text flow but it is unclear even to him what that view component represent...

packages/image/src/safeAreaView/react-native-safe-area.js
```diff
@@ -1,50 +1,14 @@
-// taken from https://github.com/react-native-community/react-native-safe-area-view
```
nathggns: Can we keep this comment in please.

packages/image/src/safeAreaView/react-native-safe-area.js
```diff
@@ -1,50 +1,14 @@
...
-    };
-
-    render() {
-      return <WrappedComponent {...this.props} {...this.state} />;
-    }
-  }
-
-  return hoistStatics(withOrientationClass, WrappedComponent);
-}
+import withOrientation from "./withOrientation";
```
nathggns: This is probably going to break. You might want to inline this file as previously.

parkjh: copied over

nathggns:
Why did you assign this to me?

parkjh:
> Why did you assign this to me?

to rereview?

nathggns:
@parkjh you need to understand what is causing the extra views before this can be approved.

parkjh:
> @parkjh you need to understand what is causing the extra views before this can be approved.

@jmars ^^^

jmars:
It's the height:0 that makes me suspicious it's related to text, but why would a reat upgrade cause this? Seems strange, looking into it now

packages/article-skeleton/__tests__/android/__snapshots__/article-with-style.android.test.js.snap
```diff
@@ -669,32 +669,7 @@ exports[`2. an article with interactives 1`] = `
 </View>
 `;
 
-exports[`3. an article with no content 1`] = `
```
parkjh: @jmars I know that this is intentional but is this the correct behaviour?

PR #2259, opened 2019-09-04
fix: vertically center text in secondary2nopicand2

apostolnikov:
[Story](https://nidigitalsolutions.jira.com/browse/REPLAT-7420)

Rules for the teaser and the star can be found in the story's comments section.

**Android(with start):** 

image
image
image
image

**=====================================================================**

**IOS (without star):** 
image
image
image


packages/edition-slices/src/tiles/shared/positioned-tile-star.js
```diff
@@ -15,13 +15,15 @@ import {
 const PositionedTileStar = ({
   centeredStar = false,
   underneathTextStar = false,
+  customStyles = {},
```
nathggns: I think this prop should just be called `style`.

packages/edition-slices/src/tiles/tile-g/styles/index.js
```diff
@@ -24,11 +24,21 @@ const defaultStyles = {
...
     flex: 1
+  },
+  innerContainer: {
+    flexDirection: "row",
+    flex: 1
+  },
+  customStarPosition: {
+    position: "relative",
+    bottom: 10,
+    marginBottom: -15
```
ValGeorgiev: use spacing here, please

packages/edition-slices/src/tiles/shared/positioned-tile-star.js
apostolnikov: I think customStyles is fine for this situation, it's a bit more descriptive and it can be easily distinguished from the native style prop.

PR #2260, opened 2019-09-04
fix: output correct metadata when lead asset is a video

nathggns:
We need to ensure image metadata is correct when the lead asset is a video. This PR ensures that.

### DIFF
--- a/packages/article-skeleton/src/head.web.js
+++ b/packages/article-skeleton/src/head.web.js
@@ -120,7 +120,11 @@ function Head({ article, paidContentClassName, faviconUrl }) {
       : null;
   const sectionname = getSectionName(article);
   const leadassetUrl = appendToUrl(
-    get(leadAsset, "crop169.url", null),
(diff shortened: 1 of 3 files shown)


packages/article-skeleton/src/head.web.js
```diff
@@ -120,7 +120,11 @@ function Head({ article, paidContentClassName, faviconUrl }) {
       : null;
   const sectionname = getSectionName(article);
   const leadassetUrl = appendToUrl(
-    get(leadAsset, "crop169.url", null),
+    get(
+      hasVideo ? get(leadAsset, "posterImage", null) : leadAsset,
```
nehasri89: `posterimage` is an object of `crops`, so you can rename this method and use `getThumbnailUrlFromVideo` which is already doing similar stuff

nathggns: 👍

PR #2264, opened 2019-09-05
chore: ResponsiveSlice render functions passed in order to cleanup test warnings

infreezer:
The change should be chore.
In order to remove the warning below - render functions for Wide and Huge in Responsive Slice changed to NOT required. And defaultProps added.
The change will improve the readability of the code with removing a duplications.

Before:
edition-slices-responsive-functions-warning

### DIFF
--- a/packages/edition-slices/src/slices/leadoneandfour/index.js
+++ b/packages/edition-slices/src/slices/leadoneandfour/index.js
@@ -75,8 +75,6 @@ class LeadOneAndFour extends Component {
       <ResponsiveSlice
         renderSmall={this.renderSmall}
         renderMedium={this.renderMedium}
-        renderWide={this.renderMedium}
-        renderHuge={this.renderMedium}
       />
     );
   }
--- a/packages/edition-slices/src/slices/leadoneandfour/index.js
+++ b/packages/edition-slices/src/slices/leadoneandfour/index.js
@@ -73,8 +73,10 @@ class LeadOneAndFour extends Component {
   render() {
     return (
       <ResponsiveSlice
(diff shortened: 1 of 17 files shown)


packages/edition-slices/src/slices/puzzle/index.js
```diff
@@ -50,8 +50,10 @@ class Puzzle extends Component {
   render() {
     return (
       <ResponsiveSlice
-        renderMedium={this.renderPuzzles}
         renderSmall={this.renderPuzzles}
+        renderMedium={this.renderPuzzles}
+        renderWide={this.renderPuzzles}
+        renderHuge={this.renderPuzzles}
       />
```
apostolnikov: Could we update the responsive-slice component so the functions that are passed are not required... or at least not all of them, so we can avoid unnecessary code repetitions like those above all over the repo?

infreezer: Sure. Work in progress.

packages/edition-slices/src/slices/leadoneandfour/index.js
```diff
@@ -73,8 +73,8 @@ class LeadOneAndFour extends Component {
   render() {
     return (
       <ResponsiveSlice
-        renderMedium={this.renderMedium}
         renderSmall={this.renderSmall}
+        renderMedium={this.renderMedium}
```
infreezer: Changing the order of props. From small to bigger.

ValGeorgiev: Game changer, thanks for that <3

PR #2274, opened 2019-09-10
Feat: Replace Grapeshot with ADmantx

aronshamash:
- https://nidigitalsolutions.jira.com/browse/REPLAT-7476
- Replaces Grapeshot with ADmantx for brand protection.


packages/ad/fixtures/check-self-contained-function.js
```diff
@@ -10,7 +10,7 @@ export const getSelfContainedFunctionErrors = f => {
     })
     .filter(e => !/'cov_\w+' is not defined./.test(e.message))
     .filter(e => !/'console' is not defined./.test(e.message))
-    .filter(e => !/'_toConsumableArray' is not defined./.test(e.message))
+    .filter(e => !/'_toConsumableArray2' is not defined./.test(e.message))
```
ValGeorgiev: Is this intentional change?

aronshamash: yes

PR #2278, opened 2019-09-11
fix: REPLAT-8454 fix slow scrolling on magazine sections

jmars:
Magazine sections have one slice with 30 tiles, this means the flatlist loading on slices doesn't work and we get really bad performance.


infreezer:
Do you see the same error? 
I'm using iPad - medium breakpoint. 
It's interesting the error isn't there in wide breakpoint.
image

jmars:
Flex doesn't work inside flatlists, I think switching at 10 items is probably a reasonable compromise? Less than 10 won't have scrolling issues and it's unlikely to see a vertical layout with 10 items on tablet?

infreezer:
> Flex doesn't work inside flatlists, I think switching at 10 items is probably a reasonable compromise? Less than 10 won't have scrolling issues and it's unlikely to see a vertical layout with 10 items on tablet?

What do you think about having two components?
And use the component with the flastlist where it's needed in this case [here](https://github.com/newsuk/times-components/blob/master/packages/edition-slices/src/slices/standard/index.js).
instead of having "if" in the vertical layout.

jmars:
Makes sense, I haven't done much slice work so hadnt thought of it, will make the changes

packages/edition-slices/src/slices/standard/index.js
```diff
@@ -39,7 +39,7 @@ class Standard extends Component {
   }
 
   renderSmall(breakpoint) {
-    return <VerticalLayout tiles={this.getTiles(breakpoint)} />;
+    return <ListVerticalLayout tiles={this.getTiles(breakpoint)} />;
```
infreezer: What is the scope of the story?
Do we need to use the ListVerticalLayout for tablet also?
Please check [this line.](https://github.com/newsuk/times-components/blob/38be789d92bf9ede7040d828cd71f977801c62ba/packages/edition-slices/src/slices/standard/index.js#L36)

infreezer:
Do we need storybook integration of the new component?

packages/slice-layout/__tests__/list.base.js
```diff
@@ -0,0 +1,65 @@
...
+        expect(wrapper).toMatchSnapshot();
+      }
+    },
+    {
+      name: "two child elements",
+      test() {
+        const wrapper = renderComponent(
+          <ListVerticalLayout
+            tiles={[
+              <test tile={{ article: { id: 'testId'}}} />,
```
ValGeorgiev: do we need this <test /> component, if yes, the test name it doesn't look appropriate

jmars:
I looked at the other slices and the other uses seem to have a fixed number of items in vertical layout? That is fine, its just the huge list we get on mobile magazine sections that doesnt work well, and that is a variable number.

PR #2280, opened 2019-09-12
feat: secondary2nopicand2 for 1366 huge tablet

apostolnikov:
[Story](https://nidigitalsolutions.jira.com/browse/REPLAT-8558)

Changes are the font size for both tiles and the padding on the right one.

image

### DIFF
--- a/packages/edition-slices/__tests__/ios/__snapshots__/tiles-with-style.test.js.snap
+++ b/packages/edition-slices/__tests__/ios/__snapshots__/tiles-with-style.test.js.snap
@@ -1493,9 +1493,9 @@ exports[`28. tile ae 1`] = `
             Object {
               "color": "#333333",
               "fontFamily": "TimesModern-Bold",
-              "fontSize": undefined,
+              "fontSize": 30,
               "fontWeight": "900",
-              "lineHeight": undefined,
+              "lineHeight": 30,
               "marginBottom": 10,
             }
           }
--- a/packages/edition-slices/__tests__/ios/__snapshots__/tiles-with-style.test.js.snap
+++ b/packages/edition-slices/__tests__/ios/__snapshots__/tiles-with-style.test.js.snap
@@ -1493,9 +1493,9 @@ exports[`28. tile ae 1`] = `
             Object {
               "color": "#333333",
               "fontFamily": "TimesModern-Bold",
-              "fontSize": 30,
+              "fontSize": undefined,
               "fontWeight": "900",
-              "lineHeight": 30,
+              "lineHeight": undefined,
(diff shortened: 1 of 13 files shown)


packages/edition-slices/__tests__/ios/__snapshots__/tiles-with-style.test.js.snap
```diff
@@ -1493,9 +1493,9 @@ exports[`28. tile ae 1`] = `
             Object {
               "color": "#333333",
               "fontFamily": "TimesModern-Bold",
-              "fontSize": 30,
+              "fontSize": undefined,
               "fontWeight": "900",
-              "lineHeight": 30,
+              "lineHeight": undefined,
```
ValGeorgiev: Did you check why snapshot is like this?

packages/edition-slices/src/tiles/tile-ae/index.js
```diff
@@ -1,36 +1,42 @@
+/* eslint-disable react/require-default-props */
```
infreezer: What's your opinion on removing this rule and moving the default value to defaultProps?

apostolnikov: The makers of React discourage the use of defaultProps, especially for function components, also it's going to be deprecated in the near future...
You can read more about here: https://twitter.com/dan_abramov/status/1133878326358171650
https://twitter.com/dan_abramov/status/1133883865930567680

You can also check this PR's comment section for what the team agreed upon: https://github.com/newsuk/times-components/pull/2094

infreezer: Great. Thanks for the information.

packages/edition-slices/src/tiles/tile-ae/index.js
```diff
@@ -1,36 +1,42 @@
...
 import React from "react";
 import PropTypes from "prop-types";
+import editionBreakpoints from "@times-components/styleguide";
```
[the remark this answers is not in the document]
apostolnikov: Good catch 👍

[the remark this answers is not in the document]
infreezer: This will also fix the "undefined" snapshot.

packages/edition-slices/__tests__/ios/__snapshots__/tiles-with-style.test.js.snap
apostolnikov: This was caused by the wrong import of **editionBreakpoints** inside **TileAE** so that the default parameter was never passed. This is fixed now.

PR #2281, opened 2019-09-12
WIP: (fix) make image loading work better

jmars:
Still work in progress, but so far:

* Lazy loads high-res images in the background using interaction manager
* Cancels high-res image loads if you scroll past the image in a flatlist
* Typescript
* Use react-native image components native properties for placeholder/lowres/highres transitions
* Use standard URL parsing for crops and offline/online instead of our own stuff


packages/ad/package.json
```diff
@@ -54,7 +54,7 @@
     "react": "16.9.0",
     "react-dom": "16.9.0",
     "react-native": "0.59.10",
-    "react-test-renderer": "16.6.3",
+    "react-test-renderer": "16.9.0",
```
parkjh: this upgrade can be a separate pr... @jmars

jmars: Yes although the tests won't pass with the older version, but can rebase once the other PR is in

parkjh: don't wait up for that...

jmars:
Not happy with how this turned out

parkjh:
> Not happy with how this turned out

what happened?

PR #2285, opened 2019-09-16
feat: ppdsc 180 add newskit tags for article extras

parkjh:


packages/article-extras/src/article-topics/index.web.js
```diff
@@ -1,15 +1,49 @@
...
+  lightTheme,
+  BorderRadiusShape
+} from "newskit";
 import TopicsContainer from "./styles/responsive";
 import styles from "./styles";
 
+const NewArticleTopics = styled(TagList)`
+  text-align: center;
+`;
+
```
infreezer: Is it better if we move the styles to separate file?

packages/article-extras/src/article-topics/index.web.js
```diff
@@ -1,15 +1,44 @@
...
 
 const ShowTopics = ({ topics }) => {
   if (topics && topics.length > 0) {
     return (
       <TopicsContainer>
-        <nav>
-          <ArticleTopics style={styles.topicsContainer} topics={topics} />
+        <nav data-cy="topic-tags">
+          <Context.Consumer>
+            {({ newskit, makeTopicUrl }) => {
```
ValGeorgiev: could we use useContext hook here please

parkjh: https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react

parkjh: @jmars another reason why we need to upgrade storybook... ^^^

PR #2287, opened 2019-09-16
chore: upgrade to RN60.4

jmars:
Upgrade times-components to work on react native 0.60.x, with this and the android work we can move to the new native runtime.


packages/article-summary/__tests__/android/__snapshots__/article-summary-with-style.android.test.js.snap
```diff
@@ -79,42 +79,7 @@ exports[`1. article summary component with a single paragraph 1`] = `
...
-                "name": "text",
-              },
-            ],
-            "name": "inline",
-          },
-        ]
-      }
-    />
-  </Text>
+  <Text />
```
ValGeorgiev: This doesn't look  good

packages/article-summary/__tests__/shared-with-style.base.js
```diff
@@ -4,7 +4,6 @@ import { iterator } from "@times-components/test-utils";
 import ArticleSummary, { ArticleSummaryContent } from "../src/article-summary";
 import defaultFixture from "../fixtures/default";
 
-jest.mock("@times-components/article-byline", () => "ArticleByline");
```
ValGeorgiev: why this is removed?

Plamen5kov:
do you know how this will affect the ios app, because there were some issues when upgrading to react native 0.60.x? Did you have the time to test the iOS app?

ValGeorgiev:
This needs to be tested on web as well with RNW

aronshamash:
We are going to also need to upgrade in render and test

Plamen5kov:
@Szymczak the latest branch on android and it seems to work ok, except for minor difficulties that are probably only happening on my machine when bundling!

packages/article-summary/__tests__/shared-with-style.base.js
jmars: There is a bug in babel somewhere that is throwing invalid set key exceptions when this is here after the upgrade.

Plamen5kov:
@jmars, I tested renderer with `?react=1` flag and it all seems in order, but it would be nice if someone double checks.

packages/article-summary/__tests__/shared-with-style.base.js
jmars: The bug was cuased by babel incorrectly generating the mock factory for jest, total insanity, I could not figure out why but manually setting the mock factory restores the mock and the snapshot. :(

parkjh: > The bug was cuased by babel incorrectly generating the mock factory for jest, total insanity, I could not figure out why but manually setting the mock factory restores the mock and the snapshot. :(

don't think this has to do with babel, rather it's the way the article byline is mocked...
see https://github.com/newsuk/times-components/pull/2287/commits/0e8e251eb3a55c69e572dc42e912d587a5d41375 which produces the same snapshot

android/app/build.gradle
```diff
@@ -74,7 +74,8 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
-    entryFile: "fructose/index.js"
+        entryFile   : "fructose/index.js",
```
ValGeorgiev: I think this line is not needed. Probably we can deleted it with next pr

lib/hermes.js
```diff
@@ -0,0 +1,47 @@
+const { exec } = require("child_process");
+const os = require("os");
+const path = require("path");
+
+const puts = (error, stdout, stderr) => {
+  console.error(stderr);
+  console.log(stdout);
```
ValGeorgiev: do we need this console logs

android/app/build.gradle
parkjh: please remove

lib/hermes.js
parkjh: please remove

parkjh:
git is freaking out... :(

parkjh:
lol https://github.com/facebook/react-native/releases/tag/v0.61.1

parkjh:
see https://react-native-community.github.io/upgrade-helper/?from=0.60.4&to=0.61.1

parkjh:
closing as it cannot be merged yet...

PR #2293, opened 2019-09-17
fix: secondary 2 no pic and 2 slice minor styling

infreezer:
Aligning the vertical lines between SecondaryTwoNoPicAndTwo and other slices.
Breakpoint affected: wide and huge.

Before:
alignment-before

After:
alignment-after

### DIFF
--- a/packages/edition-slices/__tests__/android/__snapshots__/tablet-slices-with-style.test.js.snap
+++ b/packages/edition-slices/__tests__/android/__snapshots__/tablet-slices-with-style.test.js.snap
@@ -2223,7 +2223,7 @@ exports[`29. secondary two no pic and two - wide 1`] = `
       style={
         Object {
           "flexDirection": "row",
-          "width": "65%",
+          "width": "66.6%",
         }
       }
     >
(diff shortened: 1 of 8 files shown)


infreezer:
> How did you calculate the width?
> Did you confirm with the designers that this table looking alignment is what they intended?

No need for calculation. I'm taking the width from design in pixels.
image

apostolnikov:
> > How did you calculate the width?
> > Did you confirm with the designers that this table looking alignment is what they intended?
> 
> No need for calculation. I'm taking the width from design in pixels.
> image

Taking the width of the teaser copy won't make the 66.6% calculation by itself. That's why I'm asking how did you make the calculation...
Did you notice that the container of which you are taking those 66.6% is not starting from the line where the teaser starts?

image

infreezer:
> > > How did you calculate the width?
> > > Did you confirm with the designers that this table looking alignment is what they intended?
> > 
> > 
> > No need for calculation. I'm taking the width from design in pixels.
> > image
> 
> Taking the width of the teaser copy won't make the 66.6% calculation by itself. That's why I'm asking how did you make the calculation...
> Did you notice that the container of which you are taking those 66.6% is not starting from the line where the teaser starts?
> 
> image

No calculation, just playing with the percentage numbers until the correct pixels are reached.
In this slice we have 3 tiles with the same width. By design each of them should be 1/3 width.
I noticed that all of the slices on this breakpoint doesn't start from the line.

PR #2294, opened 2019-09-17
feat: list2and6nopic for 1366px tablet

apostolnikov:
[Story](https://nidigitalsolutions.jira.com/browse/REPLAT-8559)

Font size changes for tiles: 
TileAS, TileL

image


packages/edition-slices/src/tiles/tile-l/styles/index.js
```diff
@@ -4,51 +4,25 @@ import {
...
-  ...sharedStyles
-};
-
-const mediumBreakpointStyles = {
-  ...sharedStyles,
-  headlineStyle: {
-    ...sharedStyles.headlineStyle,
-    fontSize: 18,
-    lineHeight: 18,
-    marginBottom: 0
```
infreezer: We need to set the marginBottom to 0. For medium+ breakpoint. Not sure about mobile.
image

apostolnikov: Where did you checked that?
By design, there should be margin bottom of 10px for these breakpoints.
Medium:
image
Wide:
image

Huge:
image

infreezer: Currently the total spacing is 20.
image

packages/edition-slices/src/tiles/tile-l/styles/index.js
```diff
@@ -4,51 +4,25 @@ import {
   editionBreakpoints
 } from "@times-components/styleguide";
 
-const sharedStyles = {
+const fontSizeResolver = {
+  [editionBreakpoints.small]: 18,
```
infreezer: The font size on small was 20.

PR #2295, opened 2019-09-17
feat: REPLAT-8755 add a switch to disable autoscroll on web articlelist

jmars:
Add a small flag for usages on web that don't involve autoscroll

### DIFF
--- a/packages/article-list/src/article-list.web.js
+++ b/packages/article-list/src/article-list.web.js
@@ -50,6 +50,7 @@ class ArticleList extends Component {
       imageRatio,
       onNext,
       onPrev,
+      scrollToTop = true,
       page = 0,
       pageSize,
       receiveChildList,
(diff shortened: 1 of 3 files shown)


packages/article-list/src/article-list.web.js
```diff
@@ -35,7 +35,7 @@ class ArticleList extends Component {
 
   shouldComponentUpdate(nextProps) {
     const { page } = this.props;
-    return page === nextProps.page;
+    return page !== nextProps.page;
```
aronshamash: why the reversal of this logic ?

jmars: It's wrong, returning true causes and update and false stops it, so this would update if the page was the same, but not if it changed :laughing:

aronshamash: ok a bug fix nice.

PR #2299, opened 2019-09-19
fix: react resolution

parkjh:
why?

fixes https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react
using https://yarnpkg.com/lang/en/docs/selective-version-resolutions/


ValGeorgiev:
This needs to be added in `lib/cli/package-json.js` as well

parkjh:
> This needs to be added in `lib/cli/package-json.js` as well

done

PR #2302, opened 2019-09-19
feat: daily register for 1366px

pdimova:
Updating styles for the headline icons that affect all breakpoints except small screens

**Medium**
Simulator Screen Shot - iPad Air 2 - 2019-09-19 at 14 00 34

**Wide**
Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2019-09-19 at 14 04 09

**Huge**
Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2019-09-19 at 14 05 29

### DIFF
--- a/packages/edition-slices/src/tiles/tile-s/index.js
+++ b/packages/edition-slices/src/tiles/tile-s/index.js
@@ -21,12 +21,16 @@ const TileS = ({
   logo = null
 }) => {
   const styles = styleFactory(breakpoint);
+  const headLineStyles =
+    breakpoint === editionBreakpoints.medium && logo
+      ? styles.titleWithoutMargin
+      : styles.title;
 
   return (
     <View style={styles.container}>
       <View style={styles.titleWrapper}>
         {logo}
-        <Text style={styles.title}>{tile.title}</Text>
+        <Text style={headLineStyles}>{tile.title}</Text>
       </View>
       <Text style={styles.paragraph}>{renderAst(tile.content)}</Text>
       {tile.byline &&
--- a/packages/edition-slices/src/tiles/tile-s/index.js
+++ b/packages/edition-slices/src/tiles/tile-s/index.js
@@ -26,7 +26,7 @@ const TileS = ({
     <View style={styles.container}>
       <View style={styles.titleWrapper}>
         {logo}
-        <Text style={[styles.title, styles.noMargin]}>{tile.title}</Text>
+        <Text style={styles.title}>{tile.title}</Text>
       </View>
       <Text style={styles.paragraph}>{renderAst(tile.content)}</Text>
       {tile.byline &&
--- a/packages/edition-slices/src/tiles/tile-s/index.js
+++ b/packages/edition-slices/src/tiles/tile-s/index.js
@@ -26,9 +26,7 @@ const TileS = ({
     <View style={styles.container}>
       <View style={styles.titleWrapper}>
         {logo}
-        <Text style={[styles.title, logo ? { marginBottom: 0 } : {}]}>
-          {tile.title}
-        </Text>
+        <Text style={[styles.title, styles.noMargin]}>{tile.title}</Text>
       </View>
       <Text style={styles.paragraph}>{renderAst(tile.content)}</Text>
       {tile.byline &&
--- a/packages/edition-slices/src/tiles/tile-s/index.js
+++ b/packages/edition-slices/src/tiles/tile-s/index.js
@@ -26,7 +26,9 @@ const TileS = ({
     <View style={styles.container}>
       <View style={styles.titleWrapper}>
         {logo}
-        <Text style={styles.title}>{tile.title}</Text>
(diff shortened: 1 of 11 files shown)


packages/edition-slices/src/tiles/tile-s/index.js
```diff
@@ -26,7 +26,9 @@ const TileS = ({
     <View style={styles.container}>
       <View style={styles.titleWrapper}>
         {logo}
-        <Text style={styles.title}>{tile.title}</Text>
+        <Text style={[styles.title, logo ? { marginBottom: 0 } : {}]}>
+          {tile.title}
+        </Text>
```
apostolnikov: You can go without this check for the additional style that you are adding as it won't make any change

pdimova: true there is no visual difference with or without the check, just prefer not to append additional css where it's not needed - in this case, the headlines without logo will also have uneccesary margin-bottom

apostolnikov: You prefer not to append additional css by doing it so. If there is no visual difference with or without the check why would you add additional complexity into the code logic?

packages/edition-slices/src/tiles/tile-s/index.js
```diff
@@ -26,7 +26,7 @@ const TileS = ({
     <View style={styles.container}>
       <View style={styles.titleWrapper}>
         {logo}
-        <Text style={styles.title}>{tile.title}</Text>
+        <Text style={[styles.title, styles.noMargin]}>{tile.title}</Text>
       </View>
```
apostolnikov: Why would you separate the noMargin css description outside the styles.title and then merge it back here again? If you want styles.title to have marginBottom of 0 just add it inside the styles for the title.

packages/edition-slices/src/tiles/tile-s/styles/index.js
```diff
@@ -62,15 +61,35 @@ const mediumBreakpointStyles = {
...
   },
   titleWrapper: {
     flexDirection: "row",
-    alignItems: "center"
+    alignItems: "center",
+    marginBottom: spacing(1)
+  },
+  noMargin: {
+    marginBottom: 0
+  }
```
apostolnikov: this doesn't need to be separated as it's not used anywhere else here in the styles, nor it's added into the js code by a check

packages/edition-slices/src/tiles/tile-s/index.js
```diff
@@ -26,7 +26,7 @@ const TileS = ({
     <View style={styles.container}>
       <View style={styles.titleWrapper}>
         {logo}
-        <Text style={styles.title}>{tile.title}</Text>
+        <Text style={[styles.title, styles.noMargin]}>{tile.title}</Text>
```
ValGeorgiev: Based on the snapshots it looks like title marginBottom is not used anymore, could you please merge this styles in styles.js file. There is no need of putting more complexity in the code. Thanks

pdimova: This is what we have in the designs:
-for medium breakpoint
Screenshot 2019-09-25 at 21 50 56

-for wide/huge breakpoint
Screenshot 2019-09-25 at 21 51 14

As you can see the difference comes in the margin space above the strapline text. 

Currently, in our structure we have a title wrapper that wraps the logo (if any) and the headline. 
In order to have the headline vertically centred with the logo it should not have any space around it.

The easiest approach is to control the margin of the tile wrapper - 5px for medium breakpoint and 10px for wide/huge breakpoint. And add the extra _5px bottom margin to the firs two headlines for medium breakpoint_ . Thus the following styles for medium breakpoint
```
  title: {
    color: colours.functional.brandColour,
    fontFamily: fonts.headline,
    fontSize: 20,
    lineHeight: 20,
    marginBottom: spacing(1)
  },
```
The purpose of this line `logo ? { marginBottom: 0 } : {}` was to overwrite the bottom margin for those two headlines that have logo icon on the left and should be centered 

Without this check the headlines for the last two tiles for medium breakpoint are not positioned in the middle of their parent container:
Simulator Screen Shot - iPad Air 2 - 2019-09-25 at 21 49 46
Simulator Screen Shot - iPad Air 2 - 2019-09-25 at 21 49 49

For the other breakpoints, there is not such an obstacle since the correct spacing comes out of the box once we have the title wrapper set. Indeed, the ternary doesn't do much in here.

```
const wideBreakpointStyle = {
  ...mediumBreakpointStyles,
  title: {
    ...mediumBreakpointStyles.title,
    marginBottom: 0
  },
  titleWrapper: {
    ...mediumBreakpointStyles.titleWrapper,
    marginBottom: spacing(2)
  }
};
```

I still don't think that a ternary operator is an uneccesary complexity. 

Please advise if you have a better approach to tackle the 5px difference for the smaller screens or if we are fine to go this way.

PR #2303, opened 2019-09-19
fix: REPLAT-7797 fix referral url passing

jmars:
The referral url was not being passed through to native in the same way as everything else and so they were sometimes getting null values.

https://nidigitalsolutions.jira.com/browse/REPLAT-7797


packages/article-skeleton/src/tracking/article-tracking-context.js
```diff
@@ -14,7 +14,7 @@ export default Component =>
       label: get(data, "label", ""),
       pageName: `${get(data, "slug", "")}-${get(data, "shortIdentifier", "")}`,
       publishedTime: get(data, "publishedTime", ""),
-      referralUrl,
+      referralUrl: referralUrl || get(data, "referralUrl", ""),
```
ValGeorgiev: could we add a test for that, we are trying to bring the test coverage back to 90+ %

PR #2305, opened 2019-09-20
feat: increase test coverage for slices (Mobile only !) 

pdimova:
Mobile test coverage:

Screenshot 2019-09-20 at 17 28 08

In order for the web to have 100% coverage, we will need to fix the `Dimension` mock first so we can set different breakpoints in the web tests

### DIFF
--- a/packages/edition-slices/__tests__/web/__snapshots__/tablet-slices.test.js.snap
+++ b/packages/edition-slices/__tests__/web/__snapshots__/tablet-slices.test.js.snap
@@ -435,7 +435,77 @@ exports[`14. secondary two no pic and two - medium 1`] = `
 </div>
 `;
 
-exports[`15. puzzle - medium 1`] = `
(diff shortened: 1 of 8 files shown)


packages/edition-slices/__tests__/web/__snapshots__/tablet-slices.test.js.snap
```diff
@@ -435,7 +435,77 @@ exports[`14. secondary two no pic and two - medium 1`] = `
 </div>
 `;
 
-exports[`15. puzzle - medium 1`] = `
+exports[`15. list two and six no pic - medium 1`] = `
+<div>
+  <div>
+    <div>
+      <TileAS
+        breakpoint="wide"
```
ValGeorgiev: why the breakpoint is wide, but the test is medium?

pdimova: a known issue for another ticket imo

PR #2309, opened 2019-09-24
feat: newskit client side feature flag

parkjh:

### DIFF
--- a/packages/ssr/src/client/article.js
+++ b/packages/ssr/src/client/article.js
@@ -1,3 +1,5 @@
+/* eslint-disable no-useless-escape */
+
 const article = require("../component/article");
 const runClient = require("../lib/run-client");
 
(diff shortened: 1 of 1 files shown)


packages/ssr/src/client/article.js
```diff
@@ -18,9 +20,23 @@ if (window.nuk && window.nuk.ssr && window.nuk.article) {
   } = window.nuk.article;
   const { getCookieValue } = window.nuk;
 
+  const enableNewskit = decodeURIComponent(
+    window.location.search.replace(
+      new RegExp(
+        `^(?:.*[&\\?]${encodeURIComponent("newskit").replace(
+          /[\.\+\*]/g,
+          "\\$&"
+        )}(?:\\=([^&]*))?)?.*$`,
```
ValGeorgiev: Could you explain why you need this escaping :)

parkjh: > Could you explain why you need this escaping :)

well for the regex here, eslint thinks that the escape char is not needed... which is not right

PR #2316, opened 2019-09-30
feat: add author to jsonLD schema REPLAT-7841

todorRadkov:
Adding author name into schema markup

JIRA:
https://nidigitalsolutions.jira.com/browse/REPLAT-7841

### DIFF
--- a/packages/article-skeleton/src/head.web.js
+++ b/packages/article-skeleton/src/head.web.js
@@ -159,6 +159,7 @@ function Head({ article, paidContentClassName, faviconUrl }) {
       isAccessibleForFree: false,
       cssSelector: `.${paidContentClassName}`
     },
+    author: authorName,
     image: {
       "@type": "ImageObject",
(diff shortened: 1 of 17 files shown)


packages/article-skeleton/src/head.web.js
```diff
@@ -159,6 +159,7 @@ function Head({ article, paidContentClassName, faviconUrl }) {
       isAccessibleForFree: false,
       cssSelector: `.${paidContentClassName}`
     },
+    author: authorName || "Times Editorial",
```
ValGeorgiev: Did we discussed that we don't need to put a default value ?

todorRadkov: One of the tests that had no value for byline was returning `null` as value for author and I discussed it with Jinnat Hasan to replace it with  `Times Editorial` for the schema only.

ValGeorgiev:
we have other pr for that, closing this one

PR #2319, opened 2019-10-01
fix: (REPLAT-9280) disable fontSizing on tile article headlines

jmars:
We have limited space on tiles for article headlines and allowFontSizing with a large OS size causes clipping issues :(

edit: This appears to only be a bug on react-native android, iOS text containers are flexed correctly.

android: before
Screenshot_1570017907

android:after
Screenshot_1570017968

ios: before
Screenshot_20191002_135622

ios:after - it scales fine
Screenshot_20191002_135458

### DIFF
--- a/packages/article-summary/src/article-summary-headline.js
+++ b/packages/article-summary/src/article-summary-headline.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { Text } from "react-native";
+import { Text, Platform } from "react-native";
 import PropTypes from "prop-types";
 import styles from "./styles";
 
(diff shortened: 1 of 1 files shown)


apostolnikov:
> We have limited space on tiles for article headlines and allowFontSizing with a large OS size causes clipping issues :(

Could you provide more description of the problem in order to have better understanding. How I'm able to reproduce it? Maybe it'll be helpful if you add before/after images on all platforms, with and without this prop **allowFontScaling={false}** being applied?

jmars:
You can reproduce this on android by changing the text size in the OS settings to a larger than default value (which some manufacturers do by default). The scaling functionality works fine on iOS

infreezer:
Do we have an approval from the product that dropping the scaling functionality for the headline on Android is okay?

PR #2320, opened 2019-10-01
fix: (REPLAT-9274) fix missing text

jmars:
We were losing some text due to a weird issue with nested Text views, an easy fix is to just set the height from our own layout stuff.
**BEFORE**
Screenshot_1569947518
**AFTER**
Screenshot_1569947416

### DIFF
--- a/packages/text-flow/lib/convert-font.js
+++ b/packages/text-flow/lib/convert-font.js
@@ -32,8 +32,8 @@ const convert = (filename, target) => {
     });
   });
   const fontObj = {
-    ascent: font.acender,
-    descent: font.decender,
+    ascent: font.ascender,
+    descent: font.descender,
     units: font.unitsPerEm,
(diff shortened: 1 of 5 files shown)


packages/text-flow/lib/convert-font.js
```diff
@@ -32,8 +32,8 @@ const convert = (filename, target) => {
     });
   });
   const fontObj = {
-    ascent: font.acender,
-    descent: font.decender,
```
aronshamash: was think just wrong before ?

jmars: Yeah it was a typo from when we removed python for the font processing, never became an issue as we didn't use height anywhere before this change. Ideally if react native 0.60s new text handling works better I'd still like to revert the height being set on paragraphs as we shouldnt even be measuring them.

PR #2321, opened 2019-10-02
feat: REPLAT-9309 lead 1 full width structure updated 

infreezer:
[Story.](https://nidigitalsolutions.jira.com/browse/REPLAT-9309)
[Design.](https://app.zeplin.io/project/5d2849a2b4a4605645afb4be/screen/5d28524f9cb5a063a8a20ac2)
Tile rearranged and minor styling updates for all breakpoints.

Before:
full width 768 before

After:
768:
full width 768 after

1024:
full width 1024 after

1366:
full width 1366 after


infreezer:
> Can you double check the spacing above and beneath the flag, as the one above looks too tight?

Looks like 10 for 768:
lead 1 full width 768 padding label

I'll increase it to 15px as per design for 1024 and 1366.

PR #2330, opened 2019-10-07
fix: (REPLAT-9346) fix cut off article paragraphs

jmars:


packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -1,429 +1,349 @@
...
+import Context from "@times-components/context"
 import InsetCaption from "./inset-caption";
-import styleFactory from "../styles/article-body";
-import { getDropCap, getDropCapValue } from "../dropcap-util-common";
+import styleFactory from "../styles/article-body"
+import ArticleLink from "./article-link"
+
+const makeDropCap = (scale, color, dropCapFont, paragraph) => {
+  const letter = paragraph.slice(0, 1)
+  const baseStyle = letter.attributes[0].tag.settings
```
ValGeorgiev: Are we sure that `letter` will have attributes array with an element inside? Should we do a check

packages/text-flow/lib/convert-font.js
```diff
@@ -2,6 +2,12 @@ const fs = require('fs');
 const Opentype = require('opentype.js');
 
 const convert = (filename, target) => {
+  const buffer = fs.readFileSync(filename)
+  const ints = new Int32Array(buffer.buffer)
+  const buf = []
+  for (let i = 0; i < ints.length; i++) {
+    buf[i] = ints[i]
+  }
```
apostolnikov: Do we have any particular reason why we want to convert the typed array? If we do, it's probably better to use **const buf = Array.from(ints);** instead of iterating it.

packages/ad/src/styles/index.js
```diff
@@ -57,7 +57,8 @@ const styles = StyleSheet.create({
   },
   container: {
     alignItems: "center",
-    flex: 1
+    flex: 1,
+    marginBottom: spacing(4)
   },
```
apostolnikov: With this marginBottom of the ad container, we are still following the designs right?

packages/typeset/README.md
```diff
@@ -0,0 +1,3 @@
+# @times-components/typeset
+
+A library for laying out text
```
parkjh: :(

packages/text-flow/lib/convert-font.js
jmars: Good point, changed this when deleting the old package

packages/ad/src/styles/index.js
jmars: That change should not have effected web, reversed it

packages/article-skeleton/__tests__/android/__snapshots__/article-images-with-style.android.test.js.snap
```diff
@@ -188,50 +136,24 @@ exports[`3. a secondary image 1`] = `
         <View
           style={
             Object {
-              "alignSelf": "center",
               "backgroundColor": "#ffffff",
               "maxWidth": "100%",
-              "width": 1194,
+              "overflow": "hidden",
+              "width": undefined,
```
ValGeorgiev: Could you check this, please

packages/section/__tests__/android/__snapshots__/section.test.js.snap
```diff
@@ -473,7 +473,7 @@ exports[`should render secondary 2 and 2 for mobile (small breakpoint) 1`] = `
 <RCTScrollView>
   <View>
     <View>
-      <SecondaryTwoAndTwo />
+      <SecondaryTwoNoPicAndTwo />
```
ValGeorgiev: There is a mapping for that, but it only should be applied on tablet.. this is mobile. (potential issue)

packages/edition-slices/src/edition-slices.js
```diff
@@ -19,7 +19,7 @@ import {
   PuzzleSlice
 } from "./slices";
 
-const { width } = getDimensions();
+const { width } = Dimensions.get("window");
```
ValGeorgiev: @infreezer please check this ^

packages/typeset/src/types/externs.d..ts
```diff
@@ -0,0 +1,117 @@
+type Line = ['L', number, number];
```
ValGeorgiev: this file should be `externs.d.ts`

packages/edition-slices/src/edition-slices.js
infreezer: In [the tests](https://github.com/newsuk/times-components/blob/915e84c4379a7959cbfd8423741996c9b2f1050b/packages/section/__tests__/shared.js#L90) we are mocking the utils package.
In this PR the dimensions is not part of the utils anymore. So the test should be fixed.
[The dimensions was extracted in utils 2 weeks ago in order to fix the tests on web.]([url](https://github.com/newsuk/times-components/pull/2324))
I believe deleting [dimensions-util.js](https://github.com/newsuk/times-components/blob/master/packages/utils/src/dimensions-util.js) might break a lot of tests.
What's the reason of the deletion?

packages/edition-slices/__tests__/ios/__snapshots__/tablet-slices-with-style.test.js.snap
```diff
@@ -1,251 +1,392 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`1. comment lead and cartoon - medium 1`] = `
+<View>
```
infreezer: The snapshot shows the output of small breakpoint.
The expected result is tablet.
[Mocking the dimensions here](https://github.com/newsuk/times-components/blob/fix/REPLAT-9346/packages/edition-slices/__tests__/shared-tablet-slices.base.js#L128) should be replaced with something like: 
`Dimensions.set({
      window: {
        width: tabletWidth,
        height: 640
      }
    });`

packages/edition-slices/__tests__/ios/__snapshots__/tablet-slices.test.js.snap
```diff
@@ -2,1777 +2,7716 @@
...
-    </View>
-    <View />
-    <View>
-      <TileAI
-        breakpoint="medium"
-        tileName="cartoon"
-      />
-    </View>
-  </View>
+  <TileP
```
infreezer: The snapshot is inaccurate and the reason is that
this [mock](https://github.com/newsuk/times-components/blob/fix/REPLAT-9346/packages/edition-slices/__tests__/shared-slices.base.js#L140) isn't valid anymore. 
React Native dimensions should be directly mocked. Can you check?

packages/article-skeleton/src/article-body/article-body-row.js
jmars: I added a check, it shouldnt be an issue unless markup is malformed but that can happen sometimes

packages/typeset/src/types/externs.d..ts
jmars: Good catch, im surprised the double dot worked at all, i guess typescript ignores the extra one :)

jmars:
I've restored the dimension mocks, I originally did that because jest was complaining but I found in the docs that you can define a mock as 'virtual' which restores the old behaviour

android-app/xnative/build.gradle
```diff
@@ -71,10 +71,17 @@ android {
         versionName packageInfo.version
         consumerProguardFiles 'proguard-rules.pro'
     }
+
+    packagingOptions {
+        // Make sure libjsc.so does not packed in APK
```
aronshamash: typo

ios-app/fonts.js
```diff
@@ -1,23 +1,35 @@
...
-    .default,
-  "TimesModern-Bold": require("../dist/public/fonts/TimesModern-Bold-metrics.js")
-    .default,
-  "TimesDigitalW04-Regular": require("../dist/public/fonts/TimesDigitalW04-Regular-metrics.js")
-    .default,
-  TimesDigitalW04: require("../dist/public/fonts/TimesDigitalW04-metrics.js")
-    .default,
-  "TimesModern-Regular": require("../dist/public/fonts/TimesModern-Regular-metrics.js")
-    .default
+const fonts = {
```
aronshamash: If this list of fonts is these same for ios and android we should DRY it up. But I guess that could be in different PR.

packages/article-paragraph/__tests__/android/__snapshots__/article-paragraph-with-style.android.test.js.snap
```diff
@@ -35,29 +22,16 @@ exports[`2. responsive tablet paragraph 1`] = `
   style={
     Object {
       "alignSelf": "center",
+      "height": undefined,
```
aronshamash: Can we check this is as expected ?

packages/gestures/__tests__/android/__snapshots__/gestures.android.test.js.snap
```diff
@@ -105,39 +107,40 @@ exports[`2. rotate anti-clockwise when turning fingers 1`] = `
...
 <View>
   <View
+    focusable={true}
     style={
       Object {
         "flexGrow": 1,
         "transform": Array [
           Object {
-            "translateX": -0,
+            "translateX": -15,
```
aronshamash: why are all these numbers changing ?

jmars: Looks like different versions of react-native etc can cause these values to change, see:
https://github.com/newsuk/times-components/blob/master/packages/gestures/__tests__/shared.native.js#L249

packages/article-paragraph/__tests__/android/__snapshots__/article-paragraph-with-style.android.test.js.snap
jmars: Not expected, was left over from the attempted quick fix of setting paragraph heights, removed.

ios-app/fonts.js
jmars: Good point, no reason we shouldn't do this to make sure both platforms actually do always have all fonts.

android-app/xnative/build.gradle
jmars: Copy and pasted from the react-native-v8 guide, my bad :)

parkjh:
> @parkjh :) i am approving it again, thanks for removing my approval....

no I removed my approval :p

PR #2335, opened 2019-10-09
chore: web only save-and-share bar

parkjh:
why?
- these packages are only used on the web


ValGeorgiev:
Why this is needed? We are doing a cross-device components, not web or native.. Please explain why this is needed, thanks

parkjh:
> Why this is needed? We are doing a cross-device components, not web or native.. Please explain why this is needed, thanks

this is because I am trying to incorporate newskit into this package, and given that the app share bar is in native code and there's no plans to move it I think it makes it easier to maintain? but this needs more work with styling anyways

PR #2337, opened 2019-10-11
chore: disable eslint default prop rule

parkjh:


ValGeorgiev:
We need this rule, because ppl will start adding attributes without default values

PR #2340, opened 2019-10-15
feat: HomePage skeleton and ssr implementation

apostolnikov:
[Story](https://nidigitalsolutions.jira.com/browse/REPLAT-9625)

In this PR we are adding basic skeleton for the Home Page to be visible behind **?react=true** flag in render.


packages/home-page/README.md
```diff
@@ -0,0 +1,53 @@
+# HomePage
+
+<!-- Add the description of the package here. -->
```
ValGeorgiev: Please add description

packages/home-page/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+To link the package to render project run `yarn watch`, make some code change and restart the server
+
+## Testing
+
+Testing can be done on each platform individually
+
+```
+yarn test:android
+yarn test:ios
```
ValGeorgiev: This is web package only, so these commands are not needed

packages/home-page/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+Or the tests for all platforms can be run
+
+```
+yarn test:all
+```
+
+Visit the official
+
+<!-- Add the storybook link here. -->
```
ValGeorgiev: ^^^

packages/home-page/__tests__/shared.base.js
```diff
@@ -0,0 +1,18 @@
...
+    {
+      name: "renders correctly",
+      test: () => {
+        const testInstance = TestRenderer.create(<HomePage />);
+
+        expect(testInstance.toJSON()).toMatchSnapshot();
+      }
+    }
+  ];
+  iterator(tests);
```
ValGeorgiev: Please don't use iterator

packages/home-page/__tests__/shared.native.js
```diff
@@ -0,0 +1,21 @@
+import {
```
ValGeorgiev: we don't need this file, isn't it

packages/home-page/package.json
```diff
@@ -0,0 +1,74 @@
...
+  "dev": "src/home-page",
+  "scripts": {
+    "depcheck": "depcheck --ignores='@babel/*,babel-*,depcheck,eslint,jest,prettier,webpack*' --ignore-bin-package=false --skip-missing",
+    "fmt": "eslint . --fix && prettier --write '**/*.*'",
+    "prettier:diff": "prettier --list-different '**/*.*'",
+    "lint": "eslint . && yarn prettier:diff && yarn depcheck",
+    "test:android": "jest --config='./__tests__/android/jest.config.js'",
+    "test:ios": "jest --config='./__tests__/ios/jest.config.js'",
+    "test:web": "jest --config='./__tests__/web/jest.config.js'",
+    "test:all": "yarn test:android --coverage && yarn test:ios --coverage && yarn test:web --coverage",
```
ValGeorgiev: android and ios commands are useless here

packages/home-page/src/styles/index.android.js
```diff
@@ -0,0 +1,12 @@
+import { StyleSheet } from "react-native";
```
ValGeorgiev: this file is not needed

packages/home-page/src/styles/index.web.js
```diff
@@ -0,0 +1,12 @@
+import { StyleSheet } from "react-native";
```
ValGeorgiev: we can use only styles/index.js

packages/home-page/src/styles/index.js
```diff
@@ -0,0 +1,8 @@
+import { StyleSheet } from "react-native";
+import sharedStyles from "./shared";
```
ValGeorgiev: shared file is not needed in this case

packages/ssr/src/client/home-page.js
```diff
@@ -0,0 +1,27 @@
+const homePage = require("../component/home-page");
```
ValGeorgiev: HomePage is a component, so it should be upper case

packages/ssr/src/component/home-page.js
```diff
@@ -0,0 +1,27 @@
...
+
+  return React.createElement(
+    HelmetProvider,
+    { context: helmetContext },
+    React.createElement(
+      ApolloProvider,
+      { client },
+      React.createElement(
+        ContextProviderWithDefaults,
+        { value: { makeArticleUrl, makeTopicUrl } },
```
ValGeorgiev: do we need makeTopicUrl here?

packages/ssr/src/server/home-page.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+  if (!logger) {
+    throw new Error(`Logger is required. Received ${logger}`);
+  }
+  if (!makeArticleUrl) {
+    throw new Error(
+      `Make article url function is required. Received ${makeArticleUrl}`
+    );
+  }
+  if (!makeTopicUrl) {
```
ValGeorgiev: is it required?

yarn.lock
```diff
@@ -3928,6 +3935,13 @@ acorn-jsx@4.1.1:
   dependencies:
     acorn "^5.0.3"
 
+acorn-jsx@^3.0.0:
```
ValGeorgiev: yarn.lock doesn't look good and i cannot see any new dependancies, only new package

packages/ssr/src/client/home-page.js
apostolnikov: Agree, but in this context, it will be passed as a function param at the bottom of the module which is not a good pattern to follow if it's upper case letter. Also, we are using the same pattern on all the other ssr client components.

yarn.lock
apostolnikov: There are no new external packages being added by this PR. What new dependancies you want to see in the yarn.lock?

packages/home-page/__tests__/shared.base.js
apostolnikov: Why not? What do you want to use instead?

packages/ssr/src/component/home-page.js
apostolnikov: Those are the default server options passed through the ContextProviderWithDefaults for all the ssr components. I think we need them, what do you think?

packages/ssr/src/server/home-page.js
apostolnikov: same as above..

packages/home-page/__tests__/shared.base.js
ValGeorgiev: regular jest describe and it, we want to be able to debug tests and this is breaking that. Iterator is very useless in this case

packages/ssr/src/component/home-page.js
ValGeorgiev: No idea, we need to check where we are using these functions. I am assuming that we need them for topics in article, on homepage we don't have topic links

yarn.lock
ValGeorgiev: Thats the case, we shouldn't have new dependencies in yarn.lock, but in the diff it looks like we have new dependencies

packages/home-page/__tests__/shared.base.js
apostolnikov: We should make this change on all the other components. Currently the majority of the packages are using the iterator...

yarn.lock
apostolnikov: No Idea what's causing this as no new packages are added. It mights be because of an update of the existing packages which are using new modules in their new versions...

aronshamash:
@apostolnikov what is happening with this PR. Can we close it ?

apostolnikov:
> @apostolnikov what is happening with this PR. Can we close it ?

The last update that I got was that @flashcheeks is doing POC with a different approach and we will freeze this one until he's ready.

aronshamash:
@flashcheeks & @chrishutchinson what is the status on the POC? Can we close this PR?

chrishutchinson:
> @flashcheeks & @chrishutchinson what is the status on the POC? Can we close this PR?

@aronshamash let's close for now

packages/home-page/__tests__/shared.base.js
aronshamash: I would keep iterator if it is the current way of doing it. If we we want to create a github issue with the issue to remove iterator then someone can pick that up.

packages/ssr/src/component/home-page.js
aronshamash: yes they shouldn't be needed on homepage slices

PR #2348, opened 2019-10-23
fix: article initial load

jmars:
Articles are being passed the same props multiple times, since the component is a function this causes lots of unnecessary work that doesn't change the output, so it should be memoized.

### DIFF
--- a/packages/article-skeleton/src/article-body/article-body-row.js
+++ b/packages/article-skeleton/src/article-body/article-body-row.js
@@ -136,7 +136,7 @@ const Paragraph = ({
 
   const container = new TextContainer(
     isTablet ? contentWidth : screenWidth() - spacing(4),
-    5000,
+    1000,
     0,
     0,
     dropCap ? [dropCap.exclusion] : []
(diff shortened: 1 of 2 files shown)


packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -136,7 +136,7 @@ const Paragraph = ({
 
   const container = new TextContainer(
     isTablet ? contentWidth : screenWidth() - spacing(4),
-    5000,
+    1000,
```
ValGeorgiev: this number 10000, yesterday we decrease it to 5000, now 1000 ? what is this number?

jmars: It's the max paragraph height, this PR is invalid as this approach will cause problems down the road with huge paragraphs and the maxheight/memo is not where the issue was. Going to close this in favour of a new PR that chunks huge paragraphs correctly so we won't have issues down the road with tablet and restores this number to something reasonable.

PR #2352, opened 2019-10-25
fix: REPLAT-9346 android app fails to bundle because of fonts

Plamen5kov:
https://github.com/newsuk/times-components/commit/35cee66496828bf12ad3b1a6913097fd628577ce#diff-c55d5a77a3267bfe3cf57c7c402e52cd
Broke android-app build because it can't bundle fonts it can't find

### DIFF
--- a/fonts.js
+++ b/fonts.js
@@ -1,24 +1,24 @@
 /* eslint-disable global-require */
 const fonts = {
   "CenturyGothic-Bold": () =>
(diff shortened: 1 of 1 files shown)


infreezer:
Clean the dist folder on root level and run yarn

Plamen5kov:
Local issue, because of invalid old folder state

PR #2353, opened 2019-10-25
fix: OS font scaling

jmars:
Need to pass in the OS font setting when rendering articles.


infreezer:
Can you check if the links in articles are scaled?
image

PR #2354, opened 2019-10-25
fix: Logo for seo

aronshamash:
- in schema.org metadata the logo field needs to use the dual masthead image not the favicon as per https://nidigitalsolutions.jira.com/browse/REPLAT-8274
- Render PR to pass `logoUrl` with new image coming


aronshamash:
@ValGeorgiev I ran on all packages and only `article-skeleton` needs update

ValGeorgiev:
> @ValGeorgiev I ran on all packages and only `article-skeleton` needs update

Strange, we updated schema last week and it required a lot of snapshots updates

PR #2355, opened 2019-10-25
Fix: Replat-9642 no click tracking on related articles

mikefieldmay:
[Replat-9642](https://nidigitalsolutions.jira.com/browse/REPLAT-9642)

After having a look at the Related Article Item web it looks like a HOC was created to handle the analytics events but never implemented for the web version. 
After wrapping the component the events seem to be firing properly


mikefieldmay:
Added the Do Not Merge Label as I am waiting to test this in Render properly

mikefieldmay:
Removed do not merge as it works locally

PR #2363, opened 2019-10-29
fix: image loading crops and overdraw

jmars:
Resolves the strange cropping issue (by flooring crop values as native does) and the flickering when images change from offline to online (by only updating element attributes rather than causing a reconciliation and vdom rebuild).


parkjh:
@RichardHeinz97 you enjoying snapshots I see~

PR #2364, opened 2019-10-30
feat: add article id to aid debugging and testing

aronshamash:
- Add the article id in an attribute ~`article-id`~ `data-article-identifier` to aid testing, debugging and INCA.

### DIFF
--- a/packages/article-skeleton/src/article-skeleton.web.js
+++ b/packages/article-skeleton/src/article-skeleton.web.js
@@ -95,10 +95,10 @@ class ArticleSkeleton extends Component {
     return (
       <StickyProvider>
         <article
+          data-article-identifier={article.id}
           ref={node => {
             this.node = node;
           }}
-          article-id={article.id}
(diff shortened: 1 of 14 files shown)


packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -98,6 +98,7 @@ class ArticleSkeleton extends Component {
           ref={node => {
             this.node = node;
           }}
+          article-id={article.id}
```
chrishutchinson: Can we use the same syntax as the current article HTML?

```html
<article data-article-identifier="..." />
```

aronshamash: yes sure. then no INCA change ?

chrishutchinson: Basically yes, I'm hoping no change in INCA, or in our other debugging extension

PR #2368, opened 2019-11-04
feat: section colouring of components from tiles

apostolnikov:
[Story](https://nidigitalsolutions.jira.com/browse/REPLAT-7461)
For more information you can check the related [Spike](https://nidigitalsolutions.jira.com/browse/REPLAT-9647) comments section.

Before: 
image
image

After:
after22
after11

### DIFF
--- a/packages/ssr/src/component/article.js
+++ b/packages/ssr/src/component/article.js
@@ -12,6 +12,10 @@ const {
 const { scales, themeFactory } = require("@times-components/styleguide/rnw");
 
 const scale = scales.large;
+const extractSectionNameFromTiles = article =>
+  article.tiles[0].slices[0].sections[0]
+    ? article.tiles[0].slices[0].sections[0].title.toLowerCase()
+    : "default";
 
 module.exports = (client, analyticsStream, data, helmetContext) => {
   const {
@@ -52,7 +56,10 @@ module.exports = (client, analyticsStream, data, helmetContext) => {
                 makeTopicUrl,
                 newskit: enableNewskit,
                 theme: {
-                  ...themeFactory(article.section, article.template),
+                  ...themeFactory(
+                    extractSectionNameFromTiles(article),
+                    article.template
+                  ),
                   scale: scale || defaults.theme.scale
                 },
                 user: userState


infreezer:
Is this change only for web? 
Are we going to handle the apps in another PR?

aronshamash:
> Is this change only for web?
> Are we going to handle the apps in another PR?

good point

apostolnikov:
> Is this change only for web?
> Are we going to handle the apps in another PR?

Having in mind what we discussed [here](https://nidigitalsolutions.jira.com/browse/REPLAT-9647), my understanding is that this is a temporary solution so we can release react articles(and the release is only for web). 
This solution doesn't cover all the cases and in every edition, we can see articles that have empty tiles arrays on which this logic depends. 

We have a lot of components which theme logic depends on the "section" prop and making such big of a change for a solution that works in 80% of the cases doesn't make sense, in my opinion.  

Bottom line:
1. To resolve this bug quickly(but not fully), so we can get react articles out, this is the safest approach(which can be easily modified and changed in later stage.)

2. To resolve this fully, the TPA team should put some work around the "section" prop and it's falsy nature, and this might take a while.

@infreezer do you have any suggestions for another approach?

infreezer:
> > Is this change only for web?
> > Are we going to handle the apps in another PR?
> 
> Having in mind what we discussed [here](https://nidigitalsolutions.jira.com/browse/REPLAT-9647), my understanding is that this is a temporary solution so we can release react articles(and the release is only for web).
> This solution doesn't cover all the cases and in every edition, we can see articles that have empty tiles arrays on which this logic depends.
> 
> We have a lot of components which theme logic depends on the "section" prop and making such big of a change for a solution that works in 80% of the cases doesn't make sense, in my opinion.
> 
> Bottom line:
> 
> 1. To resolve this bug quickly(but not fully), so we can get react articles out, this is the safest approach(which can be easily modified and changed in later stage.)
> 2. To resolve this fully, the TPA team should put some work around the "section" prop and it's falsy nature, and this might take a while.
> 
> @infreezer do you have any suggestions for another approach?

Can we verify that a temporary solution only for web is the expected result?

apostolnikov:
> > > Is this change only for web?
> > > Are we going to handle the apps in another PR?
> > 
> > 
> > Having in mind what we discussed [here](https://nidigitalsolutions.jira.com/browse/REPLAT-9647), my understanding is that this is a temporary solution so we can release react articles(and the release is only for web).
> > This solution doesn't cover all the cases and in every edition, we can see articles that have empty tiles arrays on which this logic depends.
> > We have a lot of components which theme logic depends on the "section" prop and making such big of a change for a solution that works in 80% of the cases doesn't make sense, in my opinion.
> > Bottom line:
> > 
> > 1. To resolve this bug quickly(but not fully), so we can get react articles out, this is the safest approach(which can be easily modified and changed in later stage.)
> > 2. To resolve this fully, the TPA team should put some work around the "section" prop and it's falsy nature, and this might take a while.
> > 
> > @infreezer do you have any suggestions for another approach?
> 
> Can we verify that a temporary solution only for web is the expected result?

Again, you can read more about in this [spike](https://nidigitalsolutions.jira.com/browse/REPLAT-9647) comments section for what was accepted and agreed upon.

infreezer:
> > > > Is this change only for web?
> > > > Are we going to handle the apps in another PR?
> > > 
> > > 
> > > Having in mind what we discussed [here](https://nidigitalsolutions.jira.com/browse/REPLAT-9647), my understanding is that this is a temporary solution so we can release react articles(and the release is only for web).
> > > This solution doesn't cover all the cases and in every edition, we can see articles that have empty tiles arrays on which this logic depends.
> > > We have a lot of components which theme logic depends on the "section" prop and making such big of a change for a solution that works in 80% of the cases doesn't make sense, in my opinion.
> > > Bottom line:
> > > 
> > > 1. To resolve this bug quickly(but not fully), so we can get react articles out, this is the safest approach(which can be easily modified and changed in later stage.)
> > > 2. To resolve this fully, the TPA team should put some work around the "section" prop and it's falsy nature, and this might take a while.
> > > 
> > > @infreezer do you have any suggestions for another approach?
> > 
> > 
> > Can we verify that a temporary solution only for web is the expected result?
> 
> Again, you can read more about in this [spike](https://nidigitalsolutions.jira.com/browse/REPLAT-9647) comments section for what was accepted and agreed upon.

Let's make sure everyone in the team is aware that the change is web only in order to get react articles out.

PR #2373, opened 2019-11-05
fix: update author structured data

aronshamash:
- change `author` to be an object of type `Person`
- https://developers.google.com/search/docs/data-types/article


RichardHeinz97:
(e.g. “example.com”)

PR #2375, opened 2019-11-06
Fix[REPLAT-10332]:/simplify query

Plamen5kov:
*Problem*: most of the standard sections have similar content

*Solution*: make **each** standard section type request the same info for all it's components

*Next steps*: make **all** standard section types request the same info

*Tested on*: hdpi, hhdpi, mdpi, including phones and tablets

Related issue: https://nidigitalsolutions.jira.com/browse/REPLAT-10332

### DIFF
--- a/packages/edition-slices/src/slices/commentleadandcartoon/index.js
+++ b/packages/edition-slices/src/slices/commentleadandcartoon/index.js
@@ -14,7 +14,9 @@ class CommentLeadAndCartoonSlice extends Component {
   renderSmall(editionBreakpoint) {
     const {
       onPress,
-      slice: { items: [lead, cartoon] }
+      slice: {
+        items: [lead, cartoon]
+      }
     } = this.props;
 
     return (
@@ -29,7 +31,9 @@ class CommentLeadAndCartoonSlice extends Component {
   renderMedium(editionBreakpoint) {
     const {
       onPress,
-      slice: { items: [lead, cartoon] }
+      slice: {
+        items: [lead, cartoon]
+      }
     } = this.props;
 
     return (
--- a/packages/edition-slices/src/slices/commentleadandcartoon/index.js
+++ b/packages/edition-slices/src/slices/commentleadandcartoon/index.js
@@ -14,7 +14,7 @@ class CommentLeadAndCartoonSlice extends Component {
   renderSmall(editionBreakpoint) {
     const {
       onPress,
-      slice: { lead, cartoon }
+      slice: { items: [lead, cartoon] }
     } = this.props;
 
     return (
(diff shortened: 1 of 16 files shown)


packages/provider-queries/src/section_fragment.graphql
```diff
@@ -225,59 +225,31 @@ fragment commonSlicesSection on StandardSectionSlice {
     }
   }
   ... on CommentLeadAndCartoonSlice {
-    lead {
+    items {
       headline
       strapline
       leadAsset {
         ...leadAsset11
+        ...leadAsset32
```
ChristianGarcia: Wouldn't the response size of this be unnecessary larger? With this new query, we'd be querying the Lead article's lead Asset on a 3:2 ratio, without actually using it. This is true, in fact, for any other item in `items`.

I'm a bit concerned about query responses growing too large as we had issues in the past with the Aws lambdas not managing to pass down the responses due to their size being too large.

CCing @peterstarling

packages/edition-slices/src/slices/commentleadandcartoon/index.js
```diff
@@ -14,7 +14,9 @@ class CommentLeadAndCartoonSlice extends Component {
   renderSmall(editionBreakpoint) {
     const {
       onPress,
-      slice: { lead, cartoon }
+      slice: {
+        items: [lead, cartoon]
```
ChristianGarcia: I know very little about JS syntax, but is this assuming the items in the response JSON's `slice.items` entry would be sorted as `lead`, `cartoon`?

IIUC, there's no guarantee from TPA that those items are sorted in that order. Am I right @peterstarling ?

Plamen5kov: If they aren't ordered I think the native query wouldn't be working either

ChristianGarcia: Out of curiosity, why wouldn't the previous one work? On the previous query, the slice items are identified by their key (`lead`, `cartoon`) and not by their position in `items`.

Is it the JS part what wouldn't work, then? Again, my lack of JS knowledge may be failing me here 😅 , but I assumed `slice: {lead, cartoon}` doesn't care about slice items order but only about their keys.

Plamen5kov: it would work, but the query is very custom, I'm trying to abstract it to be simpler to be easier to maintain and smaller.

packages/provider-queries/src/section_fragment.graphql
Plamen5kov: @ChristianGarcia keep in mind this is only the first step in the process. The end goal is to require only `1:1` crops for images and media and that they are further handled on the client device as is the case for the native query. I intentionally don't want to do the whole refactoring at once, because it's error prone.

PR #2377, opened 2019-11-08
fix: tighten up section name from tiles validation

aronshamash:
- We needed to add additional checks for getting section name from tiles arrays

### DIFF
--- a/packages/ssr/src/lib/section-from-tiles.js
+++ b/packages/ssr/src/lib/section-from-tiles.js
@@ -1,5 +1,6 @@
 module.exports = article =>
   article &&
+  article.tiles &&
   article.tiles.length > 0 &&
   article.tiles[0].slices &&
   article.tiles[0].slices.length > 0 &&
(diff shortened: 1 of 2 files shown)


packages/ssr/src/lib/section-from-tiles.js
```diff
@@ -0,0 +1,10 @@
+export default article =>
+  article &&
+  article.tiles.length > 0 &&
+  article.tiles[0].slices &&
```
infreezer: Is the default value of tiles an empty array
Can we have an article without tiles

aronshamash: maybe yes. I don't know for sure but this will work if it is not present

infreezer: If tiles is undefined .length will throw an error

aronshamash: good point. updated

PR #2378, opened 2019-11-11
fix: (REPLAT-10300) performance regression fixes, fixes from review

jmars:
Fixes:
Square images on author pages (REPLAT-10446)
Performance regressions on articles
Performance regressions on ram-bundle config
Image loading spinning T
Adds an activity indicator for the case that scrolling is faster than rendering
Break up paragraph into multiple components for readability


packages/typeset/src/LayoutManager.ts
```diff
@@ -32,170 +27,43 @@ const getSettings = (str: AttributedString) => {
...
-      if (w.string === '\n') {
-        items.push({
-          shrink: 0,
-          stretch: MAX_COST,
-          text: new AttributedString('', []),
-          type: 'glue',
+  for (const split of splits) {
+    if (isSpace(split)) {
+      if (split.string === '\n') {
+        yield {
```
ValGeorgiev: can we not use yield anywhere :D

packages/fixture-generator/package.json
```diff
@@ -50,7 +50,7 @@
     "jest": "24.8.0",
     "prettier": "1.14.3",
     "ts-jest": "23.10.4",
-    "typescript": "3.5.3",
+    "typescript": "3.7.2",
```
parkjh: very nice!

packages/typeset/src/LayoutManager.ts
```diff
@@ -32,170 +27,43 @@ const getSettings = (str: AttributedString) => {
...
-
-function layoutItemsFromString(
+function* layoutItemsFromString(
   s: AttributedString,
-  measureFn: (word: AttributedString) => number,
-  hyphenateFn?: (word: AttributedString) => AttributedString[]
-): TexItem[] {
-  const items: TexItem[] = [];
+  measureFn: (word: AttributedString) => number
+): Generator<MeasuredItem> {
```
parkjh: noice!

parkjh:
> packages/typeset/src/LayoutManager.ts needs refactoring, using yield and nested while loops will cause future issues and hard maintenance. Otherwise all goods

@ValGeorgiev  is there a bug with the implementation, or do you just mean in general generators should not be used?

PR #2380, opened 2019-11-12
fix: render watch command

ValGeorgiev:
@jmars please have a look, we have a watch command in all TC packages, we can use watch:tests for new packages to watch tests


packages/responsive-image/package.json
```diff
@@ -26,7 +26,7 @@
     "test": "jest",
     "test:lint": "tslint --project . && prettier \"src/**/*.ts\" --list-different",
     "test:unit": "jest --coverage",
-    "watch": "run-s clean build:main && run-p \"build:main -- -w\" \"test:unit -- --watch\"",
+    "watch:tests": "run-s clean build:main && run-p \"build:main -- -w\" \"test:unit -- --watch\"",
```
infreezer: `build:main` command not found.
Should we use `build:module` instead?

ValGeorgiev: No idea, I will give it a try

PR #2385, opened 2019-11-14
feat: add article image lazy loading

ValGeorgiev:
Add article image lazy loading


packages/date-publication/package.json
```diff
@@ -40,7 +40,7 @@
     "@babel/core": "7.4.4",
     "@times-components/eslint-config-thetimes": "0.8.16",
     "@times-components/jest-configurator": "2.6.10",
-    "@times-components/storybook": "4.1.6",
+    "@times-components/storybook": "4.1.7",
```
aronshamash: why all of these version bumps in this PR ?

ValGeorgiev: build failed, because of that. This is the result of deps:fix

PR #2387, opened 2019-11-15
feat: REPLAT-10475 draft article query and provider created

infreezer:
[Story.](https://nidigitalsolutions.jira.com/browse/REPLAT-10475)

Draft article query created.
Draft article provider created.
isPreview boolean consumed in ssr.

image

### DIFF
--- a/packages/ssr/src/component/article.js
+++ b/packages/ssr/src/component/article.js
@@ -45,8 +45,11 @@ module.exports = (client, analyticsStream, data, helmetContext) => {
           debounceTimeMs,
           id: articleId
         },
-        ({ article, draftArticle, isLoading, error, refetch }) => {
-          const articleData = isPreview ? draftArticle : article;
+        providerData => {
+          const { isLoading, error, refetch } = providerData;
+          const article = isPreview
+            ? providerData.draftArticle
+            : providerData.article;
 
           return React.createElement(
             ContextProviderWithDefaults,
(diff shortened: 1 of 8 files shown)


packages/ssr/src/component/article.js
```diff
@@ -37,14 +39,16 @@ module.exports = (client, analyticsStream, data, helmetContext) => {
...
-        ArticleProvider,
+        isPreview ? DraftArticleProvider : ArticleProvider,
         {
           analyticsStream,
           debounceTimeMs,
           id: articleId
         },
-        ({ article, isLoading, error, refetch }) =>
-          React.createElement(
+        ({ article, draftArticle, isLoading, error, refetch }) => {
```
aronshamash: hmm not sure I understand why we need this ?

packages/ssr/src/component/article.js
```diff
@@ -37,14 +39,16 @@ module.exports = (client, analyticsStream, data, helmetContext) => {
...
+          const articleData = isPreview ? draftArticle : article;
```
apostolnikov: I don't think that it's good idea to always have the 2 fields from the two providers as destruct parameters, as one of them will be missing all the times. Could we use this articleData logic at upper level so we can have it inside the destructuring as well?

packages/ssr/src/component/article.js
infreezer: When the draft article provider is used the data is returned in draftArticle variable

packages/ssr/src/component/article.js
aronshamash: yes my thoughts too.

infreezer: Comments addressed. Please review.

PR #2389, opened 2019-11-19
Fix/replat 7221 header ads missing firefox

mikefieldmay:
Firefox and Safari header ads weren't loading for React articles. 
Asynchronous promises with the Amazon ads were causing issues.

Tested running local copy of Times Components in Render


mikefieldmay:
@ValGeorgiev Chrome is fine as well.
I'm not sure why, but chrome always worked. It must be something to do with how the different browsers handle promises.

PR #2392, opened 2019-11-20
feat: design changes for ios release

jmars:
Some design changes that were requested by the editorial team ahead of the iOS release.
See:
https://nidigitalsolutions.jira.com/browse/REPLAT-8119
https://nidigitalsolutions.jira.com/browse/REPLAT-10181
https://nidigitalsolutions.jira.com/browse/REPLAT-10562
https://nidigitalsolutions.jira.com/browse/REPLAT-10633
https://nidigitalsolutions.jira.com/browse/REPLAT-10636
https://nidigitalsolutions.jira.com/browse/REPLAT-10568
https://nidigitalsolutions.jira.com/browse/REPLAT-10566

I have tried to avoid any changes touching web articles in order not to effect the ongoing work there, however slices has no web overrides so I have not added any there.


packages/edition-slices/src/tiles/tile-e/index.js
```diff
@@ -30,9 +40,11 @@ const TileE = ({ onPress, tile, breakpoint = editionBreakpoints.small }) => {
         style={styles.imageContainer}
         uri={crop.url}
         fill
+        hasVideo={hasVideo}
       />
       <TileSummary
         headlineStyle={styles.headline}
+        summary={getTileSummary(tile, withMoreTeaser ? 800 : 125)}
```
infreezer:
```suggestion
        summary={getTileSummary(tile, 125)}
```
We aren't using TileE with `withMoreTeaser`, so we don't need that tile to support it.

packages/edition-slices/src/tiles/tile-p/index.js
```diff
@@ -31,9 +32,8 @@ const TileP = ({ onPress, tile }) => {
         rounded
         resizeMode="cover"
       />
+      <Byline ast={tile.article.bylines} bylineStyle={styles.bylineOpinion} />
```
infreezer: According to the [story](https://nidigitalsolutions.jira.com/browse/REPLAT-10633): `Journalist Name must be in All Caps`
image

packages/edition-slices/src/tiles/tile-e/index.js
jmars: Is TileE not the one mentioned in the ticket? https://nidigitalsolutions.jira.com/browse/REPLAT-10562 , or should it only have a summary on mobile?

infreezer: It is the one. We just don't need withMoreTeaser functionality. 
For tablet we are using TileE for [support1](https://github.com/newsuk/times-components/blob/master/packages/provider-queries/src/section_fragment.graphql#L458).
Support1 doesn't have summary in the query. So there is no summary for TileE for tablet.
image

jmars: Sorry I misunderstood what you meant here but that makes sense, it doesn't ever have extended teaser content so there is no need for the prop or check. Updated the PR. :)

PR #2393, opened 2019-11-20
fix: related articles fix for old ios

ValGeorgiev:
Fix related articles for old safaries

Before:
image

After: 
image


infreezer:
Work in progress:
LeadOneAndTwoSlice
image

infreezer:
Let's make sure the related articles styling is unchanged for modern browsers. This can be easily tested in [storybook](http://components.thetimes.co.uk/?selectedKind=Composed%2FRelatedArticles&selectedStory=Lead%20and%20two%20template%20with%20three%20related%20articles&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs).

packages/card/src/card-content.js
```diff
@@ -6,14 +7,53 @@ import { cardPropTypes, cardDefaultProps } from "./card-prop-types";
 import Loading from "./card-loading";
 import styles from "./styles";
 
+const checkBrowser = () => {
```
aronshamash: Just checking this file is not used for native?

ValGeorgiev: refactored, now its not used

packages/card/src/card-content.web.js
```diff
@@ -0,0 +1,145 @@
...
+    const { isOldSafari } = this.state;
+
+    const renderImage = isTablet => {
+      if (!showImage) return null;
+
+      return (
+        <View
+          className={imageContainerClass}
+          style={[
+            isTablet ? styles.imageContainerTablet : styles.imageContainer,
```
infreezer: I think `isTablet` is always "false" for web. 
Do you think it will be better if it removed from here?

infreezer: The only pros of keeping it is that the card-content.web.js and card-content.js are synced. 
But it might create a confusion that isTablet is used in web.
What do you think?

ValGeorgiev: Done, removed

PR #2397, opened 2019-11-20
update readme

mikefieldmay:
Testing a slightly different Bitrise config that may run the tests asynchronously


mikefieldmay:
Testing complete

PR #2399, opened 2019-11-22
chore: Add test ids to aid ui testing

aronshamash:
- `save-and-share-bar` component doesn't have any class names or id in order help UI testing such as Cypress
- As part of https://github.com/newsuk/cps-content-render/pull/3101 where we need to convert some Cypress tests which were using the non-react dom structure.


apostolnikov:
I wonder why we got coverage decreased on this one...

PR #2400, opened 2019-11-25
chore: knobs added to save and share bar

infreezer:
image

### DIFF
--- a/packages/save-and-share-bar/package.json
+++ b/packages/save-and-share-bar/package.json
@@ -44,6 +44,7 @@
     "@times-components/tracking": "2.5.2",
     "@times-components/user-state": "0.1.5",
     "@times-components/utils": "5.1.9",
+    "@storybook/addon-knobs": "4.1.18",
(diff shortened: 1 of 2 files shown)


aronshamash:
@infreezer do we need to add any more snapshots for these scenarios too ?

infreezer:
> @infreezer do we need to add any more snapshots for these scenarios too ?

Sure. We can do it in another PR and merge this one.

packages/save-and-share-bar/package.json
```diff
@@ -44,6 +44,7 @@
     "@times-components/tracking": "2.5.2",
     "@times-components/user-state": "0.1.5",
     "@times-components/utils": "5.1.9",
+    "@storybook/addon-knobs": "4.1.18",
```
mikefieldmay: Just to double check, this shouldn't have caused any changes to the yarn.lock should it?

infreezer: No changes in yarn.lock

PR #2401, opened 2019-11-25
fix: remove display:none css styling

mikefieldmay:
Removes the display none styling which was stopping the header ad from being displayed on mobile web.

Have tested locally and everything works.

Header Ad container is only used in web so shouldn't cause an issue in the Native world


aronshamash:
@mikefieldmay looks fine. Just waiting for snaps then will approve and merge

PR #2418, opened 2019-12-04
fix: bookmarks of null on a preview article

apostolnikov:
Bookmarks on a draft article are currently causing error on the CI tests. 
Adding guard for the cases when article.isBookmarked is null or undefined...


aronshamash:
@apostolnikov is there a snapshot test you can add ?

apostolnikov:
> @apostolnikov is there a snapshot test you can add ?

Good point, but I don't think we need any additional snaps since there are no visual diffs. 
This only ensures that when isBookmarked is missing or it's passed as null nothing will break...

image

PR #2420, opened 2019-12-04
chore: fix broken deps

aronshamash:
- Fix broken deps


ValGeorgiev:
Not needed, deps are fixed now

PR #2430, opened 2019-12-09
fix: removed placeholder ad, removed ad loading state from storybook …

Daniel-Georgiev:
This PR relates to this [issue](https://nidigitalsolutions.jira.com/browse/REPLAT-11097).

- Removed placeholder from ad & loading state from storybook showcase.
- Updated snapshots.

It used to be the case where the placeholder rendered first and made the page jump. It can be observed in the video below:
Screen-Recording-2019-12-09-at-16 29 59

Here's the updated version:
Screen-Recording-2019-12-09-at-16 33 06


packages/ad/__tests__/ios/__snapshots__/ad-with-style.ios.test.js.snap
```diff
@@ -33,14 +33,8 @@ Array [
 
 exports[`2. placeholder when isloading 1`] = `
```
aronshamash: now we have no placeholder this test can be removed I believe

packages/ad/__tests__/android/__snapshots__/ad-with-style.android.test.js.snap
```diff
@@ -67,14 +67,8 @@ Array [
 
 exports[`2. placeholder when isloading 1`] = `
```
aronshamash: can remove this test also

Daniel-Georgiev:
@aronshamash Is it a good idea to remove it, since we're still using it as a placeholder in storybook?

aronshamash:
@Daniel-Georgiev It looks like the the placeholder used by ads is only used by ads so it can be removed from storybook too

aronshamash:
@Daniel-Georgiev run `yarn fmt` as seems to be a formatting issue somewhere

packages/ad/__tests__/ios/__snapshots__/ad-with-style.ios.test.js.snap
```diff
@@ -31,16 +31,4 @@ Array [
 ]
 `;
 
-exports[`2. placeholder when isloading 1`] = `
```
aronshamash: @Daniel-Georgiev have we checked with apps team if removing ad placeholder is ok for them too ?

Daniel-Georgiev: removed

aronshamash: do you mean that you have checked and it is fine ? @jmars

packages/ad/src/ad.js
```diff
@@ -5,7 +5,6 @@ import { Platform, View } from "react-native";
 import { screenWidth, ServerClientRender } from "@times-components/utils";
 import { getPrebidSlotConfig, getSlotConfig, prebidConfig } from "./utils";
 import adInit from "./utils/ad-init";
-import AdPlaceholder from "./ad-placeholder";
 import DOMContext from "./dom-context";
 import AdComposer from "./ad-composer";
```
apostolnikov: Do we still need the ad-placeholder.js inside the ad package? We may also need to remove the 
image dependancy inside the ad package. I think this is causing the lint issues...

packages/ad/__tests__/ios/__snapshots__/ad-with-style.ios.test.js.snap
aronshamash: Checked with Kumaran. It is fine to remove.

packages/ad/src/ad.js
aronshamash: yes that needs to go. `ad-placeholder.js` is not needed at all.

Daniel-Georgiev: Confirmed! The issue came from the watermark import.

PR #2431, opened 2019-12-10
fix: various ios

jmars:
Fixes:
* Many spacing issues on labels/flags
* Content image modals
* Interactive sizing
* A bug that caused reloads when a resize event was fired even if it was the same value
* Many image loading/cache issues


infreezer:
I haven't seen the design.
Are the spacings below between labels and flags what we expect?

image

PR #2433, opened 2019-12-10
chore: increment utils version

infreezer:
Manual bump of utils package.
Hopefully this will end the cycle of deps fixes.

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@times-components/utils",
-  "version": "5.1.17",
+  "version": "6.0.0",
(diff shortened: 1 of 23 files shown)


packages/utils/package.json
```diff
@@ -51,7 +51,7 @@
   },
   "dependencies": {
     "@times-components/schema": "0.6.23",
-    "@times-components/styleguide": "3.33.12",
+    "@times-components/styleguide": "3.33.13",
```
ValGeorgiev: The issue is with styleguide, not with utils. Bump styleguide packagr as well

infreezer: [This is what we did the last time.](https://github.com/newsuk/times-components/pull/2334/files)

PR #2434, opened 2019-12-10
feat: add basic implementation for ads

mikefieldmay:
### Description

[REPLAT-11204](https://nidigitalsolutions.jira.com/browse/REPLAT-11204)

Because adverts are currently working for regular articles, but React Articles are buggy, we have decided for go live to allow Render to handle ad loading.
This PR is accompanied by a PR in Times Components. React Article Ads will now just be div's with ID's and render will load the ads in. This PR ads a new script that is used to load the ads in.
I've added it to the static folder and webpack, but there may be a better place for it. 

Just putting this up here to make sure the implementation is okay before I continue


packages/ad/src/ad-container.js
```diff
@@ -0,0 +1,5 @@
+function AdContainer() {
```
aronshamash: maybe you don't need this file as `AdContainer` will only be used by web anyway ?

mikefieldmay: I thought so, but this seemed to be the pattern across other other repos.
I think because the Ad Component loads this in we need a false native version so that it compiles properly

jmars: You can rename this to .web.js and add a dummy component in its place. That is what we do but opposite for native only things.

packages/ad/src/ad-container.web.js
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { View, ViewPropTypes } from "react-native";
+
+import styles from "./styles";
+
+const AdContainer = ({ slotName, style }) => {
+  const adMap = {
+    header: "ad-header",
+    "inline-ad": "ad-article-inline",
```
aronshamash: for coverage of all code paths would we not need to have snapshot with each of the slot names ?

packages/ad/__tests__/web/ad-container.web.test.js
```diff
@@ -0,0 +1,27 @@
...
+  addSerializers(
+    expect,
+    compose(
+      print,
+      minimaliseTransform((value, key) => key === "style"),
+      minimalNativeTransform
+    )
+  );
+
+  it("advert placeholder", () => {
```
aronshamash: think would be better to desc as `header advert`

packages/ssr/__tests__/helpers/article-helper.js
```diff
@@ -68,13 +68,15 @@ const articleTemplateTest = template =>
...
+      cy.get("#ad-article-inline").should("exist");
+
+      cy.get("#ad-pixel").should("exist");
+
+      cy.get("#ad-pixelteads").should("exist");
 
-      cy.get("#header")
-        .get("googleQueryId")
-        .should("not.be.empty");
+      cy.get("#ad-pixelskin").should("exist");
```
mikefieldmay: Modified the tests and checked for every ad @aronshamash

PR #2436, opened 2019-12-12
fix(REPLAT-11036): Wrap sponsored article div to fix issue with commenting

aronshamash:
- Wrap sponsored article div to fix issue with commenting

### DIFF
--- a/packages/article-extras/src/article-extras.web.js
+++ b/packages/article-extras/src/article-extras.web.js
@@ -22,12 +22,14 @@ const ArticleExtras = ({
   spotAccountId,
   topics
 }) => {
-  /* Nativo inserts Sponsored Articles in this div */
-  const sponsoredArticle = (
-    <div id="sponsored-article" key="sponsored-article" />
+  /* Nativo insert Sponsored Articles after the div#sponsored-article element. They are not able to insert directly into that element hence the container div */
+  const sponsoredArticles = (
+    <div id="sponsored-article-container">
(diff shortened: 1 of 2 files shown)


packages/article-extras/src/article-extras.web.js
```diff
@@ -22,12 +22,14 @@ const ArticleExtras = ({
   spotAccountId,
   topics
 }) => {
-  /* Nativo inserts Sponsored Articles in this div */
-  const sponsoredArticle = (
-    <div id="sponsored-article" key="sponsored-article" />
+  /* Nativo insert Sponsored Articles after the div#sponsored-article element. They are not able to insert directly into that element hence the container div */
+  const sponsoredArticles = (
+    <div id="sponsored-article-container">
+      <div id="sponsored-article" key="sponsored-article" />
```
chrishutchinson: Does the `key` need to move to the top level? Or maybe we don't even need it? I'm not really clear on it's purpose here

mikefieldmay: I've noticed a lot of `key`s in TimesComponents as well. Normally with React you only ad a key to things you are iterating through so I thought it may be a React Native thing or something to do with how we test?

aronshamash: Actually I don't think it is needed. will remove.

aronshamash: @mikefieldmay @chrishutchinson I think the key is not needed in this case. I've removed now

PR #2437, opened 2019-12-12
fix: scroll to top on content load

jmars:
A state update for removing the loading spinner was causing the article content to rerender, this PR makes generous use of useMemo and useCallback to make sure that rerender only occurs when the data itself changes.

### DIFF
--- a/packages/ad/src/ad.js
+++ b/packages/ad/src/ad.js
@@ -141,6 +141,10 @@ class Ad extends Component {
   }
 
   render() {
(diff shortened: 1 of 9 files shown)


packages/ad/src/ad.js
```diff
@@ -141,6 +141,10 @@ class Ad extends Component {
   }
 
   render() {
+    const { adConfig: propAdConfig } = this.props;
```
mikefieldmay: I did not know you could de-structure this way

PR #2442, opened 2019-12-17
fix(REPLAT-10511): tabbing issue in share and save bar

aronshamash:
- Email share, copy to clipboard and save button were not focusable


packages/save-star-web/src/save-star-with-tracking.js
```diff
@@ -34,7 +34,11 @@ function SaveStar({
     const { fillColour, strokeColour } = getStyles({ saveStatus: savedStatus });
     return (
       // eslint-disable-next-line jsx-a11y/anchor-is-valid
```
ValGeorgiev: is this eslint disable not needed anymore?

aronshamash: yes still needed. It really should be a `<button>` but that would require quite a hefty refactor...

PR #2444, opened 2019-12-17
fix: (REPLAT-11373, REPLAT-11368, REPLAT-11367, REPLAT-11207)

jmars:
More spacing fixes, missing placeholders and try to make ads init faster


ValGeorgiev:
We have a separate logic for web, please ignore my last comment.  I will merge this

PR #2445, opened 2019-12-19
fix: added aria-label and removed the brightcove player

Daniel-Georgiev:
This PR is for the following [issue](https://nidigitalsolutions.jira.com/browse/REPLAT-10513).
We used to have '[title]' as a placeholder on aria-labels or no aria-labels at all.
Here are screenshots:
Screenshot at Dec 19 16-46-04
Screenshot at Dec 19 17-05-52


aronshamash:
@Daniel-Georgiev Can we have them in 2 separate PRs please.

PR #2448, opened 2019-12-20
chore: removed brightcove package and dependencies

Daniel-Georgiev:
Removed brightcove player and it's dependencies


packages/article-magazine-standard/__tests__/web/__snapshots__/semantic.web.test.js.snap
```diff
@@ -138,7 +138,7 @@ exports[`1. a full article 1`] = `
       target="_blank"
     >
       <svg
-        aria-label="[title]"
+        aria-label="icon-twitter"
```
aronshamash: I think this is for a different PR no ?

Daniel-Georgiev: Yep, it's not meant to go inside this one, will revert once I have the chance

Daniel-Georgiev: Reverted.

aronshamash:
should it be a `BREAKING CHANGE` @ValGeorgiev ?

ValGeorgiev:
> should it be a `BREAKING CHANGE` @ValGeorgiev ?

I think the title of the PR was not correct, we are not deprecating brightcove, we are removing not used package `brightcove-video`. We are still using brightcove, but in video package directly. There is no need of this package and its not used. BREAKING CHANGE will bump all packages that are using brigthcove-video package with 1 major version. Currently there are no packages that are using brigthcove-video (only several mocks, but these are not dependancies), so I don't think this is needed in that case. I will merge it

PR #2451, opened 2020-01-03
feat: REPLAT-11397 drop cap starting with quote for web

infreezer:
Capitalising the first two characters if the first one is:
image

Before:
image

After:
image

### DIFF
--- a/packages/article-skeleton/src/dropcap-util-common.js
+++ b/packages/article-skeleton/src/dropcap-util-common.js
@@ -1,4 +1,6 @@
 /* eslint-disable no-param-reassign */
+import get from "lodash.get";
+
 const templateWithDropCaps = [
   "indepth",
   "maincomment",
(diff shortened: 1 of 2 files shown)


infreezer:
Please advise if the idea to detect the first letter or digit and capitalize everything before it is okay.

packages/article-skeleton/src/dropcap-util-common.js
```diff
@@ -12,6 +12,13 @@ const splitNode = node => {
     return node;
   }
   if (children[0].name === "text") {
+    const firstLetterOrDigit = children[0].attributes.value.match(
+      "[a-zA-Z0-9]"
+    );
+    const firstLetterOrDigitIndex = firstLetterOrDigit
```
apostolnikov: Are we sure that the drop cap-ed word can only start with letter or a digit?

packages/article-skeleton/src/dropcap-util-common.js
```diff
@@ -22,15 +29,20 @@ const splitNode = node => {
           ...children[0],
           attributes: {
             ...children[0].attributes,
-            value: children[0].attributes.value.slice(0, 1),
+            value: children[0].attributes.value.slice(
+              0,
+              firstLetterOrDigitIndex + 1
+            ),
```
apostolnikov: We might want to compute the plus one adding where the constant is created, so we can avoid doing the operation twice.

apostolnikov:
> Please advise if the idea to detect the first letter or digit and capitalize everything before it is okay.

I think the approach is good, the only thing that we need to check is whether or not we can have some symbol instead of digit or letter.

packages/article-skeleton/src/dropcap-util-common.js
infreezer: Let's check this with Kumaran.

packages/article-skeleton/src/dropcap-util-common.js
infreezer: Please verify.

PR #2454, opened 2020-01-08
feat: update react dependencies

jmars:
Keep our react deps up to date so we don't have a huge upgrade later.


lib/cli/package-json.js
```diff
@@ -53,7 +53,7 @@ module.exports = variables => ({
     prettier: "1.14.3",
     react: "16.6.3",
     "react-dom": "16.9.0",
```
ValGeorgiev: Can we bump react dependencies here as well. We need to bump react-dom with react as well, same with react-test-renderer

aronshamash:
@ValGeorgiev think we should update this PR with latest dep versions as part of our perf work ?

ValGeorgiev:
> @ValGeorgiev think we should update this PR with latest dep versions as part of our perf work ?

totally agree, it's in my to do list.

ValGeorgiev:
I created a new PR for bumping react dependancies. I will close this one

PR #1931, opened 2019-05-10
chore: react native 0.59.8

tuncaulubilge:
Bumping from React Native 0.55.4 to latest (0.59.8)

Tasks:
- [x] Bump RN, Jest, React, React-Native-Web, react-native-test-renderer to latest
- [x] Bump Babel-Jest
- [x] Update metro configs for app builds
- [x] Update invalid accessibilityRoles as those are throwing errors on latest react-native
- [x] Update gradle configuration on the android-app library bundle
- [x] Update gradle configuration on the android storybook bundle
- [x] Update metro configs for native storybooks
- [x] Fix React-Native-Web issues
- [x] Fix ios storybook issues 
- [ ] Integrate to the existing ios app (Need iOS dev assistance on this one)
- [x] Fix broken tests (Jest update has some breaking changes, this might take time)

PR #2470, opened 2020-01-10
fix: broken image urls

aronshamash:
- https://nidigitalsolutions.jira.com/browse/REPLAT-11690
- Drastically improves image loading on the web.

Image urls on react pages were far too large because the `resize` param was not being correctly added to the url for the initial server rendered image before client-side lazy loading kicks in. This meant on the server the image requested was the full dimension uploaded to Methode which is some cases is a big as 4000 pixels. See below:

Screenshot 2020-01-09 at 17 08 14

`appendToURL.web.js` was created in this PR in May 2019: https://github.com/newsuk/times-components/pull/1931. 

As `new url()` doesn't work on nodejs (only on the client) the server side rendered images on the web do not get the lowRes size. I have completely removed the `appendToURL.web.js` file as I don't see a good reason to have 2 different implementations.

This affects all images on topic, author and article pages so needs to be throughly testing.


aronshamash:
> Is there a commit missing? I only see snapshot and test changes.

`packages/image/src/utils/appendToURL.web.js` deleted

PR #2472, opened 2020-01-10
feature: remove ads from react web pages

mikefieldmay:
This is a refactor of all our web components to no longer call any ads. 
All the web components will now do is return divs and everything else will be handled by the ad library in Render itself.

The vast majority of changes are just removing references to the adConfig in the .web files


aronshamash:
@mikefieldmay this needs a rebase

mikefieldmay:
Closing this temporarily as it needs to go in the same time as the ad library

PR #2473, opened 2020-01-10
fix: remove unnecessary version scripts

ValGeorgiev:
Remove unnecessary version scripts from typeset and responsive-image

### DIFF
--- a/packages/responsive-image/package.json
+++ b/packages/responsive-image/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@times-components/responsive-image",
-  "version": "0.4.5",
+  "version": "0.4.4",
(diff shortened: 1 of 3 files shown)


packages/responsive-image/package.json
```diff
@@ -29,7 +29,6 @@
     "watch:tests": "run-s clean build:module && run-p \"build:module -- -w\" \"test:unit -- --watch\"",
     "doc": "run-s doc:html && open-cli build/docs/index.html",
     "doc:publish": "gh-pages -m \"[ci skip] Updates\" -d build/docs",
-    "version": "standard-version",
```
ValGeorgiev: This is done via lerna, there is no need for using standard version

PR #2475, opened 2020-01-13
fix: (REPLAT-9831) update article/section analytics

jmars:

### DIFF
--- a/packages/article-skeleton/src/tracking/article-tracking-context.js
+++ b/packages/article-skeleton/src/tracking/article-tracking-context.js
@@ -28,10 +28,13 @@ export default Component =>
           ""
         )}`,
         edition_type: editionType,
-        published_time: get(data, "publishedTime", ""),
+        published_time: published.toISODate(),
         past_edition_date:
-          editionType === "past 6 days" ? get(data, "publishedTime", "") : null,
-        parent_site: get(data, "publicationName", ""),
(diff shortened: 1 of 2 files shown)


packages/article-skeleton/src/tracking/article-tracking-context.js
```diff
@@ -28,10 +28,13 @@ export default Component =>
...
         edition_type: editionType,
-        published_time: get(data, "publishedTime", ""),
+        published_time: published.toISODate(),
         past_edition_date:
-          editionType === "past 6 days" ? get(data, "publishedTime", "") : null,
-        parent_site: get(data, "publicationName", ""),
+          editionType === "past 6 days" ? published.toISODate() : null,
+        parent_site:
+          get(data, "publicationName", "") === "TIMES"
+            ? "the times"
```
aronshamash: do we not have this mapping of `publicationName` to analytics publication name somewhere already ?

aronshamash:
@jmars needs desc and JIRA link

packages/article-skeleton/src/tracking/article-tracking-context.js
apostolnikov: Is this trenary applicable to the web as well? What's the reason we want the parent_site to be lower case?

ValGeorgiev:
This needs to be confirmed if its needed or no

ValGeorgiev:
Not needed

PR #2480, opened 2020-01-15
fix: (REPLAT-11799) add a lower bound for images on web

jmars:
On mobile we want to display smaller images, the lower bound of 320 means that on a retina display the lower bound is actually 640 which hurts our mobile score.


aronshamash:
@jmars Can you supply the JIRA ticket this relates to please.

jmars:
https://nidigitalsolutions.jira.com/browse/REPLAT-11799

aronshamash:
Closing as we don't want to make this change.

PR #2483, opened 2020-01-17
fix: REPLAT-11812 caption appearing over images

jmars:
Before, captions were over the top of the image, this moves them to the bottom of the modal as per the designs and turns on preloading as people complained about load time in image modals. It also enabled the swipe down to close gesture so that it works the same on iOS and Android.

https://nidigitalsolutions.jira.com/browse/REPLAT-11812
Screenshot_1579280060


aronshamash:
@jmars Please add a desc and link to JIRA.

PR #2484, opened 2020-01-17
fix: REPLAT-11819 fix gaps in urls

jmars:
Screenshot_1579280139


packages/article-skeleton/src/article-body/inline-paragraph.js
```diff
@@ -11,6 +11,12 @@ import {
 } from "@times-components/typeset";
 import ArticleParagraphWrapper from "@times-components/article-paragraph";
 
+const isLink = node => {
```
aronshamash: Can you add unit test for this

packages/article-skeleton/src/article-body/inline-paragraph.js
```diff
@@ -49,7 +55,18 @@ const InlineParagraph = ({
     inlineExclusion ? [inlineExclusion.exclusion] : []
   );
 
-  const positioned = manager.layout();
+  const positioned = manager.layout().reduce((acc, current) => {
```
aronshamash: this should have some tests too

jmars:
No longer needed

PR #2485, opened 2020-01-17
fix: REPLAT-11573 drop caps with a starting quote

jmars:
Screenshot_1579280487


packages/article-skeleton/src/article-body/drop-cap.js
```diff
@@ -19,8 +22,18 @@ export default (scale, color, dropCapFont, paragraph) => {
     color
   };
   const font = FontStorage.getFont(fontSettings);
-  const glyph = font.charToGlyph(letter.string);
-  const { x1, x2, y1, y2 } = glyph.getBoundingBox();
+  const glyphs = font.stringToGlyphs(letter.string);
```
aronshamash: think this code below could be better in a helper function or util no?

aronshamash: then can add unit tests

infreezer:
Is the "T" too close to the text?
image
https://www.thetimes.co.uk/article/written-in-green-ink-the-single-letter-that-altered-a-sport-and-changed-lives-crh3dz5sk?react=1

PR #2486, opened 2020-01-20
fix: REPLAT-11787 use native ios modal for author images

jmars:


aronshamash:
@jmars please add desc to PRs with at least a link to JIRA ticket

PR #2489, opened 2020-01-22
fix(REPLAT-11906): remove keylines below header ad on indepth template

aronshamash:
- https://nidigitalsolutions.jira.com/browse/REPLAT-11906
- Indepth template should not have keyline or padding below header ad


packages/article-skeleton/src/styles/responsive.web.js
```diff
@@ -10,13 +10,13 @@ export const MainContainer = styled(View)`
...
 `;
 
-export const HeaderAdContainer = styled(View)`
+export const getHeaderAdContainer = hideKeylines => styled(View)`
   display: flex;
   border-top-color: ${colours.functional.keyline};
   border-bottom-color: ${colours.functional.keyline};
-  border-bottom-width: 1px;
+  border-bottom-width: ${hideKeylines ? "0" : "1px"};
   padding-top: ${spacing(2)};
```
apostolnikov: Nice one!

packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -28,6 +28,11 @@ const adStyle = {
   marginBottom: 0
 };
 
+const templatesWithoutHeaderAdKeylines = ["indepth"];
+
+const shouldHideHeaderAdKeyline = template =>
+  templatesWithoutHeaderAdKeylines.includes(template);
+
```
apostolnikov: Could we move this logic and the const inside the responsive.web.js because it's only styles related. We could do a compose there and use it here by passing only the template. Also, we can get rid of the array and the .includes operation since the only template for now will be indepth, we can always add the array later if we need it.

packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -76,6 +81,10 @@ class ArticleSkeleton extends Component {
       content.length > 0 &&
       insertDropcapIntoAST(content, template, dropcapsDisabled);
 
+    const HeaderAdContainer = getHeaderAdContainer(
+      shouldHideHeaderAdKeyline(template)
+    );
```
apostolnikov: const HeaderAdConteiner = getHeaderStyles(template); The function should be able to resolve the styles only by using the passed template...

packages/article-skeleton/src/styles/responsive.web.js
```diff
@@ -3,22 +3,26 @@ import styled from "styled-components";
 import { breakpoints, colours, spacing } from "@times-components/styleguide";
 import { ArticleKeylineItem } from "../keylines";
 
+export const getHeaderAdStyles = template => {
+  const hideKeylines = template === "indepth";
+
```
apostolnikov: Good job! 👍

PR #2490, opened 2020-01-22
fix: REPLAT-11684 handle responsive-graphic specially on ios

jmars:
This is for https://nidigitalsolutions.jira.com/browse/REPLAT-11684 again.
Reloading on errors did not fix the issue and it appears to be something about ios webviews. This adds a small component which does what the interactive does except without using webviews (as recommended by Chris Hutchinson).


packages/interactive-wrapper/src/responsive-image.ios.js
```diff
@@ -0,0 +1,72 @@
...
+  componentDidMount() {
+    const { deckId } = this.props;
+    if (deckId) {
+      this.getDeck(deckId);
+    }
+  }
+
+  async getDeck(id) {
+    // eslint-disable-next-line no-undef
+    const res = await fetch(deckUrl(id), {
```
ValGeorgiev: Do we have fetch on iOS?

packages/interactive-wrapper/src/responsive-image.ios.js
```diff
@@ -0,0 +1,72 @@
...
+        "Content-Type": "application/json"
+      }
+    });
+    const {
+      body: { data }
+    } = await res.json();
+    const images = data
+      .filter(dat => dat.type === "image")
+      .map(d => d.data)
+      .map(d => ({
```
ValGeorgiev: we can combine these maps in one

packages/interactive-wrapper/src/responsive-image.ios.js
jmars: Yep fetch is available on ios in the version of JSC that RN uses :)

packages/interactive-wrapper/src/responsive-image.ios.js
jmars: Good point :)

PR #2494, opened 2020-01-23
fix: sticky share and save bar on preview article

apostolnikov:
[Jira](https://nidigitalsolutions.jira.com/browse/REPLAT-11949)

Implement fix for the sticky share and save bar position on preview article.
Moving the Article Meta Banner Container inside TC so the sticky bar position can be calculated correctly. This will also fix the copy link popup.

image
image


packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -87,6 +87,30 @@ class ArticleSkeleton extends Component {
 
     return (
       <StickyProvider>
+        <div className="Container">
```
ValGeorgiev: This should be behind a preview check, isnt it ?

aronshamash:
@ValGeorgiev @apostolnikov there is alot of repetition in the snapshots. Maybe we need to stub the preview placeholder in all but 1 package ?

apostolnikov:
> @ValGeorgiev @apostolnikov there is alot of repetition in the snapshots. Maybe we need to stub the preview placeholder in all but 1 package ?

@aronshamash  Where do you think it's good idea for the placeholder to be?

The placeholder is currently inside article-skeleton, which is used by all 5 templates and that's why we have it inside those snaps.

ValGeorgiev:
> > @ValGeorgiev @apostolnikov there is alot of repetition in the snapshots. Maybe we need to stub the preview placeholder in all but 1 package ?
> 
> @aronshamash Where do you think it's good idea for the placeholder to be?
> 
> The placeholder is currently inside article-skeleton, which is used by all 5 templates and that's why we have it inside those snaps.

We can refactor later, we have a lot of repetition in all templates

packages/article-skeleton/__tests__/web/article-with-isPreview.web.test.js
```diff
@@ -0,0 +1,333 @@
...
+    const output = TestRenderer.create(renderArticle());
+
+    expect(output).toMatchSnapshot();
+  });
+
+  it("Render article without ArticleMetaBanner when isPreview is missing", () => {
+    isPreview = false;
+    const output = TestRenderer.create(renderArticle());
+
+    expect(output).toMatchSnapshot();
```
ValGeorgiev: Could you add these tests in the other file and you can pass isPreview to renderArticle function. Just to avoid repetition

PR #2498, opened 2020-01-28
fix: Date publication and publication name

ValGeorgiev:
Previous: 
image

Now: 
image

### DIFF
--- a/packages/date-publication/src/date-publication.web.js
+++ b/packages/date-publication/src/date-publication.web.js
@@ -24,8 +24,8 @@ const DatePublication = ({ publication, ...props }) => (
   </DateTime>
 );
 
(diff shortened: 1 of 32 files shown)


packages/date-publication/src/date-publication.web.js
```diff
@@ -1,19 +1,33 @@
...
+        </time>
+        {publication ? (
+          <PublicationName>{getPublicationName(publication)}</PublicationName>
+        ) : null}
       </Fragment>
     )}
   </DateTime>
 );
 
+const PublicationName = styled.p`
```
chrishutchinson: Probably shouldn’t be a `<p>` tag as it isn’t a paragraph. I’d do a block-level span

PR #2503, opened 2020-01-29
feat(REPLAT-12043): improve low res images

aronshamash:
- inline article images - width of `400 `
- related article images - width of `200 `
- article list pages - width of `200 `

and all have new `quality=3` param to keep file size down

As a result the filesize of the low res images have increased from `~3kb` to `~15kb`


infreezer:
I'm unable to see the `quality=3` for the inline article images. Can you check?
image

packages/card/__tests__/android/__snapshots__/card.android.test.js.snap
```diff
@@ -213,9 +220,10 @@ exports[`11. card should not re-render when imageratio prop is changed 2`] = `
 <View>
   <View>
     <Image
-      aspectRatio={0.6666666666666666}
+      aspectRatio={1.7777777777777777}
```
infreezer: Any ideas why the value is changed?

aronshamash: good spot! That caught a bug which I have now fixed

PR #2506, opened 2020-01-30
feat: lead asset image optimization

apostolnikov:
[Jira](https://nidigitalsolutions.jira.com/browse/REPLAT-12042)

Added Lead asset image optimisation, so it is server side rendered with shorter loading time.


packages/article-lead-asset/src/article-lead-asset-prop-types.js
```diff
@@ -16,6 +16,13 @@ const videoLeadAssetPropTypes = {
   skySports: PropTypes.bool
 };
 
+export const imageLeadAssetPropTypes = {
```
aronshamash: we have a `leadAssetImagePropTypes` and `imageLeadAssetPropTypes` in same file. a bit confusing maybe ?

aronshamash: would a `prop-types.web.js` file help here?

apostolnikov: changes applied

PR #2508, opened 2020-01-31
chore: fix deps and upgrade to node v10.13.0

aronshamash:
- fix broken deps 
- upgrade to node v10.13.0


infreezer:
Is the deletion of yarn.lock intentional?

aronshamash:
> Is the deletion of yarn.lock intentional?

no. I'll push it back.

PR #2514, opened 2020-02-04
fix: section name for analytics

aronshamash:
- https://nidigitalsolutions.jira.com/browse/REPLAT-12182
- The section name we are sending to analytics is unreliable / incorrect because we use the `section` field from `Article` in TPA response.
- Until TPA can give us this data in a "nicer" way we have to look into the `tiles` array and find it there instead. This is how we do it for the `<title>` data anyway.
- This currently always "news" in the Scotland and Ireland section.

### Scotland inside current edition

Screenshot 2020-02-04 at 11 05 16

### Scotland outside current edition

Screenshot 2020-02-04 at 11 03 53


packages/ssr/src/lib/get-analytics-section.js
```diff
@@ -0,0 +1,29 @@
...
+
+  const titles = sections.map(section => section.title);
+
+  if (titles.length === 0) {
+    return null;
+  }
+
+  const nonNewsSection = titles.filter(title => title !== "news");
+
+  return nonNewsSection.length ? nonNewsSection[0].toLowerCase() : "news";
```
ValGeorgiev: can we have some unit tests for this file

aronshamash: well  I did think of that but `ssr` doesn't have any tests.

aronshamash: option  _maybe_ to put it in `utils` package ?

ValGeorgiev: yes, we can move it to utils package

aronshamash: I'll create unit tests there

packages/ssr/src/lib/get-analytics-section.js
```diff
@@ -0,0 +1,29 @@
+module.exports = article => {
+  // Get the section for an article, preferring it not to be news
+
```
apostolnikov: exports { tiles } =>

packages/ssr/src/lib/get-analytics-section.js
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  const slices = tiles.reduce((acc, tile) => {
+    acc.push(...tile.slices);
+    return acc;
+  }, []);
+
+  const sections = slices.reduce((acc, slice) => {
+    acc.push(...slice.sections);
+    return acc;
```
apostolnikov: this one and the one above could use implicit return 
const sections = slices.reduce((acc, slice) => acc.push(...slice.sections), []);

packages/ssr/src/lib/get-analytics-section.js
aronshamash: done

packages/ssr/src/lib/get-analytics-section.js
aronshamash: yes that seems good idea, but when I tried it tests all fail. Going to merge as it is as it is fairly urgent  fix

PR #2515, opened 2020-02-04
feat: add srcSet to lead-asset-image

apostolnikov:
[Jira](https://nidigitalsolutions.jira.com/browse/REPLAT-12059)

Srcset added to the lead asset image. 
Article Skeleton refactored into functional component since there is no longer need for it to calculate the width of the article and pass it down to the lead asset image.
This is also fixing the issue caught by @chrishutchinson:
Image from iOS
With this pr, there will be only one request for a lead asset image.
Result: blazingly fast lead asset images, image loads faster or equal to the article text.

Additional notes:
Srcset cannot be added to the image component as adding it, will skip all the lazy loading logic that we currently have and make it useless.


packages/article-lead-asset/src/article-lead-asset-image.web.js
```diff
@@ -13,7 +13,16 @@ const LeadAssetImage = ({ aspectRatio, alt, uri, size }) => {
...
+        src={appendToImageURL(url, "resize", 1180)}
+        srcSet={`${appendToImageURL(
+          url,
+          "resize",
+          320
+        )} 320w, ${appendToImageURL(
+          url,
+          "resize",
+          685
+        )} 685w, ${appendToImageURL(url, "resize", 1180)} 1180w`}
```
aronshamash: I think we are going to want an additional bigger size for retina desktop screens.

apostolnikov: Good catch, we already have 2400width one, on the legacy template. The problem was, that the srcset was forcing the image to get the biggest size, even on my 13' inch laptop.

apostolnikov: So we'll end up getting 2400width images for basically all desktop requests, which is double the size of the 1180width image... I'm open to suggestions on this one

aronshamash: maybe we would have to use media queries and set the `display` property on the image eg: 
```
@media 
(-webkit-min-device-pixel-ratio: 2), 
(min-resolution: 192dpi) {
  .low-res {
    display: none;
  }

  .high-res {
    display: block;
  }
}
```

aronshamash: I think what you have here is a big improvement. Let's get it shipped and we can always iterate.

PR #2524, opened 2020-02-07
feat(TNLT-193): added newSoureData to headers

Gujibell:


Gujibell:
> Do we need the same headers and logic for the server requests [here](https://github.com/newsuk/times-components/blob/0e5e728746510af2d08ad636caf37d4804374827/packages/ssr/src/server/topic.js#L41).

We are passing the header from Render to Times Components for SSR, but are unable to do it for client side.

PR #2525, opened 2020-02-10
fix: REPLAT-12259 Show lead asset captions for mobile native apps 

infreezer:
[REPLAT-12259](https://nidigitalsolutions.jira.com/browse/REPLAT-12259)

Before:
mobile-caption-before

After:
mobile-caption-after

### DIFF
--- a/packages/article-main-standard/__tests__/android/__snapshots__/article-with-style.android.test.js.snap
+++ b/packages/article-main-standard/__tests__/android/__snapshots__/article-with-style.android.test.js.snap
@@ -12,6 +12,15 @@ exports[`full article with style on mobile 1`] = `
         }
       >
         <ModalImage />
+        <Caption
+          style={
+            Object {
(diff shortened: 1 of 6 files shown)


packages/article-main-standard/__tests__/android/__snapshots__/article-with-style.android.test.js.snap
```diff
@@ -144,6 +153,15 @@ exports[`full article with style on tablet 1`] = `
...
         <ModalImage />
+        <Caption
+          style={
+            Object {
+              "container": Object {
+                "paddingHorizontal": 10,
+              },
+            }
+          }
+        />
```
infreezer: I believe `setIsTablet(true);` in the tests is not working as expected.
Here is proof that the tablet caption styling is unchanged:

tablet-caption-unchanged

PR #2527, opened 2020-02-11
fix (REPLAT-12224): fix image urls with special characters

aronshamash:
- add unit tests to `getImageUrl`
- disable character encoding


aronshamash:
> We should probably leave the previous name of the function as it will just append something to a URL that we are passing, it's a bit confusing with the new name.

@apostolnikov done.

PR #2529, opened 2020-02-11
copy correct json

jmars:


ValGeorgiev:
@newsuk-lee-daffen  could you confirm if this is needed?

newsuk-lee-daffen:
> @newsuk-lee-daffen could you confirm if this is needed?

No, I'll close this PR

PR #2532, opened 2020-02-13
fix: REPLAT-12031 'and' keyword bylines handling

infreezer:
'and' word handling for main standard template on huge and wide breakpoints.

Before:
image

After:
and-bylines-after

### DIFF
--- a/packages/article-byline/__tests__/shared.base.web.js
+++ b/packages/article-byline/__tests__/shared.base.web.js
@@ -154,6 +154,46 @@ export default Component => {
 
         expect(testInstance).toMatchSnapshot();
       }
+    },
+    {
+      name: "main standard template - with multiple authors separated by and",
+      test: () => {
+        const testInstance = renderArticleBylineMainStandard({
+          ast: authorsFixture.multipleAuthorsAndSeparated
+        });
+
+        expect(testInstance).toMatchSnapshot();
+      }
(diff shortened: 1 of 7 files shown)


aronshamash:
@infreezer did we get Product/editorial agreement that this is what that want ? see legacy article handling: https://www.thetimes.co.uk/past-six-days/2020-02-09/news/no-deal-emergency-plan-revived-as-coronavirus-threatens-shortages-q7ld3vdkk?legacy=1. Think they might want it to be same as that. ie don't remove the `and` just remove the`,` when there is an `,`  after.

infreezer:
> @infreezer did we get Product/editorial agreement that this is what that want ? see legacy article handling: https://www.thetimes.co.uk/past-six-days/2020-02-09/news/no-deal-emergency-plan-revived-as-coronavirus-threatens-shortages-q7ld3vdkk?legacy=1. Think they might want it to be same as that. ie don't remove the `and` just remove the`,` when there is an `,` after.

Good point. Let me think about the implementation. 
There is a tricky part:
`Name1, and, Name2` in that case both commas should be removed.
Do you have an idea how to detect the 'and' while rendering `Name1,`.

infreezer:
Will be fixed by reverting [#2538](https://github.com/newsuk/times-components/pull/2538).

PR #2537, opened 2020-02-20
feat: allow native to override breakpoints

jmars:
This is so that the native side on android can set all parts of the app into tablet mode when unfolding a folding phone, which is not quite big enough to trigger the tablet breakpoints but is too large for mobile layouts to look reasonable.

### DIFF
--- a/packages/edition-slices/src/edition-slices.js
+++ b/packages/edition-slices/src/edition-slices.js
@@ -1,5 +1,6 @@
 import { getDimensions } from "@times-components/utils";
 import { tabletWidth } from "@times-components/styleguide";
+import { NativeModules } from "react-native";
 import {
   CommentLeadAndCartoonSlice,
   DailyRegisterLeadFourSlice,
@@ -19,8 +20,11 @@ import {
   PuzzleSlice
 } from "./slices";
 
+const config = (NativeModules || {}).ReactConfig;
+
 const { width } = getDimensions();
-const isTablet = width > tabletWidth;
+const isTablet =
+  (config && config.breakpoint !== "small") || width > tabletWidth;
 const SecondaryTwoAndTwoMapper = isTablet
   ? SecondaryTwoNoPicAndTwoSlice
   : SecondaryTwoAndTwoSlice;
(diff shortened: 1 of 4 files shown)


packages/edition-slices/src/edition-slices.js
```diff
@@ -19,8 +20,11 @@ import {
   PuzzleSlice
 } from "./slices";
 
+const config = (NativeModules || {}).ReactConfig;
+
 const { width } = getDimensions();
-const isTablet = width > tabletWidth;
+const isTablet =
+  (config && config.breakpoint !== "small") || width > tabletWidth;
```
ChristianGarcia: IIUC, this is considering the any values from of the `breakpoint` native constant that are not "small" to be tablet. Am I right?

However, in native we've set it up so that:
* small: small phones (mostly old very small phones)
* normal: Current typical phones (most phones nowadays tbh) ~4-6"
* large: tablet ~7"
* huge: tablet ~10"

So this would make phones show the slices as in tablet.
To sum up, `isTablet` should be `true` if breakpoint is either `"large"` of `"huge"`.

Or am I just confusing how you've named it in Times Components?

packages/responsive/src/responsive.js
```diff
@@ -9,12 +9,15 @@ import {
...
+import { NativeModules } from "react-native";
 import ResponsiveContext from "./context";
 
+const config = (NativeModules || {}).ReactConfig;
+
 const calculateState = (width, fontScale) => ({
   editionBreakpoint: getEditionBreakpoint(width),
   fontScale,
-  isTablet: width > tabletWidth,
+  isTablet: (config && config.breakpoint !== "small") || width > tabletWidth,
```
ChristianGarcia: See comment above about tablet vs phone

packages/edition-slices/src/edition-slices.js
ChristianGarcia: Happy to change this in native btw

ChristianGarcia: As discussed offline, we'll change this in native and use the same breakpoint names as Times Components.

PR #2539, opened 2020-02-25
feat: add guide to running ios app for development

jmars:


aronshamash:
Is this additional info from that given here: https://github.com/newsuk/nu-digital-projectd-times-smartphone-ios/ ?

PR #2542, opened 2020-02-27
fix: (TNLT-400) dropcap web handling

infreezer:
https://nidigitalsolutions.jira.com/browse/TNLT-400

Dropcap insertion in the first paragraph on web articles.
Fix bug with duplicated paragraph when the article start with keyfacts or interactive.

Before:
keyfacts-dropcap

After:
keyfacts-dropcap-after


packages/article-skeleton/src/dropcap-util.web.js
```diff
@@ -6,11 +6,24 @@ export default (children, template, isDropcapDisabled) => {
     template,
     isDropcapDisabled
   );
-  if (newContent.length > 0 && newContent[0].name === "dropCap") {
+  if (
+    newContent.length > 0 &&
+    newContent.find(child => child.name === "dropCap")
+  ) {
```
apostolnikov: we can extract the second conditional here as a const, than reuse it again with indexOf instead of doing two find operations

packages/article-skeleton/src/dropcap-util-common.js
```diff
@@ -100,11 +88,14 @@ const insertDropcapIntoAST = (children, template, isDropcapDisabled) => {
...
-      const child = getChild(children);
+      const child = children.find(x => x.name === "paragraph");
 
-      if (!child || child.name !== "paragraph" || child.children.length === 0) {
+      if (!child || child.children.length === 0) {
         return children;
       }
+      const firstParagraphIndex = children.findIndex(
+        x => x.name === "paragraph"
+      );
```
apostolnikov: Doing a find operation is expensive so we should avoid it where it's possible. You are doing it here twice. When you already have the child you can just check for its index in the children array: firstParagraphIndex = children.indexOf(child)

packages/article-skeleton/src/dropcap-util.web.js
infreezer: Please take a look again.

PR #2543, opened 2020-02-27
feat: improve inline images retina indepth article

apostolnikov:
In this PR we are improving the quality of the inline images on retina screens, indepth article template. 
 **Before:** 
image

**After:** 
image

### DIFF
--- a/packages/article-skeleton/src/article-body/article-body.web.js
+++ b/packages/article-skeleton/src/article-body/article-body.web.js
@@ -46,7 +46,22 @@ export const responsiveDisplayWrapper = displayType => {
   }
 };
 
-const renderers = ({ paidContentClassName }) => ({
+const highResSizeCalc = (observed, key, template) => {
+  const screenWidth =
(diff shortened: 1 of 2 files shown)


infreezer:
Do you have an idea why the `highResSize` is null in the article snaps [here](https://github.com/newsuk/times-components/blob/feat/indepth-template-retina-inline-images-quality/packages/article-skeleton/__tests__/web/__snapshots__/article.web.test.js.snap#L81).

apostolnikov:
> Good job.
> I'm curious why the `highResSize` is null in the article snaps [here](https://github.com/newsuk/times-components/blob/feat/indepth-template-retina-inline-images-quality/packages/article-skeleton/__tests__/web/__snapshots__/article.web.test.js.snap#L81)
> Can you check?

Yes, good point. It's because of the passed provider property observed. It will always be null in the snapshots as it depends on a few properties that will only live in the browser, like window.IntersectionObserver. Anyways, this does not affect us at all, putting more effort around that won't give any value for the project.

PR #2544, opened 2020-02-27
fix: (TNLT-634) fix handling of empty paragraphs

jmars:
Fixes a crash on empty paragraphs.


mikefieldmay:
@jmars Looks good, but is the yarn.lock change supposed to be there?

PR #2546, opened 2020-03-04
fix: issue with topics

mikefieldmay:
Topics had prematurely been updated to use v2. This was causing mismatches between the article and the topics pages. This PR reverts everything back to topics v1


aronshamash:
so can update to v2 soon but need to do all pages @mikefieldmay

mikefieldmay:
@aronshamash Yeah I think V2 needs editorial sign off first so it'll probably happen in the next few weeks.

PR #2547, opened 2020-03-05
fix: make sure we show the placeholder while image decode is running

jmars:
Sometimes when we switch from the placeholder to the image the image decode takes a little while and you see just the background. This moves the placeholder over the image, it is removed anyway so will not cover it once its loaded.

### DIFF
--- a/packages/responsive-image/src/index.tsx
+++ b/packages/responsive-image/src/index.tsx
@@ -212,7 +212,7 @@ const ResponsiveImage = (props: ResponsiveImageProps) => {
       fadeDuration={0}
     />
   );
-  const placeholder = showPlaceholder && (
+  const placeholder = (
     <Image
       key="placeholder"
       source={logoPath}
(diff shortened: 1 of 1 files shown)


packages/responsive-image/src/index.tsx
```diff
@@ -229,9 +229,10 @@ const ResponsiveImage = (props: ResponsiveImageProps) => {
 
   return (
     <View style={{ ...styles.style, ...propStyle, aspectRatio, borderRadius }}>
-      {placeholder}
+      {showPlaceholder && placeholder}
       {lowRes}
       {highRes}
+      {!showPlaceholder && placeholder}
```
mikefieldmay: Is this always showing placeholder now? I don't quite think I understand the logic here?

jmars: Basically, react native renders using the painters algorithm, so the top component first, then on top the next and then on top the next. The problem is that when we place the low res image over it, sometimes decode hasnt finished so it covers the placeholder with a background. So what this is doing is putting a placeholder on the top of the stack until that flag changes.

jmars: There are 2 because sometimes we get a load error which falls down to the placeholder and we still want one in that case.

jmars:
Closing as in the end this solution is not suitable

PR #2548, opened 2020-03-05
fix: (TNLT-652) use safe area view for modals

jmars:


packages/image/src/styles/index.ios.js
```diff
@@ -1,5 +1,6 @@
 import { StyleSheet } from "react-native";
 import { colours } from "@times-components/styleguide";
+import spacing from "@times-components/styleguide/dist/spacing";
```
aronshamash: why not `{ colours, spacing }` ?

mikefieldmay: That import does look slightly off

jmars: VSCode auto import fail

PR #2549, opened 2020-03-05
feat: native images add predefined widths

apostolnikov:
[Jira](https://nidigitalsolutions.jira.com/browse/TNLT-338)

It was reported that there are too many requests for different image widths to the image server, this is expensive and it is preventing us from caching the images properly, thus adding pre-defined image widths for native images.

Image widths: [160, 280, 330, 485, 600, 685, 750, 1024, 1138, 1240, 1440, 1600]
Those were taken from a report done by Andy Raines which he shared on this channel: 
https://newsuktechnology.slack.com/archives/C3V2A2SLQ/p1581359619272500

image


packages/responsive-image/src/utils/findClosestWidth.ts
```diff
@@ -0,0 +1,7 @@
+export default (width: number) => {
+  const mostUsedWidths = [160, 280, 330, 485, 600, 685, 750, 1024, 1138, 1240, 1440, 1600];
```
aronshamash: `supportedWidths` maybe ?

apostolnikov: > Looks like you need to adjust the values in the tests, but otherwise 👍

Good point 👍 , forgot to change those because I wrote the tests first and right before creating the branch, I changed the values

Issue #1384, opened 2018-10-05
Renaming of boolean props


aronshamash:
- `video` package has two props (`skySports` and `paidOnly`) that need renaming.
- `skySports` -> `isSkySports`
- `paidOnly` -> `isPaidOnly`

A corresponding change should also happen in TPA for consistency.

ValGeorgiev:
These props will be removed by https://github.com/newsuk/times-components/pull/2555 . I will close this issue

PR #2555, opened 2020-03-10
fix: Remove paidOnly and skySports video attributes

ValGeorgiev:
Issue raised: https://github.com/newsuk/times-components/issues/1384

After discussion with product, these fields are not needed and they can be removed


packages/article-lead-asset/src/article-lead-asset-video.js
```diff
@@ -32,7 +32,7 @@ const ArticleLeadAssetVideo = ({
       onVideoPress={onVideoPress}
       policyKey={brightcovePolicyKey}
       poster={{ uri }}
-      skySports={skySports}
+      isSkySports={skySports}
```
aronshamash: actually this can be removed as we no longer have a contract to be able to show EPL football clips

packages/article-lead-asset/src/article-lead-asset.web.js
```diff
@@ -30,12 +30,12 @@ const ArticleLeadAsset = ({
       height="100%"
       id={leadAsset.id}
       is360={leadAsset.is360}
-      paidOnly={leadAsset.paidOnly}
+      isPaidOnly={leadAsset.paidOnly}
```
aronshamash: same with this

packages/article-lead-asset/src/article-lead-asset-video.js
manosim: Can confirm this as well. We can remove any reference to this.

packages/article-lead-asset/src/article-lead-asset.web.js
aronshamash: @manosim can you confirm that we have no video content that should only be shown to paid subscribers. (In past videos were marked with this to show that video can't be played for RA users, only for paid subscribers)

aronshamash:
Once we have confirmation we can also remove the code from render.

packages/article-lead-asset/src/article-lead-asset.web.js
manosim: Just checked - from a product perspective, there's no such thing as videos available only to subscribers so we can remove this.

ValGeorgiev:
> Once we have confirmation we can also remove the code from render.

Do you need ticket for that? We can discuss it on the standup tomorrow. Otherwise, the PR is updated. Please review

PR #2563, opened 2020-03-16
feat: (TNLT-697) vendors.js bundle extracted from common.js

infreezer:
Dividing `common.react.bundle.js` to `vendor.bundle.js` and `common.react.bundle.js` to benefit from client level caching


ValGeorgiev:
this needs a change in render as well, when you are bumping times-components packages

PR #2568, opened 2020-03-17
feat: (TNLT-697) remove unused slices for web

infreezer:
Slice layout is used only in related articles for web.
We don't need all of the slices there.
This will decrease the slice-layout rnw.js file from **50kb** to **22kb**.

image


packages/slice-layout/src/slice-layout.web.js
```diff
@@ -0,0 +1,5 @@
+import StandardSlice from "./templates/standard";
+import LeadOneAndTwoSlice from "./templates/leadoneandtwo";
+import OpinionOneAndTwoSlice from "./templates/opiniononeandtwo";
+
+export { StandardSlice, LeadOneAndTwoSlice, OpinionOneAndTwoSlice };
```
ValGeorgiev: Do we have stories and tests for these slices on web?

infreezer: We do. [Web snaps.](https://github.com/newsuk/times-components/tree/feat/web-slice-layout-perf/packages/slice-layout/__tests__/web/__snapshots__) and stories: 
image

PR #2573, opened 2020-03-20
feat: ad component changes to integrate with ad manager library

mikefieldmay:
This PR replaces the Ad component with divs that have specific ids.
This PR will break the ads on the Topic and Profile pages without the changes to render from this PR:
https://github.com/newsuk/cps-content-render/pull/3331


mikefieldmay:
I'm just trying something

PR #2576, opened 2020-03-30
feat: improve load of interactives

apostolnikov:
[Jira](https://nidigitalsolutions.jira.com/browse/TNLT-881)

Before: 

BeforeINTERACTIVE

After:

AfterINTERACTIVE22


packages/interactive-wrapper/src/interactive-wrapper.js
```diff
@@ -24,7 +24,7 @@ class InteractiveWrapper extends Component {
   constructor() {
     super();
     this.state = {
-      height: 1
+      height: 0
```
chrishutchinson: I seem to recall there was an issue with certain interactives which required a minimum height of 1 to make sure they fired to required postMessage calls. Does your optimisation still work if we leave this as `1`?

apostolnikov: Yes, it does. But the UX will be worse. For a small amount of time, there will be 1px line of the interactive content, which will look like a border, then the whole interactive will load.

Ineractive1px

chrishutchinson: I think Datawrapper charts were the offending one. Are you able to test with a different component type? It could be that a combination of this fix, plus the one I made to the web view last week means we can leave it at `0`. It'd be good to confirm.

I can provide a mock data structure for a Datawrapper chart component if you can do that. Otherwise I can provide you an article URL that has a chart in it. Let me know which works for you.

apostolnikov: It works as expected with this article: https://www.thetimes.co.uk/edition/news/britain-faces-6-months-of-coronavirus-curbs-t5cpk33sx . No problems with the puffs nor with graphics.

PR #2577, opened 2020-04-06
Feat/remove polished lib

apostolnikov:
[Jira](https://nidigitalsolutions.jira.com/browse/TNLT-1174)

Before: 
image

After:
image


packages/article-skeleton/src/sticky-save-and-share-bar.js
```diff
@@ -27,13 +26,10 @@ const SaveShareContainer = styled.div`
       computeProgressStyles(
         progress => css`
           height: ${60 - progress * 10}px;
-          box-shadow: 0 2px 5px 0 ${rgba(0, 0, 0, 0.2 * progress)};
+          box-shadow: 0 2px 5px 0 #c9c7c7ed;
```
ValGeorgiev, line 29: can we add this hexes to styleguide

PR #2579, opened 2020-04-08
feat: ad library integration

mikefieldmay:
This PR includes the previous ad integration work as well as a fix for the pagination of topics and authors that keeps the inline ad on the page


mikefieldmay:
I'm closing this as I am unable to fix the issue with the cypress tests. 
If we use any key other than the elementId, the cypress tests fail throwing a cannot find index of undefined error. It works fine when run in render so I think it must be some weirdness with headless headless Cypress.
I've had a good go at fixing it and haven't got anywhere.

PR #2582, opened 2020-04-14
fix: TNLT-326 nested styling for inline paragraph links

infreezer:
Apply nested styling to links in inline paragraphs and paragraphs with dropcap.

Before:

nested-styling-dropcap-before

After:
nested-styling-dropcap-after


packages/article-skeleton/src/article-body/article-link.js
```diff
@@ -11,6 +11,7 @@ const ArticleLink = props => (
       const styles = styleFactory(scale);
       return (
         <TextLink
+          allowFontScaling={false}
```
infreezer, line 14: The font scaling of the text is disabled. That's why I'm disabling it for the links also. Fixing:

links-font-scale

PR #2585, opened 2020-04-27
feat(TNLT-883): Related articles clean-up

infreezer:
Remove not used slices from related articles.

[TNLT-883](https://nidigitalsolutions.jira.com/browse/TNLT-883)


packages/related-articles/src/related-articles.js
```diff
@@ -56,75 +52,17 @@ class RelatedArticles extends Component {
...
-            />
-          );
-      }
-    };
-
     return (
       <View>
         <RelatedArticlesHeading />
-        {renderSlice()}
+        <StandardSlice
```
ValGeorgiev, line 58: Every time we will render standardslice, are we sure we want that. In the previous implementation if sliceName is undefined or different than "StandardSlice" nothing will be rendered.

infreezer: I believe it's the same now. When sliceName is undefined, nothing is rendered.
[https://github.com/newsuk/times-components/blob/4095e300fdd9a6012acec1acc9bc6ee3677de355/packages/related-articles/src/related-articles.js#L19](https://github.com/newsuk/times-components/blob/4095e300fdd9a6012acec1acc9bc6ee3677de355/packages/related-articles/src/related-articles.js#L19)

infreezer: Check added.

PR #2591, opened 2020-05-05
feat: improve loading for interactives on android

apostolnikov:
[Jira](https://nidigitalsolutions.jira.com/browse/TNLT-1654)


packages/interactive-wrapper/src/interactive-wrapper.js
```diff
@@ -74,7 +74,15 @@ class InteractiveWrapper extends Component {
     } = this.props;
     const { height } = this.state;
     const uri = `${editorialLambdaProtocol}${editorialLambdaOrigin}/${editorialLambdaSlug}/${id}?dev=${dev}&env=${environment}&platform=${platform}&version=${version}`;
-    const scriptToInject = `window.postMessage = function(data) {window.ReactNativeWebView.postMessage(data);};(${webviewEventCallbackSetup})({window});`;
+    const scriptToInjectIOS = `window.postMessage = function(data) {window.ReactNativeWebView.postMessage(data);};(${webviewEventCallbackSetup})({window});`;
+    const scriptToInjectAndroid = `
+      setTimeout(function() {
+        window.ReactNativeWebView.postMessage(document.documentElement.scrollHeight);
+      }, 500);
+      true;
```
ValGeorgiev, line 82: do we need `true;` statement?

apostolnikov: Yes we do, the whole thing is a dirty hack, but it happens to works like charm on every android device that I tested it. There is nothing more we could do on this one... at least for now

PR #2594, opened 2020-05-11
chore(TNLT-1941): prevent crash of puff slice

apostolnikov:
This is preventing the apps from crashing when there is puff slice in the edition. It also includes version bumps for ios, pages and section packages. 

Linked with: https://github.com/newsuk/times-components/pull/2593

### DIFF
--- a/ios-app/package.json
+++ b/ios-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@times-components/ios-app",
-  "version": "0.17.100",
+  "version": "0.17.102",
   "license": "BSD-3-Clause",
   "scripts": {
     "build": "cat package.json | grep version | head -1 | sed 's/[\",\t ]//g' | awk -F: '{ print \"Bundle Version: \" $2 }' > ios-assets/js/version.meta",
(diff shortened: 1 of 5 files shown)


ios-app/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@react-native-community/netinfo": "5.3.3",
-    "@times-components/pages": "2.1.96",
+    "@times-components/pages": "2.1.98",
```
infreezer, line 14: Do we have to do the same in the [android-app](https://github.com/newsuk/times-components/blob/master/android-app/package.json#L15)?

ios-app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@times-components/ios-app",
-  "version": "0.17.100",
+  "version": "0.17.102",
```
infreezer, line 3: What is the latest published ios-app version?

ios-app/package.json
apostolnikov: We do not, that's why it was causing lint errors...

ios-app/package.json
apostolnikov: Currently it's, "version": "0.17.100". I expect that it will be bumped to 101 from the latest merged change

infreezer: I believe @manosim published `0.17.102` [here](https://circleci.com/gh/newsuk/times-components/9291?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

manosim: Yes, that's correct - I did publish `0.17.102` in case we had to do an emergency release this morning.

PR #2597, opened 2020-05-12
fix: Update react-native-v8 android for Android

manosim:


manosim:
Did a quick check and it's looking good!

PR #2599, opened 2020-05-13
feat(TNLT-1209) Add GraphQL service and tracking for NewsletterPuff

themischatzis:


packages/article-skeleton/__tests__/inline-newsletter-puff.base.js
```diff
@@ -1,30 +1,160 @@
...
-      headline: "Politics. Explained.",
-      copy:
-        "Sign up to receive our brilliant Red Box newsletter, Matt Chorley`s poke at politics delivered every weekday morning at 8am."
-    };
-
-    it("renders the Signup view when the user is NOT subscribed to the newsletter", () => {
-      const component = TestRenderer.create(
-        <InlineNewsletterPuff {...props} isSubscribedToNewsletter={false} />
-      );
+  describe.only("Inline Newsletter Puff", () => {
```
chrishutchinson, line 56: Stray `.only`

packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -238,21 +238,16 @@ export default ({
         );
       }
       if (element && element.value === "newsletter-puff") {
-        const copy = decodeURIComponent(element.attributes.copy);
-        const label = decodeURIComponent(element.attributes.label);
-        const headline = decodeURIComponent(element.attributes.headline);
-
         return (
           <InlineNewsletterPuff
             // When the GeaphQL service is ready newsletterId will be passed as prop
```
chrishutchinson, line 243: Can we remove this comment now?

packages/article-skeleton/src/article-body/inline-newsletter-puff.js
```diff
@@ -14,78 +18,146 @@ import {
...
-      </InpSignupContainer>
-    )}
-  </InpContainer>
-);
+            {(subscribeNewsletter, { loading: updatingSubscription }) => (
+              <InpContainer>
+                <InpImageContainer>
+                  <Image
+                    aspectRatio={1.42}
+                    uri="https://nuk-tnl-deck-prod-static.s3-eu-west-1.amazonaws.com/uploads/2aa9050e6c3d4de682f11a4802ebba96.jpg"
```
chrishutchinson, line 97: I think we should find a better URL for this. Proabbly one to either add as part of `times-components` itself, or one to upload to our `/assets/...` bucket.

packages/article-skeleton/src/article-body/inline-newsletter-puff.js
```diff
@@ -14,78 +18,146 @@ import {
...
+            <InpContainer style={{ height: 257 }}>
+              <Placeholder />
+            </InpContainer>
+          );
+        }
+
+        const { newsletter } = data;
+
+        if (newsletter.isSubscribed && !justSubscribed) {
+          return null;
```
chrishutchinson, line 82: It'd be good if we can discuss the UX for readers who are already subscribed. I believe that due to the speed of the Customer API (slow), we should make sure that the page doens't jump around after the request is finished.

packages/article-skeleton/src/article-body/inline-newsletter-puff.js
csaltmountain: Actually, no you mention it, this image URL should be passed in as a prop. The URL needs to be dynamic so it can be set in the component builder.

themischatzis: Agree. We need to figure out how the URL will be generate from the component builder or map to it by using the `newsletterId`. Also we need to make sure that the images is render properly on Androids.

packages/article-skeleton/src/article-body/inline-newsletter-puff.js
csaltmountain: There was a bit of a discussion about it here: https://newsuktechnology.slack.com/archives/CKQ9XU7MF/p1588068466004000 but no real consensus was reached about what to do. The `return null` was just a quick option. I agree the UX should be better. But who needs to make the decision over what the "already subscribed" state should be?

packages/article-skeleton/src/article-body/inline-newsletter-puff.js
```diff
@@ -1,91 +1,165 @@
...
+export const GET_NEWSLETTER = gql`
+  query GetNewsletter($code: String!) {
+    newsletter(code: $code) {
+      id
+      isSubscribed
+    }
+  }
+`;
+
+export const SUBSCRIBE_NEWSLETTER = gql`
```
ValGeorgiev, line 48: can we put graphql queries in a separate file

PR #2603, opened 2020-05-15
feat: handle crash takeoverpage ios

apostolnikov:
Handle app crash on opening the app with a takeover page link from an external app. 

The problem is coming from a deep link setup from the native ios app. Every link that starting with https://thetimes.co.uk will be fist open inside the app and when this happens with takeover page, which is not supported by the app, it will crash. 

There is no way to redirect this link from the react native side, because it will initiate the deep link "loop" again and the will end up crashing.

My fix is currently using a webview to open the article inside the app, which may lead to confusion for some users, but at least we will show the clicked article to the user. (Again, this will happen only on IOS platform with a takeoverpage link opened from an external app, which is really rare case...)

Other way of handling this, will be to show some king of pop up alert, that this article is not supported by the app, and when the user clicks okay, it will redirect to the main edition screen.

takeoverpage

### DIFF
--- a/packages/article/src/article.js
+++ b/packages/article/src/article.js
@@ -1,4 +1,5 @@
 import React from "react";
+import { Platform } from "react-native";
 import ArticleMagazineComment from "@times-components/article-magazine-comment";
 import ArticleInDepth from "@times-components/article-in-depth";
 import ArticleMagazineStandard from "@times-components/article-magazine-standard";
@@ -8,6 +9,7 @@ import Responsive from "@times-components/responsive";
 import { scales } from "@times-components/styleguide";
 import { MessageManager } from "@times-components/message-bar";
 import { getMediaList, addIndexesToInlineImages } from "./utils";
+import handleTakeoverpageUrl from "./utils/handle-takeoverpage-url";
 
 export const templates = {
   indepth: ArticleInDepth,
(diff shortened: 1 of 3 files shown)


manosim:
Shall we hold on this one as we might need to confirm with UX too? Isn't there a way to prevent the loop and just open Safari from the native side? cc @newsuk-lee-daffen

apostolnikov:
> Shall we hold on this one as we might need to confirm with UX too? Isn't there a way to prevent the loop and just open Safari from the native side? cc @newsuk-lee-daffen

Any udpates? @manosim @newsuk-lee-daffen

newsuk-lee-daffen:
@apostolnikov @manosim This looks like something we might be able to fix natively. We would need some way to detect that an article isn't supported and pop open an in-app browser view to handle the content. Is there any way to know that content isn't supported (other than using the url path)?

apostolnikov:
> @apostolnikov @manosim This looks like something we might be able to fix natively. We would need some way to detect that an article isn't supported and pop open an in-app browser view to handle the content. Is there any way to know that content isn't supported (other than using the url path)?

Well that's the thing. The url will be the same. The only thing that can be used to differentiate is the "template" field is "template === "takeoverpage"" inside the article data that is returned.

packages/article/src/article.js
```diff
@@ -28,15 +30,22 @@ const Article = props => {
   const { article, onImagePress } = props;
   const { leadAsset, template } = article || {};
   let { content } = article || {};
+
   if (template === "takeoverpage") {
+    if (Platform.OS === "ios") {
+      return handleTakeoverpageUrl(article.url);
+    }
     throw new TakeoverBailout("Aborted react render: Takeover page");
```
manosim, line 38: Does this mean that Android would still crash? The reason that it throws is so that Render can fallback to the legacy template - @chrishutchinson will be able to give us some more info soon.

apostolnikov: The android deep link config works a bit different. The user will be promped with an alert box, to choose the browser he wants to open the article with. So it is avoiding the crash. The issue is IOS only... This is synced with @newsuk-lee-daffen, he created a ticket where he will try to fix it from the native side.

PR #2623, opened 2020-06-19
fix(TNLT-2740): updates/fixes storybook on web/iOS/android

c-j-j:
I've made the following changes in order to get storybook running properly on iOS:
1. Updated the `ios`/`android` directories to be compatible with the latest version of XCode/RN 0.60.5
2. Updates storybook from v4 to v5, which fixed issues with the sidebar overlapping the story view on iOS.
3. Updated `.storybook/webpack.config.js` file to be compatible with v5 of storybook.

I wasn't able to figure out why autolinking wasn't working for android. They should link automatically now with RN60+, but I kept seeing warnings that the libraries were missing at runtime.


c-j-j:
@aronshamash @infreezer FYI since you last reviewed, I have made a few changes for iOS autolinking to work. I've added a few more RN deps to the root package.json, which means we don't need to specify things like react-native-webview in the ios/Podfile. 

Autolinking is still not working for android - we'll need to specify any new deps like we're currently doing.

I also configured storybook to use async storage - which then is used to remember the last story you were on whenever you refresh the app.

PR #2626, opened 2020-06-22
fix(TNLT-1893): indents dropcap on tablet

c-j-j:
https://nidigitalsolutions.jira.com/browse/TNLT-1893

Tested on Android Pixel C, iPhone 8, iPad mini, iPad pro

Before: 
image

After:
image


packages/article-skeleton/src/article-body/inline-paragraph.js
```diff
@@ -52,7 +52,9 @@ const InlineParagraph = ({
   const positioned = manager.layout();
 
   return [
-    dropCap && dropCap.element,
+    dropCap && (
+      <View style={{ left: gutters - spacing(1) }}>{dropCap.element}</View>
```
infreezer, line 56:
```suggestion
      <View style={{ left: gutters - spacing(2) }}>{dropCap.element}</View>
```
Do you think we have to subtract 10px (spacing(2)) in order to keep the current alignment from production:

image


image

c-j-j: Yeah good spot @infreezer - - just tested this on an iPad and definitely looks more aligned. I'll update to `spacing(2)`

ios/Podfile.lock
```diff
@@ -377,6 +377,6 @@ SPEC CHECKSUMS:
   RNSVG: be27aa7c58819f97399388ae53d7fa0572f32c7f
   Yoga: f2a7cd4280bfe2cca5a7aed98ba0eb3d1310f18b
 
-PODFILE CHECKSUM: d3dec4569b3074e3ce5e8b734ab822b73c3e9438
+PODFILE CHECKSUM: 023b69829f0d4896347e521f099c5b096cb2c1df
```
infreezer, line 380: Is this intentional?

c-j-j: Yes. It should have been included in my previous PR where I updated the iOS storybook project, somehow this change wasn't included. 

I'm really not sure what changed the checksum, to be honest, however @Geoff-Ford also had the same checksum generated on his branch too.

PR #2628, opened 2020-06-23
feat: long read flag

Geoff-Ford:
Added a new article flag for long read articles. The value is provided by the new `longRead` article property from TPA rather than as an additional `expirableFlags`.

I have added the flag logic and added/updated tests wherever I can see this must be applied (NATIVE ONLY) and also added a story for this too. I have also added `longRead` to the graphql provider-queries so pay special attention to those to make sure I haven't put something where it shouldn't be.

As the BE work to request `longRead` in Contemporary Beta hasn't been done yet, I don't have a decent screen shot of this flag in place but here is a screenshot from storybook of all flags together.

Screenshot 2020-06-23 at 11 16 01

Have some screenshots now:

Screenshot 2020-06-25 at 12 40 07

Screenshot 2020-06-25 at 13 41 31


packages/article-main-standard/src/article-header/article-header.js
```diff
@@ -27,20 +28,20 @@ const ArticleHeader = ({
       selectable
       style={[
         styles.articleHeadLineText,
-        !(getActiveFlags(flags).length > 0 || standfirst) &&
+        !(getActiveFlags(flags).length > 0 || longRead || standfirst) &&
```
c-j-j, line 31: `getActiveFlags(flags).length > 0 || longRead` is duplicatd 3 times here - could be extracted to a var called something like `shouldRenderFlags`?

aronshamash:
@geoff looks good. Can you add some more screenshots to description for example of article pages ?

Geoff-Ford:
> @geoff looks good. Can you add some more screenshots to description for example of article pages ?

As mentioned in PR description above, I can't get that at the moment as I need Dan's work to bring in `longRead` from TPA. We have been trying to work out where this needs changing so I might get some screen shots this afternoon.

packages/article-main-standard/src/article-header/article-header.js
Geoff-Ford: This was existing code but yeah, makes sense.

PR #2631, opened 2020-06-29
fix: Increase summary length in tile X

aronshamash:
- https://nidigitalsolutions.jira.com/browse/TNLT-2945
- Increase length of summary text in Tile X from `800` to `1000` to reduce white space
- @mrdanb this will require a TC bump + query update for iOS.

Screenshot 2020-06-29 13 52 17


aronshamash:
> Looks good to me - I assume we'll need to update the iOS project with the new `teaser1000`/`summary1000` fields before this is merged?

I thought we merge _then_ update the query in iOS after no? @Geoff-Ford  ?

packages/provider-queries/src/section_fragment.graphql
```diff
@@ -440,11 +440,11 @@ fragment commonSlicesSection on StandardSectionSlice {
         strapline
         summary125: summary(maxCharCount: 125)
         summary300: summary(maxCharCount: 300)
-        summary800: summary(maxCharCount: 800)
+        summary1000: summary(maxCharCount: 1000)
```
Geoff-Ford, line 443: I see a bunch of tiles that will still be requesting `getTileSummary(tile, 800)`. Won't replacing this break those instance?

Geoff-Ford:
> > Looks good to me - I assume we'll need to update the iOS project with the new `teaser1000`/`summary1000` fields before this is merged?
> 
> I thought we merge _then_ update the query in iOS after no? @Geoff-Ford ?

Well I only asked @mrdanb to change the iOS query in Beta first so I could confirm the UI changes were as expected. I had already modified the query in my PR which Dan referenced when making his changes.

packages/provider-queries/src/section_fragment.graphql
aronshamash: this is just for the lead tile in the `leadtwonopicandtwo` slice, which is only `tile X` and `tile F` so this shouldn't break any other instances.

Geoff-Ford: Right, gotcha. I can see that now when I look at the whole of the changed file.

Geoff-Ford: Cool, and I see `tile F` only uses `125` anyway.

PR #2632, opened 2020-07-01
fix: register fonts for storybook native

c-j-j:
Some of the article rendering logic on native requires fonts to be registered with `FontStorage`.

### DIFF
--- a/storybook-native/index.js
+++ b/storybook-native/index.js
@@ -1,6 +1,15 @@
 import "core-js";
 import "regenerator-runtime/runtime";
 import { AppRegistry } from "react-native";
+import { FontStorage } from "@times-components/typeset";
 import StorybookUIRoot from "./storybook";
+import ttf from "../fonts";
+
+Object.keys(ttf).forEach(fontName => {
+  FontStorage.registerFont(fontName, ttf[fontName]);
+});
+
+// Suppress all warnings (WIP solution until issues are correctly addressed)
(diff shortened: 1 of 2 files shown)


storybook-native/index.js
```diff
@@ -1,6 +1,15 @@
...
 import { AppRegistry } from "react-native";
+import { FontStorage } from "@times-components/typeset";
 import StorybookUIRoot from "./storybook";
+import ttf from "../fonts";
+
+Object.keys(ttf).forEach(fontName => {
+  FontStorage.registerFont(fontName, ttf[fontName]);
+});
+
+// Suppress all warnings (WIP solution until issues are correctly addressed)
```
Geoff-Ford, line 12: You want to leave this in?

c-j-j: I don't mind... It's useful for when you're developing with storybook, as it's impossible with all the existing warnings taking up most of the screen.

Geoff-Ford: True. Just there is always a risk it will slip through in a PR uncommented. Which wouldn't be the end of the world but would mean we wouldn't see warnings and miss something important. I agree at the moment it is super annoying and you would probably miss any new warnings anyway unless you read every single one after a change.
🤷‍♂️

c-j-j: That's true - I'll remove this, as it's not too difficult to add back in

PR #2634, opened 2020-07-01
fix(TNLT-1331): add subsection to targeting attribute path

infreezer:
[TNLT-2614](https://nidigitalsolutions.jira.com/browse/TNLT-2614)

Fix for Ad Section based targeting.
The variable `section` in Ad props was undefined.
Updated to use section name from ad config slot targeting.

ad-section-fix
ad-section-timeline


apostolnikov:
Migrated to times-components-native https://github.com/newsuk/times-components-native/pull/5

PR #2642, opened 2020-09-02
feat(TNLT-3719): add once to spot IM comment call

mikefieldmay:
This PR adds the once option to the comments event listener as it was firing multiple times.
After investigating, the event listener was only registered once, so multiple firing must be to do with slightly dodgy event listener firing. Once option should limit the event listener to firing just once

https://nidigitalsolutions.jira.com/browse/TNLT-3719


chrishutchinson (approved):

chrishutchinson:
I'm not sure the ticket you've linked is the right one @mikefieldmay

aronshamash (approved):

PR #2648, opened 2020-10-02
feat/scb-525: implement native ad slots replacing nativo on article pages

nikkibyrne:
Replacing the current nativo ads with native ones served via GAM, div and styling to match existing nativo ad on the article pages after 9th para.

To render the ads to see them, you need to run newsUkAdLibrary.render() in the console.

[SCB-525](https://nidigitalsolutions.jira.com/browse/SCB-525)

Screenshot 2020-10-15 at 18 23 44

Screenshot 2020-10-15 at 18 52 15


packages/article-skeleton/src/native-ad.web.js
```diff
@@ -0,0 +1,36 @@
...
+	// Counts the number of paragraphs and returns the index
+	paywallChildren.map((item, i) => {
+		if (item.name === 'paragraph') {
+			paraCount += 1;
+			if (paraCount === 7) index = i;
+		}
+	});
+
+	// if nativeAd does not exist then place in the paragraph index number
+	const nativeAd = paywallChildren.find(item => item.name === 'nativeAd');
```
mikefieldmay, line 23: We're adding the nativeAd ourselves right? Not quite sure what the check is for?

packages/article-skeleton/src/native-ad.web.js
```diff
@@ -0,0 +1,36 @@
...
+	const child = children.find(item => item.name === 'paywall');
+	
+  if (!child) {
+    return children;
+  }
+
+	const clonedChildren = [...children];
+	const paywallChildren = child.children;
+	
+	let paraCount = 0;
```
mikefieldmay, line 11: We tend to use `const` values instead of lets so that we aren't mutating the value and it's easier to read.

```suggestion
	const paragraphItems = paywallChildren.map((item, index) => ({...item, index})).filter(item => item.name === 'paywall')
	const indexToAdd = paragraphItems[6] ? item.index : null
```

packages/article-skeleton/src/native-ad.web.js
```diff
@@ -0,0 +1,36 @@
+const insertNativeAd = (children) => {
+	const child = children.find(item => item.name === 'paywall');
+	
+  if (!child) {
+    return children;
+  }
+
+	const clonedChildren = [...children];
```
mikefieldmay, line 8: I would move this to line 2 and then do 
```suggestion
	const child = clonedChildren.find(item => item.name === 'paywall');
	if (!child) {
    return clonedChildren;
  }
```

packages/article-skeleton/src/article-body/article-body.web.js
```diff
@@ -80,6 +81,15 @@ const renderers = ({ paidContentClassName, template, analyticsStream }) => ({
       </Context.Consumer>
     );
   },
+  nativeAd(key, attrs, children) {
+    return (
+      <NativeAd key={key}>
```
mikefieldmay, line 86: Do we need an ID on this block?
What happens if the user is using add block or something like that?
It may be worth setting this div to display none initially and to then change that in the ad library if the ads load?

packages/article-skeleton/__tests__/web/native-ad.web.test.js
```diff
@@ -0,0 +1,30 @@
...
+const { content, contentWithAd } = nativeAdFixture;
+
+describe('native-ad', () => {
+
+	it('it adds the native ad block after the 7th paywall paragraph', () => {
+		expect(insertNativeAd(content)).toStrictEqual(contentWithAd);
+	});
+
+	test('Checks nativeAd does not already exist and that Paywall does exist', () => {
+		const content = [
```
mikefieldmay, line 12:
```suggestion
		const content = [
		{
				name: 'paywall',
				children: []
			}]
```

packages/article-skeleton/__tests__/web/native-ad.web.test.js
```diff
@@ -0,0 +1,30 @@
...
+import nativeAdFixture from "../../fixtures/native-ad";
+const { content, contentWithAd } = nativeAdFixture;
+
+describe('native-ad', () => {
+
+	it('it adds the native ad block after the 7th paywall paragraph', () => {
+		expect(insertNativeAd(content)).toStrictEqual(contentWithAd);
+	});
+
+	test('Checks nativeAd does not already exist and that Paywall does exist', () => {
```
mikefieldmay, line 11:
```suggestion
	it('does not insert ad if less than 7 paragraphs in paywall', () => {
```

packages/article-skeleton/__tests__/web/native-ad.web.test.js
```diff
@@ -0,0 +1,30 @@
...
+			{
+				name: 'paywall',
+				children: []
+			},
+			{
+				name: 'paragraph',
+				children: []
+			}
+		]
+		expect(content).toEqual(
```
mikefieldmay, line 22: This seems to be asserting against the object you created above without running the `insertNativeAd` on it>?
Maybe: 
```
const contentWithoutParagraphs = [
		{
				name: 'paywall',
				children: []
			}
]

expect(insertNativeAd(content)).toEqual(content)
```

packages/article-skeleton/__tests__/web/native-ad.web.test.js
```diff
@@ -0,0 +1,30 @@
...
+			expect.arrayContaining([
+				expect.objectContaining({name: 'paywall'}),
+				expect.not.objectContaining({name: 'nativeAd'})
+			])
+		);
```
mikefieldmay, line 27: I would add a third test where if paywall doesn't exist the original content is just returned returned

packages/article-skeleton/src/native-ad.web.js
mikefieldmay: Array indexes start at 0, so the seventh paragraph element should be at position 6?

mikefieldmay (changes requested):
A couple of changes to keep it in line with our codebase.
Might be worth thinking about what happens to the div if the use has ad blocker or things along those lines. Maybe don't display and use the ad library to display when the ads have loaded

packages/article-skeleton/src/native-ad.web.js
```diff
@@ -0,0 +1,26 @@
+const insertNativeAd = (children) => {
+	const clonedChildren = [...children];
+	const child = children.find(item => item.name === 'paywall');
```
mikefieldmay, line 3: this should be:
```suggestion
	const child = clonedChildren.find(item => item.name === 'paywall');
```

packages/article-skeleton/src/native-ad.web.js
nikkibyrne: have updated

packages/article-skeleton/src/native-ad.web.js
nikkibyrne: has been updated

packages/article-skeleton/src/article-body/article-body.web.js
nikkibyrne: have added a class for the hidden styling

packages/article-skeleton/__tests__/web/native-ad.web.test.js
nikkibyrne: Have added an extra test for the paywall

mikefieldmay (approved):

PR #2652, opened 2020-11-12
feat(TNLT-5628): added a width to avoid string coming out of the topic button

MariaRomero:
added a width to the topic tags to spot text overflowing from them

https://nidigitalsolutions.jira.com/browse/TNLT-5628


mikefieldmay (approved):
Looks good!

PR #2662, opened 2021-01-13
fix(TNLT-2359): add new user state check for comments readOnly state

chrishutchinson:
When we render comments as read only, or read/write, we used to check the `UserState.loggedIn` state. This check effectively looks at whether the request came in through a non-`_tp_` or non-`_fp_/open_anonymous` route  in Render.

The downside to thios approach, is that for readers that come as logged out with a `?shareToken=` param,  they also get routed through the full article path in Render. This ultimately means that despite being logged out, they are actually treated as 'logged in'.

This means the check we make on comments isn't valid, as it's trying to make them writable for readers that should only get read-only access.

Instead, I have introduced a new state checker, with the catchy name: `isLoggedInOrLoggedInAndNotShared`. This effectively does the following:
* If the reader is logged in, return `true`
* If the reader is logged in, and has a share token, return `true`
* If the reader is logged out, return `false`
* If the reader is logged out, and has a share token, return `false`

This is different from the previous behaviour, where now an additional explicit check is made to the customer's share token status, in addition to the route them came through on Render. The presence of a share token is checked client-side by inspecting the page URL, which cannot be done server side as Render does not get given the information to determine if the request was a share token request.

### DIFF
--- a/packages/article-extras/__tests__/shared.web.js
+++ b/packages/article-extras/__tests__/shared.web.js
@@ -33,7 +33,10 @@ export default () => {
     {
       name: "renders correctly",
       test: () => {
-        UserState.mockStates = [UserState.fullArticle, UserState.loggedInOrLoggedInAndNotShared];
+        UserState.mockStates = [
+          UserState.fullArticle,
+          UserState.loggedInOrLoggedInAndNotShared
+        ];
         const testInstance = TestRenderer.create(
           <ArticleExtras
             analyticsStream={() => {}}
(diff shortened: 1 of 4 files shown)


chrishutchinson:
In case, like me, you're struggling to get your head around it, basically the `isLoggedIn` check isn't actually checking if the reader is logged in, instead "did they come through a full article route in Render"

mikefieldmay (approved):
Looks good

PR #2666, opened 2021-03-18
fix(TNLT-6962): Add set heights to ads times components

mikefieldmay:
This PR adds set height to the article, topic and profile adverts.


c-j-j (approved):
All looks good to me 👍

PR #2670, opened 2021-03-23
fix(tnlt-000): slight tweaks to ad styles

mikefieldmay:
This pr adds the correct alignment to the ad header


packages/article-skeleton/src/styles/responsive.web.js
```diff
@@ -8,6 +8,7 @@ export const getHeaderAdStyles = template => {
 
   return styled(View)`
     box-sizing: content-box;
+    justif-content: center;
```
c-j-j, line 11: justify-content?

mikefieldmay: Good spot

c-j-j (approved):

PR #2681, opened 2021-04-14
feat(TE-35): add auto newsletter puff - refactored to use display:non…

vinhlam:
Using display:none instead of rendering null when set to not view.
because of this, added an empty <div/> for the intersectionobserver as display:none wont trigger an intersection event.


packages/article-skeleton/src/article-body/view-count-wrapper.js
```diff
@@ -56,7 +65,7 @@ const ViewCountWrapper = ({
           entries => {
```
flashcheeks, line 65: Do we need to limit the curation of the observer to only once?

```
observer =
    observer === undefined &&
    typeof window !== "undefined" &&
```

flashcheeks (approved):

PR #2682, opened 2021-04-14
feat(TE-33): Additional Related Articles rail

vinhlam:
Additional Related Articles rail functionality added. Enabled using the `additionalRelatedArticlesFlag` flag which should be passed in through render as a feature flag


flashcheeks (approved):

packages/article-extras/src/article-extras-content.js
```diff
@@ -35,6 +36,12 @@ const ArticleExtrasContent = ({
                 onPress={onRelatedArticlePress}
                 slice={relatedArticleSlice}
               />
+              <RelatedArticles
```
flashcheeks, line 39: Is this file change needed? Is this not app code?

vinhlam: Ive reverted this as it is used for native

PR #2686, opened 2021-04-21
feat(TE-64): algolia related articles

flashcheeks:
[TE-64](https://nidigitalsolutions.jira.com/browse/TE-64?atlOrigin=eyJpIjoiZjZjMTkxNDJkOTQzNDk1NTk5MDZmMGQ0NTkwODE0N2QiLCJwIjoiaiJ9): Load relevant articles from Algolia and display in `Additional Featured Articles` slice

* created `AlgoliaSearchProvider` and `useAlgoliaSearch` context to expose search to article components
* created `searchRelatedArticles` func to load/format article for use in `RelatedArticles` component
* pass `AlgoliaSearchKeys` down from Render

Logic for the search query is a combination of the article `label`, `section` and `topics`, excluding the current article being read.


packages/utils/src/algolia-related-articles.web.js
```diff
@@ -0,0 +1,77 @@
...
+  if (article.topics)
+    query = article.topics.reduce((acc, t) => acc.concat(" ", t.name), query);
+
+  return query.trim();
+};
+
+const searchRelatedArticles = async (index, article) => {
+  if (index) {
+    const query = formatQuery(article);
+    console.log("searchRelatedArticles", query); // eslint-disable-line no-console
```
flashcheeks, line 62: `console.log` will be used in production (behind feature flag) to view Aloglia query.

packages/utils/src/algolia-search.web.js
```diff
@@ -0,0 +1,50 @@
+import React, { createContext, useMemo, useCallback, useContext } from "react";
+import PropTypes from "prop-types";
+import algoliasearch from "algoliasearch";
+
+import searchRelatedArticles from "./algolia-related-articles.web";
+
+const applicationId = "PZGYBTWG3J";
+const apiKey = "572d633b038f582c813e45a798b94238";
+const indexName = "prod_articles_by_published_date_desc";
```
flashcheeks, line 9: I did want to store these values as `process.env` but they seem to exist in [cps-content-render](https://github.com/newsuk/cps-content-render/blob/develop/packages/render/src/config/prod.json#L40) so they are probably fine to exist here.

packages/utils/src/algolia-related-articles.web.js
```diff
@@ -0,0 +1,78 @@
...
+  if (article.topics)
+    query = article.topics.reduce((acc, t) => acc.concat(" ", t.name), query);
+
+  return query.trim();
+};
+
+const searchRelatedArticles = async (index, article) => {
+  if (index) {
+    const query = formatQuery(article);
+    console.log("AlgoliaSearch query", query); // eslint-disable-line no-console
```
flashcheeks, line 62: console.log will be used in production (behind feature flag) to view Aloglia query.

packages/utils/src/algolia-search.web.js
mikefieldmay: My only concern with this is that render is a private repo where theoretically times components is open to the public

mikefieldmay (changes requested):
everything looks good, but I'm not sure about storing any keys in the code as Times Components is a public repo

flashcheeks:
> everything looks good, but I'm not sure about storing any keys in the code as Times Components is a public repo

Good point, I will remove

mikefieldmay (approved):

PR #2692, opened 2021-05-04
feat(TE-43): Typescript in article puffs with fixed dependencies

bibicollins1:
### Description

This PR includes: 
- Create packages ts-components for new typescript components, ViewCountWrapper has been added to this package
- Create the InArticlePuff component with tests using React testing library

Screenshot 2021-05-04 at 14 02 00
Screenshot 2021-05-04 at 14 02 38
Screenshot 2021-05-04 at 14 03 51
Screenshot 2021-05-04 at 14 03 37


packages/ts-components/src/__tests__/ViewCountWrapper.test.tsx
```diff
@@ -0,0 +1,180 @@
+/**
+ * @jest-environment jsdom
```
vinhlam, line 2: this no longer require as we added it to one of the config files

packages/ts-components/src/in-article-puff/styles.ts
```diff
@@ -0,0 +1,114 @@
...
+export const MainContentContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: space-evenly;
+`;
+
+export const LinkWrapper = styled.a<ContainerType>`
+  display: flex;
+  flex-direction: row;
+  align-items: top;
```
vinhlam, line 98: `top` is not a valid value for align-items, its been complaining in my IDE

vinhlam (approved):
happy for you to not fix these as i can do them in my PR

packages/ts-components/CHANGELOG.md
```diff
@@ -0,0 +1,1565 @@
+# Change Log
```
flashcheeks, line 1: We should delete this file.

packages/ts-components/README.md
```diff
@@ -0,0 +1,3 @@
+# @times-components/responsive-image
```
flashcheeks, line 1: We should update this file with relevant info/commands to help explain the components and example usage.

packages/ts-components/package.json
```diff
@@ -0,0 +1,92 @@
...
+    "@times-components/storybook": "4.1.60",
+    "@times-components/test-utils": "2.3.8",
+    "@types/jest": "24.9.1",
+    "@types/react-dom": "17.0.3",
+    "@types/react-test-renderer": "16.9.0",
+    "@types/styled-components": "5.1.9",
+    "jest": "24.8.0",
+    "metro-react-native-babel-preset": "0.53.1",
+    "prettier": "1.14.3",
+    "react-native": "0.61.5",
```
flashcheeks, line 73: Do we need native libraries?

packages/ts-components/tsconfig.json
```diff
@@ -0,0 +1,46 @@
+{
+  "compilerOptions": {
+    "target": "esnext",
+    "outDir": "dist",
+    "rootDir": "src",
+    "jsx": "react-native",
```
flashcheeks, line 6: Should this be `react`?

packages/article-skeleton/src/article-body/article-body.web.js
```diff
@@ -158,6 +160,17 @@ const renderers = ({
             label={decodeURIComponent(label)}
           />
         );
+      case "in-article-puff":
+        return inArticlePuffFlag ? (
+          <InArticlePuff
```
flashcheeks, line 165: Looking at the current/old `in-article-puff` component I can see that it receives some props but also loads data from an external .json file (receives id then loads `image`, `copy`, `url` etc).

It is probably worth us chatting about our desired data approach.

flashcheeks (changes requested):

packages/ts-components/package.json
bibicollins1: @vinhlam

vinhlam: Ill have a check to see if it causes any issues

packages/ts-components/CHANGELOG.md
vinhlam: does it not automatically get generated when there are changes and version updates?

packages/ts-components/package.json
vinhlam: I can remove it from the dependences and the babel config but if I try to do any `yarn add` it complains. I've removed it, and it builds and passes all tests

flashcheeks (approved):

PR #2699, opened 2021-05-12
feat(TE-43): InArticlePuff - fetch data from API

flashcheeks:
Update `InArticlePuff` to use data fetch from Deck API

#### DONE

* Wrap `InArticlePuff` in `FetchProvider` to request data
* Use fetch context within `InArticlePuff` to return data
* Add loading placeholder state for server-side render
* Polyfill fetch with `isomorphic-unfetch`
* Update storybook and fixtures data
* Update tests

#### TO DO

* Mock fetch for storybook
* Add test for `FetchProvider`


packages/article-skeleton/src/article-body/article-body.web.js
```diff
@@ -160,22 +164,18 @@ const renderers = ({
...
-                imageUri={imageUri}
-                headline={headline}
-                copy={copy}
-                link={link}
-                linkText={linkText}
-                sectionColour={sectionColour || colours.section.default}
-              />
+            {({ theme }) => (
+              <FetchProvider url={deckApiUrl + attributes["deck-id"]}>
+                <InArticlePuff sectionColour={theme.sectionColour} />
```
vinhlam, line 173: If we're moving all of the props out of the InArticlePuff, we should probably pull the `secitionColour` out as well, as we can just get it using `useContext` within the component instead of `Context.Consumer`

vinhlam (approved):

packages/article-skeleton/src/article-body/article-body.web.js
flashcheeks: Added to the todo list

PR #2702, opened 2021-05-14
added storybook for ts-components

vinhlam:
ts-components typescript files will render in storybook without having to do a 'yarn build'. This is still required if a typescript component is used by a module that is not typescript


bibicollins1 (approved):
Works well nice one!

PR #2703, opened 2021-05-14
feat: TE89 - Add Tracking to InArticlePuff

vinhlam:
created 2 wrapper components. Look in the InArticlePuff component for an example of usage

`TrackingContextProvider` -> adds tracking context and analytics stream. these can be nested.
```
       <TrackingContextProvider
          context={{
            articleHeadline: 'articleHeadline',
            section: 'section'
          }}
          analyticsStream={analyticsStream}
        >
```

`TrackScrolledIntoView` -> triggers an analytics event to the componet assigned the intersectObserverRef via a ref
```
       <TrackScrolledIntoView analyticsEvent={scrollEvent}>
          {({ intersectObserverRef }) => (
            <div
              ref={intersectObserverRef}
```


packages/ts-components/src/components/in-article-puff/__tests__/InArticlePuff.test.tsx
```diff
@@ -117,4 +119,135 @@ describe('InArticlePuff', () => {
...
+      fireEvent.click(getByText('Read the full article'));
+
+      expect(analyticsStream).toHaveBeenCalledWith({
+        articleHeadline: 'articleHeadline',
+        component_name: 'Where can I get a Covid vaccine in England?',
+        component_type: 'in-article-component : puff : interactive',
+        event_navigation_action: 'navigation',
+        event_navigation_browsing_method: 'click',
+        event_navigation_name:
+          'button : Read the full articleIconForwardChevron',
```
vinhlam, line 189: This is a bit weird, but I think its just the way the tests bundle the Icon. It should just use the `.textContent` of the link.

packages/article-skeleton/src/article-skeleton.web.js
```diff
@@ -96,122 +98,132 @@ const ArticleSkeleton = ({
...
-                type="button"
-                className="ArticleMetaBanner-button"
-                data-clipboard-target="#ArticleMetaBanner-uuid"
-              >
-                Copy
-              </button>
+      <TrackingContextProvider
+        context={{
+          article_name: headline || shortHeadline || "",
+          section_details: section
```
flashcheeks, line 104: Do we need to add context here eg: `"section : <section-name>"`

packages/ts-components/src/components/in-article-puff/InArticlePuff.tsx
```diff
@@ -17,11 +20,30 @@ import {
...
 }> = ({ sectionColour }) => {
   const [colour, setColour] = useState('#bf0000');
 
+  const handleClick = (
+    event: MouseEvent<HTMLAnchorElement>,
+    fireAnalyticsEvent: (evt: any) => void
+  ) => {
+    const clickTarget = event.currentTarget.textContent;
+    const buttonType =
+      clickTarget !== '' && clickTarget !== null ? clickTarget : 'image';
```
flashcheeks, line 44: Looking at the analytics spec, I don't think they want to pass the headline text, but actually the value `headline`

packages/ts-components/src/components/in-article-puff/InArticlePuff.tsx
```diff
@@ -48,38 +70,64 @@ export const InArticlePuff: React.FC<{
...
-          {copy && <Copy>{copy}</Copy>}
-        </div>
+              <ContentContainer hasImage={hasImage}>
+                <div>
+                  <Label hasImage={hasImage} style={{ color: sectionColour }}>
+                    {label}
+                  </Label>
+                  <a
+                    href={link}
+                    onClick={evt => handleClick(evt, fireAnalyticsEvent)}
```
flashcheeks, line 106: Maybe we can just pass (hardcode) `headline`, `image` and `Read more` as the button type rather than pass the event

packages/ts-components/src/components/in-article-puff/InArticlePuff.tsx
```diff
@@ -48,38 +70,64 @@ export const InArticlePuff: React.FC<{
   const hasImage = Boolean(image);
 
   return (
-    <Container
-      style={{ borderTop: `2px ${sectionColour} solid` }}
-      data-testid="InArticlePuff - Container"
+    <TrackingContextProvider
+      context={{
+        component_type: 'in-article-component : puff : interactive',
```
flashcheeks, line 75: I think the string `in-article-component` is different to the spec where they asked for `in-article component`

packages/article-skeleton/src/article-skeleton.web.js
vinhlam: in the doc, it say `section_details : <section-title>` is this different to `section : <section-name>`?

bibicollins1 (approved):
LGTM!

PR #2704, opened 2021-05-18
feat(TE-43): InArticlePuff - cleaner section colour implementation

flashcheeks:
This PR updates how the `InArticlePuff` uses `sectionColour`

* Remove inline styles
* Pass `sectionColour` to component styles

This approach does mean we are unable to unit test the `sectionColour` is rendering correctly? Is this a problem?


bibicollins1 (approved):
👍 I imagine the actual section colour mapping has been tested, so all we are doing is passing it through so guess shouldn't a problem not having those tests?

PR #2705, opened 2021-05-24
feat(TE-99): channel slices

flashcheeks:
Add channel slices to Times Components


bibicollins1 (approved):
Looks really good! 👍

vinhlam (approved):
Looks good, but the only thing i would be worried about is keeping this version and the `render` version in sync. If we needed a new type of slice we need to make sure its added to both?

PR #2708, opened 2021-05-26
feat: POC force aspect ratio

flashcheeks:
POC for forcing the aspect ratio of an image in the `InArticlePuff` component


packages/ts-components/src/components/in-article-puff/InArticlePuff.stories.tsx
```diff
@@ -36,7 +36,19 @@ const showcase = {
...
-      name: 'In Article Puff',
+      name: 'In Article Puff - Image',
+      type: 'story'
+    },
+    {
+      component: () => (
+        <FetchProvider url="https://gobble.timesdev.tools/deck/api/deck-post-action/41548">
+          <InArticlePuff
+            sectionColour="#13354e"
+            forceImageAspectRatio={'3:2'}
```
flashcheeks, line 47: If we enable storybook knobs, this could be a dynamic value

packages/ts-components/src/components/in-article-puff/InArticlePuff.stories.tsx
```diff
@@ -36,7 +36,16 @@ const showcase = {
...
         </FetchProvider>
       ),
-      name: 'In Article Puff',
+      name: 'In Article Puff - Image',
+      type: 'story'
+    },
+    {
+      component: () => (
+        <FetchProvider url="https://gobble.timesdev.tools/deck/api/deck-post-action/41548">
+          <InArticlePuff sectionColour="#13354e" forceImageAspectRatio="3:2" />
```
flashcheeks, line 45: If we enable storybook knobs, this could be a dynamic value

packages/ts-components/src/components/aspect-ratio/styles.ts
```diff
@@ -0,0 +1,33 @@
+import styled, { css } from 'styled-components';
+
+const ratioToPadding = new Map<string, string>([
```
vinhlam, line 3: cant this just be a standard map?
```
const ratioToPadding = {
  "2:3": "150%",
  "4:5": "125%",
  "1:1": "100%",
  "4:3": "75%",
  "3:2": "66.66%",
  "16:9": "56.25%",
};
```

then you can type the keys and use it to ensure only valid aspect ratios are used
```
type AspectRatios = keyof typeof ratioToPadding;

export const AspectRatioContainer = styled.div<{ ratio?: AspectRatios }>`
```

flashcheeks: @vinhlam good shout, I have pulled these out of the styles into `types/` and `utils/`

vinhlam (approved):

bibicollins1 (approved):

PR #2709, opened 2021-05-27
fix:(TE-119) strip html tags from in article puff

bibicollins1:
### Description ###
Import sanitize-html library to strip HTML tags from copy in InArticlePuff component. 
Add sanitiseHtml property to make html tag stripping optional.


bibicollins1:
Pulled from another branch instead of master

PR #2710, opened 2021-05-27
fix:(TE-119) Strip HTML from copy in InArticlePuff

bibicollins1:
### Description ###

Add property sanitiseHtml which will strip the HTML tags from the copy in the InArticlePuff component.


bibicollins1:
Has pulled in some work from another branch

PR #2712, opened 2021-05-28
feat(TE-119): InArticlePuff update to sanitiser

flashcheeks:


packages/ts-components/src/components/in-article-puff/InArticlePuff.tsx
```diff
@@ -124,21 +121,30 @@ export const InArticlePuff: React.FC<{
...
-                  fireAnalyticsEvent,
-                  linkText ? linkText : 'Read more'
-                )
+                handleClick(fireAnalyticsEvent, linkText || 'Read more')
               }
               onMouseOver={() => setColour('#696969')}
               onMouseLeave={() => setColour('#BF0000')}
             >
-              <LinkText>{linkText ? linkText : 'Read more'}</LinkText>
+              <LinkText>{linkText || 'Read more'}</LinkText>
```
vinhlam, line 147: {linkText ?? 'Read more'} would be better but I don't think its supported in our version of typescript, same for the above one.

vinhlam (approved):

PR #2717, opened 2021-06-09
feat(TD-49): added meta description to author and topic pages

liam-lewis-news-uk:
**Background**

Currently on topic and author page types the meta description is populated using either the page introduction (topic) or the bio (author)

Those are also shown to readers which means they cannot be optimised for SEO.

Topic Donald Trump | The Times & The Sunday Times 

Profile/author page Mike Atherton | The Times & The Sunday Times 

**Work done**

[Topic & author pages] Updated metadescription to source data from new a new metadescription field in TPA


flashcheeks (commented):
Can you also update the storybook to show the new field in both the topic and author pages

PR #2718, opened 2021-06-09
feat(TE): OPTA football widgets

flashcheeks:

### DIFF
--- a/packages/article-skeleton/src/article-body/article-body.web.js
+++ b/packages/article-skeleton/src/article-body/article-body.web.js
@@ -24,7 +24,9 @@ import {
   InlineNewsletterPuff,
   PreviewNewsletterPuff,
   AutoNewsletterPuff,
-  OptaFootballStandings
+  OptaFootballFixtures,
+  OptaFootballStandings,
+  OptaFootballSummary
 } from "@times-components/ts-components";
 
 import ArticleLink from "./article-link";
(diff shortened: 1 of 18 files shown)


packages/article-skeleton/src/article-body/article-body.web.js
```diff
@@ -178,7 +209,6 @@ const renderers = ({
                   <InArticlePuff
                     sectionColour={theme.sectionColour}
                     forceImageAspectRatio="3:2"
```
vinhlam, line 211: is the santisedhtml required? It doesn't have a default for true does it?

packages/ts-components/src/components/opta/football/summary/OptaFootballSummary.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { showcaseConverter } from '@times-components/storybook';
+
+import { ArticleHarness } from '../../../../fixtures/article-harness/ArticleHarness';
+import { OptaFootballSummary } from './OptaFootballSummary';
+
+const showcase = {
```
vinhlam, line 7: Not for this PR but maybe convert this to a standard storybook so we don't have to rely on `@times-components/storybook`  (using StoryOf works but not the modern format).

packages/ts-components/src/components/opta/utils/config.ts
```diff
@@ -0,0 +1,79 @@
+declare global {
```
vinhlam, line 1: does this break other things in the window? Should it be in a `*.d.ts` file?

vinhlam (approved):
Nothing that needs to be done to merge.

bibicollins1 (approved):
LGTM!

packages/article-skeleton/src/article-body/article-body.web.js
flashcheeks: I'll add it back

packages/ts-components/src/components/opta/utils/config.ts
flashcheeks: It does not seem to work in the `*.d.ts` file? Not sure if I need to import/export though 😕

PR #2729, opened 2021-06-22
Feat:(TE-54) in article galleries analytics

vinhlam:
https://nidigitalsolutions.jira.com/browse/TE-54


bibicollins1 (approved):
Just wondering @vinhlam whether there was a requirement to distinguish between regular, large and small image carousel? Perhaps a question for Ian later down the line. Otherwise looks really good!

vinhlam:
> Just wondering @vinhlam whether there was a requirement to distinguish between regular, large and small image carousel? Perhaps a question for Ian later down the line. Otherwise looks really good!

I think Ian was going to go back and ask about the other question i head about the heading as well as the clicks on the actual `pages` as well as the `left` and `right` buttons, so probably ask him to add it to the list

PR #2732, opened 2021-07-06
fix(TE-60): opta football updates

flashcheeks:
* Add `fullWidth` styling
* Update `useEffect` dependencies


packages/ts-components/src/components/opta/football/match-stats/OptaFootballMatchStats.tsx
```diff
@@ -17,45 +17,49 @@ export const OptaFootballMatchStats: React.FC<{
...
-        setIsReady(true);
-      }
-    });
-  }, []);
+          initComponent();
+          setIsReady(true);
+        }
+      });
+    },
+    [ref]
```
vinhlam, line 58: does this need to be a callback `ref` for it to work? the value of the ref itself never changes.

flashcheeks: The value of `ref` changes when I use the component in Render. I think there is re-rendering in the parent component that would cause `ref` to change. I will memorize the component instead.

flashcheeks: @vinhlam yes using `React.memo` works. When importing in Render I do not get re-rendering. Would you mind reviewing again please.

vinhlam (approved):

PR #2733, opened 2021-07-12
feat:Text Cropping Component

vinhlam:
So we can develop with the figma text cropping design system


bibicollins1 (approved):
Looks good!!

PR #2734, opened 2021-07-16
fix (TSEO-2): Issue where SEO description is not showing up on articl…

virendrapatelnuk:
…e pages

Issue where SEO description is not showing up on article pages


packages/article-skeleton/src/head.web.js
```diff
@@ -102,7 +102,7 @@ function Head({ article, logoUrl, paidContentClassName }) {
 
   const publication = PUBLICATION_NAMES[publicationName];
   const authorName = getAuthorAsText(article);
-  const desc =
+  const desc = article.seoDescription ? article.seoDescription :
```
vinhlam, line 105: Feels like something that requires a unit test. This is also breaking linting, so make sure you fix that as well.

vinhlam: change for linting good, but I am expected a test in `head.test.web.js` with something that covers your new case.
eg.
```
    const testRenderer = TestRenderer.create(
      <Head
        article={{...article, seoDescription:'somevalue' }}
```

virendrapatelnuk: ok got it.

flashcheeks (approved):

packages/article-skeleton/src/head.web.js
flashcheeks: Very minor: You could drop `article` from `article.seoDescription` if you add `seoDescription` to the destructuring expression above.

packages/article-skeleton/src/head.web.js
```diff
@@ -102,10 +102,8 @@ function Head({ article, logoUrl, paidContentClassName }) {
 
   const publication = PUBLICATION_NAMES[publicationName];
   const authorName = getAuthorAsText(article);
-  const desc =
-    Array.isArray(descriptionMarkup) && descriptionMarkup.length
-      ? renderTreeAsText({ children: descriptionMarkup })
-      : null;
+  const desc = article.seoDescription || (Array.isArray(descriptionMarkup) && descriptionMarkup.length
```
flashcheeks, line 105: Very minor: You could drop article from article.seoDescription if you add seoDescription to the destructuring expression above.

vinhlam (approved):

packages/article-skeleton/src/head.web.js
virendrapatelnuk: yes! thanks, @flashcheeks  I'll remember this next time

virendrapatelnuk: fixed

PR #2735, opened 2021-07-16
Fix/carousel data structure

bibicollins1:
In this PR: 
- Update the Carousel to take the data structure from deck tooling.
- Update storybook to use deck data and no longer use showcase converter.
- Update InArticlePuff to also no longer use showcase converter.
- Add sanitiseCopy function to helpers as is now used in InArticlePuff and GalleryCarousel 
- Add gallery carousel to article body interactive switch case


packages/ts-components/src/components/carousel/GalleryCarousel.tsx
```diff
@@ -10,31 +10,23 @@ import {
...
-  credit: string;
-  image: string;
-};
-
 const CustomPagination: React.FC<{
   activePage: number;
   current: number;
   onClick: (current: number, label?: string) => number;
-  data: DataObj;
+  data: CarouselDataObj[];
```
vinhlam, line 29: You'll need to update this when you integration with the new fetch, but I think you should call this `GalleryData`

packages/ts-components/src/components/carousel/types.ts
```diff
@@ -0,0 +1,32 @@
+export type DataObj = {
```
vinhlam, line 1: Not sure if this is the same as the previous one by should be `GalleryDeckData` `GalleryDeckField` etc. Have a look at what I did there. I have a type called DeckData, and GalleryDeckData should be `DeckData<GalleryFields,GalleryData>`  or something like that. I don't like the use of `Obj` in the type name, usually redundant.

packages/ts-components/src/components/carousel/GalleryCarousel.tsx
```diff
@@ -71,20 +64,49 @@ const CustomPagination: React.FC<{
...
+
+  if (error) {
+    return null;
+  }
+
+  const { headline, label, size } = data.fields;
+  const carouselData = data.body.data;
+
+  const isSmall = (carouselSize: string) => {
+    if (carouselSize === '4033') {
```
vinhlam, line 97: Are these the same as the `Infocard` sizes? If so, should share the same `enum` as them

vinhlam (approved):

PR #2736, opened 2021-07-19
Fix(TE-135): Productionise in article puffs

bibicollins1:


vinhlam (approved):
Code changes look good, just fix the failing test then good to go.

PR #2744, opened 2021-07-22
Fix/TSEO-15 article page schema updates

virendrapatelnuk:


packages/article-skeleton/src/head.web.js
```diff
@@ -35,21 +35,79 @@ function getSectionName(article) {
...
       acc.push(...byline.byline);
     } else {
       acc.push(byline.byline);
     }
     return acc;
   }, []);
+}
 
-  return renderTreeAsText({ children });
+function getAuthorsAndInlines(
```
virendrapatelnuk, line 56: could update graphql query to get authors instead of manipulating bylines

flashcheeks (approved):

PR #2749, opened 2021-07-26
fix(te-000): update olympic schedule scroll

bibicollins1:
Enable scroll on the olympics schedule widget. -- Remove show all button as is no longer required, and adjust footer height. 
Screenshot 2021-07-26 at 11 33 44
Screenshot 2021-07-26 at 11 43 04

### DIFF
--- a/packages/ts-components/src/components/olympics/__tests__/__snapshots__/OlympicsSchedule.test.tsx.snap
+++ b/packages/ts-components/src/components/olympics/__tests__/__snapshots__/OlympicsSchedule.test.tsx.snap
@@ -22,7 +22,7 @@ exports[`<OlympicsSchedule> renders 1`] = `
 }
 
 .c0 .pa-schedule .pa_UnitListView_ctr {
-  max-height: 400px;
+  height: 400px;
 }
 
 .c0 .pa-schedule .pa_LoadingOverlayContainer_ctr .pa_UnitListView_ctr {
@@ -116,7 +116,7 @@ exports[`<OlympicsSchedule> renders outside of article 1`] = `
 }
 
 .c0 .pa-schedule .pa_UnitListView_ctr {
-  max-height: 400px;
+  height: 400px;
 }
 
 .c0 .pa-schedule .pa_LoadingOverlayContainer_ctr .pa_UnitListView_ctr {
--- a/packages/ts-components/src/components/olympics/__tests__/__snapshots__/OlympicsSchedule.test.tsx.snap
+++ b/packages/ts-components/src/components/olympics/__tests__/__snapshots__/OlympicsSchedule.test.tsx.snap
@@ -146,19 +146,18 @@ exports[`<OlympicsSchedule> renders 1`] = `
   font-size: 16px;
 }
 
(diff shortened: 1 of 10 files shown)


packages/ts-components/src/components/olympics/__tests__/__snapshots__/OlympicsSchedule.test.tsx.snap
```diff
@@ -269,6 +114,10 @@ exports[`<OlympicsSchedule> renders with wrapHelmetProvider 1`] = `
   font-size: 16px;
 }
 
+.c0 .pa-schedule .pa_UnitListView_ctr {
+  max-height: 400px;
```
vinhlam, line 118: just wondering whether it can just be height, so there is no layout shift going between no-data and with data

packages/ts-components/src/components/olympics/schedule/OlympicsSchedule.stories.tsx
```diff
@@ -17,7 +17,7 @@ storiesOf('Typescript Component/Olympics', module)
         staging: 'https://olympics-embed-staging.pamedia.io',
         prod: 'https://olympics-embed.pamedia.io'
       },
-      'https://olympics-embed-staging.pamedia.io'
+      'https://olympics-embed-staging.pamedia.io/'
```
vinhlam, line 20: I think i add a `/` and not check if there is one, so this might result in an extra `/` in the url

vinhlam (approved):

packages/ts-components/src/components/olympics/schedule/OlympicsSchedule.stories.tsx
vinhlam: not really a problem as its only for storybook

packages/ts-components/src/components/olympics/__tests__/__snapshots__/OlympicsSchedule.test.tsx.snap
bibicollins1: So I only added this to avoid 
Screenshot 2021-07-26 at 11 33 16
Unless we can make the no events part more central?

packages/ts-components/src/components/olympics/schedule/OlympicsSchedule.stories.tsx
bibicollins1: Ah okay will update that 👍

packages/ts-components/src/components/olympics/__tests__/__snapshots__/OlympicsSchedule.test.tsx.snap
bibicollins1: Or happy to leave as above if that works better for switching between days. Maybe something to bring up with Scott?

packages/ts-components/src/components/olympics/schedule/OlympicsSchedule.stories.tsx
vinhlam: sorry, I did it in my commit as I had to change that file anyway

PR #2750, opened 2021-07-27
Fix(TSEO-21):Added sameAs attributes to author schema markup

virendrapatelnuk:
the author schema markup is lacking the sameAs attributes. It should contain the Twitter handler and link to the profile


packages/article-skeleton/src/head.web.js
```diff
@@ -56,13 +56,26 @@ function getAuthors({ bylines }) {
...
   return bylines
-    ? getAuthors(article).map(({ name, jobTitle, twitter, slug }) => ({
-        "@type": "Person",
-        name,
-        jobTitle,
-        url: `https://thetimes.co.uk/profile/${slug}`,
-        sameAs: `https://twitter.com/${twitter}`
-      }))
+    ? getAuthors(article).map(({ name, jobTitle, twitter, slug }) => {
+        const person = {
```
flashcheeks, line 60: You can simplify this:
```
const url = `https://thetimes.co.uk/profile/${slug}`;

return {
  "@type": "Person",
  name,
  jobTitle,
  sameAs: twitter ? [url, `https://twitter.com/${twitter}`] : url
})
```

flashcheeks (changes requested):

packages/article-skeleton/src/head.web.js
virendrapatelnuk: yes, Thanks! I could do that.

flashcheeks (approved):

PR #2752, opened 2021-07-27
Fix/(TSEO-7):Added Video Schema Markup

virendrapatelnuk:


packages/article-skeleton/src/head.web.js
```diff
@@ -173,6 +174,22 @@ function Head({ article, logoUrl, paidContentClassName }) {
     Object.assign(jsonLD, { author: authors });
   }
 
+  let videoJsonLD = null;
```
flashcheeks, line 177: Can we make this a `const` and the value conditional, for example:

`const videoJsonLD = (hasVideo) ? {...} : null;`

flashcheeks (changes requested):

packages/article-skeleton/src/head.web.js
virendrapatelnuk: done

flashcheeks (approved):

PR #2757, opened 2021-07-29
fix(te-000): Revert medal position change 

bibicollins1:
### Description
Now we can see live data on the Olympics schedule, can see that the medal in the right hand side overlaps both 'Live' status and 'finished' status so have to move it back to the original left position. 
Also the text in the 'Live' status was very high up so have centralised it. -- More styling issues in my opinion with 'Live' and 'Finished' but that can be added later. 

### Before 
Screenshot 2021-07-29 at 09 11 06
### After
Screenshot 2021-07-29 at 09 24 39
Screenshot 2021-07-29 at 09 24 42

### DIFF
--- a/packages/ts-components/src/components/olympics/schedule/styles.ts
+++ b/packages/ts-components/src/components/olympics/schedule/styles.ts
@@ -55,13 +55,11 @@ export const Container = styled.div<{
         .pa_UnitListView_unit-time {
           color: ${olympicColour};
           line-height: 30px;
-          position: absolute;
         }
 
         .pa_UnitListView_unit-text {
           font-family: ${fonts.supporting};
           font-size: 16px;
-          margin-right: 50px;
         }
 
         .pa_UnitListView_discipline {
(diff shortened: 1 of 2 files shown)


packages/ts-components/src/components/olympics/schedule/styles.ts
```diff
@@ -70,15 +68,9 @@ export const Container = styled.div<{
...
-          position: relative;
-          left: calc(100% - 65px);
-          align-self: center;
-          @media only screen and (min-width: 321px) {
-            left: calc(100% - 75px);
-          }
-        }
+      }
+      .pa_UnitListView_status-live {
+        padding: 9px 20px 5px 20px;
```
vinhlam, line 73: 20px on the end here is redundant

vinhlam (approved):

packages/ts-components/src/components/olympics/schedule/styles.ts
bibicollins1: Oh yeah true! 👍

flashcheeks (approved):

PR #2765, opened 2021-08-06
feat/TE-155_infocard_bulletpoints

Srinivasan-Ramasamy:
Please review and suggest if we need any changes.


packages/ts-components/src/components/in-article-info-card-bulletpoints/InfoCardBulletPoints.tsx
```diff
@@ -0,0 +1,93 @@
...
+    );
+  }
+
+  if (error || data === undefined) {
+    return null;
+  }
+
+  const { headline, label } = data.fields;
+  const infoCardData = data.body.data;
+  const [showAll, setReadMore] = useState(false);
```
vinhlam, line 48: The setter of `showAll` should be `setReadMore`, this will make the code confusing to read

packages/ts-components/src/components/in-article-info-card-bulletpoints/InfoCardBulletPoints.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const handleReadMore = () => {
+    setReadMore(!showAll);
+  };
+
+  const readMoreRef = useRef<HTMLDivElement>();
+  const [readMore, showReadMore] = useState(false);
```
vinhlam, line 54: Same as previous but this will actually confuse even further. `readMore` value should be set by a setter called `setReadMore`

packages/ts-components/src/components/in-article-info-card-bulletpoints/InfoCardBulletPoints.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const maxHeight = 350;
+  useEffect(() => {
+    const listContainer = readMoreRef.current;
+    if (listContainer) {
+      showReadMore(listContainer.clientHeight > maxHeight);
```
vinhlam, line 59: Do to the issues above, you are not showing read more here

packages/ts-components/src/components/in-article-info-card-bulletpoints/styles.ts
```diff
@@ -0,0 +1,169 @@
...
+    &:active {
+      border: 1px solid ${colours.functional.action};
+      color: ${colours.functional.action};
+    }
+  }
+`;
+
+export const ListContainer = styled.div<{
+  showAll: boolean;
+  ref: any;
```
vinhlam, line 115: should be typed, we need to avoid using `any` when possible

packages/ts-components/src/components/in-article-info-card-bulletpoints/InfoCardBulletPoints.tsx
vinhlam: based on reading the code below, I think it should be 
```
const [showReadMore, setShowReadMore] = useState(false);
```

packages/ts-components/src/components/in-article-info-card-bulletpoints/InfoCardBulletPoints.tsx
Srinivasan-Ramasamy: Thanks Vinh, I've done the changes to show the differentiate in naming convention and functionality. Sorry.

packages/ts-components/src/components/in-article-info-card-bulletpoints/InfoCardBulletPoints.tsx
Srinivasan-Ramasamy: Thanks Vinh, done the changes as per your suggestion.

packages/ts-components/src/components/in-article-info-card-bulletpoints/InfoCardBulletPoints.tsx
Srinivasan-Ramasamy: Thanks Vinh, done the changes.

packages/ts-components/src/components/in-article-info-card-bulletpoints/styles.ts
Srinivasan-Ramasamy: Thanks Vinh, ref is not using in style file, so removed and done the necessary changes.

packages/ts-components/src/components/in-article-info-card-bulletpoints/__tests__/InfoCardBulletPoints.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+  it('should render the initial loading state correctly', () => {
+    (useFetch as jest.Mock).mockReturnValue({ loading: true });
+    const { asFragment } = render(
+      <InfoCardBulletPoints sectionColour="#636C17" />
+    );
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('should render the initial read more state correctly', () => {
+    (useFetch as jest.Mock).mockReturnValue({ showAll: true });
```
bibicollins1, line 101: Do you use this? Can't find it anywhere else in the code.

packages/ts-components/src/components/in-article-info-card-bulletpoints/styles.ts
```diff
@@ -0,0 +1,168 @@
...
+  font-family: ${fonts.supporting};
+  font-weight: 500;
+  line-height: 20px;
+  border: 1px solid ${colours.functional.primary};
+  background: transparent;
+  margin: 15px 0;
+  padding: 10px 12px 5px;
+  cursor: pointer;
+
+  @media (hover: hover) {
```
bibicollins1, line 98: Do you need the @media(hover: hover) here? Think it will still work with just &:hover {}? Might be something I am unaware of though

bibicollins1: And for active below same applies

packages/ts-components/src/components/in-article-info-card-bulletpoints/__tests__/InfoCardBulletPoints.test.tsx
Srinivasan-Ramasamy: thanks bibi, this is suppose to be removed after covering the Vinh's feedback changes. I've done the changes.

packages/ts-components/src/components/in-article-info-card-bulletpoints/styles.ts
Srinivasan-Ramasamy: Thanks Bibi, actually this code written to differentiate the touch and normal device hover/active style changes as per the Danielle design feedback.

Now I've handled in other way. Changes are done.

packages/ts-components/src/components/in-article-info-card-bulletpoints/styles.ts
```diff
@@ -94,19 +94,9 @@ export const ReadMoreButton = styled.button`
...
-      color: ${colours.functional.action};
-    }
-  }
-
-  @media (hover: none) {
-    &:active {
-      border: 1px solid ${colours.functional.action};
-      color: ${colours.functional.action};
-    }
+  &:active {
```
bibicollins1, line 97: Not sure if I misunderstood what you meant in the previous comments about this, so do you not want a hover state? - What was Danielle's requirement?

Srinivasan-Ramasamy: Actually, this button will visible in only mobile view. In Mobile no hover behaviour, it will be tap (touch). So if we give hover then, hover style will sticky until further tap. That's why I've used active, it will behave as they expect on tap [style will apply only on tap]. 

Sorry, I may confuse you :). If you need more detail, I can explain you again in details.

bibicollins1 (approved):
Looks good to me, forgot that you wouldn't see read more in desktop etc so using active makes sense. 👍

PR #2766, opened 2021-08-06
feat(TE-125) - Inline Related Article placement

vinhlam:


packages/article-skeleton/src/article-body/article-body.web.js
```diff
@@ -310,6 +311,19 @@ const renderers = ({
       />
     );
   },
+  autoInlineRelatedArticles(key, { element }) {
```
bibicollins1, line 314: Just a quick thought, are we calling these InlineRelatedArticles or InArticleRelatedArticles? Just wondering if we want to make it consistent?

packages/article-skeleton/__tests__/web/inline-related-article.test.js
```diff
@@ -0,0 +1,168 @@
...
+    });
+
+    it("outside paywall", () => {
+      const fourParagraphContenet = [
+        fakeContent(),
+        fakeContent(),
+        fakeContent(),
+        fakeContent()
+      ];
+      expect(contentReducer(fourParagraphContenet)).toEqual(
```
bibicollins1, line 163: Tiny thing, but think there is a misspelling in fourParagraphContenet -- fourParagraphContent?

packages/article-skeleton/src/contentModifiers/inline-related-article.js
```diff
@@ -0,0 +1,95 @@
...
+  },
+  children: []
+});
+
+const insertInlineRelatedArticles = (
+  relatedArticleSlice,
+  inlineRelatedArticlesFlag,
+  { afterParagraph = 7, paragraphPadding = 2 } = {}
+) => articleContent => {
+  console.log(
```
bibicollins1, line 19: Do we need this log?

bibicollins1: And the ones below.. is it for more visibility in prod articles while testing this?

bibicollins1 (approved):
Couple of comments about the logs, but otherwise works great! Good to see lots of article examples and test how the placement works in them 👍

packages/article-skeleton/src/contentModifiers/inline-related-article.js
vinhlam: I think I accidentally added it while debugging some build issues. Will remove, don't think the build will let it through

PR #2767, opened 2021-08-09
Feat/te 133 in article related articles analytics

bibicollins1:
Add the analytics for in article related articles. 
Spreadsheet can be found here for comparison https://docs.google.com/spreadsheets/d/14JamPsFxWCg2gLXZeYSIPAFaJgthWXeU/edit#gid=1612776471

### DIFF
--- a/packages/ts-components/src/components/in-article-related-articles/RelatedArticle.tsx
+++ b/packages/ts-components/src/components/in-article-related-articles/RelatedArticle.tsx
@@ -6,7 +6,7 @@ import {
 } from './styles';
 import { InArticleLink } from '../in-article-link/InArticleLink';
 import { AspectRatio } from '../aspect-ratio/AspectRatio';
-import { TrackingContextProvider } from '../../helpers/tracking/TrackingContextProvider';
+import { useTrackingContext } from '../../helpers/tracking/TrackingContextProvider';
 import { handleClick } from './tracking-helpers';
 
 export type RelatedArticleType = {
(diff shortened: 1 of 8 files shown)


packages/ts-components/src/components/in-article-related-articles/RelatedArticle.tsx
```diff
@@ -27,23 +29,45 @@ export const RelatedArticle = ({
...
-        <RelatedArticlesImageContainer>
-          <a href={link}>
-            <AspectRatio ratio="16:9">
-              <img src={image} />
-            </AspectRatio>
-          </a>
-        </RelatedArticlesImageContainer>
+}: RelatedArticleProps) => {
+  return (
+    <TrackingContextProvider>
```
vinhlam, line 34: if you just need to use `fireAnalyticsEvent`, you can use 
```
  const { fireAnalyticsEvent } = useTrackingContext();
```
instead of using `<TrackingContextProvider>`

packages/ts-components/src/components/in-article-related-articles/SingleRelatedArticle.tsx
```diff
@@ -19,26 +21,42 @@ export const SingleRelatedArticle = ({
...
-        <DatePublication date={publishedTime!} showDay={false} />
-      </div>
-      <div className="byline">{byline}</div>
-    </section>
-  </SingleRelatedArticleContainer>
-);
+}: RelatedArticleProps) => {
+  return (
+    <TrackingContextProvider>
+      {({ fireAnalyticsEvent, intersectObserverRef }) => (
```
vinhlam, line 27: is this `intersectObserverRef` required as its already done in the `InArticleRelatedArticles` component  which encompasses this one? Iff thats the case the the `useTrackingContext` would be prefered for the fireEvent

packages/ts-components/src/components/in-article-related-articles/RelatedArticle.tsx
bibicollins1: Ah great will use that!

vinhlam (approved):

PR #2768, opened 2021-08-10
fix(TSEO-26):added default author

virendrapatelnuk:
When there isn't an author name present, using the organization name as an author.


vinhlam (approved):
Looks good. You may get merge conflicts with one of my changes as I may have renamed `head.web.js` to just `head.js`

PR #2772, opened 2021-08-17
feat(TE-000): add previewData to FetchProvider

flashcheeks:
`FetchProvider` now accepts `previewData` as a short-circuit to fetching.

`previewData` can be used to render Deck components without calling the Deck API.


packages/ts-components/src/helpers/fetch/FetchProvider.tsx
```diff
@@ -2,8 +2,9 @@ import fetch from 'isomorphic-unfetch';
 import React, { createContext, useState, useEffect, useContext } from 'react';
 
 type FetchProviderProps = {
-  url: string;
+  url?: string;
   options?: { [key: string]: any };
+  previewData?: any;
```
vinhlam, line 7: I think `previewData` is of type  `DeckData<F, B>`

in gallery we have
`type GalleryCarouselDeckData = DeckData<GalleryCarouselFields, CarouselDataObj>;`

Is there is way to infer the type from the prop?

vinhlam (approved):

PR #2773, opened 2021-08-18
Feat/te 51 infocard quotes

budalanaveen:
Adding the changes for InfoCard Quote story


packages/ts-components/src/components/in-article-info-card/Quote.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export const Quote = () => {
```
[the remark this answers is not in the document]
vinhlam: I also see this error in the console, but may not be due to your component
```
Error: <svg> attribute height: Expected length, "auto".
```

packages/ts-components/src/components/in-article-info-card/__tests__/__snapshots__/InfoCard.test.tsx.snap
```diff
@@ -49,6 +49,23 @@ exports[`InfoCard renders 1`] = `
                       >
                         Birch
                       </div>
+                      <div
```
vinhlam, line 52: I am a bit concerned that all the existing snapshots for the `text/image infocard` have changed and have a quote added to them.

vinhlam (commented):
There is a bit of refactor I think you need to do because the `infocard.tsx` file is getting too big. This means that the extracted component needs to have its own test. You didn't the abstraction of the component using generics but it did parameterised, which for now is ok.

packages/ts-components/src/components/in-article-info-card/__tests__/__snapshots__/InfoCard.test.tsx.snap
budalanaveen: Hi Vinh. Yes. We have added one flag to all the existing snapshots and based on the flag we will identify and hide the quotes accordingly

bibicollins1: Yeah I think on this, I had imagined you would have an 'InfoCard' or a 'QuoteCard' either of which type would be rendered depending on the data provided. Where this is one dynamic card right? If this is preferred @vinhlam happy to leave it, but wondering whether it is a bit confusing having 3 states for one card? Just as the tooling is quite separate between the two types do we want the cards to be different components.

budalanaveen: Yes Bibi. Initially I have created a new component where there is lot of code duplicated. So I have merged it as a single component.

bibicollins1: Was that what we spoke about last time? Having a new carousel component for it? Or just the card within?

budalanaveen: Yes. Its the same that we have talked earlier. Though I have created the new component for the Quotes and there is lot of duplicate in that Vinh has suggested to use the existing component only

vinhlam: The info card carousel part of it should be a generic component that will hold all the common code. Each of the components eg. TextImageInfoCard and Quotes should use it. There are several ways to do this but I have suggested from the start that it is done using` generics`, i.e. `InfoCard<QuoteCard>`  if possible.

budalanaveen: Hi @vinhlam . So do you want me to create a seperate component for Infocard quote story ?

PR #2774, opened 2021-08-23
fix(TSEO-32): Added page number in meta data

virendrapatelnuk:
For any search query in topic, page numbers are changing but the title and description remain the same. To differentiate and let the search engines know that the page has changed, we suggest adding page numbers at the end in the title and description fields.

Currently if topic is India URL https://www.thetimes.co.uk/topic/india Title: India | The Times & The Sunday Times Description: Discover expert India articles from The Times and The Sunday Times. 

*Suggested: * Title: India | The Times & The Sunday Times | Page Number

Description: Discover expert India articles from The Times and The Sunday Times. Page Number of total pages  

[JIRA LINK](https://nidigitalsolutions.jira.com/browse/TDP-172)


bibicollins1:
Hi @virendrapatelnuk Just wondering about the sentence structure on the description, maybe it's more a case of getting key words if so it's fine, but if not I would think 'Discover expert articles about India from The Times and The Sunday Times. Page Number of total pages' might work a bit better as it would work for topics like Politics but also place names and people names etc? I might just be being pedantic though.

packages/topic/__tests__/web/__snapshots__/topic-loading.web.test.js.snap
```diff
@@ -4,11 +4,10 @@ exports[`1. an article list loading 1`] = `
 <topicArticlesProvider>
   <Helmet>
     <title>
-      
        | The Times & The Sunday Times
     </title>
     <meta
-      content="Discover expert  articles from The Times and The Sunday Times."
+      content="Discover expert  articles from The Times and The Sunday Times. Page 1 of 0"
```
bibicollins1, line 10: Should this be page 1 of 1?

packages/topic/__tests__/web/__snapshots__/topic.web.test.js.snap
```diff
@@ -48,15 +47,14 @@ exports[`3. an article list 1`] = `
 <topicArticlesProvider>
   <Helmet>
     <title>
-      Test Name
-       | The Times & The Sunday Times
+      Test Name | The Times & The Sunday Times | Page 2
     </title>
     <meta
-      content="Test italic bold text."
+      content="Test italic bold text. Page 2 of 1"
```
bibicollins1, line 53: Same here, there are some others on lines 138 and 228

packages/topic/__tests__/web/__snapshots__/topic.web.test.js.snap
```diff
@@ -6,11 +6,10 @@ exports[`2. a loading state 1`] = `
 <topicArticlesProvider>
   <Helmet>
     <title>
-      
        | The Times & The Sunday Times
     </title>
     <meta
-      content="Discover expert  articles from The Times and The Sunday Times."
+      content="Discover expert  articles from The Times and The Sunday Times. Page 1 of 1"
```
bibicollins1, line 12: If there is no topic do we want to remove the empty string space here?

packages/topic/__tests__/web/__snapshots__/topic-loading.web.test.js.snap
virendrapatelnuk: @bibicollins1 Yes! it's my bad. I'll change it.

virendrapatelnuk: @bibicollins1 
Done, kindly review.

PR #2789, opened 2021-09-02
revert: remove sanitize html

flashcheeks:
revert the previous commit and add tests


vinhlam (approved):

packages/ts-components/src/helpers/text-formatting/__tests__/SanitiseCopy.test.tsx
```diff
@@ -0,0 +1,23 @@
+import { sanitiseCopy } from '../SanitiseCopy';
+
+describe('<SanitiseCopy>', () => {
```
vinhlam, line 3: I usually only put the `<>` if its a react component.

flashcheeks: 👍 yes, copy and paste typo

PR #2790, opened 2021-09-07
Feat/TDP-121 Inarticle Bignumbers Analytics

Srinivasan-Ramasamy:
Please review and suggest if we need to do any changes. Thank you.

### DIFF
--- a/packages/ts-components/src/components/in-article-big-numbers/__tests__/BigNumbers.test.tsx
+++ b/packages/ts-components/src/components/in-article-big-numbers/__tests__/BigNumbers.test.tsx
@@ -3,6 +3,7 @@ import mockDate from 'mockdate';
 import { cleanup, fireEvent, render } from '@testing-library/react';
 import { BigNumbers } from '../BigNumbers';
 import { useFetch } from '../../../helpers/fetch/FetchProvider';
+import FakeIntersectionObserver from '../../../test-utils/FakeIntersectionObserver';
 import '@testing-library/jest-dom';
 import 'regenerator-runtime';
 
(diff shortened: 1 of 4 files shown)


packages/ts-components/src/components/in-article-big-numbers/__tests__/BigNumbers.test.tsx
```diff
@@ -178,12 +179,27 @@ describe('BigNumbers', () => {
...
   });
 
-  it('click show all', async () => {
-    const { asFragment, getByText, findByText } = render(
-      <BigNumbers sectionColour="#636C17" />
-    );
-    fireEvent.click(getByText('Show all'));
-    await findByText('Collapse');
-    expect(asFragment()).toMatchSnapshot();
+  describe('tracking', () => {
```
bibicollins1, line 182: This doesn't actually seem to be testing the analytics calls?

bibicollins1 (changes requested):
I don't see any analytics calls happening in the storybook, if you check the other components in the actions tab on the right you should be getting an object back called analytics-action which will contain the tracking. Maybe adding some tests for the tracking will help debug this

packages/ts-components/src/components/in-article-big-numbers/__tests__/BigNumbers.test.tsx
Srinivasan-Ramasamy: Hi @bibicollins1, I've implemented the same way which other story implementation. Analytics call is happening Bibi. Please check this screen shot and please correct me is I'm wrong.


Screenshot from 2021-09-13 15-11-59
Screenshot from 2021-09-13 15-12-16
Screenshot from 2021-09-13 15-12-09

bibicollins1: Oh strange they weren't happening with mine when I ran it, will try again. Could we still get some tests for the tracking though? The describe block is tracking but I think we are only testing when you click read more button

bibicollins1, line 182: Resolved part of this with the analytics showing in storybook but could we still get some tests on the tracking as you have added a tracking describe block. I don't think click show all needs to be in the tracking describe block and can go back to where it was before.

Srinivasan-Ramasamy: Hi **@bibicollins1**, Thanks you. UI Changes are done, updated the tests on the tracking part. Please review and suggest if you need any more changes.

bibicollins1: Tests look good thanks @Srinivasan-Ramasamy, would it be possible to add a test to show the tracking shows interactive when there is a show all button?

Srinivasan-Ramasamy: Thanks @bibicollins1, Its need some more logic has to be done in unit testing.  Its more tricky one.

Srinivasan-Ramasamy: Hi @bibicollins1, Good Morning.
Reg#2790 Feedback comments "Add a test to show the tracking shows interactive when there is a show all button". I've tried all the way, still facing issues to get more details to the unit test file to show interactive/static. Its needs a calculation to get clientHeight and card size -  but in unit test file not able to get all those details. Please suggest?

bibicollins1 (approved):

PR #2793, opened 2021-09-23
chore: remove commentCount

flashcheeks:
* Remove `commentCount` from TPA request
* Remove prop from components
* Remove prop from fixtures and tests


bibicollins1 (approved):
LGTM!

PR #2794, opened 2021-09-23
feat(scb-1193): add aditional in article ad slots

nikkibyrne:
Remove hidden styling of the new ad slots so the container does not collapse and respects the min height


bibicollins1 (approved):
Looks good to me

flashcheeks (approved):

PR #2814, opened 2021-11-12
feat(TDP-478): Article LIVE flag

flashcheeks:


packages/ts-components/src/components/article-flag/LiveArticleFlag.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { Container, BulletContainer, ContentContainer, Title } from './styles';
+
+export const LiveArticleFlag: React.FC = () => (
```
vinhlam, line 4: This needs to be tested. Not just a snapshot test. Check that when it has been mounted that it contains the word `LIVE` in it.

vinhlam (changes requested):

packages/ts-components/src/components/article-flag/LiveArticleFlag.tsx
othetgyi: I've added a unit test to check for the word `LIVE`.

PR #2823, opened 2021-12-08
TDP-1024 - update im posted event name and added new tests for it

vinhlam:
Based on this, I think we were listening on the wrong event
https://developers.openweb.com/docs/event-listener-reference


bibicollins1 (approved):
LGTM 👍

PR #2827, opened 2021-12-21
feat(TDP-588): replaced 2 packages with packages that we already use …

adamosborne-tnl:
…in render and TC and updated to packages used to the latest versions

replaced sanitize-html with xss, the same as we did with CPS-Render
replaced luxon with date-fns
updated date-fns to latest versions


yarn.lock
```diff
@@ -24729,6 +24665,14 @@ xregexp@4.0.0:
   resolved "https://registry.yarnpkg.com/xregexp/-/xregexp-4.0.0.tgz#e698189de49dd2a18cc5687b05e17c8e43943020"
   integrity sha512-PHyM+sQouu7xspQQwELlGwwd05mXUFqwFYfqPO0cC7x4fxyHnnuetmQr6CjJiafIDoH4MogHb9dOoJzR/Y4rFg==
 
+xss@^1.0.10:
```
vinhlam, line 24668: any reason why there are two versions of xss in the lock file?

flashcheeks (approved):

yarn.lock
adamosborne-tnl: one version is the one I stuck in as part of my changes, unfortunately the other one is a dependency of **apollographql**

PR #2831, opened 2022-01-05
feat(TDP-479): Add article breaking flag

othetgyi:
This PR adds the breaking flag as well as refactors the existing article flag into a TypeScript component within the `ts-components` package. The original `article-flag` package has been deleted. 

[TDP-479](https://nidigitalsolutions.jira.com/browse/TDP-479)

Breaking flag in Storybook:
image


packages/article-flag/README.md
```diff
@@ -1,40 +0,0 @@
-# Article Flag
```
bibicollins1, line 1: Hey! Just wondering whether some of this information might be useful in the ts-components package where they now live - Maybe just the initial explanation?

packages/provider-test-tools/fixtures/article.json
```diff
@@ -1125,25 +1125,29 @@
     }
   ],
   "expirableFlags": [
+    {
+      "type": "BREAKING",
+      "expiryTime": "2030-03-13T13:00:00.000Z"
```
DMackintosh-TNL, line 1130: Why are the expiry times so far in the future?
Why do we have expiry times at all?

Would it make more sense in the future to have these based more around a user based interaction?

DMackintosh-TNL (approved):
WOW! This is great, nice work consolidating the code. I'm not fully up to speed with this aspect of the codebase, but this looks good from a coding perspective :)

packages/article-flag/README.md
DMackintosh-TNL: Good point :)

othetgyi: Good point - I've added the explanatory bit about the article flags to the `ts-components` package's README.

packages/provider-test-tools/fixtures/article.json
othetgyi: The expiry times in the article mimics the data that would be received from TPA or CAPI so we can test the flags show or don't show as appropriate. I've extended the `expiryTime` in keeping with how the expiry times were set three years ago to make sure the flags are all in date.

Since the flags are set by editorial in their CMS we don't control when a flag is considered `new` or `live` so are limited to making sure the flags show up when they come through in article data from TPA or CAPI.

bibicollins1 (approved):

PR #2833, opened 2022-01-06
feat(TDP-805): No commenting for RA users

vinhlam:


vinhlam:
> Store URLs should probably come from Render, but hopefully won't be changed any time soon!! 😬

I originally was going to do the passing from rendering but decided not to as it would generate a lot of clutter across times components just to get the data to the comments, we need to find a better way to pass data from render to times components. The store urls are not all pulling from the config in render consistently either, eg. the modals use their own versions. Maybe do a refactor to make all of these consistent in the future.

PR #2834, opened 2022-01-07
fix(TPD-997): Remove inline related articles flag

bibicollins1:
## Description

- The inline related articles feature flag is only working on the origin server and not through the proper prod urls. The query string is ignored by the CDN and you are served a cached page
- So for the experiment they asked that we remove the flag and instead wrap it in a display: none that can then be targeted with a class name 'inlineRelatedArticles' and set display: true in their variations.

[TPD-997](https://nidigitalsolutions.jira.com/browse/TDP-997)


packages/article-skeleton/src/article-body/article-body.js
```diff
@@ -365,11 +365,13 @@ const renderers = ({
     return (
       <Context.Consumer key={key}>
         {({ theme }) => (
-          <InArticleRelatedArticles
-            heading="Related Articles"
-            relatedArticles={element.attributes.relatedArticles}
-            sectionColour={theme.sectionColour}
-          />
+          <div style={{ display: "none" }} className="inlineRelatedArticles">
```
vinhlam, line 368: Wondering whether this would be better as a StyledComponent. I'd prefer not to have inline styles.  you could just call it `HiddenDiv` and just have `styled.div`display:none` etc. I know this is only for an expermiment though so might not be worth the hassle. Ill pre-approve. You can merge either way.

vinhlam (approved):

packages/article-skeleton/src/article-body/article-body.js
vinhlam: I think if you specifically specifcy a className, it shouldn't `hash` it

PR #2849, opened 2022-02-10
chore(TDP-759): Add hover and active state to links in key moments component

bibicollins1:
### Description

- Design wanted to add a hover colour state to the links in the key moments component


bibicollins1:
> Does it not affect any snapshots?

@vinhlam Seemingly not...

othetgyi (approved):

PR #2860, opened 2022-02-24
fix/tdp-760-article-divider-headline-fix

bibicollins1:
Add a try catch around headline decoding as sometimes the input isn't encoded properly


d3vAdv3ntur3s (approved):

themischatzis (approved):

flashcheeks (approved):
@bibicollins1 If you want to make this function reusable you could create `ts-components/src/utils/safeDecodeURI.ts` and import it. I think we decode in other components too 😄

PR #2861, opened 2022-02-24
chore(TDP-000): Share safeDecodeURIComponent function

bibicollins1:
We use the safeDecodeURIComponent across other areas of times-components, so have made the try catch function a shared util to avoid the same errors as we had in article header in the other components.


flashcheeks (approved):
💥 looking good!

PR #2867, opened 2022-03-02
feat(TDP-1351): fix article keylines responsive bug

SunshineEleda:
Bug fix for article key lines at 1024px

[JIRA-1351](https://nidigitalsolutions.jira.com/jira/software/c/projects/TDP/boards/1491?modal=detail&selectedIssue=TDP-1351)

Screenshot 2022-03-02 at 13 57 27

Screenshot 2022-03-02 at 14 02 34

### DIFF
--- a/packages/article-main-standard/src/styles/responsive.js
+++ b/packages/article-main-standard/src/styles/responsive.js
@@ -56,9 +56,10 @@ export const MetaContainer = styled(View)`
   width: 20.8333%;
   z-index: 1;
   top: 100%;
+  display: none;
 
-  @media (max-width: ${breakpoints.wide - 1}px) {
-    display: none;
+  @media (min-width: ${breakpoints.wide}px) {
+    display: block;
   }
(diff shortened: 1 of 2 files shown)


DMackintosh-TNL (approved):
We may need to review our breakpoints at a later date... But good work Adele :)

packages/article-main-standard/src/styles/responsive.js
```diff
@@ -57,7 +57,7 @@ export const MetaContainer = styled(View)`
   z-index: 1;
   top: 100%;
 
-  @media (max-width: ${breakpoints.wide}px) {
+  @media (max-width: ${breakpoints.wide - 1}px) {
```
bibicollins1, line 60: Just wondering, know you didn't add it but given everything else is min-width - is it worth swapping it to a min-width wide with all the styling, and then default display: none? Maybe more readable. But also this works so can go ahead with either

PR #2870, opened 2022-03-08
feat(TDP-1139):debugging

SunshineEleda:

### DIFF
--- a/packages/ts-components/src/utils/intersectObserverHook.ts
+++ b/packages/ts-components/src/utils/intersectObserverHook.ts
@@ -5,6 +5,10 @@ export function useIntersectionObserver<T extends Element | null>(
   onIntersect: () => void,
   options: IntersectionObserverInit
 ): void {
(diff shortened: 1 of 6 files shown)


packages/ts-components/src/utils/intersectObserverHook.ts
```diff
@@ -13,6 +17,10 @@ export function useIntersectionObserver<T extends Element | null>(
           (typeof window !== 'undefined' &&
             window.IntersectionObserver &&
             new window.IntersectionObserver(entries => {
+              console.log('XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
+              console.log('XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
+              console.log('XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
+              console.log('intersect obs entries', entries, entries[0], entries[0].isIntersecting)
```
SunshineEleda, line 23: entries[0[.intersecting is returning false which is causing the current issue

PR #2876, opened 2022-03-09
feat(TDP-1472): Small changes to Live/Breaking flag UI 

othetgyi:
This PR makes small adjustments to the LIVE and BREAKING flags on article pages to make the padding above and below the text the same, and make the padding right of the text and left of the icon equal.

[TDP-1472](https://nidigitalsolutions.jira.com/browse/TDP-1472)

Breaking flag on article before UI change:
image

Breaking flag on article after UI change:
image


packages/ts-components/src/helpers/fetch/__tests__/FetchProvider.test.tsx
```diff
@@ -53,14 +53,12 @@ describe('<FetchProvider>', () => {
...
 
-    const { asFragment, findByText } = render(
+    const { findByText } = render(
       <FetchProvider url="http://thetimes.co.uk">
         <TestComponent />
       </FetchProvider>
     );
     await findByText('Loading');
-
-    expect(asFragment()).toMatchSnapshot();
```
flashcheeks, line 63: @othetgyi You shouldn't need to remove this snapshot, what was the issue?

othetgyi: @flashcheeks The snapshot was previously passing incorrectly  -- instead of showing 'loading' it would show 'boo', so we though testing that 'loading' would show would be ok. Also, the test was a bit flaky in that the test seemed to be failing just because of a timing issue.

flashcheeks (approved):

PR #2877, opened 2022-03-10
feat(TDP-1139): fixing article name value

SunshineEleda:

### DIFF
--- a/packages/key-facts/src/key-facts-text.js
+++ b/packages/key-facts/src/key-facts-text.js
@@ -4,6 +4,14 @@ import coreRenderers from "@times-components/markup";
 import props from "./key-facts-text-prop-types";
 import { Text, KeyFactTextLink, BulletContainer, Bullet } from "./styles";
 
+const getTitle = data => {
+  if (data.children.length === 1) return data.children[0].attributes.value;
+
+  const linkText = data.children.map(child => child.attributes.value);
+  const title = linkText.join(" ");
+  return title.lenghth > 0 ? title : " ";
+};
+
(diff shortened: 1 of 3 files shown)


packages/key-facts/src/key-facts-text.js
```diff
@@ -15,17 +15,20 @@ const KeyFactsText = ({ listIndex, keyFactItem, fireAnalyticsEvent }) => (
             ...coreRenderers,
             link(key, attributes, renderedChildren) {
               const { href: url } = attributes;
+              const title = data.children[0].attributes.value
```
DMackintosh-TNL: What happens if this value is empty or null?

DMackintosh-TNL (changes requested):

packages/key-facts/src/key-facts-text.js
SunshineEleda: have added added an or to send an empty string

DMackintosh-TNL (approved):
Brilliant, nice work thank you Adele :)

packages/key-facts/src/key-facts-text.js
```diff
@@ -15,6 +15,7 @@ const KeyFactsText = ({ listIndex, keyFactItem, fireAnalyticsEvent }) => (
             ...coreRenderers,
             link(key, attributes, renderedChildren) {
               const { href: url } = attributes;
+              const title = data.children[0].attributes.value || " ";
```
flashcheeks, line 18: When there is multiple children this will only return the first value. Should you concatenate the values together?

PR #2878, opened 2022-03-10
feat(TDP-1452): updated live blog schema to have live blog updates

adamosborne-tnl:
I have the **LiveBlogPosting** schema to include the incremental **liveBlogUpdate** values in the schema. This is mostly the ArticleHeader divider data provided by these headers 
image
There is also additional information provided by the subsequent paragraphs, images or videos included in the update.

you should be able to validate the schema at https://validator.schema.org/ by looking for the **liveBlogUpdate** data
image


packages/ts-components/src/components/article-header/ArticleHeader.tsx
```diff
@@ -38,13 +38,13 @@ const ArticleHeader: React.FC<{
   const isLessThan1Hour = diffInSeconds < 60 * 60;
   const isLessThan13Hours = diffInSeconds < 60 * 60 * 13;
   const isDaysAgo = differenceInCalendarDays(currentDateTime, updatedDate) >= 1;
-
+  const anchorPoint = `u_${window.btoa(updated)}`;
```
flashcheeks, line 41: `updated` might not be unique. maybe we can add the index value?

packages/article-skeleton/src/head.js
```diff
@@ -130,6 +130,62 @@ const getThumbnailUrl = article => {
   return crop ? crop.url : "";
 };
 
+const getLiveBlogUpdates = (article, publisher, author) => {
+  const updates = [];
+  let update;
+
+  for (let i = 0; i < article.content.length; i += 1) {
```
flashcheeks, line 137: `article.content` also contains a `paywall` object that you need to parse, or has this been normalised already?

flashcheeks (changes requested):

packages/ts-components/src/components/article-header/ArticleHeader.tsx
adamosborne-tnl: Only needs to be unique enough for the page and I somewhat doubt that they would have multiple headlines to the second on the same blog page. I use the same value in the head.js for the schema so didn't want to curry the value that may not be the same elsewhere

packages/article-skeleton/src/head.js
adamosborne-tnl: Hmm, only working on storybook, got any ideas on how i could check that?

adamosborne-tnl: would i need to check though `content` and `paywalledContent` ?

adamosborne-tnl: ok, updated for `paywall`

packages/ts-components/src/components/article-header/ArticleHeader.tsx
DMackintosh-TNL: Should it be back ticked?

adamosborne-tnl: yes, but that's an old line now.

DMackintosh-TNL (approved):
Looks good mate :)

flashcheeks (approved):

PR #2898, opened 2022-04-04
feat(TDP-1747):  commented out message so an override cant display it…

adamosborne-tnl:
… before it's ready

Commented out the Real Name Commenting Message, but left it in the code, so that it can be uncommented when it's needed. It seems that the display:none on the hidden div wasn't sufficient as an unknown override has displayed it to a customer, maybe optomizely

### DIFF
--- a/packages/article-comments/__tests__/web/__snapshots__/shared.web.test.js.snap
+++ b/packages/article-comments/__tests__/web/__snapshots__/shared.web.test.js.snap
@@ -6,64 +6,6 @@ exports[`Render comments label, when comments are loaded 1`] = `
   display: none;
 }
 
-.c2 {
-  display: -webkit-box;
-  display: -webkit-flex;
-  display: -ms-flexbox;
-  display: flex;
-  margin-top: 50px;
-}
-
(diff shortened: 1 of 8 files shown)


bibicollins1:
Just wondering if this solves the issue entirely - comments aren't showing when the banner is so will this just remove the banner also and not re-enable the comments?

adamosborne-tnl:
@bibicollins1 unless i read it wrong, the issue is just the banner being displayed and we aren't ready for that yet, and the comments themselves haven't been affected?

bibicollins1:
Oh, so I saw the banner and think I didn't have access to comments - though I now can't recreate it... sorry!

bibicollins1:
Will keep checking and maybe try to follow up on reports to see if they could see comments or not

bibicollins1:
@adamosborne-tnl Sorry think I misunderstood - have asked Ian just incase whether they still saw the comments when they reported it.

bibicollins1 (approved):
Lgtm!

flashcheeks (approved):

PR #2900, opened 2022-04-05
feat(TDP-1639): live flag analytics param renaming

SunshineEleda:
Storing analytics to 'article_flag' from 'other_details'

[JIRA-1639](https://nidigitalsolutions.jira.com/jira/software/c/projects/TDP/boards/1491?modal=detail&selectedIssue=TDP-1639)
-->

### DIFF
--- a/packages/ts-components/src/components/in-article-puff/InArticlePuff.tsx
+++ b/packages/ts-components/src/components/in-article-puff/InArticlePuff.tsx
@@ -48,7 +48,7 @@ const clickEvent = (buttonLabel: string, isLiveOrBreaking?: string) => ({
   attrs: {
     event_navigation_name: `button : ${buttonLabel}`,
     event_navigation_browsing_method: 'click',
-    other_details: isLiveOrBreaking
+    article_flag: isLiveOrBreaking || 'no flag'
   }
 });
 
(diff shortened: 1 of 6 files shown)


packages/ts-components/src/components/in-article-puff/InArticlePuff.tsx
```diff
@@ -57,12 +57,14 @@ export const InArticlePuff: React.FC<{
   forceImageAspectRatio?: AspectRatios;
   isLiveOrBreaking?: string;
 }> = ({ sectionColour, forceImageAspectRatio, isLiveOrBreaking }) => {
+  const articleFlag = isLiveOrBreaking || 'no flag';
   const handleClick = (
     fireAnalyticsEvent: (evt: TrackingContext) => void,
-    buttonLabel: string
+    buttonLabel: string,
+    articleFlag: string
```
flashcheeks, line 64: @SunshineEleda this function is within the main scope of the component so it already has access to `articleFlag` there is no need to pass it.

flashcheeks (changes requested):
@SunshineEleda Are `key-facts` and `article-puff` the only components that need this analytics?

SunshineEleda:
> @SunshineEleda Are `key-facts` and `article-puff` the only components that need this analytics?

some other components use it but for logic not analytics

PR #2908, opened 2022-04-06
feat(TDP-1714): remove react native styleguide

SunshineEleda:
### Changes
- Created new TcScrollView component to replace react native ScrollView
- Adapted TcView and TcText to show styles passed in as props
- Amendment to tests

[JIRA-1714](https://nidigitalsolutions.jira.com/jira/software/c/projects/TDP/boards/1491?modal=detail&selectedIssue=TDP-1714)


packages/styleguide/__tests__/shared.js
```diff
@@ -1,10 +1,18 @@
-import { Text } from "react-native";
+import 'jest-styled-components'
+import { TcText} from "@times-components/utils";
+import { mount } from "enzyme";
```
fayeez-tnl, line 3: Hey @SunshineEleda was it easier to use enzyme here rather than use react testing library?

packages/styleguide/__tests__/shared.js
```diff
@@ -39,13 +47,14 @@ const tests = [
...
-      const testInstance = TestRenderer.create(
+      const testInstance = mount(
         <Animations.FadeIn>
-          <Text>Hello World</Text>
+          <TcText>Hello World</TcText>
         </Animations.FadeIn>
       );
 
-      expect(testInstance.toJSON()).toMatchSnapshot();
+ expect(testInstance.find('animations__FadeIn-sc-2iiu4y-0 iZkZiB'))
```
fayeez-tnl, line 56: Hmm, I'm not sure about this selector since it's just a hashed classname, could you add a testID instead or use a more robust one?

fayeez-tnl (commented):
Thanks @SunshineEleda ! Just have a question and a suggestion.

packages/styleguide/__tests__/shared.js
SunshineEleda: yes, I didn't realise it was a requirement at this stage but if it would be better moving forward would be fine with updating the testing framework

fayeez-tnl (approved):

bibicollins1 (approved):

PR #2909, opened 2022-04-06
feat(TDP-1705): remove react native from link

bibicollins1:
Remove react native code and dependencies from the link package.
Future to do: 
- Remove 'web' from test naming, file paths and yarn test commands - probably best to wait until all the packages don't contain react-native as the jestConfigurator is used in the majority of these packages.
- Change babel presets to not be react-native.
- Change onPress property names to onClick - have changed in Link and TextLink so onClick is called, however the function passed down is still named onPress as would require changes in lots of packages.


fayeez-tnl (commented):
Great work for adding react-testing-library in here! One thing I'd like to try and get us to think about more is when to use snapshot tests - in general they can be really brittle to maintain and not easy to understand the intent of them all of the time. Good article by Kent C Dodds goes into some nice use cases for it: https://kentcdodds.com/blog/effective-snapshot-testing

adamosborne-tnl (approved):

PR #2916, opened 2022-04-19
fix(TDP-1777): Make analytics article_flag data lower case

bibicollins1:
The value for article_flag if a live or breaking article was coming through as uppercase. This PR changes it to lower case.
[TDP-1777](https://nidigitalsolutions.jira.com/browse/TDP-1777)

### DIFF
--- a/packages/article-skeleton/src/tracking/article-tracking-context.js
+++ b/packages/article-skeleton/src/tracking/article-tracking-context.js
@@ -50,7 +50,9 @@ export default Component =>
         template: get(data, "template", "Default"),
         registrationType: getRegistrationType(),
         shared: getSharedStatus(),
-        article_flag: getIsLiveOrBreakingFlag(flags) ? getIsLiveOrBreakingFlag(flags).toLowerCase() : "no flag",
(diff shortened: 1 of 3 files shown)


othetgyi (approved):

packages/article-skeleton/src/tracking/article-tracking-context.js
```diff
@@ -50,7 +50,9 @@ export default Component =>
         template: get(data, "template", "Default"),
         registrationType: getRegistrationType(),
         shared: getSharedStatus(),
-        article_flag: getIsLiveOrBreakingFlag(flags) || "no flag",
+        article_flag: getIsLiveOrBreakingFlag(flags)
+          ? getIsLiveOrBreakingFlag(flags).toLowerCase()
+          : "no flag",
         article_template_name: getIsLiveOrBreakingFlag(flags)
```
flashcheeks, line 56: We are now calling `getIsLiveOrBreakingFlag` three time in this file. We should put this into a const before the return statement.

```
const isLiveOrBreaking = getIsLiveOrBreakingFlag(flags);
```

flashcheeks (changes requested):

PR #2936, opened 2022-05-10
feat(TDP-1705): fixed css issue from recent change

adamosborne-tnl:
fixed issue where links were appearing with underlines 
image
 and now they don't
image


adamosborne-tnl:
helps when you leave the right screenshot

image

bibicollins1 (approved):
Nice one!

PR #2942, opened 2022-05-13
feat (tdp-1688): remove react native from article skeleton

bibicollins1:
## Description

Remove React native from article-skeleton package.


adamosborne-tnl (commented):
dang, must remember to press 'submit review'

adamosborne-tnl (approved):

PR #2958, opened 2022-05-31
fix(TDP-1825): Live News: Update Timestamp Logic

DMackintosh-TNL:
### Description
Change #1:
New logic for the timestamp on the Article Divider: Examples below;

All timestamps shall remain the same with the same logic barring the following exceptions
- The 24hr timestamp shall display the correct timezone for non-UK viewers of the page

Currently 12.55 BST OR in winter 12.55 GMT
**Note:** This logic should already exist and be in use for international readers on the publication date

Change #2:
The date stamp logic shall remain as-is, the only change is that we are removing the year;
Currently: mmm dd yyyy (May 31 2022)
Should be: mmm dd (May 31)

[TDP-1825](https://nidigitalsolutions.jira.com/browse/TDP-1825)

### Checklist

Screenshots (from StoryBook):
**GMT**
Screenshot 2022-06-07 at 07 34 13

**BST**
Screenshot 2022-06-07 at 07 34 45


packages/ts-components/src/components/article-header/ArticleHeader.tsx
```diff
@@ -92,12 +93,12 @@ const ArticleHeader: React.FC<{
           <UpdatedTime
             isLessThan13Hours={!isLessThan1Minute && isLessThan13Hours}
           >
-            {format(parsedDate, 'h.mmaaa')}
+            {format(parsedDate, 'h.mmaaa z')}
```
flashcheeks, line 96: Will this display the timezone for all users not just non-UK users?

flashcheeks (changes requested):

packages/ts-components/src/components/article-header/ArticleHeader.tsx
```diff
@@ -1,13 +1,14 @@
 import React from 'react';
 import {
   // parse,
```
othetgyi, line 3: I noticed in other PRs there are some commented out imports, but wasn't sure why we need them?

packages/ts-components/src/components/article-header/ArticleHeader.tsx
```diff
@@ -1,13 +1,14 @@
...
-  format,
+  // format,
   differenceInSeconds,
   differenceInCalendarDays,
   formatDistanceStrict
 } from 'date-fns';
 import {
   // zonedTimeToUtc,
```
othetgyi, line 10: Same comment as above

packages/ts-components/src/components/article-header/ArticleHeader.tsx
DMackintosh-TNL: Removed

packages/ts-components/src/components/article-header/ArticleHeader.tsx
DMackintosh-TNL: Works now, this might be suitable for moving to UTILS, but that is out of scope for this story :)

othetgyi (approved):

flashcheeks (approved):

PR #2959, opened 2022-05-31
fix(TDP-2011): Lead asset src set

flashcheeks:
Update lead asset sizes in article pages [TDP-2011](https://nidigitalsolutions.jira.com/browse/TDP-2011)

New source set size are `[360, 520, 680, 860, 1200, 1500]` - please comment if you disagree or to recommend other values.


adamosborne-tnl:
I was going to ask the same question as Bibi :)

themischatzis (approved):

flashcheeks:
@bibicollins1 @adamosborne-tnl My thinking was that most low width devices (mobile) were high density resolutions meaning they would render the larger x2 or x3 sizes. That said, there is no reason not to add a low width version.

PR #2965, opened 2022-06-06
fix(TDP-1520): add articleSection to live articles

othetgyi:
This PR adds the `articleSection` property to the JSON LD of live articles (`articleSection` was previously added to the JSON LD of articles but not of live articles)

[TDP-1520](https://nidigitalsolutions.jira.com/browse/TDP-1520)


adamosborne-tnl (approved):

DMackintosh-TNL (commented):
Looks good :)

PR #2980, opened 2022-06-09
fix(TDP-2101): remove articleSection "article" and "current edition" prefix

othetgyi:
This PR removes the logic to add an "article" or "current edition" prefix before the section name in the JSON LD as that work is no longer required. 

[TDP-2101](https://nidigitalsolutions.jira.com/browse/TDP-2101)


DMackintosh-TNL (approved):
Looks ok to me

PR #2996, opened 2022-06-21
feat(TDP-1721): Remove React-Native from Ad package

othetgyi:
This PR removes React-Native from the ad package.

[TDP-1721](git push --set-upstream origin feat/TDP-1721-remove-react-native-from-ad-package)


adamosborne-tnl (approved):
looks good to me 👍

PR #2997, opened 2022-06-22
feat(TDP-1419): Remove react-native from message-bar package

bibicollins1:
### Description 

Remove React Native from message-bar package
[TDP-1419](https://nidigitalsolutions.jira.com/browse/TDP-1419)
Due to the message bar using TouchableOpacity and Animate particularly from react-native we had to refactor the MessageBar component quite a bit to recreate the animation. 
To test the behaviour locally best component is the save-share-bar if you click the icon to copy the link you should see the message bar, and it will automatically close after 3000ms or you can click the close button and it will close. - Compare against the built component [here](https://components.thetimes.co.uk/?path=/story/composed-save-and-share-bar--save-and-share-bar)

Big thank you to @adamosborne-tnl and @DMackintosh-TNL for the assist on this one!

### DIFF
--- a/packages/message-bar/src/message-bar.js
+++ b/packages/message-bar/src/message-bar.js
@@ -56,7 +56,7 @@ class MessageBar extends Component {
           <TcView style={styles.messageBarBody}>
             <TcText style={styles.messageBarText}>{message}</TcText>
               <CloseButton style={styles.messageBarCloseButton} className={this.state.closeActive ? ' active' : ''} onClick={this.closeMessage}>
-                <CloseIcon width="28" height="28" onPress={this.closeMessage}/>
+                <CloseIcon width="28" height="28" onClick={this.closeMessage}/>
               </CloseButton>
             </TcView>
         </TcView>
@@ -88,7 +88,7 @@ const StyledAnimation = styled(TcView)`
   }
 `;
 
-const CloseButton = styled.button`
+export const CloseButton = styled.button`
   cursor: pointer;
   &.active {
     opacity: 0.5
--- a/packages/message-bar/src/message-bar.js
+++ b/packages/message-bar/src/message-bar.js
@@ -2,22 +2,21 @@ import React, { Component } from "react";
 import { TcView, TcText } from "@times-components/utils";
 import PropTypes from "prop-types";
 import { CloseIcon } from "@times-components/icons";
-import styleFactory from "./styles";
-import { CloseButton, StyledAnimation } from './styles'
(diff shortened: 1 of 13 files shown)


packages/message-bar/src/styles/shared.js
```diff
@@ -21,7 +21,10 @@ const messageBarCloseButton = {
   justifyContent: "center",
   marginLeft: "auto",
   marginRight: 20,
-  width: 28
+  width: 28,
+  padding: 0,
+  display: "inherit",
+  border: "0px"
```
othetgyi, line 27: Should we omit the unit of measure since the value is 0?

bibicollins1: I think it's because the button has a default border around it, so this removes it. But good spot I'll test it now and get back to you 👍

packages/message-bar/src/message-bar.js
```diff
@@ -47,66 +38,44 @@ class MessageBar extends Component {
...
+  closeMessage() {
     const { close } = this.props;
-    clearTimeout(this.timeout);
-    this.timeout = null;
-    this.animateClosed(() => {
+    this.setState({ closeActive: true });
+    this.timeout = setTimeout(() => {
       close();
-    });
+    }, 250);
```
othetgyi, line 46: Just wondering where the 250 comes from?

packages/message-bar/src/styles/shared.js
bibicollins1: @othetgyi just tried it, and the button has default css applied to it that adds a big ol border. So this removes that, essentially just making the icon a button but keeping the icon style.

othetgyi (commented):
Nice work, Bibi! Looked like there was a ton of refactoring, but the component works just fine. I just had a teensy comment about a unit and a question about a value but other than that, looks good!

othetgyi (approved):

packages/message-bar/src/message-bar.js
bibicollins1: @othetgyi Really good spot again. So this one is a bit of a weird one, especially because the units are different. But essentially it matches the time we set for the animation durations in the StyledAnimation component (0.25s) So it allows the animation to go through. If you remove the timeout it zips away the message bar when you click on it from the save share bar for example, because it calls that closeMessage function - If that makes sense?

othetgyi: That does make sense, thank you!

PR #3010, opened 2022-07-05
feat(TDP-1694): removed react native from save and share bar

adamosborne-tnl:
removed react native from save and share bar

### DIFF
--- a/packages/save-and-share-bar/__tests__/mocks.js
+++ b/packages/save-and-share-bar/__tests__/mocks.js
@@ -12,11 +12,3 @@ jest.mock("@times-components/icons", () => ({
   IconSaveBookmark: "IconSaveBookmark",
   IconTwitter: "IconTwitter"
 }));
-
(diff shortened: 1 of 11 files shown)


packages/save-and-share-bar/__tests__/mocks.js
```diff
@@ -10,13 +10,6 @@ jest.mock("@times-components/icons", () => ({
   IconEmail: "IconEmail",
   IconFacebook: "IconFacebook",
   IconSaveBookmark: "IconSaveBookmark",
-  IconTwitter: "IconTwitter"
+  IconTwitter: "IconTwitter",
+  IconActivityIndicator: "IconActivityIndicator"
```
flashcheeks, line 14: @adamosborne-tnl just added `IconActivityIndicator` to the mocks - seems to pass now!

flashcheeks (approved):

PR #3012, opened 2022-07-05
feat(TDP-1422): remove react native from svgs

adamosborne-tnl:
which ultimately means the removal of the svgs package as was only there to produce RN svgs

[TDP-1422](https://nidigitalsolutions.jira.com/browse/TDP-1422)

FOR QA:
this ultimately changes the icons (as was only real thing using it), pagination arrows and watermark logo. these should look no different from before


packages/article-in-depth/__tests__/web/__snapshots__/semantic.web.test.js.snap
```diff
@@ -113,12 +113,10 @@ exports[`1. a full article 1`] = `
         aria-label="icon-email"
         height={16}
         role="img"
+        title="Share by email"
         viewBox="0 0 22 16"
         width={22}
       >
-        <title>
-          Share by email
-        </title>
```
flashcheeks, line 121: @adamosborne-tnl title attributes are ignored in svgs so we are losing some tooltip information. Is there a way we can keep the `title` as a tag?

adamosborne-tnl: OK, got the <title> tags back. The RN Svg lib, put them in automatically. I didn't notice that, so they're in now.

flashcheeks (approved):

PR #3025, opened 2022-07-12
feat(TDP-1713): remove react native from showcase

adamosborne-tnl:
removed react native from showcase project

[TDP-713](https://nidigitalsolutions.jira.com/browse/TDP-1713)


othetgyi (approved):
LGTM!

PR #3049, opened 2022-07-25
feat(TDP-1712): create configs for failing packages

bibicollins1:


bibicollins1:
Closed as have made changes to the jest-configurator-web instead

PR #3056, opened 2022-07-26
feat(TDP-1712): update packages to use new jest config web

bibicollins1:
### Description 

Update packages to use the latest jest-configurator-web. Some packages still had some react-native so have removed that where needed. And updated some babel.config files too.


adamosborne-tnl (approved):
LGTM 👍

PR #3057, opened 2022-07-26
feat(TPD-1712): updated jest config 

adamosborne-tnl:
have updated some packages to use updated jest config 0.7.0


bibicollins1 (approved):
Looks good! Only thing is the yarn.lock with version 0.6.0 of the configurator?

PR #3074, opened 2022-08-01
feat(TDP-1712): save and share bar

bibicollins1:
Update babel config
Use new jest-configurator-web
Update tests as they were failing with the mounting of the testInstance


packages/save-and-share-bar/__tests__/shared-tracking.js
```diff
@@ -111,6 +113,7 @@ export default () => {
     });
 
     xit("when press copy to clipboard", () => {
```
adamosborne-tnl, line 115: can you change this xit to it as I think this was me before for some reason

adamosborne-tnl (changes requested):

adamosborne-tnl (approved):

PR #3109, opened 2022-08-09
feat(TDP-2233): Trigger real name commenting banner

othetgyi:
This PR gets the user's display name, and triggers the real name commenting banner.

[TDP-2233](https://nidigitalsolutions.jira.com/browse/TDP-2233)


packages/article-comments/src/comments.js
```diff
@@ -84,11 +89,38 @@ class Comments extends Component {
...
+        }
+      },
       { once: true }
     );
+
+    document.addEventListener("spot-im-user-auth-success", async event => {
+      if (isFeatureFlagEnabled) {
+        const { displayName } = event.detail;
+
+        const shouldShowBanner = await userShouldUpdateName(displayName);
```
flashcheeks, line 118: Please can we move the call to `userShouldUpdateName` into the `spot-im-current-user-typing-start` event listener.

packages/article-comments/src/comments.js
```diff
@@ -84,11 +89,38 @@ class Comments extends Component {
...
+
+        this.setState({ shouldUpdateName: shouldShowBanner });
```
flashcheeks, line 120: Switch to storing the display name as state `this.setState({ displayName });`

flashcheeks (changes requested):

flashcheeks (approved):

PR #3130, opened 2022-08-15
fix (TDP-2233): get user details from local storage instead of openweb

SunshineEleda:
When a user initially logs into an article page, dummy user details are used which means an incorrect display name is passed which will not trigger the functionality of the smart banner.

The issue of the dummy data is initially resolved and the correct values are stored into local storage. This pr simply reads the local storage data as a source of truth rather than data we get back from spotim which is dummy data.

https://nidigitalsolutions.jira.com/jira/software/c/projects/TDP/boards/1491?modal=detail&selectedIssue=TDP-2233

### DIFF
--- a/packages/article-comments/__tests__/web/utils.test.js
+++ b/packages/article-comments/__tests__/web/utils.test.js
@@ -1,7 +1,52 @@
-import { userShouldUpdateName } from "../../src/utils";
+/* global window */
+
+import {
+  userShouldUpdateName,
+  getDisplayNameFromLocalStorage
+} from "../../src/utils";
 
 const unmockedFetch = global.fetch;
 let mockFetchResponse = {};
+
+const localStorageMock = (function() {
+  const store = {};
+  return {
+    getItem(key) {
+      return store[key];
+    },
+    setItem(key, value) {
+      store[key] = value.toString();
+    }
+  };
+})();
+
+Object.defineProperty(window, "localStorage", { value: localStorageMock });
+
+describe("", () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+});
+describe("getDisplayNameFromLocalStorage()", () => {
+  it("should return false if user it not signed in", () => {
+    expect(getDisplayNameFromLocalStorage()).toEqual(false);
+  });
+
+  it("should return false if there is no display name", () => {
+    window.localStorage.setItem(
+      "SPOTIM_CURRENT_USER",
+      '{"data":{"id":"u_sgCrqrs7KNLv","imageId":"#Grey-Cactus","username":"JohnSmith750","isRegistered":true,"isCommunityModerator":false,"isJournalist":false,"isModerator":false,"isAdmin":false,"isSuperAdmin":false,"reputation":{"total":1},"ssoData":{},"email":"","isEmailVerified":false}}'
+    );
+    expect(getDisplayNameFromLocalStorage()).toEqual(false);
+  });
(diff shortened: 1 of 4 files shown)


packages/article-comments/__tests__/web/utils.test.js
```diff
@@ -1,8 +1,53 @@
...
+  });
+
+  it("should return false if there is no display name", () => {
+    window.localStorage.setItem(
+      "SPOTIM_CURRENT_USER",
+      '{"data":{"id":"u_sgCrqrs7KNLv","imageId":"#Grey-Cactus","username":"JohnSmith750","isRegistered":true}}'
+    );
+    expect(getDisplayNameFromLocalStorage()).toEqual(false);
+  });
+  it.only("should return the display name ", () => {
```
othetgyi, line 42: `.only` can be removed.

packages/article-comments/src/utils.js
```diff
@@ -29,6 +29,19 @@ export const userShouldUpdateName = async username => {
   return isPseudonym;
 };
 
+export const getDisplayNameFromLocalStorage = () => {
+  const spotimUserDetails = window.localStorage.getItem("SPOTIM_CURRENT_USER");
+
+  if (!spotimUserDetails) return false;
```
othetgyi, line 35: Should this just do an early `return` if there are no `spotimUserDetails`? Or what benefit would it have to return `false`?

packages/article-comments/src/utils.js
```diff
@@ -29,6 +29,19 @@ export const userShouldUpdateName = async username => {
...
+
+  const spotimUserDetailsJSON =
+    spotimUserDetails && JSON.parse(spotimUserDetails);
+
+  const { displayName } = spotimUserDetailsJSON.data;
+
+  return displayName || false;
```
othetgyi, line 42: Is the option to return `false` here in case there is no `displayName`? Is returning `false `meant to keep the rest of the code from breaking if there is no `displayName`?

SunshineEleda: yes just as an extra precaution

packages/article-comments/src/utils.js
SunshineEleda: yes by returning false it would return early, just keeping the return message consistent to when it returns if there is no display name

packages/article-comments/__tests__/web/utils.test.js
```diff
@@ -1,8 +1,53 @@
...
+    setItem(key, value) {
+      store[key] = value.toString();
+    }
+  };
+})();
+
+Object.defineProperty(window, "localStorage", { value: localStorageMock });
+
+describe("", () => {
+  afterEach(() => {
```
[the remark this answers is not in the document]
othetgyi: Sorry, what I meant was, can't the `afterEach` stand on its own outside the `describe` block? Then it will clear the mocks for all the tests. If it's in it's `describe` block it will clear the mocks only for the tests inside that block, and this block doesn't have any.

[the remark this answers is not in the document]
SunshineEleda: So I created a a larger describe that should clear all mocks for all the tests, just keeping in mind you will also be adding in some stuff with local storage mocks as well. So 1 describe to encapsulate all of them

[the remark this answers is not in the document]
othetgyi: Oh I see! Looks good. :)

othetgyi (approved):

PR #3131, opened 2022-08-16
feat(TDP-2203): Reauthenticate users to real name commenting

othetgyi:
This PR reauthenticates users who are logged into the old commenting system into the new system. As part of that work, it adds a localStorage token to the user's browser when they log into the new commenting system, and checks for that token before reauthenticating users.

[TDP-2203](https://nidigitalsolutions.jira.com/browse/TDP-2203)


packages/article-comments/src/utils.js
```diff
@@ -29,6 +29,20 @@ export const userShouldUpdateName = async username => {
...
+  if (realNameToken) {
+    console.log('***There is a real name token***');
+    return
+      }
+      window.localStorage.removeItem("SPOTIM_DEVICE_V2");
+      window.localStorage.removeItem("SPOTIM_CURRENT_USER");
+      window.localStorage.removeItem("SPOTIM_ACCESS_TOKEN");
+      window.localStorage.removeItem("SPOT_AB");
+      window.localStorage.removeItem("SPOTIM_DEVICE_UUID_V2");
+
```
SunshineEleda, line 43: I recall CPS saying we don't need to remove the local storage spotim stuff anymore. You just need to fire the executeSSOTransaction if there is not a `realNameToken` and this should take care of that.

packages/article-comments/src/comments.js
```diff
@@ -3,7 +3,7 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import { CommentContainer } from "./styles/responsive";
-import executeSSOtransaction from "./comment-login";
+import executeSSOtransaction, { setNewUserToken } from "./comment-login";
```
SunshineEleda, line 6: You won't need to import `setNewUserToken` here if you just call it in the `executeSSOTransaction` function in `comment-login.js`

bibicollins1 (commented):
Functionally looks good to me, just wondering whether the naming could be improved? Find myself getting a bit confused with the shouldUserReauthenticate/shouldReauthenticateUser. Saying that I don't yet have a suggestion that would be better but will have a think and maybe chat about it tomorrow?

gczene (approved):

packages/article-comments/src/comments.js
```diff
@@ -147,7 +150,10 @@ class Comments extends Component {
         executeSSOtransaction(() => {});
       } else {
         document.addEventListener("spot-im-api-ready", () => {
-          executeSSOtransaction(() => {});
+          const shouldUserReauthenticate = shouldReauthenticateUser();
```
SunshineEleda, line 153: With what @bibicollins1  said I think it might make it less confusing to rename `shouldUserReauthenticate` to `shouldReauthenticateUser` 

https://github.com/newsuk/times-components/pull/3131#pullrequestreview-1076079284

SunshineEleda (changes requested):

packages/article-comments/src/comment-login.js
```diff
@@ -19,13 +18,17 @@ const ssoCallback = (codeA, completeSSOCallback) =>
...
 
+const setNewUserToken = () => {
+  window.localStorage.setItem("isUsingRealNameCommenting", true);
+};
+
 const executeSSOtransaction = callback => {
   if (window.SPOTIM && window.SPOTIM.startSSO) {
     window.SPOTIM.startSSO(ssoCallback);
-
+    setNewUserToken();
```
SunshineEleda, line 28: Also can we put this behind the feature flag `enableRealNameCommenting`

packages/article-comments/src/comments.js
```diff
@@ -147,7 +150,10 @@ class Comments extends Component {
...
+          if (shouldUserReauthenticate) {
+            executeSSOtransaction();
+          }
```
SunshineEleda, line 156: Can we put this behind the feature flag enableRealNameCommenting

SunshineEleda (changes requested):

othetgyi:
> Functionally looks good to me, just wondering whether the naming could be improved? Find myself getting a bit confused with the shouldUserReauthenticate/shouldReauthenticateUser. Saying that I don't yet have a suggestion that would be better but will have a think and maybe chat about it tomorrow?

@bibicollins1 I've removed the shouldUserReauthenticate variable, hopefully that makes it less confusing.

othetgyi:
I've added the "DO NOT MERGE" label until the code freeze on Times Components is over and the big React Native PR has been merged in and tested.

packages/article-comments/src/comments.js
```diff
@@ -147,7 +150,13 @@ class Comments extends Component {
...
       } else {
         document.addEventListener("spot-im-api-ready", () => {
-          executeSSOtransaction(() => {});
+          if (isFeatureFlagEnabled) {
+            if (shouldReauthenticateUser()) {
+              executeSSOtransaction();
+            }
+          } else {
+            executeSSOtransaction(() => {});
+          }
```
SunshineEleda, line 159: It would be a bit cleaner to do something like this since the executeSSoTransaction will fire regardless ```suggestion
          if (isFeatureFlagEnabled) {
            shouldReauthenticateUser()
            }

            executeSSOtransaction(() => {});
          
```

File: packages/article-comments/src/utils.js [hunk 7]
```diff
@@ -29,6 +28,21 @@ export const userShouldUpdateName = async username => {
   return isPseudonym;
 };
 
+export const shouldReauthenticateUser = () => {
+  const isUsingRealNameCommenting = window.localStorage.getItem(
+    "isUsingRealNameCommenting"
+  );
+  if (isUsingRealNameCommenting) {
+    return false;
+  }
```
SunshineEleda, line 37: Based on [this](https://github.com/newsuk/times-components/pull/3131#discussion_r949245978) you could change this to:


```suggestion
  if (isUsingRealNameCommenting) return 
```

packages/article-comments/src/utils.js
```diff
@@ -29,6 +28,21 @@ export const userShouldUpdateName = async username => {
...
+  window.localStorage.removeItem("SPOTIM_DEVICE_V2");
+  window.localStorage.removeItem("SPOTIM_CURRENT_USER");
+  window.localStorage.removeItem("SPOTIM_ACCESS_TOKEN");
+  window.localStorage.removeItem("SPOT_AB");
+  window.localStorage.removeItem("SPOTIM_DEVICE_UUID_V2");
+  return true;
```
SunshineEleda, line 43: And then you can get rid of this since we don't need this function to return a boolean anymore

packages/article-comments/__tests__/web/shared.web.test.js
```diff
@@ -130,6 +129,16 @@ it("single comment", () => {
...
+  const mockLocalStorage = {
+    storage: {},
+    setItem: jest.fn((key, value) => {
+      mockLocalStorage.storage[key] = value;
+    })
+  };
+
+  Object.defineProperty(global.window, "localStorage", {
+    value: mockLocalStorage
+  });
```
SunshineEleda, line 141: Can also get rid of this since it's not being used for anything

SunshineEleda (approved):

PR #3134, opened 2022-08-19
fix(TDP-2385): Fix object,Object error

bibicollins1:
### Description 

We were getting an [object, Object] error rendering on the page in render when testing our RN removal changes
This change fixes that error, we have tested running render locally with the change and the [object, Object] is no longer present. 
Screenshot 2022-08-19 at 15 01 10


adamosborne-tnl (approved):
LGTM

dimitrianoudi (approved):

PR #3135, opened 2022-08-24
fix(TDP-2433): fix native ad visibility

SunshineEleda:
Add display:none to fix hidden native ad

https://nidigitalsolutions.jira.com/jira/software/c/projects/TDP/boards/1491?modal=detail&selectedIssue=TDP-2433


adamosborne-tnl:
@SunshineEleda I take it the Ad team are aware of the display:none; and would re-enable the display? How does this affect the Cumulative Layout Shift if the Ad's are then inserted and the page dimensions change, which is why I thought it was visibility:hidden to begin with?

SunshineEleda:
> 

I was just adding the `display:none` to the class called `hidden` so I assume that is the toggle they would be using but not directly setting the `display` or `visibility` styles. The class does not seem to be affected by page dimension change but I will run this by the commercial team.

PR #3140, opened 2022-09-06
fix(TDP-2475): Add event listener to real name commenting reauthentication

othetgyi:
This PR adds another event listener to real name commenting reauthentication that will check for the localStorage token dropped when then user previously logged into commenting.


packages/article-comments/src/comments.js
```diff
@@ -145,15 +145,25 @@ class Comments extends Component {
...
+      const realNameCommentingCheck = shouldReauthenticateUser();
+      if (isFeatureFlagEnabled) {
+        if (realNameCommentingCheck) {
+          window.localStorage.removeItem("SPOTIM_DEVICE_V2");
+          window.localStorage.removeItem("SPOTIM_CURRENT_USER");
+          window.localStorage.removeItem("SPOTIM_ACCESS_TOKEN");
+          window.localStorage.removeItem("SPOT_AB");
+          window.localStorage.removeItem("SPOTIM_DEVICE_UUID_V2");
+        }
+        executeSSOtransaction(() => {});
```
flashcheeks, line 158: Call to `executeSSOtransaction()` needs to be inside the conditional

packages/article-comments/src/comments.js
```diff
@@ -145,15 +145,25 @@ class Comments extends Component {
     document.addEventListener("spot-im-share-type", event =>
       getShareEvent(event)
     );
+    document.addEventListener("spot-im-user-auth-success", () => {
+      const realNameCommentingCheck = shouldReauthenticateUser();
+      if (isFeatureFlagEnabled) {
+        if (realNameCommentingCheck) {
```
flashcheeks, line 151: Can you check your logic here. When the user does not have a localStorage value we should be reauthenticating.

flashcheeks (changes requested):

packages/article-comments/src/comments.js
```diff
@@ -89,16 +89,20 @@ class Comments extends Component {
       }
     }
 
-    const isFeatureFlagEnabled = window.location.search.includes(
+    const isRealNameCommentingEnabled = window.location.search.includes(
       "enableRealNameCommenting"
     );
 
+    const isReauthenticationEnabled = window.location.search.includes(
+      "enableReauthentication"
```
SunshineEleda, line 97: Would be good to make it more explicit what it is i.e. `enableRealNameReauthentication`

SunshineEleda (approved):

PR #3143, opened 2022-09-06
fix(TDP-2399): Related-Article image sizes need optimising

DMackintosh-TNL:
PR to fix the bloated Related Article image sizes...


flashcheeks (commented):
@DMackintosh-TNL I closed because I have already removed this code. Please shout when you have time to discuss.

PR #3170, opened 2022-10-20
feat(TDP-2618): updated topic page to use the lcpItem class for the f…

adamosborne-tnl:
…irst item only

[TDP-2618](https://nidigitalsolutions.jira.com/browse/TDP-2618) updated topic page to have the first image item to have the LCP class applied


packages/image/src/image.js
```diff
@@ -125,7 +126,12 @@ class TimesImage extends Component {
     }
 
     return (
-      <TcView onLayout={onLayout} style={styles} testID="Image">
+      <TcView
+        onLayout={onLayout}
+        style={styles}
+        testID="Image"
+        className={`${isLcpItem === true ? "lcpItem" : ""}`}
```
shubhambajpaiNUK, line 133: instead of ternary operation can we use `isLcpItem && "lcpItem" || ""`

shubhambajpaiNUK (approved):

packages/image/src/image.js
adamosborne-tnl: good point, i am only testing a boolean

bibicollins1 (approved):

PR #3211, opened 2023-01-11
chore(TDP-000): Create Newskit package

flashcheeks:


packages/ts-newskit/README.md
```diff
@@ -0,0 +1,16 @@
+# Typescript Components
```
bibicollins1, line 1: Does this want to be Newskit components or something? or maybe I'm confusing this as a title

bibicollins1 (approved):

.circleci/config.yml
```diff
@@ -12,7 +12,7 @@ commands:
         steps:
             - run:
                   name: Install Dependencies
-                  command: npx yarn@1.22.15 install --frozen-lockfile
+                  command: npx yarn@1.22.15 install --frozen-lockfile --ignore-engines
```
bibicollins1, line 15: Just curious why we are adding this?

flashcheeks: One of the NK dependancies requires node version `12.22` we are currently using `12.15` so `yarn install` will fail without `--ignore-engines`. This is obviously not a good solution long term, but as this package is a NK temp sandbox, thought is was ok for now. The correct solution is to upgrade the node version.

packages/ts-newskit/README.md
flashcheeks: Yep, good spot 👍

.circleci/config.yml
bibicollins1: Ah okay 👍 Thanks!

PR #3223, opened 2023-01-29
Fix/tdp 2980 legacy type error

douglasmik:
Addition of more defensive code to clear TypeError's in trying to access data


packages/author-profile/__tests__/web/__snapshots__/author-profile.web.test.js.snap
```diff
@@ -154,8 +151,7 @@ exports[`5. fetches more articles 1`] = `
 <authorArticlesWithImagesProvider>
   <Helmet>
     <title>
-      Name
-       | The Times & The Sunday Times
```
bibicollins1, line 158: Just wondering why these are rendering differently now?

adamosborne-tnl: like it was due to the way `packages/author-profile/src/head.js` was changed, but it's a good thing being `&amp;`

douglasmik: Think this is probably due to wrapping in template literal to cover another TypeError, but would rather leave as '&amp;'.
Good spot though! 👍

adamosborne-tnl (approved):

adamosborne-tnl (approved):

ibrahim-kurhan (approved):

PR #3225, opened 2023-01-31
feat(TDP-000) POC - Import independent newskit components into ts-newskit package

bibicollins1:
## Description 

When we import newskit it doubles the bundle size in render for times-components 
This is a test to see if we import the components independently rather than `import { TextBlock } from 'newskit'` we hope that it will significantly reduce it.


flashcheeks (approved):

adamosborne-tnl (approved):
If you can import those modules directly, wonderful 👍

PR #3251, opened 2023-02-28
fix(TDP-3079): remove section name from page meta title

ibrahim-kurhan:
We’d like to remove the page section name which is appended to page meta titles

Hypothesis that this will improve the UX for readers presented with a link to our content when using Google

Context to request and example of page meta title:

User story:

Given I’m a web user
When I’m using Google
And I’m presented with a link to content from The Times
Then The meta title doesn’t include the section name

[JIRA-3079](https://nidigitalsolutions.jira.com/browse/TDP-3079)

### DIFF
--- a/packages/article-skeleton/src/head.js
+++ b/packages/article-skeleton/src/head.js
@@ -382,10 +382,7 @@ function Head({
         jsonLD.mainEntityOfPage["@id"] = makeArticleUrl(article);
         return (
           <Helmet encodeSpecialCharacters={false}>
-            <title>
-              {title} | {sectionname ? `${sectionname} | ` : ""}
-              {publication}
-            </title>
+            <title>{title}</title>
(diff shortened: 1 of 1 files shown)


bibicollins1 (approved):

packages/article-skeleton/src/head.js
```diff
@@ -382,10 +382,7 @@ function Head({
         jsonLD.mainEntityOfPage["@id"] = makeArticleUrl(article);
         return (
           <Helmet encodeSpecialCharacters={false}>
-            <title>
-              {title} | {sectionname ? `${sectionname} | ` : ""}
-              {publication}
-            </title>
+            <title>{title}</title>
```
douglasmik, line 385: are we also to remove the publication name?

douglasmik (approved):

PR #3254, opened 2023-02-28
feat(TDP-3290): add ThemeProvider wrapper component

douglasmik:
Packaging of NewskitProvider component to allow for exporting of this and associated ThemeProvider

### DIFF
--- a/packages/ts-newskit/src/components/utils/__tests__/TCThemeProvider.test.tsx
+++ b/packages/ts-newskit/src/components/utils/__tests__/TCThemeProvider.test.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import '@testing-library/jest-dom';
+import { render, screen } from '@testing-library/react';
+import { SectionHeader } from '../../section-header';
+import { TCThemeProvider } from "../TCThemeProvider"
+
+const renderComponent = () => render(
+  <TCThemeProvider>
+    <SectionHeader title="test" />
+  </TCThemeProvider>
+);
+
+describe('<TCThemeProvider>', () => {
+  it('should match snapshot', () => {
+    const wrapper = renderComponent();
+
+    expect(wrapper).toMatchSnapshot();
+  })
+  it('should wrap child components with theme and pass styling', () => {
+    renderComponent();
+
+    const title = screen.getByText("test");
+
+    expect(title).toHaveStyle({
+      font: "inherit",
+      display: "inline-block",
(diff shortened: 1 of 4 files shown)


packages/ts-newskit/src/components/utils/__tests__/__snapshots__/TCThemeProvider.test.tsx.snap
```diff
@@ -0,0 +1,85 @@
...
+    >
+      <h1
+        class="nk-headline-heading css-e5fswc"
+      >
+        test
+      </h1>
+    </section>
+  </div>,
+  "debug": [Function],
+  "findAllByAltText": [Function],
```
bibicollins1, line 34: Wondering how come you are getting all this in your snapshot?

bibicollins1: Oh actually probably because not destructuring in the same way - np

packages/ts-newskit/src/components/utils/__tests__/TCThemeProvider.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+);
+
+describe('<TCThemeProvider>', () => {
+  it('should match snapshot', () => {
+    const wrapper = renderComponent();
+
+    expect(wrapper).toMatchSnapshot();
+  })
+  it('should wrap child components with theme and pass styling', () => {
+    renderComponent();
```
bibicollins1, line 20: Does const { getByText } = renderComponent(); still work here?

ibrahim-kurhan (approved):

packages/ts-newskit/src/components/utils/__tests__/__snapshots__/TCThemeProvider.test.tsx.snap
douglasmik: It's returning the complete `render` object, which includes all of those methods. Will update to just return fragment, using `asFragment` method.

packages/ts-newskit/src/components/utils/__tests__/TCThemeProvider.test.tsx
douglasmik: yes, that should still be possible. Using `screen` here, as best practice, though makes no difference in this scenario as only single test.

bibicollins1 (approved):

PR #3264, opened 2023-03-06
fix(TDP-3100-Cypress cases): fixed the cypress scripts for ssr

npriya21:


gczene (approved):

shubhambajpaiNUK (approved):

packages/ssr/__tests__/helpers/article-helper.js
```diff
@@ -41,13 +41,13 @@ const articleTemplateTest = (template, options = {}) => {
         .scrollIntoView()
         .then(() => {
           // wait for the image to transition and be removed (unfortunately Cypress doesn't auto wait for this)
-          cy.wait(2000);
+          // cy.wait(2000);
```
shubhambajpaiNUK, line 44: You can remove this line if it is not in use. 

Can we use intercept or something similar to it? I am mentioning this because as a part of flakiness sometimes image loads with a bit of delay and tests start running earlier resulting into failure.

PR #3274, opened 2023-03-13
chore(TDP-000): Export drawer

bibicollins1:
Another one to test why GlobalNav not working in render

### DIFF
--- a/packages/ts-newskit/src/components/global-nav/index.tsx
+++ b/packages/ts-newskit/src/components/global-nav/index.tsx
@@ -39,10 +39,11 @@ export const GlobalNav = ({ isLoggedIn, isSunday, data }: GlobalNavProps) => {
 export const TSNewskitDrawer: React.FC<{
   children: ReactChildren;
   open: boolean;
-}> = ({ children, open }) => {
+  onDismiss: () => void;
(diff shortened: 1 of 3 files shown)


douglasmik (approved):

packages/ts-newskit/src/components/global-nav/index.tsx
```diff
@@ -34,3 +35,16 @@ export const GlobalNav = ({ isLoggedIn, isSunday, data }: GlobalNavProps) => {
...
   );
 };
+
+export const TSNewskitDrawer: React.FC<{
+  children: ReactChildren;
+  open: boolean;
+  onDismiss: () => void;
+}> = ({ children, open, onDismiss }) => {
+  return (
+    // tslint:disable-next-line: no-console
```
ibrahim-kurhan, line 45: do you need to disable tslint?

bibicollins1: Yeah was not allowing console.log - only using it for debugging purposes won't keep it in

ibrahim-kurhan: got it

PR #3277, opened 2023-03-13
fix(TDP-3223): Create custom drawer component

bibicollins1:
### Description 
Still only going to be used for testing purposes but have created a custom Drawer component to see whether this can be used in place of Newskit Drawer as that is breaking when imported to render.

### DIFF
--- a/packages/ts-newskit/src/components/global-nav/HamburgerMenuContainer.tsx
+++ b/packages/ts-newskit/src/components/global-nav/HamburgerMenuContainer.tsx
@@ -1,19 +1,19 @@
-import React, {ReactNode} from 'react';
-import styled from 'styled-components';
+import React, { ReactNode } from 'react';
+import { styled } from 'newskit';
 
-const MenuDrawer = styled.div<{open: boolean}>`
+const MenuDrawer = styled.div<{ open: boolean }>`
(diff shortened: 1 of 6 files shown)


ibrahim-kurhan (approved):

packages/ts-newskit/src/components/global-nav/HamburgerMenuContainer.tsx
```diff
@@ -0,0 +1,42 @@
...
+  transition: transform 0.3s ease-in-out;
+  width: 100%;
+  height: 100%;
+  @media screen and (min-width: 780px) {
+    width: 320px;
+  }
+`;
+
+const Overlay = styled.div<{ open: boolean }>`
+  z-index: 99;
```
douglasmik, line 17: Does this interfere with TopNav and other page items being 'higher' on the page

douglasmik (approved):

packages/ts-newskit/src/components/global-nav/HamburgerMenuContainer.tsx
bibicollins1: Doesn't seem to as rendered outside of those but yeah can align later

PR #3296, opened 2023-03-22
feat(TDP-3413): add journalist quote component 

douglasmik:
- Addition of `JournalistQuote` component for Content Bucket 1 slice.
- Removal of breakpoint dependant style on topnav, using `useBreakpointKey` hook
- Update Storybook directory structure


packages/ts-newskit/src/components/slices/journalist-quote/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <QuoteText
+          $color={textColour}
+          typographyPreset={{
+            xs: 'editorialItalic060',
+            sm: 'editorialItalic050'
+          }}
+        >
+          {quote}
+        </QuoteText>
+      </Block>
```
ibrahim-kurhan, line 70: are QuoteText, JournalistName , JournalistImage supposed to be link? because for lead story image heading and summary were link

packages/ts-newskit/src/components/slices/slice-header/index.tsx
```diff
@@ -14,7 +14,7 @@ export const SliceHeader = ({ title, color }: HeaderProps) => {
           xs: 'newPreset070',
           lg: 'newPreset060'
         }}
-        color={color}
+        $color={color}
```
ibrahim-kurhan, line 17: just out of curiousty what difference does this $ sign make becasue i didnt use it for color prop in lead story

ibrahim-kurhan (approved):

packages/ts-newskit/src/components/slices/slice-header/index.tsx
douglasmik: There are 'transient props' that prevent them coming through to the DOM. Will probably need to look through to see if we have this elsewhere also.
FYI - https://jakemccambley.medium.com/transient-props-in-styled-components-3105f16cb91f

PR #3302, opened 2023-03-28
feat(TDP-0000): add code addon

ibrahim-kurhan:
I have added code addon and its working perfectly fine with the ts-component, with the ts-newskit instead of showing the full source code it is only showing the imported part of the component as below 
Screenshot 2023-03-31 at 11 55 36
to be able to get the full source code we need to make some changes in mdx file  after the changes in mdx file thats how it looks like 
Screenshot 2023-03-31 at 12 00 17
for the rest of the packages it is not working because we are using showcase its still doable but we need to convert all the packages which I thing will be time consuming 

I also tried add performance addon as well latest version is not compatible with our node version I tried to install the older version but its clashing with custumRender util that we use for testing and reason for that is The performance addon for Storybook includes a tool called "react-
perf-tool', which can cause type inference issues when used with "@testing-
library/react'

I labeled it as DO NOT MERGE becasue I just wanted to get your opinion on this

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -12,7 +12,8 @@ module.exports = {
     "@storybook/addon-knobs",
     "@storybook/addon-options",
     "@storybook/addon-viewport",
-    "storybook-addon-paddings"
+    "storybook-addon-paddings",
+    "@storybook/addon-storysource",
   ],
   framework: '@storybook/react',
 };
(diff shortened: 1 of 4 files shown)


douglasmik (commented):
Great work! Thinking it may not be best to show story code in any event, as it could be a bit misleading, where we'd want them to use the code we show in the docs instead? For example, we wouldn't want them to change logical props, presets, etc

PR #3312, opened 2023-04-13
fix(TMRX-1148): fonts displaying bold when they dont need tobe

ibrahim-kurhan:
On article pages, the headline and the standfirst are displaying bold, but they don’t need to be.
image-20230404-141122

[JIRA-1148](https://nidigitalsolutions.jira.com/browse/TMRX-1148)
From the image above, you can see the fonts on the left, without bold applied, and on the right as they currently are, but are incorrect.

Both the H1 and the H2 on the template are using the browser default stylesheet formatting of using bold for Hx tags, bold doesn’t need to be applied as it is already bold in the fonts and a font weight of normal(400) should be used.

We need to apply font-weight: 400 to both of them so they display as on the left.

Acceptance Criteria

The font-weight is applied so that the fonts are displayed correctly.

Check all article templates and apply same if needed.

### DIFF
--- a/packages/article-in-depth/__tests__/web/__snapshots__/article-with-style.web.test.js.snap
+++ b/packages/article-in-depth/__tests__/web/__snapshots__/article-with-style.web.test.js.snap
@@ -425,6 +425,7 @@ exports[`full article with style 1`] = `
   padding: 0px;
   white-space: pre-wrap;
   overflow-wrap: break-word;
+  font-weight: 400;
   font-size: 30px;
   line-height: 33px;
   font-weight: 400;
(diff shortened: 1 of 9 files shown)


douglasmik (approved):

douglasmik (commented):
Good work! 
Will this be the case for all headings? Is it worth having a 'reset' of sorts?

ibrahim-kurhan (commented):
I am not sure if its gonna be for all the Headings @adamosborne-tnl?

adamosborne-tnl:
I wouldn't do a reset, as we would need to re-evaluate the entire site. This is just a mini-reset of the h1 & h2 in articles

PR #3332, opened 2023-04-25
feat(TMRX-1199): create feedback component

ibrahim-kurhan:
Description

For the feedback component, see the two states (on load, and when the user interacts with the :thumbsup: :thumbsdown:  icon buttons) will be the same styling, but would just drop off the thumb icon buttons, and change the text for the state of 
the feedback component when a user has clicked on either icon button. The close icon button on the top right would cause the feedback component to be hidden, and the content below on the page shifted up.

See the handoff here for this: https://www.figma.com/file/6TI11VNNjZdG8DRXbZTlVR/TML-Audio-Tests?node-id=2733%3A153958&t=QDFglhff0dihhqGB-1 - Connect your Figma account 

Acceptance Criteria

The feedback feature is presented to the user if (after) they’ve interacted with the play button (and have started to listen to the audio)

[JIRA-765](https://nidigitalsolutions.jira.com/browse/TMRX-765)


ibrahim-kurhan (commented):
as you can see I am not really good at naming variables and file names:) any file and variable naming suggestion more than welcome

douglasmik (approved):

packages/ts-newskit/src/components/article-audio/feedback/message.tsx
```diff
@@ -0,0 +1,32 @@
...
+        {message}
+      </TextBlock>
+      <IconButton
+        overrides={{
+          stylePreset: 'feedbackIconButtonOutlineSecondary',
+          iconSize: 'iconSize080',
+          marginInline: 'space000'
+        }}
+        onClick={() => setClosed(!closed)}
+        aria-label="Thank you Close"
```
bibicollins1, line 26: Just wondering why this is Thank you Close?

bibicollins1:
@ibrahim-kurhan Looks good! Just one minor comment on the aria label. Also noted that on the docs tab of the story the icons don't seem to be loading in properly? - Minor just wondering if theres a reason. Also lastly were there meant to be any hover states on the thumbs up/down icons?

packages/ts-newskit/src/components/article-audio/feedback/message.tsx
ibrahim-kurhan: changed to close

ibrahim-kurhan (commented):
@bibicollins1 I haven't seen any hover state in design also I have also noticed the issue on docs tab related to icons but I have no idea why it does that

bibicollins1:
@ibrahim-kurhan Okay cool to ignore the docs one for now. Might be worth looking into why at some point though. Might be worth asking design about any hover state but doesn't have to block this if they haven't already provided it.

bibicollins1 (approved):

PR #3338, opened 2023-04-27
fix(TMRX-0000): add missing article type to lead story

ibrahim-kurhan:
This PR is to add missing article type to lead story of the slice 1


packages/ts-newskit/src/components/slices/lead-story/index.tsx
```diff
@@ -85,13 +88,35 @@ export const LeadStory = ({
...
-        </TextBlock>
+        <Block>
+          <ColouredText
+            typographyPreset="articleListArticleType"
+            as="span"
+            $color={color}
+            paddingBlockStart="space010"
+          >
+            {articleType}
+          </ColouredText>
```
douglasmik, line 99: dos `articleType` also need some logic applied with it being conditional?

packages/ts-newskit/src/components/slices/lead-story/index.tsx
```diff
@@ -85,13 +88,35 @@ export const LeadStory = ({
...
+          {articleType &&
+            readingTime && (
+              <ContainerInline>
```
douglasmik, line 102: could these styles be applied to the `Block` component above `ColouredText`?

douglasmik (approved):

packages/ts-newskit/src/components/slices/lead-story/index.tsx
```diff
@@ -85,13 +88,35 @@ export const LeadStory = ({
...
+                  }}
+                />
+              </ContainerInline>
+            )}
+          <TextBlock
+            typographyPreset="newPreset100"
+            stylePreset="leadStoryTextColor"
+            as="span"
+            paddingBlockStart="space010"
+          >
```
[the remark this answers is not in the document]
ibrahim-kurhan: well actually these TextBlocks doesn't need to be coloured it is just wrong naming I will change the preset name

packages/ts-newskit/src/components/slices/lead-story/index.tsx
ibrahim-kurhan: it is good that you mentioned I found that some styles are not needed and I removed them also I tried to apply to block but as not of the styles are being shared between them I believe we cant apply them to Block

packages/ts-newskit/src/components/slices/lead-story/index.tsx
ibrahim-kurhan: as articleType is optional it would not effect the UI if we don't have it then it would just remove the divider

bibicollins1 (approved):

PR #3340, opened 2023-04-28
feat(TNLT-10974): Remove ads from coronation article

bibicollins1:
### Description 

Editorial would like us to remove the inline ads from this article http://preview.thetimes.co.uk/preview/article/c8bf6998-d498-11ed-b5c3-54651fc826e9 
Currently we don't have a good solution for this, ideally they would be able to set a flag on the article to disable ads which we would then retrieve from the data and do this dynamically. Optimizely also isn't the perfect solution as it will remove the ads after they have already been loaded in on the client side. 
So with the view that we will be looking to implement the full end to end solution this is a workaround to disable the ads for this particular article. 
Very much open to discussion on whether we should be doing this at all or whether we would prefer to push back on it, so let me know!


packages/article-skeleton/src/article-body/article-body.js
```diff
@@ -100,35 +100,36 @@ const renderers = ({
...
   isLiveOrBreaking,
   section,
-  articleHeadline
+  articleHeadline,
+  id
 }) => ({
   ...coreRenderers,
   ad(key) {
-    return (
+    return id === 'c8bf6998-d498-11ed-b5c3-54651fc826e9' ? null : (
```
douglasmik, line 108: could we reference an array, which id's can be added to, for this? We could then just manage the list, as opposed to needing to update the logic here.

bibicollins1: Good shout 👍

douglasmik (approved):

PR #3348, opened 2023-05-05
fix(fonts): Fixed fonts for all Puzzles components

lukakereceni:


josipVu (approved):

ibrahim-kurhan (commented):
I think these fonts have already been added by @douglasmik https://github.com/newsuk/times-components/blob/master/.storybook/preview-head.html its worth to have a look

douglasmik (approved):

PR #3374, opened 2023-05-23
feat(TMRX-1272): resolve newskit typography

douglasmik:
Update Newskit themes with function to map to Render fonts and a base font size of 10px.

### DIFF
--- a/packages/ts-newskit/src/utils/theme-formatting.ts
+++ b/packages/ts-newskit/src/utils/theme-formatting.ts
@@ -7,19 +7,29 @@ const modifyBaseRemValue = (multiplier: number, valueToUpdate: string) => {
   return `${updatedVal}rem`
 }
 
+const modifyFontFamilyValue = (family: string, weight?: number) => {
+  let updatedFont = family;
+  switch (family) {
+    case "Roboto":
+      updatedFont = weight && weight > 500 ? "Roboto-Medium" : "Roboto-Regular"
+      break;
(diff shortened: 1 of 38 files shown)


ibrahim-kurhan (approved):

packages/ts-newskit/src/utils/theme-formatting.ts
```diff
@@ -0,0 +1,111 @@
+import { StylePreset, Theme, TypographyPreset } from 'newskit';
+
+const modifyBaseRemValue = (multiplier: number, valueToUpdate: string) => {
+  const getVal = valueToUpdate.slice(0, -3);
```
adamosborne-tnl, line 4: is it always 'rem' in there, would it be easier to use parseFloat(valueToUpdate)?

adamosborne-tnl (commented):
one very minor question

packages/ts-newskit/src/utils/theme-formatting.ts
douglasmik: Yeah, this is only called if there's a 'rem' in the returned value

adamosborne-tnl (approved):

packages/ts-newskit/src/utils/theme-formatting.ts
douglasmik: thinking, I might move the logic to inside the function though to cut down on repetition?

PR #3380, opened 2023-05-25
fix(TMRX-1272): font resizing

douglasmik:
Revert font family mapping to resolve discrepancy in font metrics from existing fonts.
Will need to resolve fonts to remove any that are unnecessary in feedback ticket.


.storybook/preview-head.html
```diff
@@ -1,4 +1,12 @@
 <style>
+  @font-face {
+    font-family: "Times Modern";
```
ibrahim-kurhan, line 3: i guess it doesn't matter whether they are in string or not

.storybook/preview-head.html
```diff
@@ -1,4 +1,12 @@
...
+    src: url(./fonts/TimesModern-Bold.woff2) format("woff2"),
+      url(./fonts/TimesModern-Bold.woff) format("woff");
+    font-weight: 600;
+    font-style: normal;
+  }
+
   @font-face {
     font-family: TimesModern-Bold;
```
ibrahim-kurhan, line 11: i guess it doesn't matter whether they are in string or not

adamosborne-tnl (approved):

ibrahim-kurhan (approved):

PR #3411, opened 2023-06-22
fix(TMRX-0000): apply design feedback

ibrahim-kurhan:
This PR addresses an issue with the height of the divider component on the Safari browser. Previously, the divider had a fixed height of 100% as the default, which caused rendering issues specifically on Safari. To resolve this issue, I have made a modification by adding height: auto to the divider component.

### DIFF
--- a/packages/ts-newskit/src/components/slices/lead-article/index.tsx
+++ b/packages/ts-newskit/src/components/slices/lead-article/index.tsx
@@ -47,7 +47,7 @@ export interface LeadArticleProps {
   loadingAspectRatio?: string;
   marginBlockStart?: string;
   textBlockMarginBlockStart?: string;
-  tagAndFlagBlockMarginBlockStart?: string;
+  tagAndFlagMarginBlockStart?: string;
   listData?: ListData[];
   showTagL1?: boolean;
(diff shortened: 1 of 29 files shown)


packages/ts-newskit/src/components/slices/lead-article/index.tsx
```diff
@@ -59,14 +65,32 @@ export const LeadArticle = ({
...
+  imageMarginBlockStart = 'space000',
+  textBlockMarginBlockStart = 'space040',
+  tagAndFlagMarginBlockStart = 'space040',
   listData,
-  showTagL1
+  showTagL1,
+  hideImage
 }: LeadArticleProps) => {
+  const cardImage = image &&
+    image.src !== '' && {
```
douglasmik, line 78: could use optional chaining? A few more examples below
e.g. `image?.src && { ...`

ibrahim-kurhan: thats what i did first but typescript doesn't like it so i had to change it

douglasmik (approved):
Nice!

adamosborne-tnl (approved):

PR #3416, opened 2023-07-05
Feat(TMRX-1250): Live story new update notification

bibicollins1:
## Description

Created a new component that will alert users to a new update in a live story. Clicking on the button will trigger a page refresh which will pull in the latest version of the article. The component should display when there is an update, and after 8 seconds will disappear, or after a click will disappear. 
I've created a component `DelayComponent` (Happy for naming inputs here) which is passed an initial state and then a delay, so you could trigger a component to initially not display and render after an amount of time, or vice versa. 
I have wrapped the created an updateButton component without the delay, and another updateButtonWithDelay which includes this `DelayComponent` 

[TMRX-1250](https://nidigitalsolutions.jira.com/browse/TMRX-1250)
[Design](https://www.figma.com/file/0M40HSI31kibPGe3enuqCR/TNL---UI-Breaking-news-MVP?type=design&node-id=4140-233923&mode=design&t=Q0xrTbBXepLcYh6n-0)

[TMRX-1250]: https://nidigitalsolutions.jira.com/browse/TMRX-1250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ts-newskit/src/assets/FilledArrowIcon.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+const FilledArrowIcon: React.FC = (props: any) => (
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    width={17}
+    height={16}
+    fill="none"
```
ibrahim-kurhan, line 8: do you think we should replace the "none" to "currentcolor"?

packages/ts-newskit/src/components/misc/delayed-component/delayed-component.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { useEffect } from 'react';
+
+export const DelayedComponent: React.FC<{
+  delay: number;
+  initialState: boolean;
+  children: any;
+}> = ({ delay, initialState, children }) => {
+  const [showElement, setShowElement] = React.useState(initialState);
```
ibrahim-kurhan, line 8: do you think it would be more readable to import useState from react?

ibrahim-kurhan (commented):
run it local as well looks great but just left a couple of small comments

packages/ts-newskit/src/assets/FilledArrowIcon.tsx
douglasmik: think that may be the background with `currentcolor` being applied to the path?

douglasmik (approved):

ibrahim-kurhan (approved):

packages/ts-newskit/src/components/misc/delayed-component/delayed-component.tsx
bibicollins1: Oh yes definitely 👍

PR #3419, opened 2023-07-10
feat(TMRX-1250): Pass hasUpdate from render

bibicollins1:
## Description

We will be polling for the update time current vs latest in render, to render the update button when an update is available I have passed through the boolean hasUpdate which will render the component if true.


adamosborne-tnl (approved):
LGTM, does the button clear when clicked and scrolls to top and starts delay again?

PR #3429, opened 2023-07-17
feat(TMRX-1250): Live updates behind a flag

bibicollins1:
### Description 

Add fetching logic to UpdateButtonWithDelay. For now will be behind a feature flag in render 'showUpdate' as I cannot test this from storybook. 
I have ignored a test as will fix it if the functionality works and I go to remove the feature flag.


packages/ts-newskit/src/components/misc/update-button/update-button-with-delay.tsx
```diff
@@ -17,16 +20,41 @@ export const UpdateButtonWithDelay = ({
...
+  arrowUp,
+  updatedTime,
+  articleId
 }: UpdateWithDelayProps) => {
+  const [hasUpdate, setUpdate] = useState(false);
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const response = await fetch(`/api/article-update-time/${articleId}`);
```
bibicollins1, line 32: If I'm on an article page would this go to the right url or would it append

flashcheeks (approved):

ibrahim-kurhan (approved):

PR #3431, opened 2023-07-18
feat(TMRX-1250): Updates to update button

bibicollins1:
Description

Minor updates to reset state after delay has passed so that it should catch the next update and render the button again. 
Having a problem with caching so have passed true into reload() to hopefully break that.

### DIFF
--- a/packages/article-skeleton/src/article-skeleton.js
+++ b/packages/article-skeleton/src/article-skeleton.js
@@ -82,8 +82,6 @@ const ArticleSkeleton = ({
     isCommentEnabled
   } = article;
 
-  console.log(article.publishedTime, 'PUBLISHED TIME');
-
   const articleContentReducers = [
     insertDropcapIntoAST(template, dropcapsDisabled),
     insertNewsletterPuff(section, isPreview, expirableFlags),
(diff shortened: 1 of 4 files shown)


packages/article-skeleton/src/article-skeleton.js
```diff
@@ -82,6 +82,8 @@ const ArticleSkeleton = ({
     isCommentEnabled
   } = article;
 
+  console.log(article.publishedTime, 'PUBLISHED TIME');
```
ibrahim-kurhan, line 85: console.log?

adamosborne-tnl (approved):

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -260,7 +260,7 @@ const ArticleSkeleton = ({
                         label="New update"
                         handleClick={() => scrollToTopAndRefresh(window)}
                         arrowUp
-                        updatedTime={article.updatedTime}
+                        updatedTime={article.publishedTime}
```
ibrahim-kurhan, line 263: just out of curiosity why are you using publishedTime instead of updatedTime?

ibrahim-kurhan (approved):

packages/article-skeleton/src/article-skeleton.js
bibicollins1: So turns out the updateTime changes automatically at any update, e.g. fixing a typo. Editorial has more control over publishedTime so they can change that when they make meaningful updates instead.

bibicollins1: Apps currently have the feature pointing to publishedTime

PR #3435, opened 2023-07-20
feat(TMRX-1250): remove update button feature flag

bibicollins1:
## Description 

Remove feature flag `showUpdate` and increase the poll time for fetching data to 2 minutes.

### DIFF
--- a/packages/ts-newskit/src/components/misc/update-button/update-button-with-delay.tsx
+++ b/packages/ts-newskit/src/components/misc/update-button/update-button-with-delay.tsx
@@ -38,7 +38,7 @@ export const UpdateButtonWithDelay = ({
     };
     const interval = setInterval(async () => {
       (await fetchData()) > updatedTime && setUpdate(true);
-    }, 30000);
+    }, 120000);
 
     return () => clearInterval(interval);
   }, []);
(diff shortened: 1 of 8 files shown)


ibrahim-kurhan (approved):

adamosborne-tnl (approved):

packages/ts-newskit/src/components/misc/update-button/update-button-with-delay.tsx
```diff
@@ -38,7 +38,7 @@ export const UpdateButtonWithDelay = ({
     };
     const interval = setInterval(async () => {
       (await fetchData()) > updatedTime && setUpdate(true);
-    }, 30000);
+    }, 120000);
```
flashcheeks, line 41: Unrelated to this PR but found this hard to read - maybe a cleaner way is:

```
useEffect(() => {
  const fetchData = async () => {
    try {
      const response = await fetch(`/api/article-update-time/${articleId}`);
      const json = await response.json();
      if (json.article.publishedTime > updatedTime) {
        setUpdate(true);
      }
    } catch (err) {
      // tslint:disable-next-line:no-console
      console.log(err);
    }
  };
  const interval = setInterval("fetchData", 120000); 
  return () => clearInterval(interval);
}, [updatedTime]);
```

bibicollins1: @flashcheeks Yeah that's definitely a good point! We will be updating the component with some animations etc soon so will make it part of that work. Thanks!

PR #3460, opened 2023-07-26
feat(TMRX-1314): add credits lead image articles

ibrahim-kurhan:
### Description

add credits lead image articles

[JIRA-1314](https://nidigitalsolutions.jira.com/browse/TMRX-1314)

### Checklist

Screenshot 2023-07-26 at 19 05 47


packages/ts-newskit/src/components/slices/lead-article/index.tsx
```diff
@@ -139,6 +141,7 @@ export const LeadArticle = ({
                 typographyPreset="utilityMeta010"
               >
                 {images && images.caption}
+                {images && <StyledSpan>{images.credits}</StyledSpan>}
```
bibicollins1, line 144: Same here - not sure if credits always come through in the data. Sometimes can be null from what I remember.

bibicollins1: Oh actually unless your clearCreditsAndCaption function returns them as empty strings that should be fine I think

bibicollins1 (approved):

packages/ts-newskit/src/components/slices/lead-article/index.tsx
douglasmik: Ideally, would have check here also to not return empty element if string is empty

douglasmik (approved):

packages/ts-newskit/src/components/slices/lead-article/index.tsx
ibrahim-kurhan: added extra check also i have changed hasCaption to hasCaptionOrCredits and checking both

PR #3477, opened 2023-08-03
feat(TMRX-1312): Add tracking to Slice Header

bibicollins1:
### Description

Add tracking to Slice Header component.

[TMRX-1312](https://nidigitalsolutions.jira.com/browse/TMRX-1312)

### Checklist

[TMRX-1312]: https://nidigitalsolutions.jira.com/browse/TMRX-1312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ts-newskit/src/slices/fixtures/analytics-actions.ts
```diff
@@ -0,0 +1,7 @@
+import { action } from '@storybook/addon-actions';
+
+export default (event: any | any[]) => {
```
yukiinasawa, line 3: Just for own understanding - if event could be an array then we need to map to call each action?

bibicollins1: Oh I think that's from the original component which also handled scroll interactions 
Can remove that type from this

bibicollins1: Do have a point though, in this case would just be passing the single event so don't need the array type

yukiinasawa: Ah okay - so `ts-components` is handling both cases. - thanks 👍

packages/ts-newskit/package.json
```diff
@@ -54,11 +54,13 @@
     "@emotion/styled": "11.11.0",
     "@newskit-themes/the-times": "0.1.1",
     "isomorphic-unfetch": "3.1.0",
+    "lodash.merge": "4.6.2",
```
yukiinasawa, line 57: Are we using this somewhere in this PR?

yukiinasawa (approved):

ibrahim-kurhan (approved):

PR #3525, opened 2023-09-11
feat(TMRX-1300): create article byline block

milanharia-tm:
### Description

Context

An update to the current live file separating block with the rule and fine and date to include a byline name and pic.

Purpose

We need this as the current way of using datawrapper byline pics get very heavy from a page load point of view so we are hoping this way would be much lighter

Requirements

It’d be great if this was really easy to update - so if we had some analysis from William Hague for example (!), we could quickly add his byline embed.

It would be great if the second line had enough space for us to be able to include a sentence of description with the author byline, thus helping google discover authors who are experts in their fields. 

Technical implementation: 
You can expect two new fields in the parameters data for the live dividers. One will be the slug name of the author, and another will be an update type.

With the slug name of the author this will be used to query the additional author information (name, image and jobTitle) 

For this query as it is within times-components we may need to use a similar approach to the one we have for live update button where you create a new route in cps-content-render /api which you pass the slug of the author to fetch this data. 

For the update type you should be able to just take the value and render that on the UI component. 

N.B. both fields should be OPTIONAL 

Designs

Figma: [Breaking & Live News Prototype](https://www.figma.com/file/9W57iKCgqQstLzTOxI1AAr/Breaking-%26-Live-News-Prototype?type=design&node-id=0-1&t=2lrHQgXHrGJx8r1h-0) 

https://docs.google.com/document/d/1J-axkSya1ytBrtrhKYqiBveIHo_VgeWdT2REJD49I4Q/edit

Stakeholders

Anthony Cappaert (Design Editor)
Andrew Keys (Acting Creative Director)
Barney Henderson (Assistant Editor)
Tim Arbabzadah (Chief Sub Editor)

[JIRA-1300](https://nidigitalsolutions.jira.com/browse/TMRX-1300)

### Checklist

Screenshot 2023-09-11 at 12 19 47

### DIFF
--- a/packages/ts-components/src/components/article-header/ArticleBylineBlock.tsx
+++ b/packages/ts-components/src/components/article-header/ArticleBylineBlock.tsx
@@ -12,10 +12,10 @@ import {
 } from './styles';
 
 export const ArticleBylineBlock: React.FC<{
-  data?: ArticleByline | null;
+  data?: ArticleByline;
   description?: string;
 }> = ({ data, description }) => {
-  if (!(data && data.name && data.image) && !description) {
+  if (!((data && data.name) || (data && data.image)) && !description) {
     return null;
   }
 
--- a/packages/ts-components/src/components/article-header/ArticleBylineBlock.tsx
+++ b/packages/ts-components/src/components/article-header/ArticleBylineBlock.tsx
@@ -1,0 +1,51 @@
+import React from 'react';
+import { ArticleByline } from '../../types/related-article-slice';
+import {
+  BylineBlockContainer,
+  BylineBlockImgContainer,
+  BylineBlockImg,
+  BylineBlockContent,
+  BylineBlockAuthorContent,
+  BylineBlockAuthorName,
+  BylineBlockAuthorJobTitle,
+  BylineBlockAuthorDescription
+} from './styles';
+
+export const ArticleBylineBlock: React.FC<{
+  data?: ArticleByline | null;
+  description?: string;
+}> = ({ data, description }) => {
+  if (!(data && data.name && data.image) && !description) {
+    return null;
+  }
+
+  return (
+    <BylineBlockContainer>
+      {data &&
+        data.image && (
+          <BylineBlockImgContainer>
+            <BylineBlockImg src={data.image} alt={data.name} />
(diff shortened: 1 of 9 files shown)


packages/ts-components/src/components/article-header/ArticleHeader.tsx
```diff
@@ -102,6 +125,7 @@ const ArticleHeader: React.FC<{
       </UpdatesContainer>
 
       {headline && <Headline>{safeDecodeURIComponent(headline)}</Headline>}
+      <ArticleBylineBlock data={bylineData} description={description} />
```
bibicollins1, line 128: Just wondering if we should update the naming here so it's more clear that it's byline data? Having data and description as the props is slightly confusing. So could just have authorData or bylineData replacing the data prop?

packages/ts-components/src/components/article-header/ArticleBylineBlock.tsx
```diff
@@ -0,0 +1,51 @@
...
+            )}
+          {data &&
+            data.jobTitle && (
+              <BylineBlockAuthorJobTitle>
+                {data.jobTitle}
+              </BylineBlockAuthorJobTitle>
+            )}
+        </BylineBlockAuthorContent>
+        {description && (
+          <BylineBlockAuthorDescription>
```
bibicollins1, line 44: Given we are splitting the author data from the description data wonder if we could just call this Description? As is separate to the author. Just naming suggests it's a description of the author.

packages/ts-components/src/components/article-header/ArticleHeader.tsx
```diff
@@ -102,6 +125,7 @@ const ArticleHeader: React.FC<{
       </UpdatesContainer>
 
       {headline && <Headline>{safeDecodeURIComponent(headline)}</Headline>}
+      <ArticleBylineBlock authorData={authorData} description={description} />
```
bibicollins1, line 128: Do we want to render this conditionally?

yukiinasawa (approved):

ibrahim-kurhan (approved):

bibicollins1 (approved):

PR #3575, opened 2023-10-11
feat(TMRX-000): Handle TakeoverPage content and scripts

flashcheeks:
### Description

Takeover pages are articles without normal text and image content. Instead html and scripts are loaded from TPA and embedded into the article body.

This PR adds `embeddedContent` to the TPA query and injects the content and scripts.

### Checklist


.nvmrc
```diff
@@ -1 +1 @@
-v12.22.0
+v16.14.2
```
flashcheeks, line 1: Looks like this is now the version used in the CI pipeline

PR #3577, opened 2023-10-11
feat(TMRX-000): Handle TakeoverPage content and scripts

flashcheeks:
### Description

Takeover pages are articles without text and images. Instead html and javascript is return by TPA and embedded into the article body. This PR adds `embeddedContent` to the TPA query and renders the returned html and javascript.

If TC articles can handle takeover pages, we can delete the legacy template in Render 🎉🎉🎉🎉🎉🎉🎉 

### Checklist


.nvmrc
```diff
@@ -1 +1 @@
-v12.22.0
+v16.14.2
```
flashcheeks, line 1: It seems this is the version of Node running in the CI pipeline

packages/fixture-generator/src/mock-article.ts
```diff
@@ -61,6 +61,7 @@ class MockArticle {
         .addParagraphs(3)
         .addAds()
         .get(),
+      embeddedContent: null,
```
flashcheeks, line 64: If we don't specify a `null` value - the fixture generator creates one, breaking all integrations tests

packages/provider-test-tools/fixtures/article.json
```diff
@@ -1369,6 +1369,7 @@
       ]
     }
   ],
+  "embeddedContent": null,
```
flashcheeks, line 1372: If we don't specify a `null` value - the fixture generator creates one, switching all articles to takeovers and breaking storybook previews

PR #3589, opened 2023-10-13
feat(TMRX-1494): update mobile web tokens

milanharia-tm:
### Description

Tokens are updated for mobile web where applicable

[JIRA-1494](https://nidigitalsolutions.jira.com/browse/TMRX-1494)

### Checklist

Screenshot 2023-10-13 at 10 16 44


packages/ts-newskit/src/components/slices/shared/tag-and-flag.tsx
```diff
@@ -53,8 +61,16 @@ export const TagAndFlag = ({
 
       {flag && (
         <TextBlock
-          typographyPreset="utilityMeta005"
-          stylePreset="inkSubtle"
+          typographyPreset={
+            flagOverrides && flagOverrides.typographyPreset
```
bibicollins1, line 65: Quick question so are the flagOverrides different depending on which slice they are in?

packages/ts-newskit/src/slices/lead-story-3/article-stack.tsx
```diff
@@ -26,14 +26,21 @@ export const ArticleStack = ({ leadArticles, clickHandler }: ArticlesProps) => {
                   textBlockMarginBlockStart: 'space050',
                   headlineTypographyPreset: {
                     xs: 'editorialHeadline040',
-                    sm: 'editorialHeadline050',
                     md: 'editorialHeadline060'
                   },
+                  tagAndFlagMarginBlockStart: {
```
bibicollins1, line 31: Seems like these values are pretty consistent for the tag and flag component? Is it better to define them in the tag and flag?

packages/ts-newskit/src/components/slices/shared/tag-and-flag.tsx
milanharia-tm: It's because the component is being used for the comments as well, so I had to add an extra prop to override styles to change it just for the articles.
Screenshot 2023-10-16 at 12 14 36

packages/ts-newskit/src/slices/shared/comment-stack.tsx
```diff
@@ -15,10 +15,29 @@ interface CommentStackProps {
...
+            }}
+          />
+        </Visible>
+        <Hidden xs sm>
+          <Divider
+            overrides={{
+              marginBlock: 'space040',
+              stylePreset: 'dashedDivider'
+            }}
+          />
```
douglasmik, line 39: could we use media query for `stylePreset` here instead of having the 2 components in Visiible/Hidden?
e.g. stylePreset: {{xs: 'lightDashedDivider', md: 'dashedDivider' }}

douglasmik (commented):
there's a few instances where we're using `Visible`/`Hidden` where we could use a single component with `MQ` values in props instead. This would help reduce the number of elements being output to the DOM

packages/ts-newskit/src/slices/shared/comment-stack.tsx
milanharia-tm: I did this originally with the MQ values, but for some reason the styling got messed up and the dividers became thicker (because the border width was no longer being set properly) so I had to use `Visible`/`Hidden` elements as a workaround for this which I agree is not ideal.

packages/ts-newskit/src/components/slices/shared/tag-and-flag.tsx
bibicollins1: Ah okay makes sense - would it be possible to set a default on them for the most frequent - I assume articles? And then the overrides for the comment ones?

milanharia-tm: Yes that's how I did it, this prop is only being used in the `CommentCard` component to overwrite the default styling.

packages/ts-newskit/src/slices/content-bucket-1/index.tsx
```diff
@@ -97,13 +99,28 @@ export const ContentBucket1 = ({
...
+            paddingInlineStart={{
+              xs: 'space045',
+              md: 'space000'
+            }}
+          >
+            <Visible xs sm>
+              <Divider
+                overrides={{
+                  marginBlock: 'space040',
+                  stylePreset: 'lightDashedDivider'
```
douglasmik, line 112: with `MQ` based values could use a single `Divider` here

packages/ts-newskit/src/slices/content-bucket-3/__tests__/index.test.tsx
```diff
@@ -38,8 +39,8 @@ describe('Render Content Bucket 3 Slice', () => {
 describe('Content Bucket 3 Articles list ', () => {
   test('articleBorder render after odd numbered items', () => {
     (useBreakpointKey as any).mockReturnValue('xl');
```
douglasmik, line 41: shouldn't need `useBreakpointKey` mocks anymore with `renderComponentWithBreakPoint`

packages/ts-newskit/src/slices/content-bucket-3/index.tsx
```diff
@@ -130,19 +131,44 @@ export const ContentBucket3 = ({
...
-          <ArticleStack
-            articles={articles}
-            breakpoint={currentBreakpoint}
-            clickHandler={clickHandler}
-          />
+          <Visible xs sm>
+            <ArticleStack
+              articles={articles.map(article => ({
+                ...article,
+                topBorderStyle: 'lightDashedDivider'
```
douglasmik, line 156: could use a single component here and pass different border presets in `ArticleStack` using `MQ` values

packages/ts-newskit/src/slices/shared/comment-stack.tsx
douglasmik: Is it passing the correct values on inspect? If so, I would be inclined to leave, or review

packages/ts-newskit/src/slices/shared/comment-stack.tsx
```diff
@@ -29,13 +47,25 @@ export const CommentStack = ({ comments, clickHandler }: CommentStackProps) => {
...
+              <Hidden xs sm>
+                <AvatarDivider
+                  overrides={{
+                    marginBlock: 'space000',
+                    stylePreset: 'lightDivider'
+                  }}
+                  vertical
+                />
+              </Hidden>
+            </>
```
douglasmik, line 68: as above

packages/ts-newskit/src/slices/stacked-module-1/index.tsx
```diff
@@ -29,13 +35,29 @@ const articleStack = ({
...
+          />
+        </Visible>
+        <Visible md lg xl>
+          <Divider
+            overrides={{
+              stylePreset: 'dashedDivider',
+              marginBlockStart: marginBlockStart || 'space000'
+            }}
+          />
+        </Visible>
```
douglasmik, line 60: as above

packages/ts-newskit/src/slices/stacked-module-1/index.tsx
```diff
@@ -62,20 +84,40 @@ const articleStack = ({
...
-                clickHandler={clickHandler}
-              />
+              <Visible xs sm>
+                <Article
+                  article={{
+                    ...clearCreditsAndCaption(article),
+                    hideImage: hasImage || (hasImage || isDesktop),
+                    isLeadImage: isMob && articleIndex === 0,
+                    hasTopBorder: hasImage,
+                    topBorderStyle: 'lightDashedDivider',
```
douglasmik, line 94: can handle this with MQ value in `Article` component

packages/ts-newskit/src/slices/shared/comment-stack.tsx
bibicollins1: Looks like it's doubling up the dividers worth checking if there are two rendering when you use the media query values

milanharia-tm: With the `MQ` it's passing:
```
border-style: dashed;
border-width: 1px;
border-color: #01000d;
```

Without the `MQ` it passes:
```
border-style: dashed;
border-width: 1px;
border-color: #01000d;
border-width: 0px;
margin: 0;
border-top-width: 1px;
width: 100%;
margin-block: 16px;
```

Not entirely sure where the extra styling is coming from?

packages/ts-newskit/src/slices/stacked-module-1/index.tsx
milanharia-tm: Again was done to work around the divider issue

packages/ts-newskit/src/slices/content-bucket-3/__tests__/index.test.tsx
milanharia-tm: Left this in as there is still breakpoint logic in `ContentBucket3` so decided this can be remedied in the refactor work instead of this ticket?

douglasmik (approved):

adamosborne-tnl (approved):

PR #3621, opened 2023-10-30
fix(TMRX-1616): Content bucket article stack headline margin

milanharia-tm:
### Description

The top margin remains even though the article doesn't have a label on LG breakpoint. This affects both `ContentBucket1` and `ContentBucket3`.

image

[JIRA-1616](https://nidigitalsolutions.jira.com/browse/TMRX-1616)

### Checklist

**BEFORE**
Screenshot 2023-10-30 at 12 03 35

**AFTER**
Screenshot 2023-10-30 at 12 03 54

### DIFF
--- a/packages/ts-newskit/src/components/slices/article/index.tsx
+++ b/packages/ts-newskit/src/components/slices/article/index.tsx
@@ -93,7 +93,8 @@ export const Article = ({
       getActiveArticleFlags(expirableFlags) &&
       expirableFlags.length > 0) ||
     label ||
-    contentType;
+    contentType ||
+    hasVideo;
 
   const cardImage = !hideImage &&
     imageWithCorrectRatio && {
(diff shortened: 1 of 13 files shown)


packages/ts-newskit/src/components/slices/article/index.tsx
```diff
@@ -187,13 +188,15 @@ export const Article = ({
           )}
 
         <ArticleTileInfo
+          className={hasArticleTileInfo ? 'article-info' : ''}
```
milanharia-tm, line 191: Only apply class if we have these details. Wondering whether it's worth refactoring this component so it doesn't render unless it has the data? Could we conditionally render based on `hasArticleTileInfo`?

douglasmik: agree, if we don't have data no value in rendering this 👍

yukiinasawa (approved):

bibicollins1 (approved):

douglasmik (approved):

PR #3632, opened 2023-11-06
feat(TMP-506): Update hardcoded references to thetimes.co.uk in Times component: group 1

hmaracicTNL:
### Description

[TMP-506](https://nidigitalsolutions.jira.com/browse/TMP-506)

### Checklist

[TMP-464]: https://nidigitalsolutions.jira.com/browse/TMP-464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TMP-506]: https://nidigitalsolutions.jira.com/browse/TMP-506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/utils/src/constants.js
+++ b/packages/utils/src/constants.js
@@ -1,0 +1,3 @@
+export const BASE_DOMAIN = "thetimes.co.uk";
+export const IMAGE_BASE_URL = "https://www.thetimes.co.uk";
+export const BASE_URL = "https://www.thetimes.co.uk";
--- a/packages/context/src/constants.js
+++ b/packages/context/src/constants.js
@@ -1,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export const BASE_URL = "https://www.thetimes.co.uk";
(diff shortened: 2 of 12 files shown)


packages/utils/src/constants.js
```diff
@@ -0,0 +1,3 @@
+export const BASE_DOMAIN = "thetimes.co.uk";
+export const IMAGE_BASE_URL = "https://www.thetimes.co.uk";
+export const BASE_URL = "https://www.thetimes.co.uk";
```
douglasmik, line 3: do we need both if the values are the same?

hmaracicTNL: I think the idea was that image server might not be on the same domain at the same time as the main server, I don't know if that could be true. It was not really much more work when modifying the code to introduce the second variable for this and it added some flexibility. If it's not needed I can remove it and make use of the same one for both.

PR #3639, opened 2023-11-07
fix(TMRX-1635): Fix lead story article headline margin

peacewilliams:
### Description
The top margin remains even though the article doesn't have a label on LG breakpoint for stacked lead articles. This same issue was fixed for content bucket articles in TMRX-1616: The top margin remains even though the article doesn't have a label on LG breakpoint

[TMRX-1635](https://nidigitalsolutions.jira.com/browse/TMRX-1635)

### Checklist

#### Correct spacing shown below on LG Breakpoint
image

[TMRX-1635]: https://nidigitalsolutions.jira.com/browse/TMRX-1635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ts-newskit/src/components/slices/article/index.tsx
```diff
@@ -99,6 +99,10 @@ export const Article = ({
     hasVideo;
 
   const marginBlockStart = imageRight || hideImage ? 'space000' : 'space040';
+  const hasNoProfileAndFlag =
+    !expirableFlags && !contentType && !label && !hasVideo
+      ? 'space000'
+      : 'space040';
```
douglasmik, line 105: do we need this? looks like we should be able to update logic in `CardHeadlineLink` as per comment below

packages/ts-newskit/src/components/slices/article/index.tsx
```diff
@@ -213,7 +217,9 @@ export const Article = ({
           role="link"
           overrides={{
             typographyPreset: titleTypographyPreset,
-            marginBlockStart: hasArticleTileInfo ? 'space030' : marginBlockStart
+            marginBlockStart: hasArticleTileInfo
+              ? 'space030'
+              : hasNoProfileAndFlag
```
douglasmik, line 222: do we need this `hasNoProfileAndFlag`? looks like this is being checked on line 220, where this could be `space000`?

douglasmik: Think it was correct with the value before butt we need to use targeted styling for this slice at this breakpoint (between LG-XL) to remove the spacing for when there isn't an image as we done here (packages/ts-newskit/src/slices/shared-styles/index.ts)

packages/ts-newskit/src/components/slices/article/index.tsx
```diff
@@ -213,7 +221,10 @@ export const Article = ({
           role="link"
           overrides={{
             typographyPreset: titleTypographyPreset,
-            marginBlockStart: hasArticleTileInfo ? 'space030' : marginBlockStart
+            marginBlockStart: {
+              xs: 'space030',
+              md: hasArticleTileInfo ? 'space030' : 'space000'
+            }
           }}
```
milanharia-tm, line 228: @peacewilliams The issue with changing margins on the `Article` component is that it will likely affect other slices using it. This change in particular seems to affect `ContentBucket1` on the `xl` breakpoint for example (see the articles on the right).

Screenshot 2023-11-14 at 11 20 47

I think a better approach would be to use CSS (i.e a styled component wrapper) for the lead story slices and not modify the article component directly so that the styles do not leak into other components? A similar approach was used for the content bucket components.

yukiinasawa: Have you addressed this one? Seems still it sets to `'space000'` 👀

peacewilliams: @yukiinasawa you were right! `marginBlockStart` has been removed now as `marginBlockEnd` will now handle all the spacing for the articles

yukiinasawa (approved):
Tested locally and looks nice 👍

PR #3689, opened 2023-12-12
feat(TMRX-1479): develop list slice article

ibrahim-kurhan:
### Description

As a web reader
I want to view stories laid out in a specified format
So that I can make a decision on the content I want to read
Acceptance Criteria

UI

List View slice is implemented in storybook as per design file

Pagination

Desktop: directional arrows and numeric page links after 10th article and before ad slot

Mobile web: Load more CTA to load the next 10 articles (with ahref link not javascript so that google bot can crawl)

Ad slots are positioned in line with design:

Mobile web: Ad after 10th article

Desktop: Right rail ad should be sticky on LG and XL breakpoints

Please inform @Iliana Avramova what id we have given the containing div so that the ads team are able to target it. 

Tracking

Tracking spec is implemented on List View slice as per spec: https://docs.google.com/spreadsheets/d/14JamPsFxWCg2gLXZeYSIPAFaJgthWXeU/edit?usp=sharing&ouid=103585823547702063140&rtpof=true&sd=trueConnect your Google account (tab: Event Navigation)

Columns - EA (title bock) and EB (Article category link)

Column - BZ (article headlines or image on channel pages)

Data team are able to track interactions with the elements on List View slice on a test channel page

Description

Build List View slice UI as per design file: [TML_Sport_Himesh](https://www.figma.com/file/C1K0ePqSX0irobX6lAMq7h/Sport-section?type=design&node-id=29-60480&mode=design&t=CFgvRKEUwr0mEaOQ-0) 

Pagination

Desktop: directional arrows and numeric page links after 10th article and before ad slot

Mobile web: Load more CTA to load the next 10 articles (with ahref link not javascript so that google bot can crawl)

Ad placement is in the correct position

Mobile web: Ad after 10th article

Desktop: Right rail ad should be sticky on LG and XL breakpoints

Implement tracking spec on List View: https://docs.google.com/spreadsheets/d/14JamPsFxWCg2gLXZeYSIPAFaJgthWXeU/edit#gid=1392155890Connect your Google account 

One pager: https://docs.google.com/document/d/1oZwpBfFbpSFmWXFswFDY7b_dpZnqBd8Q-QqVJUFdtFk/edit#heading=h.dndrkh2u2gmpConnect your Google account 

Spike Documentation: [Spike: TMRX-1664 Define technical approach to List View Slice](https://nidigitalsolutions.jira.com/wiki/spaces/TNLDigital/pages/4621369460/Spike+TMRX-1664+Define+technical+approach+to+List+View+Slice) 

NB: We do not currently have the schema for this, but expect article data to follow existing data structure. Will need to resolve data for pagination, where existing topic slice may serve as the best example for this. Please catchup with @Tim Snow @Moomba Michelo Muloba 

Refinement notes:

When we use the pagination function, are we updating the whole page or just the list slice?

Should each paginated page have a unique URL or should the URL remain static at all times?

Update URL but page not refresh

When you use the pagination, should you be navigated to the top of the list slice view or stay at the bottom?
[JIRA-1446](https://nidigitalsolutions.jira.com/browse/TMRX-1446)

### Checklist

https://github.com/newsuk/times-components/assets/139121277/b28a7590-afa8-415e-8fe5-c56c926157b7


packages/ts-newskit/src/slices/list-view-slice/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Visible } from 'newskit';
+import { ListViewSliceMobile } from './mobile';
+import { ListViewSliceDesktop } from './desktop';
+
+export type ListViewSliceProps = {
+  leadArticles: LeadArticleProps[];
+  clickHandler: ClickHandlerType;
+  articleWithAdSlot?: LeadArticleProps;
+  currentPage?: number;
+  handlePageChange: (page: number) => void;
```
yukiinasawa, line 14: I might be wrong but the reason why I made `onMobileClick` and `onDesktopclick` separately was that I thought render needs to know whether if it's for mobile or desktop as need to dispatch different actions. Do we need to pass props? i.e.) ` handlePageChange: (page: number, isMobile: boolean) => void;`

packages/ts-newskit/src/slices/list-view-slice/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+    isLeadImage: false
+  }));
+
+  const sliceProps = {
+    leadArticles: modifiedLeadArticles,
+    clickHandler,
+    currentPage,
+    handlePageChange,
+    itemsPerPage,
+    totalItems: leadArticles.length,
```
yukiinasawa, line 43: `LeadArticle.length` will be 10 in our case but we need know `totalItems` for pagination component.

packages/ts-newskit/src/slices/list-view-slice/index.tsx
douglasmik: should be the same event - where the 'mobileClick' is the same as next page on desktop. `handlePageChange` is condensed to a single function that receives a page number, which is set from the button being clicked on, for pagination

packages/ts-newskit/src/slices/list-view-slice/index.tsx
douglasmik: not sure I follow? `totaltems` is equal to the total amount of articles here. Pagination now uses this and `itemsPerPage` to calculate how many pages to display, where it's now dynamic

yukiinasawa: If I understood our logic in render correctly, let's say page is 3 then we dispatch action then return articles 21-30 which we receive `leadArticles` in TC so the `LeadArticle.length` would be `10` here, however, `totaltems` should be num of all articles (not only 21-30) so I think it's not equal to `LeadArticle.length` if that makes sense.

packages/ts-newskit/src/slices/list-view-slice/index.tsx
yukiinasawa: mobile -> action takes page number and returns articles till that page number (if page is 3 then 0-30)
desktop -> action takes page number and return only 10 articles (if page is 3 then 21-30)
so I thought we need to know but I might be wrong

packages/ts-newskit/src/slices/list-view-slice/index.tsx
douglasmik: understand, makes sense - I think we currently get all articles and then use `articleCursor` and `articlesPerPage` to dictate the paginated articles that are displayed, so `leadArticles.length` is total amount of articles

packages/ts-newskit/src/slices/list-view-slice/index.tsx
douglasmik: understand, we could do it this way but think we'll follow the current approach of iterating through received articles (all articles) for both desktop/mobile, which would enable us to use the same function - won't have to fetch new data each time

packages/ts-newskit/src/slices/list-view-slice/index.tsx
yukiinasawa: Ah okay cool that makes sense! Thanks for clarifying 👍

packages/ts-newskit/src/slices/list-view-slice/index.tsx
yukiinasawa: Ah okay cool that makes sense! Thanks for clarifying 👍

yukiinasawa (approved):

ibrahim-kurhan (commented):
I can't approve it as I created the PR but great work @yukiinasawa @douglasmik

peacewilliams (approved):
LGTM - Great work!

PR #3710, opened 2024-01-09
fix(TMRX-1806): remove byline from tag and flag

ibrahim-kurhan:
### Description

the byline is currently appearing in the Tag and Flag section below the summary or short summary. The byline should be removed specifically from these sections, excluding the List View slice.

Acceptance Criteria

Byline should not be displayed in the Tag and Flag section(below the summary , short summary or headline) in slices other than the List View.
[JIRA-1806](https://nidigitalsolutions.jira.com/browse/TMRX-1806)

### Checklist

### DIFF
--- a/packages/ts-newskit/src/components/slices/lead-article/index.tsx
+++ b/packages/ts-newskit/src/components/slices/lead-article/index.tsx
@@ -41,6 +41,7 @@ export interface LeadArticleProps {
     label: string;
     href: string;
   };
+  isListView?: boolean;
   imageTop?: boolean;
   isLeadImage?: boolean;
   byline?: string;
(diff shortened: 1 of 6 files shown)


packages/ts-newskit/src/components/slices/lead-article/index.tsx
```diff
@@ -234,7 +237,7 @@ export const LeadArticle = ({
         <TagAndFlag
           tag={tag}
           flag={flag}
-          byline={byline}
+          byline={isListView ? byline : ''}
```
bibicollins1, line 240: This would still render the component just with an empty string right? Is it worth using the boolean flag to wrap the render of the byline component instead so we just don't render it at all?

ibrahim-kurhan: yes you are right, just pushed a commit to fix this issue

bibicollins1 (approved):

PR #3798, opened 2024-03-12
feat(TMRS-88): article domain reference updates

douglasmik:
### Description

Update hardcoded domain references to be dynamic, depending on host (.com/.co.uk)

[TMRS-88](https://nidigitalsolutions.jira.com/browse/TMRS-88)

### Checklist

[TMRS-88]: https://nidigitalsolutions.jira.com/browse/TMRS-88?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ts-components/src/components/recommended-articles/RecommendedArticles.tsx
```diff
@@ -26,6 +28,10 @@ export const RecommendedArticles: React.FC<{
   if (!articles || !articles.length) {
     return null;
   }
+  const transformedArticles = transformDomainCom(
+    articles,
+    'https://www.thetimes.com'
```
yukiinasawa, line 33: QQ: Why is it hardcoded as .com? should it be hostName?

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -194,6 +197,9 @@ const ArticleSkeleton = ({
   const isSharingSavingEnabledByTPA = savingEnabled || sharingEnabled;
   const isSharingSavingEnabled =
     isSharingSavingEnabledByTPA && isSharingSavingEnabledExternal;
+  const domainSpecificUrl = hostName
```
HarisHodzicnt, line 200: would it be easier if we set:
if (host?.includes('thetimes.com'))  in the getDomainSpecificUrl rather than having the check here ?

packages/ts-components/src/components/recommended-articles/RecommendedArticles.tsx
```diff
@@ -14,6 +15,7 @@ import { Header } from './styles';
 
 export const RecommendedArticles: React.FC<{
   heading: string;
+  domainSpecificUrl: string;
```
HarisHodzicnt, line 18: do we need domainSpecificUrl ?

packages/ts-components/src/components/recommended-articles/RecommendedArticles.tsx
douglasmik: It should! Was using this for testing and forgot to update!

packages/article-skeleton/src/article-skeleton.js
douglasmik: We actually have that `.com` check in the `getDomainSpecificUrl` function already, this check is if is `hostName`  isn't passed.

HarisHodzicnt: yes but if we have if (host?.includes('thetimes.com')) and hostname is falsy it will return the default url passed as a second param

HarisHodzicnt (approved):

yukiinasawa (approved):

packages/article-skeleton/src/article-skeleton.js
douglasmik: TC doesn't support optional chaining and host is a required parameter. This will be refactored shortly to receive function from Render where we could review perhaps

PR #3800, opened 2024-03-18
feat(TMD-11): redesigned article share bar

milanharia-tm:
### Description

Implements redesign of the article share bar.

[TMD-11](https://nidigitalsolutions.jira.com/browse/TMD-11)

### Checklist

### Screenshots:

https://github.com/newsuk/times-components/assets/142982056/bd35b899-00a7-40a3-9f87-64ab2a3be255

[TMD-11]: https://nidigitalsolutions.jira.com/browse/TMD-11?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ibrahim-kurhan (approved):
@milanharia-tm looks really good but would it be better to build this component in `ts-newskit` by using function based component other than its existing package

milanharia-tm:
> @milanharia-tm looks really good but would it be better to build this component in `ts-newskit` by using function based component other than its existing package

I thought about doing this originally, however, because I built on top of a lot of the existing components and logic I decided it made more sense to just add it to the `save-and-share` package. I think when we work one the next iteration (which is a bigger change) we should move this into `ts-newskit`.

bibicollins1 (approved):

amerkadicE (approved):

PR #3854, opened 2024-05-17
chore(TMD-336): Remove newskit from job title component

amerkadicE:
### Description

Removing newskit from job title component.

Also removed JobTitle integration because it is not being used currently. 
We will leave JobTitle implementation as it could be required in the feature. It will require adding a floating element to fix the popover being out of screen issue.

Source: https://newsuktechnology.slack.com/archives/C06MX1T0DGQ/p1716549169811519

[TMD-336](https://nidigitalsolutions.jira.com/browse/TMD-336)

### Checklist

New in the first column:
imageimage

imageimage

[TMD-336]: https://nidigitalsolutions.jira.com/browse/TMD-336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


milanharia-tm (commented):
Popover placement seems to be broken on author profile
Screenshot 2024-05-17 at 16 41 14

packages/author-profile/src/author-profile-head-base.js
```diff
@@ -20,8 +19,8 @@ const AuthorProfileHeadBase = memo(
     twitter,
     contractualTitle
   }) => {
-    const breakpoint = useBreakpointKey();
-    const isLargeDevice = breakpoint === "lg" || breakpoint === "xl";
+    const isLargeDevice =
+      typeof window !== "undefined" && window.innerWidth >= 1440;
```
milanharia-tm, line 23: Potential problem with this code is that it will not re-render if the breakpoint changes so you can end up with the tooltip on mobile if you resize. Are we happy with this behaviour?

packages/ts-newskit/src/components/misc/job-title/styles.ts
```diff
@@ -0,0 +1,154 @@
...
+
+export const Popover = styled.div`
+  position: relative;
+  display: inline-block;
+`;
+export const PopoverContext = styled.div<{ isOpen?: boolean }>`
+  display: ${({ isOpen }) => (isOpen ? 'grid' : 'none')};
+  position: absolute;
+  top: -55px;
+  left: 4px;
```
milanharia-tm, line 32: Positioning/sizing the element this way will have issues with smaller viewports since we are not also controlling width or altering the positioning based on screen size. 

Screenshot 2024-05-22 at 15 45 23

packages/author-profile/src/author-profile-head-base.js
amerkadicE: Most users won't resize their screen in the middle of the surfing. I think it is worth discussing best behavior.

packages/ts-newskit/src/components/misc/job-title/styles.ts
amerkadicE: It has the same behavior as the old Popover. I can update PR to support your suggestion if it is our priority.

milanharia-tm: The old popover adjusted its positioning if there was not enough space:

Screenshot 2024-05-23 at 10 29 35

amerkadicE: JobTitle component will be removed from the author profile page for now as discussed on slack.
https://newsuktechnology.slack.com/archives/C06MX1T0DGQ/p1716549169811519

packages/author-profile/.eslintrc.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": ["@times-components/thetimes"]
```
bibicollins1, line 2: Should be able to remove this change also @amerkadicE

packages/author-profile/src/author-profile-head-base.js
bibicollins1: @amerkadicE Think we can delete this block as won't be used if not integrated

bibicollins1 (approved):

ibrahim-kurhan (approved):

PR #3858, opened 2024-05-24
feat(TMRS-170): euros ticker replace flags

douglasmik:
Add function to replace flags for national competitions with references to flag images in S3
- add options to select competition in Storybook

[TMRS-170](https://nidigitalsolutions.jira.com/browse/TMRS-170)

### Checklist

[TMRS-170]: https://nidigitalsolutions.jira.com/browse/TMRS-170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ts-components/tsconfig.json
```diff
@@ -38,7 +38,7 @@
 
     "skipLibCheck": true,
 
-    "lib": ["es6", "dom", "es2017"],
+    "lib": ["es6", "dom", "es2017", "dom.iterable"],
```
yukiinasawa, line 41: Do we need this?

yukiinasawa (approved):

packages/ts-components/tsconfig.json
douglasmik: Needed to iterate over `HTMLCollection` that we have from `getElementsByClassName` method

bibicollins1 (approved):

packages/ts-components/src/components/opta/utils/replaceFlags.ts
```diff
@@ -0,0 +1,25 @@
+export const replaceFlags = async (
+  element: HTMLCollectionOf<Element> | undefined
+) => {
+  const replaceImages = await setInterval(() => {
```
HarisHodzicnt, line 4: Do we need await here ? 
setInterval should not be awaited since it does not return a promise

packages/ts-components/src/components/opta/utils/replaceFlags.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (element && element.length > 0) {
```
HarisHodzicnt, line 5: we can use element?.length > 0 

but if element can be null and there is a chance that element is missing all the time, we will have memory leak.
setInterval will never be closed, because clearInterval is inside this if check.

We should have at least counter, if element is null after 20 tries, clearInterval.

packages/ts-components/src/components/opta/utils/replaceFlags.ts
```diff
@@ -0,0 +1,25 @@
...
+      for (let optaFlagContainer of element) {
+        const country =
+          optaFlagContainer && (optaFlagContainer as HTMLElement).innerText;
+
+        if (!optaFlagContainer.querySelector('img')) {
```
HarisHodzicnt, line 10: I can see you have condition on line 8. optaFlagContainer && 
but on line 10 you dont have optaFlagContainer?.querySelector

if optaFlagContainer can be null/undefined, we will get error that we are trying to access property querySelector of undefined.

HarisHodzicnt (changes requested):

packages/ts-components/src/components/opta/utils/replaceFlags.ts
douglasmik: we don't need await, was left from earlier function 👍

packages/ts-components/src/components/opta/utils/replaceFlags.ts
douglasmik: would be 'falsy' in both scenarios, which is fine - with `if` statement wrapping this, `optaFlagContainer &&` is no longer needed. Can remove

packages/ts-components/src/components/opta/utils/replaceFlags.ts
douglasmik: package doesn't support optional chaining atm unfortunately

HarisHodzicnt (approved):

PR #3887, opened 2024-07-11
feat(TMD-539): Move the remaining components to ts-components

YusufYF:
### Description

`UpdateButton`, `UpdateButtonWithDelay`, `Banner`, and `DelayedComponent` have moved and still work as normal in: 

`import { UpdateButton, UpdateButtonWithDelay, Banner, DelayedComponent } from "@times-components/ts-components";`

⚠️ Original ticket was just to move the button, confirmed the rest with Adam. The icons were moved over to ts-components as well.

[TMD-539](https://nidigitalsolutions.jira.com/browse/TMD-539)

### Checklist

[TMD-539]: https://nidigitalsolutions.jira.com/browse/TMD-539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ts-components/src/components/update-button/update-button-with-delay.tsx
```diff
@@ -24,18 +35,8 @@ export const UpdateButtonWithDelay = ({
...
-        // tslint:disable-next-line:no-console
-        console.log(err);
-      }
-    };
     const interval = setInterval(async () => {
-      (await fetchData()) > updatedTime && setUpdate(true);
+      (await fetchData(articleId)) > updatedTime && setUpdate(true);
     }, 120000);
 
     return () => clearInterval(interval);
```
milanharia-tm, line 42: It is bad practice to have a useEffect whose dependency array does not reference its external dependencies, here we can either move the `fetchData` function back into the useEffect which will mean we can omit it from the dependency array, or wrap the `fetchData`  function in a `useCallback` hook so its reference remains constant between renders and add it to the dependency array. See the [react docs](https://react.dev/reference/react/useCallback) for more information.

milanharia-tm (approved):

PR #3913, opened 2024-08-04
feat(TMRS-268): update queries and article list component to use upda…

douglasmik:
feat(TMRS-268): update queries and article list component to use updated url field, which uses mapped `categoryPath` value
[TMRS-268](https://nidigitalsolutions.jira.com/browse/TMRS-268)

### Checklist

[TMRS-268]: https://nidigitalsolutions.jira.com/browse/TMRS-268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/article-list/src/article-list-item.js
+++ b/packages/article-list/src/article-list-item.js
@@ -107,32 +107,28 @@ const ArticleListItem = props => {
 
   return (
     <Context.Consumer>
-      {({ makeArticleUrl }) => {
-        const canonicalUrl = makeArticleUrl(props.article);
-
-        return (
-          <Link url={canonicalUrl}>
-            <ListItemWrapper>
-              <Card
-                contentContainerClass="articleListContent"
-                fadeImageIn={fadeImageIn}
-                highResSize={highResSize}
-                imageAccessibilityLabel={imageAccessibilityLabel}
-                imageContainerClass="articleListImage"
-                imageRatio={imageRatio}
-                imageUri={imageUri}
-                isLoading={isLoading}
-                lowResQuality={lowResQuality}
-                lowResSize={lowResSize}
-                showImage={showImage}
-                isLcpItem={index === 0}
-              >
-                <Animations.FadeIn>{children}</Animations.FadeIn>
-              </Card>
-            </ListItemWrapper>
-          </Link>
-        );
-      }}
+      {() => (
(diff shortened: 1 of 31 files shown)


packages/article-list/src/article-list-item.js
```diff
@@ -107,32 +107,28 @@ const ArticleListItem = props => {
...
-                isLcpItem={index === 0}
-              >
-                <Animations.FadeIn>{children}</Animations.FadeIn>
-              </Card>
-            </ListItemWrapper>
-          </Link>
-        );
-      }}
+      {() => (
+        <Link url={props.article.url}>
```
yukiinasawa, line 111: Isn't it `props.article.categoryPath` ?

packages/fixture-generator/src/types.ts
```diff
@@ -1351,6 +1351,8 @@ export interface Article extends ArticleInterface {
   /** Topics that the requested article belong to */
   topics?: (Topic | null)[] | null;
 
+  categoryPath?: Url | null;
```
yukiinasawa, line 1354: Isn't it `categoryPath: Url` as per prev our conversation earlier?

HarisHodzicnt (approved):

packages/article-list/src/article-list-item.js
douglasmik: I've mapped the `categoryPath` field to the `url` in the query;
e.g. `url: categoryPath`

packages/fixture-generator/src/types.ts
douglasmik: followed type for `url` where this is for the fixture generator, that is used to mock different components, etc - they all have conditional `null` values

packages/article-list/src/article-list-item.js
yukiinasawa: If i understand correctly - we receive `categoryPath` as `url` so then we don't need to add `categoryPath` fixtures you've added into fixtures files?

douglasmik: fixtures seemed to need it, as they were not 'querying' for the data but working from a static object

yukiinasawa (approved):

PR #4001, opened 2024-12-10
feat(TMD-1182): Update to TPA field for image alt text

amerkadicE:
### Description

Update the Image component to use the title property as alt.

[TMD-1182](https://nidigitalsolutions.jira.com/browse/TMD-1182)

### Checklist

BEFORE:
image

AFTER:
image

[TMD-1182]: https://nidigitalsolutions.jira.com/browse/TMD-1182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


milanharia-tm (approved):

bibicollins1 (commented):
One small thing - I think the title field will return an empty string if there is no title, so we might need to check whether it has an actual value and then fallback to caption if not? Worth checking TPA to see if it returns null or ''

ibrahim-kurhan (approved):

amerkadicE:
@bibicollins1 Shouldn't it fallback to the caption whatever it is null or '' because both are falsy. 
Or are you suggesting that if the title is '' that alt should be '' as well?

bibicollins1:
> @bibicollins1 Shouldn't it fallback to the caption whatever it is null or '' because both are falsy. Or are you suggesting that if the title is '' that alt should be '' as well?

@amerkadicE Oh yeah meant that it should fall back to caption in both cases so if that's all good we are covered 👍

bibicollins1 (approved):

PR #4016, opened 2024-12-20
Feat/tmpz 776 create user feedback box

josipVuko:
### Description

Created sticky note for feedback box for audio player

[TMPZ-776](https://nidigitalsolutions.jira.com/jira/software/c/projects/TMPZ/boards/1656?selectedIssue=TMPZ-776)

### Checklist


dbuljanNewsUk (approved):

suneethanagella (approved):

packages/save-and-share-bar/src/save-and-share-bar.js
```diff
@@ -247,6 +247,11 @@ function SaveAndShareBar(props) {
           )}
         </>
       ) : null}
+      {process.env.NODE_ENV !== "test" && (
```
adamosborne-tnl, line 250: can we also add a boolean here for `audioEnabled` and have it's default as false, and then in the tests and showcase we can enable it by setting it to true, much like we do with the `savingEnabled` or `sharingEnabled` ?

PR #4029, opened 2025-01-08
feat(TMG-000): testing pipeline error

sauravarora04:
### Description

[JIRA-1234](https://nidigitalsolutions.jira.com/browse/JIRA-1234)

### Checklist


packages/article-comments/src/article-comments.js
```diff
@@ -15,54 +13,68 @@ const ArticleComments = ({
...
-  const [userEntitlementData, setUserEntitlementData] = useState(undefined);
+  const [hasCommentingEntitlement, setHasCommentingEntitlement] = useState(
+    undefined
+  );
+
+  useEffect(() => {
+    const convertBase64JSONCookie = cookieValue => {
+      try {
+        return cookieValue
+          ? JSON.parse(Buffer.from(cookieValue, "base64").toString())
```
lukejsikuade, line 26: This isn't really safe. As far as I know Buffer is a Node package and not browser compatible.

`atob` is both browser safe and node16+ safe.

packages/article-comments/src/article-comments.js
```diff
@@ -15,54 +13,68 @@ const ArticleComments = ({
...
-    <DisabledComments />
-  );
+    fetchClientSideCookie();
+  }, []);
+
+  let content;
+  if (!isEnabled && !isCommentEnabled) {
+    content = <DisabledComments />;
+  } else if (hasCommentingEntitlement) {
+    content = (
```
lukejsikuade, line 62: Why not just use conditional rendering? This pattern is... unusual

sauravarora04: The purpose of this pattern was to wrap different condition results within a user state wrapper.

othetgyi: What are the benefits of using the user state wrapper versus conditional rendering?

sauravarora04: So, the user-state component is dealing with all user states (like hasAccess, isLoggedin, etc.). I've just tried to re-use the existing pattern.

othetgyi (approved):

.circleci/config.yml
```diff
@@ -110,7 +110,7 @@ jobs:
       - run:
           name: Run Web Tests
           command: |
-            lerna run test:web --stream --since -- -- --ci --bail --coverage
+            lerna run test:web --since -- -- --ci --coverage --verbose --runInBand --logHeapUsage
```
lukejsikuade, line 113: So  runInBand will probably help with memory usage, but it should also mean the tests run slower. Was this a required change?

lukejsikuade (approved):
One comment, nothing major

PR #4188, opened 2025-07-17
fix(TMPR-299): inline video url undefined

ibrahim-kurhan:
### Description

A large amount of video articles are using the incorrect Video URL (showing up as 'undefined' when we check this with the Schema Builder Chrome extension). 

Example page: https://www.thetimes.com/world/russia-ukraine-war/article/putin-peace-deal-talks-v6pwvvnmj. 
We need to ensure that the correct video URL is being used instead of the undefined value.

We can see that this is happening mostly for Video Articles which do not have a video as a lead asset. In the example above the lead asset is an image and the video is inline.

The following articles are populated correctly but these are both lead video:
https://www.thetimes.com/uk/religion/article/video-how-the-new-pope-will-be-chosen-tctndshtrbf 
https://www.thetimes.com/culture/tv-radio/article/actor-donald-sinden-dies-at-90-8zg57wbft9t 

To find examples of the articles with undefined,  we can use Google Search Console: and the video indexing error report.
https://search.google.com/search-console/video-index/drilldown?resource_id=sc-domain%3Athetimes.com&item_key=CAQYJiAD

Drilling into the  Video isn't on a watch page errors, you can see many  videos where the  Video URLs is marked as undefined and if you open these articles you can see that this appears to only happen when the  Video is inline

NOTE - the index error itself that is causing the errors in GSC is to do with the articles not being watch articles (ie video only) - that is a separate issue and wont be resolved until we decide to use video templates that are mainly just video with a small amount of text.

Issue created in Slack from a [message](https://times.slack.com/archives/C08S9MYUJNA/p1747298950543829?thread_ts=1747297598.812249&cid=C08S9MYUJNA).

Acceptance Criteria

- Ensure that the correct Video ID and URL is used so that the videos can be indexed correctly. 

- The Video URL should be displayed correctly in Google Search Console and no video should be displayed with the undefined value. If a video is underfined, it should not have the Video tag.
[JIRA-299](https://nidigitalsolutions.jira.com/browse/TMPR-299)

### Checklist

### DIFF
--- a/packages/article-skeleton/src/head.js
+++ b/packages/article-skeleton/src/head.js
@@ -8,6 +8,22 @@ import { appendToImageURL } from "@times-components/utils";
 
 const SYNDICATED_ARTICLE_IDS = ["37a19ac4-1cbb-11ee-8198-bf96b6365670"];
 
+function extractVideoFromContent(content = []) {
+  for (let i = 0; i < content.length; i += 1) {
+    if (
+      content[i].name === "video" &&
+      content[i].attributes.brightcoveVideoId
+    ) {
+      return content[i].attributes;
+    }
+    if (content[i].name === "paywall" && Array.isArray(content[i].children)) {
+      const nested = extractVideoFromContent(content[i].children);
+      if (nested) return nested;
+    }
+  }
+  return null;
+}
+
 function getIsLiveBlogExpiryTime(articleFlags = []) {
   let time = "";
   if (articleFlags !== undefined) {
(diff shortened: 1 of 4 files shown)


yukiinasawa (approved):

packages/article-skeleton/src/head.js
```diff
@@ -359,8 +375,13 @@ function Head({
     };
   }
 
-  const videoJsonLD = hasVideo
-    ? {
+  const videoFromContent = extractVideoFromContent(article.content);
```
HarisHodzicnt, line 378: should we call extractVideoFromContent always or just if hasVideo is true ?
Plus we can make logic to be less complex something like:
const bAccountId = brightcoveAccountId || videoFromContent.brightcoveAccountId;
const bVideoId = brightcoveVideoId ||  videoFromContent.brightcoveVideoId;

if(bAccountId && bVideoId ){
    videoJsonLD = {
        "@context": "https://schema.org",
        "@type": "VideoObject",
        name: leadAsset && leadAsset.title ? leadAsset.title : (videoFromContent.caption || title),
        uploadDate: dateModified,
        thumbnailUrl,
        description:
          Array.isArray(descriptionMarkup) && descriptionMarkup.length
            ? renderTreeAsText({ children: descriptionMarkup })
            : seoDescription || videoFromContent.caption || title,
        contentUrl: `https://players.brightcove.net/${
          videoFromContent.brightcoveAccountId
        }/default_default/index.html?videoId=${
          videoFromContent.brightcoveVideoId
        }`
      };
}

